PR #1, opened 2019-04-20
Config object with placeholder for schema validation

liljenback:
First version of config object

## Description

Config object to provide some abstraction helper methods to get configuration entries and prepare for module configuration schema validation support.

## Related Issue

CORE-27731

## Motivation and Context

Makes it easier for developers to access configurations and to ensure required configuration parameters is made available.

## How Has This Been Tested?

New unit tests were added and validated that sandbox still runs properly

## Types of changes

## Checklist:


liljenback:
@wgamboa I added a unit test for the undefined check.

Aaronius:
I have a few thoughts here, but I'm not sure how to express them. I think some of my thoughts may be in contrast to what Joe detailed in the JIRA issue, which may be a difference of opinion or perhaps a misunderstanding. I feel like we're trying to do too much here for the amount of value added, or at least that we're adding things before we know they will be needed maybe?

That said, here are some thoughts on the code itself:

* If we feel like we need generic utilities like for settings/getting properties of objects using dot-notation strings (`"a.b.c"`), I would recommend we just expose them like we do the rest of our utilities (in the utils directory) rather than putting them directly on the configuration object itself. Components could use the utilities with any object, including the configuration object. By putting them directly on the configuration object, I feel like it's a bit of a slippery slope to start making "smart objects" for all the data we pass around, which I don't think is necessary.
* For some of the logic, like the `get` method, we could have components use the nullish coalescing operator (https://github.com/tc39/proposal-nullish-coalescing) and configure our babel to transpile appropriately.
* For `extendSchema`, were you thinking about using JSON schema? If so, I don't think bundling a JSON schema validator would be worth the weight for validation. It might make more sense for each component to expose a validate method (rather than schema) that has their own validation logic. Personally, I think it's sufficient and simplest for components to just validate configuration data when the config object is passed into the component.
* With some of the validation logic we do, like `isObject(cfgAdd)) {  Object.assign(cfg, cfgAdd); }`, I'd rather straight up fail if `cfgAdd` isn't an object. Ignoring the call like we're doing currently is probably bad because the problem may go unnoticed for longer than it should, and adding an `else` with logging IMO isn't work the file size cost. It wouldn't help the Alloy user and the component developer could figure out the issue pretty easily with whatever the default error is that would be thrown. There's similar logic in `extendSchema`.
* I'd like to make sure we get enough value out of this code to make it worth the cost of file size and the increased knowledge needed to understand how to integrate a component.
* I don't think you'll be able to use `Set` in IE without a ponyfill, which I don't think is worth including.
* For `Object.assign`, you'll need to use our `assign` utility to make it work in IE.

liljenback:
Having a clear definition for configurations that can be enforced has a lot of value that we have just recently started to benefit from (in other projects). However, we have realized the value for years. Validation does add weight and we need to decide if there is enough benefit to be gained (I believe there is).

The schema definition I am thinking of would be a slimmed down version of a full-fledged JSON schema. Having individual components validate configurations will result in different interpretations on what that validation means. Some more advanced validation could still occur in each module if needed but a schema will provide common ways of introducing and enforcing default and required settings, etc. This in turn will let us more easily output a fully expanded configuration that the library is running on.

I could try moving the "a.b.c" feature to a utility. It becomes more useful when we later introduce more nested configurations.

PR #3, opened 2019-04-24
Handle Cookie destinations

wgamboa:
## Description
Handle Cookie destinations and adjust url destinations.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-26796

## How Has This Been Tested?
Manual and unit tests.

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Audiences/processDestinationsFactory.js
+++ b/src/components/Audiences/processDestinationsFactory.js
@@ -1,0 +1,63 @@
+/*
+Copyright 2019 Adobe. All rights reserved.
+This file is licensed to you under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License. You may obtain a copy
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+import createDestinations from "../../utils/createDestinations";
+import { isObject, isNonEmptyString, cookie } from "../../utils";
+
+export default ({ destinations, config, logger }) => {
+  const urlDestinations = destinations.reduce((arr, dest) => {
+    if (isObject(dest) && dest.type === "url" && isObject(dest.spec)) {
+      arr.push({
+        url: dest.spec.url,
+        hideReferrer: dest.spec.hideReferrer
+      });
+    }
+
+    return arr;
+  }, []);
+
+  if (
+    urlDestinations.length &&
+    (config.destinationsEnabled === undefined || config.destinationsEnabled)
+  ) {
+    const destsUtil = createDestinations({ logger });
+
+    destsUtil.fire(urlDestinations);
+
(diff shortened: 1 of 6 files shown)


src/components/Audiences/processDestinationsFactory.js
```diff
@@ -0,0 +1,63 @@
...
+    if (isNonEmptyString(dest.name)) {
+      cookie.set(dest.name, dest.value || "", {
+        domain: dest.domain || "",
+        expires: dest.ttl ? dest.ttl : 13 * 30
+      });
+    } else {
+      logger.error("Cookie destination had an invalid or no name.");
+    }
+  });
+};
```
Aaronius: I don't think this module would be considered a "factory" since it doesn't [return a "power constructor"](https://adobetechops.slack.com/archives/CG7S5ENS2/p1554382461016400).

src/components/Audiences/processDestinationsFactory.js
```diff
@@ -0,0 +1,63 @@
...
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+import createDestinations from "../../utils/createDestinations";
+import { isObject, isNonEmptyString, cookie } from "../../utils";
+
+export default ({ destinations, config, logger }) => {
+  const urlDestinations = destinations.reduce((arr, dest) => {
+    if (isObject(dest) && dest.type === "url" && isObject(dest.spec)) {
```
Aaronius: By ignoring the `dest` when it's not an object, I'm afraid a bigger problem will go unnoticed for too long. I'd rather a customer come to us saying "Hey, we're getting a 'Cannot read property 'type' of undefined' error" as soon as possible so we can fix it promptly for all our customers. Thoughts?

src/components/Audiences/processDestinationsFactory.js
```diff
@@ -0,0 +1,63 @@
...
+      arr.push({
+        url: dest.spec.url,
+        hideReferrer: dest.spec.hideReferrer
+      });
+    }
```
Aaronius: Is there any reason to not just `arr.push(dest.spec)`? If not, and considering my comment above, I think we could shrink this all down to:

```js
const urlDestinations = destinations.filter(dest => dest.type === "url").map(dest => dest.spec);
```

src/components/Audiences/processDestinationsFactory.js
```diff
@@ -0,0 +1,63 @@
...
+      arr.push({
+        name: dest.spec.name,
+        value: dest.spec.value,
+        domain: dest.spec.domain,
+        ttl: dest.spec.ttl
+      });
+    }
+
+    return arr;
+  }, []);
```
Aaronius: My comments above apply here as well.

src/components/Audiences/processDestinationsFactory.js
wgamboa: Thanks, I changed it. I'll consult with Joe to get better educated about these differences.

src/components/Audiences/processDestinationsFactory.js
wgamboa: Joe agrees.

wgamboa:
Thanks @Aaronius . I implemented all your suggestions.

PR #4, opened 2019-04-25
CORE-28595 - adding awaitSelector with some DOM utils

artur-ciocanu:
Adding a DOM polling utility named `awaitSelector` along with some DOM helpers.

## Description

Personalization module requires DOM polling. IFRAME destinations also requires some logic around DOM polling to figure out when BODY tag is available. `awaitSelector` tries to cover both use cases.

## Related Issue

NONE

## Motivation and Context

`awaitSelector` abstracts away all the logic related to DOM polling. It provides a nice API that plays well with promises. Also it allows some customizations like:
- how long to poll
- which method to use for polling

## How Has This Been Tested?

Unit tests have been added.

## Types of changes

## Checklist:


artur-ciocanu:
@wgamboa since I added some DOM helpers, could you please double check the destinations logic, to make sure I didn't mess it up.

src/utils/delay.js
```diff
@@ -0,0 +1,31 @@
...
+}
+
+/**
+ * Cancels an in progress delayed func
+ *
+ * @param {Number} The timer ID
+ */
+export function cancelDelay(id = -1, win = window) {
+  win.clearTimeout(id);
+}
```
Aaronius: It seems like the validation/conversion/fallback logic is pretty much how it behaves natively. In other words, if you pass `null` or `NaN` as the timeout to `setTimeout`,  internally it converts it to 0 automatically. `clearTimeout` is similarly lenient. I think we could remove some of the logic:

```
export default function delay(func, wait, win = window) {
  return win.setTimeout(func, wait);
}
```

and `cancelDelay` to:

```
export function cancelDelay(id, win = window) {
  win.clearTimeout(id);
}
```

But at that point I don't know what value the utility provides. I can see how it might be useful when mocking `window.setTimeout()` and `window.clearTimeout()`, but I think `jasmine.clock()` would suffice for those cases?

src/utils/dom/appendNode.js
```diff
@@ -0,0 +1,15 @@
...
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+export default function appendNode(parent, node) {
+  return parent.appendChild(node);
+}
```
Aaronius: I think this relates to my other comment talking about what should/should not be a utility and the slippery slope.

wgamboa:
@artur-ciocanu it looks like your changes to destinations make sense, but I used https://gist.github.com/wgamboa/024037380cd6ba4d61178e38266cb2db and I'm getting

alloy.js:962 Uncaught (in promise) TypeError: parent.appendChild is not a function
    at appendNode (alloy.js:962)
    at createPair (alloy.js:2088)

  function createPair(body) {
    var iframe = createNode(IFRAME_TAG, IFRAME_ATTRS);
    return {
      body: body,
      iframe: appendNode(body, iframe)
    };
  }

src/utils/dom/awaitSelector.js
```diff
@@ -0,0 +1,133 @@
...
+const VISIBLE = "visible";
+const DELAY = 100;
+const MAX_POLLING_TIMEOUT = 5000;
+
+function createError(selector) {
+  return new Error(`Could not find: ${selector}`);
+}
+
+function createPromise(func) {
+  return new Promise(func);
```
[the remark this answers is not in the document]
artur-ciocanu: Renamed `funct` to `executor`

src/utils/dom/awaitSelector.js
```diff
@@ -0,0 +1,133 @@
...
+  if (canUseMutationObserver(win)) {
+    return awaitUsingMutationObserver(win, doc, selector, timeout, selectFunc);
+  }
+
+  if (canUseRequestAnimationFrame(doc)) {
+    return awaitUsingRequestAnimation(win, selector, timeout, selectFunc);
+  }
+
+  return awaitUsingTimer(selector, timeout, selectFunc);
+}
```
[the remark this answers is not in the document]
artur-ciocanu: Renamed

src/utils/delay.js
artur-ciocanu: @Aaronius I gave it a second thought and I removed this utility. `setTimeout` is good enough there is no need to abstract it away.

src/utils/dom/appendNode.js
artur-ciocanu: I just wanted to be consistent with the other utility `createNode`, name wise, hence this almost trivial utility. Also while I don't have any checks here, I was thinking of adding `null` checks to make it more robust.

src/utils/keys.js
```diff
@@ -0,0 +1,21 @@
...
+*/
+
+/**
+ * Returns an array whose items are the provided object's own enumerable
+ * property keys.
+ * @param {Object} obj
+ * @returns {Array}
+ */
+export default obj => {
+  return Object.keys(obj);
```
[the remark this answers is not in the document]
artur-ciocanu: I have removed to `keys` utility for now. The only argument I have regarding transforming method invocations like `Object.keys()`, `Array.forEach()` etc to utilities is that you can have robust validation of input parameters. On the other hand if we control the whole stuff end to end probably this is an overkill.

artur-ciocanu:
@wgamboa good catch, I have fixed the issue.

src/utils/dom/awaitSelector.js
```diff
@@ -0,0 +1,132 @@
...
+  return createPromise((resolve, reject) => {
+    function execute() {
+      const nodes = selectFunc(selector);
+
+      if (isNonEmptyArray(nodes)) {
+        resolve(nodes);
+        return;
+      }
+
+      win[RAF](execute);
```
Aaronius: Selecting elements on every animation frame without any delay seems like it could be expensive and impact the user experience since it would try to run 60 times a second. I see it being recommended online without much mention of degraded performance though, so maybe I'm just not understand something fundamental or maybe element selection is just always fast enough to not be noticeable? There's some mention of combining `requestAnimationFrame` with a delay, like https://gist.github.com/chrisjhoughton/7890303#gistcomment-2334190. Anyway, I don't know that anything needs to be done here; I just thought I'd express my initial reaction.

Aaronius:
Thanks for the changes @artur-ciocanu. I re-approve!

src/utils/dom/awaitSelector.js
```diff
@@ -0,0 +1,132 @@
...
+
+const MUTATION_OBSERVER = "MutationObserver";
+const RAF = "requestAnimationFrame";
+const MUTATION_OBSERVER_CONFIG = { childList: true, subtree: true };
+const VISIBILITY_STATE = "visibilityState";
+const VISIBLE = "visible";
+const DELAY = 100;
+const MAX_POLLING_TIMEOUT = 5000;
+
+function createError(selector) {
```
Aaronius: I just noticed this and thought I'd bring up a discussion from a prior PR: https://git.corp.adobe.com/Activation/a-tag/pull/34#discussion_r1725571

I don't know how much the team cares about inconsistencies between function declarations vs function expressions, but if we want to stick to one or the other, we should probably add https://eslint.org/docs/rules/func-style to our eslint rules. I can do that in a separate PR and update any inconsistencies.

wgamboa:
> @wgamboa good catch, I have fixed the issue.

Thanks. I have confirmed the fix.

src/utils/dom/removeNode.js
```diff
@@ -0,0 +1,15 @@
...
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+export default function removeNode(parent, node) {
+  return parent.removeChild(node);
```
Aaronius: What do you think about doing:

```
if (node.parentNode) {
  node.parentNode.removeChild(node);
}
```

Aaronius: That would probably clean up some of the `createPair` stuff in `createDestinations` as well.

src/utils/dom/awaitSelector.js
artur-ciocanu: @Aaronius yeah, initially I was thinking that using RAF might be too aggressive, but we are using a similar approach in Adobe Target and it works great. Also this is mostly a fallback mechanism for IE 10, ideally we should use MutationObserver all the time.

src/utils/dom/removeNode.js
artur-ciocanu: Good call, adjusted.

src/utils/dom/awaitSelector.js
artur-ciocanu: @Aaronius let's decide and stick to one format, I agree that we should be consistent, I don't mind using func expressions. If we can enforce it via ESLint rules even better.

PR #5, opened 2019-04-26
Tweak base code for handling multiple instances.

Aaronius:
## Description

At the cost of slightly longer base code, this makes it simpler for someone to create multiple instances of Alloy. Previously, if a consumer wanted to create multiple instances of alloy, they'd basically have to paste the base code twice, like this:

```html
<script>
      (function(a,b){if(!a[b]){a.__adobeNS=a.__adobeNS||[];
      a.__adobeNS.push(b);a[b]=function(){a[b].q.push(arguments)};a[b].q=[]}})
      (window,'alloy');
      (function(a,b){if(!a[b]){a.__adobeNS=a.__adobeNS||[];
      a.__adobeNS.push(b);a[b]=function(){a[b].q.push(arguments)};a[b].q=[]}})
      (window,'example');
</script>
<script src="alloy.js" async></script>
```

I thought that was kind of lame. This tweaks the base code to the following:

```html
<script>
  (function(a,b){b.forEach(function(c){if(!a[c]){a.__alloyNS=a.__alloyNS||[],
  a.__alloyNS.push(c),a[c]=function(){a[c].q.push(arguments)},a[c].q=[]}})})
  (window,['alloy'])
</script>
<script src="alloy.js" async></script>
```

and if they want to add another instance, they just add the name to the array:

```html
<script>
  (function(a,b){b.forEach(function(c){if(!a[c]){a.__alloyNS=a.__alloyNS||[],
  a.__alloyNS.push(c),a[c]=function(){a[c].q.push(arguments)},a[c].q=[]}})})
  (window,['alloy','example'])
</script>
<script src="alloy.js" async></script>
```

Thoughts? I'll update the docs if this gets merged.

Btw, it seems like it would be nice to put unminified base code in a file somewhere and then minify it automatically. Unfortunately, I haven't found a good way to minify it properly without it trying to restructure the code in an undesirable way (e,g, moving the "alloy" string" to a different area, proper line breaks). I've just been editing it as minified code, but if you have better ideas let me know. **EDIT: I think I'll be able to work something out, but I'll submit it in a separate PR.**

## Related Issue

https://jira.corp.adobe.com/browse/CORE-28774

## Motivation and Context

The old way felt awkward.

## How Has This Been Tested?

Manually.

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

### DIFF
--- a/sandbox/public/index.html
+++ b/sandbox/public/index.html
@@ -14,8 +14,8 @@
     <title>Mock website hosting the Alloy</title>
 
     <script>
-      (function(a,b){b.forEach(function(c){if(!a[c]){a.__adobeNS=a.__adobeNS||[],
-      a.__adobeNS.push(c),a[c]=function(){a[c].q.push(arguments)},a[c].q=[]}})})
+      (function(a,b){b.forEach(function(c){if(!a[c]){a.__alloyNS=a.__alloyNS||[],
+      a.__alloyNS.push(c),a[c]=function(){a[c].q.push(arguments)},a[c].q=[]}})})
       (window,['alloy','ebda'])
     </script>
     <script src="alloy.js" async></script>
--- a/sandbox/public/index.html
+++ b/sandbox/public/index.html
@@ -14,14 +14,9 @@
     <title>Mock website hosting the Alloy</title>
 
     <script>
-      (function(a,b){if(!a[b]){a.__adobeNS=a.__adobeNS||[];
-      a.__adobeNS.push(b);a[b]=function(){a[b].q.push(arguments)};a[b].q=[]}})
-      (window,'alloy');
-
-      // For Testing multiple instances
-      // (function(a,b){if(!a[b]){a.__adobeNS=a.__adobeNS||[];
-      // a.__adobeNS.push(b);a[b]=function(){a[b].q.push(arguments)};a[b].q=[]}})
-      // (window,'ebda');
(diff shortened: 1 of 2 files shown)


sandbox/public/index.html
```diff
@@ -14,8 +14,8 @@
     <title>Mock website hosting the Alloy</title>
 
     <script>
-      (function(a,b){b.forEach(function(c){if(!a[c]){a.__adobeNS=a.__adobeNS||[],
-      a.__adobeNS.push(c),a[c]=function(){a[c].q.push(arguments)},a[c].q=[]}})})
+      (function(a,b){b.forEach(function(c){if(!a[c]){a.__alloyNS=a.__alloyNS||[],
+      a.__alloyNS.push(c),a[c]=function(){a[c].q.push(arguments)},a[c].q=[]}})})
       (window,['alloy','ebda'])
```
jfkhoury: Since most customers will have a single instance, should we support passing the namespace as a string, and only use an array if you need multiple?

```
...(window,'alloy')...

&
...(window,['alloy','ebda'])...
```

jfkhoury: The more I think about this, it doesn't seem necessary to change the array. Actually, defaulting to an array might be a good thing, because it'll educate the user that this library supports multiple instances.

Aaronius: I went through the same thought process. I actually built it that way initially (string unless you have multiple), but in the end thought it might be more straightforward for a user to just tack on another string rather than understand they need to change it to an array.

PR #6, opened 2019-04-29
Switched from callback pattern to promise pattern.

Aaronius:
## Description

It's been decided that even if we need to support Internet Explorer, we can require our customer to have `Promise` polyfilled on their site. With this decision made, we can now return promises rather than users needing to pass in a `callback` option. This pull request makes that happen.

There is one significant difference in behavior with this approach. With the previous approach, we could ensure that errors only hit the developer console if Alloy debugging is enabled. Now that we're returning a promise from each call and we reject the promise if there's an error, if the consumer doesn't handle the error, then the error will be reported to the console. This is just how promises work natively and we can't change that behavior. I think the only thing we could do is, in the case of an error, don't reject the promise returned to the consumer. I would think the consumer would always expect the promise to be rejected on an error though, regardless of whether Alloy debugging is enabled.

## Related Issue

https://jira.corp.adobe.com/browse/CORE-28780

## Motivation and Context

Simplified API for the user. Also allows us to make [more beneficial APIs](https://git.corp.adobe.com/gist/aahardy/f60aa30c361a7bf078af76fee13d39b0).

## How Has This Been Tested?

Manually 

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] (I'll update it after the PR) I have updated the documentation accordingly.


src/baseCode/index.js
```diff
@@ -0,0 +1,31 @@
...
+      window[instanceNamespace] = function() {
+        var userProvidedArgs = arguments;
+        return new Promise(function(resolve, reject) {
+          window[instanceNamespace].q.push([resolve, reject, userProvidedArgs]);
+        });
+      };
+      window[instanceNamespace].q = [];
+    }
+  });
+})(window, ["alloy"]);
```
Aaronius: This is the unminified source of our base code. I added an npm script in `package.json` for minifying.

liljenback: It probably make sense to add some comments/docs in this file describing what it does.

Aaronius:
Yeah, I agree. Let's see if it comes up from our users.

src/baseCode/index.js
Aaronius: Yeah, good idea. I will do so.

PR #7, opened 2019-04-30
Added correlation ID support.

Aaronius:
## Description

Added correlation ID support. Provides a `createCorrelationID` command that will return a promise resolving to a v4 UUID-compliant ID. The promise, or the ID the promise resolves to, can then be passed on `viewStart` and `event` commands.

## Related Issue

https://jira.corp.adobe.com/browse/CORE-27093

## Motivation and Context

https://wiki.corp.adobe.com/pages/viewpage.action?spaceKey=Activation&title=Universal+Event+Correlation

## How Has This Been Tested?

Manually and automated.

## Types of changes

## Checklist:

- [ ] (I will next) I have updated the documentation accordingly.

### DIFF
--- a/src/components/Correlation/index.js
+++ b/src/components/Correlation/index.js
@@ -10,25 +10,25 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import createPayloadItemMerger from "./createPayloadItemMerger";
+import { uuid } from "../../utils";
 
-export default () => {
-  const content = {};
-
+const createCorrelation = () => {
   return {
-    addIdentity(namespaceCode, identity) {
-      content.identityMap = content.identityMap || {};
-      content.identityMap[namespaceCode] =
-        content.identityMap[namespaceCode] || [];
-      content.identityMap[namespaceCode].push(identity);
+    lifecycle: {
+      onBeforeEvent(event, options) {
+        return Promise.resolve(options.correlationID).then(correlationID => {
+          if (correlationID !== undefined) {
+            event.correlationID = correlationID;
+          }
+        });
+      }
     },
-    addEvent(event) {
-      content.events = content.events || [];
-      content.events.push(event);
-    },
-    mergeMeta: createPayloadItemMerger(content, "meta"),
-    toJSON() {
-      return content;
+    commands: {
+      createCorrelationID: uuid
     }
   };
 };
+
+createCorrelation.namespace = "Correlation";
+
+export default createCorrelation;
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -8,6 +8,7 @@ module.exports = {
   plugins: ["prettier", "testcafe"],
   rules: {
     "prettier/prettier": "error",
+    "no-param-reassign": ["error", { props: false }],
     // Turning this off allows us to import devDependencies in our build tools.
     // We enable the rule in src/.eslintrc.js since that's the only place we
     // want to disallow importing extraneous dependencies.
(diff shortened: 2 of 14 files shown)


src/components/Correlation/index.js
```diff
@@ -0,0 +1,34 @@
...
+
+import { uuid } from "../../utils";
+
+const createCorrelation = () => {
+  return {
+    lifecycle: {
+      onBeforeEvent(event, options) {
+        return Promise.resolve(options.correlationID).then(correlationID => {
+          if (correlationID !== undefined) {
+            event.correlationID = correlationID;
```
Aaronius: ESLint will throw an error on this line without the `no-param-reassign` adjustment I've added to `.eslint.rc`. I feel like assigning a property to event will be something we'll be doing regularly, so I decided to adjust it in `.eslint.rc`. Let me know if you disagree with this.

src/components/Correlation/index.js
```diff
@@ -0,0 +1,34 @@
...
+governing permissions and limitations under the License.
+*/
+
+import { uuid } from "../../utils";
+
+const createCorrelation = () => {
+  return {
+    lifecycle: {
+      onBeforeEvent(event, options) {
+        return Promise.resolve(options.correlationID).then(correlationID => {
```
Aaronius: Note that we needed to get `correlationID` from the `options` object in order to handle it and attach it appropriately to the event. Previously, components didn't have access to `options` from within their `onBeforeEvent`, so I changed DataCollector to pass `options` along.

The alternative to this is to make DataCollector in charge of handling the `correlationID` and attaching it appropriately to the event. The role and responsibility of the DataCollector isn't well-defined to me so I wasn't really sure if that's what we want to do. It seemed to me like breaking out correlation ID handling into its own component made sense. Thoughts?

liljenback:
Wouldn't we need to be able to handle multiple concurrent correlation ID's at any time/event?

Aaronius:
I asked this question on [the wiki](https://wiki.corp.adobe.com/display/Activation/Universal+Event+Correlation). The latest from Justin was that we were not going to support multiple correlation IDs per event: https://wiki.corp.adobe.com/display/Activation/Universal+Event+Correlation?focusedCommentId=1647813818#comment-1647813818

Aaronius:
Since creating this pull request, I have added corresponding documentation to the user docs: https://github.com/AdobeDocs/alloy-docs/pull/1

src/components/Correlation/index.js
```diff
@@ -0,0 +1,34 @@
...
+          if (correlationID !== undefined) {
+            event.correlationID = correlationID;
+          }
+        });
+      }
+    },
+    commands: {
+      createCorrelationID: uuid
```
jfkhoury: I think we're going to need a `key` or `consumer` of each correlation ID. For example, a customer can generate multiple correlation IDs on the same page/view for different flows, and might want to get them by `key` and share them with 3rd party vendors later on in the page lifecycle:

```
const loginCorrelationID = correlation.createCorrelationID("login");
const checkoutCorrelationID = correlation.createCorrelationID("checkout");
....
....
const checkoutCorrelationID = correlation.getCorrelationID("checkout");
```

or maybe `createCorrelationID` creates an ID if it doesn't exist, or creates a new one and returns.

Let's confirm with Daniel and Lubby.

Aaronius: Yeah, that's interesting. You're thinking they'll want to retrieve them by a key because they won't want to pass the correlation ID around?

Would you like me to wait until we figure this out before merging?

jfkhoury: Yeah let's wait to learn more. I will start the conversation with Daniel and team as soon as possible and include you.

Aaronius:
There are more conflicts than I want to deal with. I'm going to close this PR and build off master for the planned changes.

PR #13, opened 2019-05-06
CORE-28203 Implement destroying the destination iFrame when done 

wgamboa:
## Description
Waits for all images to complete (loaded/errored/aborted) before removing the iframe.

## Related Issue
CORE-28203

## Motivation and Context
To complete destination processing.

## How Has This Been Tested?
Manual and unit.

## Types of changes

## Checklist:


Aaronius:
@wgamboa how were you manually testing this? I have some ideas for improvement, but I'm not sure if they'll work without some testing. If you could share any manual testing code you might have, that would be helpful.

src/utils/fireDestinationsFactory.js
```diff
@@ -1,45 +1,80 @@
...
+
+    if (processed === numberOfDestinations) {
+      logger.log(
+        `All destinations processed. Loaded: ${
+          processedDests.loaded.length
+        }; errored: ${processedDests.errored.length}; aborted: ${
+          processedDests.aborted.length
+        }`
+      );
+      destinationsProcessedDeferred.resolve(processedDests);
```
Aaronius: If I'm reading this right, it seems like `fireDestinations` only supports being called a single time. I believe this is the case because `destinationsProcessedDeferred` will be resolved after the first call to `fireDestinations` once all the destinations have fired. A promise can only be resolved once, so if `fireDestinations` were called again, resolving `destinationsProcessedDeferred` again would do nothing.

This spurs a few questions:

Is `fireDestinations` only intended to be called once? If YES, then why do we expose an `end` method from `createDestinations`, when it could handle "ending" internally after `fire` is called and is completed. If NO, then have a single promise that is resolved after the first `fire` call finishes doesn't seem rational. Instead, it would seem like we would want to return a promise from each `fire` call.

src/utils/fireDestinationsFactory.js
```diff
@@ -1,45 +1,80 @@
...
-import isObject from "./isObject";
 import isNonEmptyString from "./isNonEmptyString";
 import fireImage from "./fireImage";
 
 const fireOnPage = fireImage;
 
-export default ({ iframe, logger }) => {
-  const fireInIframe = ({ url }) => {
+const processedDests = {
+  loaded: [],
```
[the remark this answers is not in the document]
wgamboa: @jfkhoury this separation will be useful for ID syncs. loaded gets added to the tracking cookie, while the other two we might report to the server.

src/components/Audiences/processDestinations.js
```diff
@@ -16,7 +16,15 @@ import { isNonEmptyString, cookie } from "../../utils";
...
     .filter(dest => dest.type === "url")
-    .map(dest => dest.spec);
+    .map(dest => {
+      const data = {
+        id: dest.id
+      };
+
+      Object.assign(data, dest.spec);
+
+      return data;
```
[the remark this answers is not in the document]
wgamboa: Thanks.

src/utils/fireDestinationsFactory.js
```diff
@@ -1,45 +1,80 @@
...
+
+export default ({ iframe, logger, destinationsProcessedDeferred }) => {
+  const checkForCompletion = () => {
+    const processed =
+      processedDests.loaded.length +
+      processedDests.errored.length +
+      processedDests.aborted.length;
+
+    if (processed === numberOfDestinations) {
+      logger.log(
```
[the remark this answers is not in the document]
wgamboa: @jfkhoury I thought the log is turned on for debugging purposes.

src/utils/fireDestinationsFactory.js
wgamboa: @Aaronius thanks for the great catch. I took out end, and return a promise from each fire call. I tested manually on sandbox with logger.log and console.log.

wgamboa:
@Aaronius @jfkhoury please review again.

src/utils/fireDestinationsFactory.js
Aaronius: Okay, so I think we're saying that `fire` is only intended to be called once. If that's the case, I think there's a bunch more we can simplify here. For example, why do we need to call `createDestinations`, then call the `fire` method, then watch the `destinationsProcessedPromise` to be notified when it's done. Why not just create a module that is a function (say, `fireDestinations`) and when the function is called, it returns a promise that's resolved when destination firing is complete, and then goes to work on firing destinations? I'll try to code up a branch to demonstrate what I'm talking about.

src/utils/fireDestinationsFactory.js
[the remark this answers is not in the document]
Aaronius: There's a difference between us debugging and our customers debugging. I think Joe is saying that if it's for us to debug and not the customer, then we should remove the code.

src/utils/fireDestinationsFactory.js
Aaronius: Here's what I'm trying to say: https://github.com/adobe/alloy/pull/15

wgamboa:
@Aaronius does this have your approval?

src/utils/fireDestinationsFactory.js
wgamboa: fire is intended to be called once per onResponse, assuming there is data. I tested this successfully.

src/utils/dom/createNode.js
```diff
@@ -21,7 +21,14 @@ export default function createNode(
   const result = doc.createElement(tag);
 
   Object.keys(attrs).forEach(key => {
-    result.setAttribute(key, attrs[key]);
+    const value = attrs[key];
+
+    if (typeof value === "function") {
```
Aaronius: Hey @wgamboa, I just saw this. I'm not sure how I feel about it, since if the value is a function, then it technically isn't an attribute, but a property instead. I almost wonder if we should rename `attrs` to `props` and apply all the key-value pairs using `result[key] = value`.

wgamboa: @Aaronius you can change it in your next pr if you want.

PR #15, opened 2019-05-07
Proposal for fireDestinations

Aaronius:
Draft changes for https://github.com/adobe/alloy/pull/13#discussion_r281770342


Aaronius:
@wgamboa, I've done zero testing with this, but I wanted to convey what I was thinking.

package.json
```diff
@@ -35,13 +35,13 @@
...
-    "lint"
+//    "precommit-msg",
+//    "format",
+//    "lint"
   ],
   "pre-push": [
-    "prepush-msg",
-    "test"
+//    "prepush-msg",
+//    "test"
```
Aaronius: I didn't mean to commit this. If you merge this branch, be sure to uncomment these when tests are fixed.

wgamboa:
@Aaronius thank you for your proposal. However, there are certain things missing that I need. First, I need to keep track of which destinations loaded, errored, and aborted. This is going to be for ID syncs, Where successful fires Will have their IDs put into a cookie with a timestamp and unsuccessful fires might be reported back to the server. It can't Fail on unsuccessful fires because each fire has to be accounted for in order to remove the iframe at the end. The iframe needs to be removed because there will be separate instances for audiences and identity.

cc @jfkhoury

Aaronius:
@wgamboa 👍 No problem. What you've said makes sense and you should feel free to add what you need, I just wanted to convey my general thoughts on simplifying createDestinations, fireDestinations, and fireDestinationsFactory. Regarding iframe removal, I just forgot to add that; it wasn't intentional.

PR #16, opened 2019-05-08
Promises are failing on IE 11, removing it from the SL test suite

shammowla:
## Description

The unit tests are currently failing on IE 11. Removing it from the saucelabs tests list to avoid having a failed status on master.

Do we need Polyfilling for IE 11?


jfkhoury:
@shammowla Please merge this PR and start with ticket CORE-29371 to polyfill Promise and add IE 10 to the list of browsers we test on.

Thanks!

PR #21, opened 2019-05-09
CORE-28203 Make fireImage return a promise, simplify fireDestinations.

Aaronius:
## Description

This builds on https://github.com/adobe/alloy/pull/20. It changes `fireImage` to return a promise. Without doing so, it's likely that every consumer of `fireImage` will need to provide `onload`, `onerror`, and `onabort` attributes, which makes `fireImage` not particularly useful. This change does make the assumption that we don't need to distinguish between failed image requests and aborted image requests. I can't think of any use case where that distinction would be useful, but maybe you can. If the distinction is necessary, we can tweak some things while still returning a promise from `fireImage`.

Another theme of this PR is that it's my take (and we've discussed it as a group) that we shouldn't be adding logic that validates that data coming back from the server is in a valid state or that data from a component to a utility or another component is in a valid state. I believe we should typically make the assumption that the data is well-formed and blow up if it isn't. I believe validation should typically be reserved for user input/configuration where we don't own the input.

## Related Issue
CORE-28203

## Motivation and Context
Simplifying code. Make fireImage more useful.

## How Has This Been Tested?
Manually and automated.

## Types of changes

## Checklist:

### DIFF
--- a/src/utils/fireDestinations.js
+++ b/src/utils/fireDestinations.js
@@ -1,5 +1,3 @@
-import isNil from "./isNil";
-import isNonEmptyString from "./isNonEmptyString";
 import fireImage from "./fireImage";
 import { appendNode, awaitSelector, createNode, removeNode } from "./dom";
 
@@ -13,7 +11,8 @@ const IFRAME_ATTRS = {
   style: "display: none; width: 0; height: 0;"
 };
 
-const createFilterResultByStatus = status => result => result.status === status;
+const createFilterResultBySucceeded = succeeded => result =>
+  result.succeeded === succeeded;
 const mapResultToDest = result => result.dest;
 
 export default ({ logger, destinations }) => {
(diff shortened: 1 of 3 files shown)


src/utils/fireDestinations.js
```diff
@@ -29,73 +28,46 @@ export default ({ logger, destinations }) => {
...
-          } else {
-            logger.error(
-              `Destination hideReferrer property is not defined for url ${
-                dest.url
-              } .`
-            );
-          }
-        } else {
-          logger.error("Destination url is not a populated string.");
-        }
```
Aaronius: If we don't end up merging this PR, we need to add this code to each of the `else` statements:

```js
resolve({	
  status: "errored",	
  dest	
});
```

Otherwise, an error will be thrown down on line 92 when we try to filter results.

src/utils/fireDestinations.js
```diff
@@ -29,73 +28,46 @@ export default ({ logger, destinations }) => {
...
-      });
+      return imagePromise
+        .then(() => {
+          logger.log(`Destination failed: ${dest.url}`);
```
jfkhoury: Shouldn't this be `Destination succeeded`?

Aaronius: Definitely! Thanks for the catch!

Aaronius:
@wgamboa did you want to review this?

PR #22, opened 2019-05-09
Added configuration validation

liljenback:
## Description

This PR adds configuration validation and support for default settings.

## Motivation and Context

Makes it easier to troubleshoot configuration problems and present a the runtime configuration.

## How Has This Been Tested?

Unit tests, sandbox tests.

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Context/createComponent.js
+++ b/src/components/Context/createComponent.js
@@ -10,26 +10,20 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-export default (config, logger, availableContexts, defaultContextNames) => {
+export default (config, logger, availableContexts) => {
   let configuredContexts;
   return {
     namespace: "Context",
     lifecycle: {
       onComponentsRegistered() {
-        let configuredContextNames;
+        let configuredContextNames = [];
 
-        if (config.context) {
-          if (Array.isArray(config.context)) {
-            configuredContextNames = config.context;
-          } else {
-            logger.warn(
-              `Invalid configured context. Please specify an array of strings.`
-            );
-            configuredContextNames = [];
-          }
+        if (Array.isArray(config.context)) {
+          configuredContextNames = config.context;
         } else {
-          logger.log(`No configured context. Using default context.`);
-          configuredContextNames = defaultContextNames;
+          logger.warn(
+            `Invalid configured context. Please specify an array of strings.`
+          );
         }
 
         configuredContexts = configuredContextNames
--- a/src/core/createConfig.js
+++ b/src/core/createConfig.js
@@ -12,6 +12,9 @@ governing permissions and limitations under the License.
 
 import { assign, getNestedObject, setNestedObject } from "../utils";
 
+const CONFIG_DOC_URI =
+  "https://launch.gitbook.io/adobe-experience-platform-web-sdk/get-started/getting-started#configuration";
+
 const createConfig = config => {
   const cfg = {
     /**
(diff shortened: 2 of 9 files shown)


src/core/createConfig.js
```diff
@@ -48,18 +51,46 @@ const createConfig = config => {
...
+        const validator = cfg.validators[key];
+        const currentValue = cfg.get(key);
+        if (
+          !validator.validate &&
+          currentValue == null &&
+          validator.defaultValue
+        ) {
+          cfg.set(key, validator.defaultValue);
+        } else if (validator.validate) {
+          const errorMessage = validator.validate(
```
jfkhoury: Does the `validate`  function need the default value? Doesn't seem needed in `required` for example, which should have the same signature.

src/components/Context/createComponent.js
```diff
@@ -10,26 +10,20 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
-        if (config.context) {
-          if (Array.isArray(config.context)) {
-            configuredContextNames = config.context;
-          } else {
-            logger.warn(
-              `Invalid configured context. Please specify an array of strings.`
-            );
-            configuredContextNames = [];
-          }
+        if (Array.isArray(config.context)) {
```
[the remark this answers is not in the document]
liljenback: It can (should) be moved now, but I didn't want to modify existing code too much in this PR.

src/core/createConfig.js
```diff
@@ -80,3 +111,12 @@ const createConfig = config => {
 };
 
 export default createConfig;
+export const required = () => {
```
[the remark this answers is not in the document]
liljenback: I thought about taking "common validators" out of createConfig, but we only have that one right now. When we get more we could move it.

src/core/createConfig.js
liljenback: If there is a default defined for a whole object structure you might want to access it in the validate function. That is, if someone has defined several configurations but not specified one default value it could be picked up and set by getting it from the default value passed in to validate.

liljenback:
> On the right track, I love that we're not going the schema route :)
> 
> One thing I would add: I don't know if we need those nested object helpers.
> 
> I think if we pass `options` into create config and `Object.assign` them into config, you would get the same structure, and users can do: `config.key1.subKey2...`. It would save us from implementing those helpers. Thoughts?
> 
> Update: I see we are using those helpers to set the default values. Might be needed.

Yea, I'm using the helpers when setting defaults and getting values. I also think it will be useful when accessing nested configuration values (not having to check if each nested object exist).

src/core/createConfig.js
```diff
@@ -48,18 +51,46 @@ const createConfig = config => {
...
-      // TODO: Validate existing configuration against defined schema.
-      return true;
+      const keys = Object.keys(cfg.validators);
+      const errors = keys.reduce((ac, key) => {
+        const validator = cfg.validators[key];
+        const currentValue = cfg.get(key);
+        if (
+          !validator.validate &&
+          currentValue == null &&
+          validator.defaultValue
```
Aaronius: This should probably be `validator.hasOwnProperty('defaultValue')` to account for default values of `0`, empty string, etc.

src/core/createConfig.js
Aaronius: If I read this right, if the validator has both `defaultValue` and `validate` defined, the `defaultValue` won't be set on the configuration automatically like it would be if `validate` were not defined. Is that intended? If so, why?

Aaronius: Sorry, it seems I hijacked the thread unintentionally.

src/core/initializeComponentsFactory.js
```diff
@@ -20,13 +20,11 @@ export default (componentCreators, logger, getNamespacedStorage) => config => {
   componentCreators.forEach(createComponent => {
     const { namespace } = createComponent;
 
-    // TODO: Get the default config from the component
-    // const { componentConfigSchema } = createComponent;
-    // TODO: Extend the config with the component config schema
+    const { configValidators } = createComponent;
```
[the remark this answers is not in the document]
liljenback: I moved the validation out before module initialization.

src/core/createConfig.js
Aaronius: After your change, it now seems like `validator.validate` will only get called if default value is **not** applied. Is that intentional?

liljenback: Yes, if the default value is applied then validate is not called. That is if a default value has been specified (within a validator) I would expect that value to be valid.

Aaronius: That makes sense. Thanks for explaining.

PR #23, opened 2019-05-09
Cleaned up user-facing configuration options.

Aaronius:
## Description

Removed configuration options that will not be used. Moved configuration options used for testing out of the sandbox and into the library itself, so that it's not consumer-facing.

## Motivation and Context
Don't confuse users with configuration options that don't apply to them.

## Types of changes

## Checklist:

### DIFF
--- a/sandbox/public/index.html
+++ b/sandbox/public/index.html
@@ -23,13 +23,7 @@
 
     <script>
       alloy("configure", {
-        orgID: 12345,
         propertyID: 9999999,
-        dataSet: "dataset1",
-        shouldStoreCollectedData: 1,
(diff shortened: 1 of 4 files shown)


liljenback:
You could just add these as temp defaults now that the configuration code has been merged. But why not keep the test config separate from the code base?

Aaronius:
I've rebased on master and moved the values to validator defaults. @liljenback, the reason we're moving these options out of the sandbox is because we don't want them to keep showing up in demos. They only mean something to the engineering team and were confusing other people. We discussed this in yesterday's sprint planning. Let us know if you have a suggestion of a better place to put them that would resolve the concerns.

PR #24, opened 2019-05-15
Polyfill for IE 11 

shammowla:
## Description
Promises in unit tests were failing on IE 11. 

## Motivation and Context
Added "@babel/polyfill" as dependency and imported into unit tests.

## How Has This Been Tested?
Ran cross browser unit tests on Saucelabs.

### DIFF
--- a/karma-saucelabs.conf.js
+++ b/karma-saucelabs.conf.js
@@ -20,6 +20,12 @@ module.exports = function(config) {
       base: "SauceLabs",
       browserName: "firefox",
       version: "66.0"
+    },
+    sl_ie_11: {
+      base: "SauceLabs",
+      browserName: "internet explorer",
+      platform: "Windows 10",
+      version: "11.285"
     }
   };
 
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -32,6 +32,7 @@ module.exports = config => {
     ],
     // list of files / patterns to load in the browser
     files: [
+      "node_modules/@babel/polyfill/dist/polyfill.js",
       {
         pattern: "test/unit/**/*.spec.js",
         watched: false // The preprocessor will use its own watcher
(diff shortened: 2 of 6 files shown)


Aaronius:
@shammowla Can you fill me in a little more on the strategy here? Why did you choose to use @babel/polyfill in one test and core-js/modules/es6.promise in another? Why did you decide to import the polyfill in each test instead of including a promise polyfill in the [karma file configuration](http://karma-runner.github.io/4.0/config/files.html)?

shammowla:
@Aaronius In the last commit I was testing the commonjs vs the @babel/polyfill package. I added the @babel/polyfill to the Karma config. Should I make @babel/polyfill  a dev dependency?

Aaronius:
Thanks @shammowla. This is better, but using @babel/polyfill poses a problem since it polyfills more than just `Promise`. We're trying to mimic a real customer's website as closely as possible when we test. In our user documentation, we will tell our customers that if they need to support Internet Explorer, they will need to have polyfilled `window.Promise`. We don't say they need to polyfill anything else though, and if they don't polyfill anything else, then we shouldn't polyfill anything else in our tests either.

For example, if our library tries to use `Set`, which is not supported in Internet Explorer, tests should fail, because Alloy will also likely fail in Internet Explorer on our customer's websites. By introducing `@babel/polyfill` into our tests through Karma, they won't fail (a bad thing), because `Set` will have been polyfilled.

Instead of `@babel/polyfill`, I recommend using the Promise polyfill from https://www.npmjs.com/package/promise-polyfill. I recommend installing it as a dev dependency.

shammowla:
@Aaronius I switched to promise-polyfill. The "fireDestinations" spec is failing on IE11, but the other tests are looking good.

Aaronius:
👍 That's good! We're catching bugs!

PR #25, opened 2019-05-16
CORE-29236 Trigger ID syncs

wgamboa:
## Description
New config: idSyncsEnabled; defaults to true

## How Has This Been Tested?
Manual and unit.

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Audiences/processDestinations.js
+++ b/src/components/Audiences/processDestinations.js
@@ -18,41 +18,38 @@ import {
 } from "../../utils";
 
 export default ({ destinations, config, logger }) => {
-  const urlDestinations = destinations
-    .filter(dest => dest.type === "url")
-    .map(dest =>
-      assign(
-        {
-          id: dest.id
-        },
-        dest.spec
-      )
-    );
+  if (config.destinationsEnabled === undefined || config.destinationsEnabled) {
+    const urlDestinations = destinations
+      .filter(dest => dest.type === "url")
+      .map(dest =>
+        assign(
+          {
+            id: dest.id
+          },
+          dest.spec
+        )
+      );
 
-  if (
-    urlDestinations.length &&
-    (config.destinationsEnabled === undefined || config.destinationsEnabled)
-  ) {
-    fireDestinations({
-      logger,
-      destinations: urlDestinations
-    }).then(result => {
-      console.log(result);
+    if (urlDestinations.length > 0) {
+      fireDestinations({
+        logger,
+        destinations: urlDestinations
+      });
+    }
+
+    const cookieDestinations = destinations
+      .filter(dest => dest.type === "cookie")
+      .map(dest => dest.spec);
+
+    cookieDestinations.forEach(dest => {
+      if (isNonEmptyString(dest.name)) {
+        cookie.set(dest.name, dest.value || "", {
+          domain: dest.domain || "",
+          expires: dest.ttl ? dest.ttl : 6 * 30 // default of 6 months
+        });
+      } else {
+        logger.error("Cookie destination had an invalid or no name.");
+      }
     });
   }
-
-  const cookieDestinations = destinations
-    .filter(dest => dest.type === "cookie")
-    .map(dest => dest.spec);
-
-  cookieDestinations.forEach(dest => {
-    if (isNonEmptyString(dest.name)) {
-      cookie.set(dest.name, dest.value || "", {
-        domain: dest.domain || "",
-        expires: dest.ttl ? dest.ttl : 6 * 30 // default of 6 months
-      });
-    } else {
-      logger.error("Cookie destination had an invalid or no name.");
-    }
-  });
 };
--- a/src/components/Audiences/processDestinations.js
+++ b/src/components/Audiences/processDestinations.js
@@ -10,15 +10,12 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import {
-  assign,
-  isNonEmptyString,
-  cookie,
-  fireDestinations
-} from "../../utils";
+import { assign, cookie, fireDestinations } from "../../utils";
 
 export default ({ destinations, config, logger }) => {
-  if (config.destinationsEnabled === undefined || config.destinationsEnabled) {
+  const { destinationsEnabled = true } = config;
+
+  if (destinationsEnabled) {
     const urlDestinations = destinations
       .filter(dest => dest.type === "url")
       .map(dest =>
@@ -30,7 +27,7 @@ export default ({ destinations, config, logger }) => {
         )
       );
 
-    if (urlDestinations.length > 0) {
+    if (urlDestinations.length) {
       fireDestinations({
         logger,
         destinations: urlDestinations
@@ -42,14 +39,10 @@ export default ({ destinations, config, logger }) => {
       .map(dest => dest.spec);
 
     cookieDestinations.forEach(dest => {
-      if (isNonEmptyString(dest.name)) {
-        cookie.set(dest.name, dest.value || "", {
-          domain: dest.domain || "",
-          expires: dest.ttl ? dest.ttl : 6 * 30 // default of 6 months
-        });
-      } else {
-        logger.error("Cookie destination had an invalid or no name.");
-      }
+      cookie.set(dest.name, dest.value || "", {
+        domain: dest.domain || "",
+        expires: dest.ttl ? dest.ttl : 6 * 30 // default of 6 months
+      });
     });
   }
 };
--- a/src/components/Audiences/processDestinations.js
+++ b/src/components/Audiences/processDestinations.js
@@ -13,9 +13,7 @@ governing permissions and limitations under the License.
 import { assign, cookie, fireDestinations } from "../../utils";
 
 export default ({ destinations, config, logger }) => {
-  const { destinationsEnabled = true } = config;
-
-  if (destinationsEnabled) {
+  if (config.destinationsEnabled) {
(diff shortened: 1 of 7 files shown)


src/components/Audiences/processDestinations.js
```diff
@@ -18,41 +18,38 @@ import {
...
+      .map(dest => dest.spec);
+
+    cookieDestinations.forEach(dest => {
+      if (isNonEmptyString(dest.name)) {
+        cookie.set(dest.name, dest.value || "", {
+          domain: dest.domain || "",
+          expires: dest.ttl ? dest.ttl : 6 * 30 // default of 6 months
+        });
+      } else {
+        logger.error("Cookie destination had an invalid or no name.");
```
Aaronius: How/when would the destination not have a name? I think this is one of those things that should be validated when the user is saving destinations and not when we're using them. We should be able to assume that if we get destinations from the gateway, that they are valid and well-structured.

src/components/Identity/index.js
```diff
@@ -24,10 +25,22 @@ const addIdsContext = (payload, ecid) => {
...
-const createIdentity = () => {
+const createIdentity = ({ config, logger }) => {
   let ecid = getEcid();
   let deferredForEcid;
 
+  const onBeforeEvent = (event, isViewStart) => {
+    if (!isViewStart) {
+      return;
+    }
+    logger.log("onBeforeEvent");
```
Aaronius: Can we remove these logs?

src/components/Identity/processIdSyncs.js
```diff
@@ -0,0 +1,72 @@
...
+};
+
+const setControlObject = obj => {
+  cookie.set("adobeIdSyncControl", JSON.stringify(obj), {
+    expires: 6 * 30 // 6 months
+  });
+};
+
+export default ({ destinations, config, logger }) => {
+  if (config.idSyncsEnabled === undefined || config.idSyncsEnabled) {
```
Aaronius: Another place where you should be able to use destructuring defaults if you want:

```
const {
  idSyncsEnabled = true
} = config;
```

src/components/Audiences/processDestinations.js
```diff
@@ -18,41 +18,38 @@ import {
...
-  const cookieDestinations = destinations
-    .filter(dest => dest.type === "cookie")
-    .map(dest => dest.spec);
-
-  cookieDestinations.forEach(dest => {
-    if (isNonEmptyString(dest.name)) {
-      cookie.set(dest.name, dest.value || "", {
-        domain: dest.domain || "",
-        expires: dest.ttl ? dest.ttl : 6 * 30 // default of 6 months
+    if (urlDestinations.length > 0) {
```
Aaronius: Can we just do `if (urlDestinations.length) {` like you did in `processIdSyncs.js`?

src/components/Identity/processIdSyncs.js
```diff
@@ -0,0 +1,72 @@
...
+
+    Object.keys(controlObject).forEach(key => {
+      if (controlObject[key] < now) {
+        delete controlObject[key];
+      }
+    });
+
+    const idSyncs = destinations
+      .filter(
+        dest => dest.type === "url" && controlObject[dest.id] === undefined
```
Aaronius: It seems like you could change `controlObject[dest.id] === undefined` to `controlObject[dest.id]` without causing problems since the value of `controlObject[dest.id]` won't ever be `0`.

src/components/Identity/processIdSyncs.js
```diff
@@ -0,0 +1,72 @@
...
+        destinations: idSyncs
+      }).then(result => {
+        const timeStamp = Math.round(
+          new Date().getTime() / millisecondsPerHour
+        ); // hours
+
+        result.succeeded.forEach(idSync => {
+          const ttl = (idSync.ttl || 7) * 24; // hours
+
+          if (idSync.id !== undefined) {
```
Aaronius: When would `idSync` not have an ID? This is probably another one of those cases where we should be able to trust our own data.

test/unit/components/Identity/processIdSyncs.spec.js
```diff
@@ -0,0 +1,56 @@
...
+        123: Math.round(new Date().getTime() / 1000 / 60 / 60) - 10
+      }),
+      {
+        expires: 6 * 30 // 6 months
+      }
+    );
+
+    let obj = getControlObject();
+
+    expect(obj[123]).toBeDefined();
```
Aaronius: I think you're testing your test? Probably unnecessary, but 🤷‍♂.

src/components/Identity/processIdSyncs.js
```diff
@@ -0,0 +1,72 @@
...
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+import { assign, cookie, fireDestinations } from "../../utils";
+
+const millisecondsPerHour = 60 * 60 * 1000;
+
+const getControlObject = () => {
+  return JSON.parse(cookie.get("adobeIdSyncControl") || "{}");
```
Aaronius: Let's make sure we use @jileeshs "alloy cookie" when it's finished (https://github.com/adobe/alloy/pull/9). Otherwise, we should use the namespace we're using elsewhere (https://github.com/adobe/alloy/blob/master/src/constants/namespace.js).

src/components/Identity/processIdSyncs.js
Aaronius: I imagine some would argue that what you've put here is preferable over my suggestion, so don't feel like you have to do this.

test/unit/components/Identity/processIdSyncs.spec.js
wgamboa: I am doing that to show the difference in the same test later on after ID syncs are processed.

src/components/Identity/processIdSyncs.js
wgamboa: AAM's id is 0.

src/components/Audiences/processDestinations.js
```diff
@@ -18,41 +18,38 @@ import {
...
-    .filter(dest => dest.type === "url")
-    .map(dest =>
-      assign(
-        {
-          id: dest.id
-        },
-        dest.spec
-      )
-    );
+  if (config.destinationsEnabled === undefined || config.destinationsEnabled) {
```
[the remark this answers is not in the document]
wgamboa: Thanks.

src/components/Identity/processIdSyncs.js
wgamboa: AAM id is 0.

src/components/Identity/processIdSyncs.js
```diff
@@ -11,21 +11,24 @@ governing permissions and limitations under the License.
 */
 
 import { assign, cookie, fireDestinations } from "../../utils";
+import namespace from "../../constants/namespace";
 
 const millisecondsPerHour = 60 * 60 * 1000;
 
 const getControlObject = () => {
-  return JSON.parse(cookie.get("adobeIdSyncControl") || "{}");
+  return JSON.parse(cookie.get(`${namespace}idSyncControl`) || "{}");
```
jfkhoury: Let's add a TODO to use the `Alloy` cookie [PR26](https://github.com/adobe/alloy/pull/26) is merged.

src/components/Identity/processIdSyncs.js
```diff
@@ -0,0 +1,75 @@
...
+import namespace from "../../constants/namespace";
+
+const millisecondsPerHour = 60 * 60 * 1000;
+
+const getControlObject = () => {
+  return JSON.parse(cookie.get(`${namespace}idSyncControl`) || "{}");
+};
+
+const setControlObject = obj => {
+  cookie.set(`${namespace}idSyncControl`, JSON.stringify(obj), {
```
jfkhoury: Let's try this format so we can save more space:

```
{
    syncs: ID-TS_ID-TS_ID-TS
}
```
Since `encodeURIComponent("-_")` outputs `-_`.

src/components/Identity/processIdSyncs.js
```diff
@@ -0,0 +1,94 @@
...
+import { assign, cookie, fireDestinations } from "../../utils";
+import namespace from "../../constants/namespace";
+
+const millisecondsPerHour = 60 * 60 * 1000;
+
+// TODO: use alloy cookie once https://github.com/adobe/alloy/pull/26 is merged
+const getControlObject = () => {
+  const obj = {};
+  const val = cookie.get(`${namespace}idSyncControl`) || "";
+  let arr = [];
```
jfkhoury: Optional: you use `reduce` instead of `forEach`, and use `const arr = val ? val.split("_") : [];`.

src/components/Identity/processIdSyncs.js
```diff
@@ -0,0 +1,94 @@
...
+
+  Object.keys(obj).forEach(id => arr.push(`${id}-${obj[id]}`));
+
+  cookie.set(`${namespace}idSyncControl`, arr.join("_"), {
+    expires: 6 * 30 // 6 months
+  });
+};
+
+export default ({ destinations, config, logger }) => {
+  const { idSyncsEnabled = true } = config;
```
jfkhoury: Let's add `idSyncsEnabled` to the configuration map in the `Identity Component` and default it there to `true` instead of doing it here.

src/components/Identity/processIdSyncs.js
Aaronius: But can we just remove the `if` statement altogether? When would `idSync` have an `undefined` value?

PR #26, opened 2019-05-17
feature/CORE-27968 adding alloy cookie util

jileeshs:
Cookie util for writing and reading from common alloy cookie

## Description

Cookie util for writing and reading from common alloy cookie

## Related Issue

https://jira.corp.adobe.com/browse/CORE-26063

## Motivation and Context
https://jira.corp.adobe.com/browse/CORE-26063

## How Has This Been Tested?

Written unit tests

## Types of changes

## Checklist:

### DIFF
--- a/src/constants/cookieDetails.js
+++ b/src/constants/cookieDetails.js
@@ -1,0 +1,16 @@
+/*
+Copyright 2019 Adobe. All rights reserved.
+This file is licensed to you under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License. You may obtain a copy
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+export default {
+  ALLOY_COOKIE_NAME: "TEMP_ALLOY_COOKIE",
+  ALLOY_COOKIE_TTL: 180
+};
--- a/src/constants/cookieDetails.js
+++ b/src/constants/cookieDetails.js
@@ -11,6 +11,6 @@ governing permissions and limitations under the License.
 */
 
 export default {
-  ALLOY_COOKIE_NAME: "TEMP_ALLOY_COOKIE",
+  ALLOY_COOKIE_NAME: "TEMP_ALLOY_COOKIE", // TODO: Rename this cookie
   ALLOY_COOKIE_TTL: 180
 };
--- a/src/core/createCookie.js
+++ b/src/core/createCookie.js
@@ -11,6 +11,7 @@ governing permissions and limitations under the License.
 */
 import cookie from "../utils/cookie";
 import cookieDetails from "../constants/cookieDetails";
+import getTopLevelCookieDomain from "../utils/getTopLevelCookieDomain";
 
 const { ALLOY_COOKIE_NAME, ALLOY_COOKIE_TTL } = cookieDetails;
 
@@ -31,7 +32,7 @@ const createCookie = (prefix, id, cookieDomain = "") => {
       );
     },
     /**
-     * Updates and returns values from AlloyCookie for Components
+     * Updates the value  of a key from AlloyCookie for a prefix
      * @param {...*} arg Strings with key and value to be stored in alloy cookie.
      */
     set(key, value) {
@@ -46,7 +47,7 @@ const createCookie = (prefix, id, cookieDomain = "") => {
 
       cookie.set(cookieName, updatedCookie, {
         expires: ALLOY_COOKIE_TTL,
-        domain: cookieDomain // TODO use getTopLevelCookieDomain here
+        domain: cookieDomain || getTopLevelCookieDomain(window, cookie)
       });
     },
     /**
@@ -61,7 +62,7 @@ const createCookie = (prefix, id, cookieDomain = "") => {
         delete currentCookieParsed[prefix][key];
         cookie.set(cookieName, currentCookieParsed, {
           expires: ALLOY_COOKIE_TTL,
-          domain: cookieDomain // TODO use getTopLevelCookieDomain here
+          domain: cookieDomain || getTopLevelCookieDomain(window, cookie)
         });
       }
     }
--- a/src/core/createCookie.js
+++ b/src/core/createCookie.js
@@ -22,13 +22,13 @@ const createCookie = (prefix, id, cookieDomain = "") => {
      */
     get(name) {
       const cookieName = `${ALLOY_COOKIE_NAME}_${id}`;
-      let currentCookie;
-      try {
-        currentCookie = cookie.getJSON(cookieName);
-        return currentCookie[prefix][name];
-      } catch (error) {
-        return currentCookie;
-      }
+      const currentCookie = cookie.get(cookieName);
+      const currentCookieParsed = currentCookie && JSON.parse(currentCookie);
+      return (
+        currentCookieParsed &&
+        currentCookie[prefix] &&
+        currentCookie[prefix][name]
+      );
     },
     /**
      * Updates and returns values from AlloyCookie for Components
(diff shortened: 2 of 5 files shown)


src/core/createCookie.js
```diff
@@ -0,0 +1,79 @@
...
+     * @param {...*} arg String key stored in alloy cookie under a component prefix.
+     */
+    remove(name) {
+      const cookieName = `${ALLOY_COOKIE_NAME}_${id}`;
+      let currentCookie;
+      try {
+        currentCookie = cookie.getJSON(cookieName);
+        delete currentCookie[prefix][name];
+        cookie.set(cookieName, currentCookie, {
+          expires: ALLOY_COOKIE_TTL,
```
liljenback: Shouldn't this be configurable also?

src/core/initializeComponentsFactory.js
```diff
@@ -24,12 +29,20 @@ export default (componentCreators, logger, getNamespacedStorage) => config => {
   config.validate();
   componentCreators.forEach(createComponent => {
     const { namespace } = createComponent;
-    const storage = getNamespacedStorage(config.orgID);
```
liljenback: This doesn't look right. Did you pull the latest changes?

src/core/createCookie.js
jileeshs: We can make it configurable and give a default as well.
May be in the next iteration

src/core/initializeComponentsFactory.js
jileeshs: I have updated the files.

src/core/createCookie.js
```diff
@@ -0,0 +1,79 @@
...
+     * Updates and returns values from AlloyCookie for Components
+     * @param {...*} arg Strings with key and value to be stored in alloy cookie.
+     */
+    set(key, value) {
+      const cookieName = `${ALLOY_COOKIE_NAME}_${id}`;
+      let currentCookie;
+      try {
+        currentCookie = JSON.parse(cookie.get(cookieName));
+      } catch (error) {
+        console.log("catch");
```
jfkhoury: Let's not use `console` anymore since we have a logger.
If parsing the cookie is throwing, I would think we would want to know why, so maybe let's log an error at least.

src/core/createCookie.js
```diff
@@ -0,0 +1,79 @@
...
+        currentCookie = {};
+      }
+      console.log(2, JSON.stringify(currentCookie));
```
jfkhoury: Same here.. Let's remove all those fun `consoles` ;)

src/core/createCookie.js
```diff
@@ -0,0 +1,79 @@
...
+      const updatedCookie = {
+        ...currentCookie,
```
jfkhoury: Did we make sure this works in IE without adding an extra Babel plugin?

src/core/createCookie.js
jfkhoury: Let's add a TODO here that we need a Cookie lifetime config.

src/core/createCookie.js
```diff
@@ -0,0 +1,79 @@
...
+      cookie.set(cookieName, updatedCookie, {
+        expires: ALLOY_COOKIE_TTL,
+        domain: cookieDomain // TODO use getTopLevelCookieDomain here
+      });
+    },
+    /**
+     * Returns the value from AlloyCookie for a prefix and a key.
+     * @param {...*} arg String key stored in alloy cookie under a component prefix.
+     */
+    remove(name) {
```
jfkhoury: By name you mean the key of the object under the prefix? Do we need to name it better?

I didn't see a unit test for this `remove` API so it wasn't clear ;)

src/core/createCookie.js
```diff
@@ -0,0 +1,79 @@
...
+const { ALLOY_COOKIE_NAME, ALLOY_COOKIE_TTL } = cookieDetails;
+
+const createCookie = (prefix, id, cookieDomain = "") => {
+  return {
+    /**
+     * Returns the value from AlloyCookie for a prefix and a key.
+     * @param {...*} arg String key stored in alloy cookie under a component prefix.
+     */
+    get(name) {
+      const cookieName = `${ALLOY_COOKIE_NAME}_${id}`;
```
jfkhoury: I would place that variable above the `return` statement and reuse it everything since it doesn't look like it's changing.

src/core/createCookie.js
```diff
@@ -0,0 +1,79 @@
...
+        domain: cookieDomain // TODO use getTopLevelCookieDomain here
+      });
+    },
+    /**
+     * Returns the value from AlloyCookie for a prefix and a key.
+     * @param {...*} arg String key stored in alloy cookie under a component prefix.
+     */
+    remove(name) {
+      const cookieName = `${ALLOY_COOKIE_NAME}_${id}`;
+      let currentCookie;
```
jfkhoury: I think you can use `const` here, and place it inside the `try` clause.

src/core/createCookie.js
```diff
@@ -0,0 +1,79 @@
...
+        [prefix]: { ...currentCookie[prefix], [key]: value }
+      };
+      console.log(1, JSON.stringify(updatedCookie));
+      cookie.set(cookieName, updatedCookie, {
+        expires: ALLOY_COOKIE_TTL,
+        domain: cookieDomain // TODO use getTopLevelCookieDomain here
+      });
+    },
+    /**
+     * Returns the value from AlloyCookie for a prefix and a key.
```
jfkhoury: Let's fix this comment. This API returns a boolean indicating if the remove was successful or not.

src/constants/cookieDetails.js
```diff
@@ -0,0 +1,16 @@
...
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+export default {
+  ALLOY_COOKIE_NAME: "TEMP_ALLOY_COOKIE",
```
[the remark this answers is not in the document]
jileeshs: Added a TODO comment

src/core/createCookie.js
```diff
@@ -0,0 +1,79 @@
...
+
+const createCookie = (prefix, id, cookieDomain = "") => {
+  return {
+    /**
+     * Returns the value from AlloyCookie for a prefix and a key.
+     * @param {...*} arg String key stored in alloy cookie under a component prefix.
+     */
+    get(name) {
+      const cookieName = `${ALLOY_COOKIE_NAME}_${id}`;
+      let currentCookie;
```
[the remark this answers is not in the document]
jileeshs: Removed the try catch block as it was to catch when no cookie is present.
It need not be logged as it is not an error  or warning

src/core/createCookie.js
jileeshs: Removed

src/core/createCookie.js
jileeshs: Removed ;)

src/core/createCookie.js
jileeshs: It is passing the tests on IE

src/core/createCookie.js
```diff
@@ -22,55 +22,47 @@ const createCookie = (prefix, id, cookieDomain = "") => {
...
-        currentCookie = cookie.getJSON(cookieName);
-        return currentCookie[prefix][name];
-      } catch (error) {
-        return currentCookie;
-      }
+      const currentCookie = cookie.get(cookieName);
+      const currentCookieParsed = currentCookie && JSON.parse(currentCookie);
+      return (
+        currentCookieParsed &&
+        currentCookie[prefix] &&
```
jfkhoury: Shouldn't these be `currentCookieParsed[prefix] && currentCookieParsed[prefix][name]`?

Doesn't `JSON.parse` throw if `currentCookie` is `undefined` or an empty string? I think we still need to `try / catch` the `JSON.parse` expressions. You can even create a small helper, `safeJsonParse` that does the wrapping for you.

src/core/createCookie.js
```diff
@@ -22,55 +22,47 @@ const createCookie = (prefix, id, cookieDomain = "") => {
...
-      let currentCookie;
-      try {
-        currentCookie = JSON.parse(cookie.get(cookieName));
-      } catch (error) {
-        console.log("catch");
-        currentCookie = {};
-      }
-      console.log(2, JSON.stringify(currentCookie));
+      const currentCookie = cookie.get(cookieName)
+        ? JSON.parse(cookie.get(cookieName))
```
jfkhoury: Same here.

src/core/createCookie.js
jileeshs: Yes you are right, it should be `currentCookieParsed`
It wouldn't go to `JSON.parse` if `currentCookie` is `undefined`.

src/core/createCookie.js
jileeshs: here also if `cookie.get(cookieName)` is `undefined` it will not do `JSON.parse`

PR #41, opened 2019-05-23
Remove `viewStart` command and pass it as type instead.

jfkhoury:
## Description

We have been discussing this for a while now, @Aaronius brought it up again last week, so here's our chance to see it in action.

How do you like it?

## How Has This Been Tested?

Manual in Sandbox.

## Types of changes

- [ x ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

### DIFF
--- a/sandbox/public/index.html
+++ b/sandbox/public/index.html
@@ -30,9 +30,9 @@
       });
 
       alloy("event", {
-        // NOTE: `viewstart` is a special type of events.
+        // NOTE: `viewStart` is a special type of events.
         // View docs for more info.
-        type: "viewstart",
+        type: "viewStart",
         data: {
           "xdm:URL": [location.href],
           "xdm:name": [location.pathname.substring(1) || "home"]
--- a/sandbox/public/index.html
+++ b/sandbox/public/index.html
@@ -29,8 +29,10 @@
         debug: true
       });
 
-      alloy("viewStart", {
-        type: "view:start",
+      alloy("event", {
+        // NOTE: `viewstart` is a special type of events.
+        // View docs for more info.
+        type: "viewstart",
         data: {
           "xdm:URL": [location.href],
           "xdm:name": [location.pathname.substring(1) || "home"]
@@ -46,18 +48,6 @@
         idSyncsEnabled: true,
         debug: true
       });
-      //
-      // ebda("viewStart", {
-      //   data: {
-      //     pageName: ["landing"]
-      //   }
-      // });
-      //
-      // ebda("event", {
-      //   data: {
-      //     snack: ["oreos"]
-      //   }
-      // });
     </script>
   </head>
   <body>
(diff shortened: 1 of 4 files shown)


src/components/DataCollector/index.js
```diff
@@ -13,6 +13,8 @@ governing permissions and limitations under the License.
 import createRequest from "../../core/createRequest";
 import createEvent from "../../core/createEvent";
 
+const VIEW_START_EVENT = "viewstart";
```
jfkhoury: We will have to decide on a naming convention here, so `viewstart` might become.

sandbox/public/index.html
```diff
@@ -29,8 +29,10 @@
         debug: true
       });
 
-      alloy("viewStart", {
-        type: "view:start",
+      alloy("event", {
+        // NOTE: `viewstart` is a special type of events.
+        // View docs for more info.
+        type: "viewstart",
```
jonsnyder: This is the reason why we had a special "viewstart" command in the first place.  Namely, we didn't want to have to dictate that this type parameter be passed here.  It would be more explicit if we had a flag to tell personalization to do its "special" logic for viewstart.  i.e. we could have a flag like `fetchPersonalization: true`.  What do you think?

Aaronius: Is there a reason we wouldn't want to dictate that the `type` parameter be passed here? It seems like whether `viewstart` is indicated as a separate command, `type` option, or `fetchPersonalization` option, we're going to be dictating it regardless. The only thing I could see potentially wrong with using `type` is if a consumer would want to provide a unique `type` value when sending an event for a "view start"....but the purposes and usage of the `type` option is still bit vague to me.

src/components/DataCollector/index.js
Aaronius: What do you think about going camelCase with these event type values (e.g., `viewStart`)? I think I'd prefer it, probably because more complicated values will be easier to read.

src/components/DataCollector/index.js
```diff
@@ -26,8 +28,10 @@ const createDataCollector = ({ config }) => {
     );
   };
 
-  const createEventHandler = isViewStart => options => {
+  const createEventHandler = options => {
     const event = createEvent();
+    const isViewStart = options.type && options.type === VIEW_START_EVENT;
```
Aaronius: `options.type && ` shouldn't be necessary.

src/components/DataCollector/index.js
jfkhoury: I agree.

sandbox/public/index.html
jfkhoury: After talking with the edge team, product and few Alpha customers, it seems like `viewStart` might be the first of few `built-in` events that the Adobe solutions can leverage.

There maybe be some use cases where you might call `alloy("event"..` and not want to pass a `type` but I really can't think of any. If the visitor has triggered a specific event like maybe `logIn` or `addToCart`.. or if the Activity collector has triggered some click on a link or button.. or a passive event has occurred like a video finished playing.. you will always need a type for all those events.

If most of us agree, I'd like to have this change be part of the next Alpha-2 release and get feedback in the next Alpha meeting.

Aaronius: 👍 from me.

PR #44, opened 2019-05-29
Feature/collection url validation

liljenback:
Added support to make sure we have a secure collectionUrl endpoint.

## Description

I extended the configuration validation to support multiple validations per setting. I wrote a regexp validator to support collectionUrl validation.

## Related Issue

CORE-30039

## Motivation and Context

Avoiding configuration problems. 

## How Has This Been Tested?

Unit tests, sandbox.

## Types of changes

## Checklist:

### DIFF
--- a/src/components/DataCollector/index.js
+++ b/src/components/DataCollector/index.js
@@ -11,6 +11,10 @@ governing permissions and limitations under the License.
 */
 
 import createEvent from "../../core/createEvent";
+<<<<<<< HEAD
+=======
+import { secureUrl } from "../../utils/config-validators";
+>>>>>>> Moved configuration validators to cozy folder
 
 const VIEW_START_EVENT = "viewStart";
 
@@ -61,4 +65,18 @@ const createDataCollector = () => {
 
 createDataCollector.namespace = "DataCollector";
 
+<<<<<<< HEAD
+=======
+createDataCollector.configValidators = {
+  collectionUrl: {
+    defaultValue: "https://edgegateway.azurewebsites.net",
+    validate: secureUrl
+    // defaultValue: "http://ex-edge.stable-stage.aam-npe.adobeinternal.net/v1"
+  },
+  device: {
+    defaultValue: "UNKNOWN-DEVICE"
+  }
+};
+
+>>>>>>> Moved configuration validators to cozy folder
 export default createDataCollector;
--- a/src/components/DataCollector/index.js
+++ b/src/components/DataCollector/index.js
@@ -10,11 +10,7 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import createEvent from "../../core/createEvent";
-<<<<<<< HEAD
-=======
-import { secureUrl } from "../../utils/config-validators";
->>>>>>> Moved configuration validators to cozy folder
+import createEvent from "./createEvent";
 
 const VIEW_START_EVENT = "viewStart";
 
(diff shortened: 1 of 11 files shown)


src/components/DataCollector/index.js
```diff
@@ -51,7 +52,8 @@ createDataCollector.namespace = "DataCollector";
 
 createDataCollector.configValidators = {
   collectionUrl: {
-    defaultValue: "https://edgegateway.azurewebsites.net"
+    defaultValue: "https://edgegateway.azurewebsites.net",
+    validate: [required, matchesRegEx("^https://[^s/$.?#].[^s]*$")]
```
jonsnyder: Why do you have required here?  It isn't required because there is a default value.

jonsnyder: Also, did you mean to escape the s in your regex to match spaces?  and should it be a case insensitive match?

liljenback: I'm not sure if we will have a default, but I can remove the required for now

liljenback: Looks like it was automatically stripped of some characters during commit

src/core/createConfig.js
```diff
@@ -118,3 +129,11 @@ export const required = (config, key, currentValue) => {
   }
   return err;
 };
+export const matchesRegEx = ex => (config, key, currentValue) => {
```
jfkhoury: Couple of things:

1. This validator is too low level; no real point of a helper if we'll be asking the components' devs to bring in their own regex. Let's make this a private function and use it to expose a `matchUrl` validator.

`const matchUrl = createRegexValidator(regex);`

2. Since we have more than 1 validator now, let's move those bad boys out of here into their own cozy `validators` folder.

src/core/createConfig.js
```diff
@@ -63,24 +63,35 @@ const createConfig = config => {
...
-          const errorMessage = validator.validate(
-            cfg,
-            key,
-            currentValue,
-            validator.defaultValue
-          );
-          if (errorMessage) {
-            ac.push(errorMessage);
+          const validators = [];
+          if (Array.isArray(validator.validate)) {
```
jfkhoury: Following up on @jonsnyder's comment about removing the `required` validator, it seems like we won't need this array of validators for now. Let's take it out; and if we need to use more than 1 validators, I should definitely go the composition route.

src/core/createConfig.js
liljenback: I removed the regex validator (for now), and added a specific one for validating the secure URL. I also created a .. cozy validation folder that can host the validation we need going forward.

src/core/createConfig.js
liljenback: I'm keeping the support for stringing together validators. I think we have sufficient functionality to validate pretty much anything we need going forward.

src/components/DataCollector/index.js
```diff
@@ -51,7 +52,8 @@ createDataCollector.namespace = "DataCollector";
 
 createDataCollector.configValidators = {
   collectionUrl: {
-    defaultValue: "https://edgegateway.azurewebsites.net"
+    defaultValue: "https://edgegateway.azurewebsites.net",
+    validate: secureUrl
```
Aaronius: Should `collectionUrl` be `collectionDomain` instead (I think so)? And if that's the case, should we validate that it's a domain and not any beyond that?

liljenback: Good point. The only thing we get out of enforcing that https is specified is that it is clear that the end-point needs to be SSL. I don't know if that is enough to having to specify the protocol. If not this would instead be highlighted by requests failing.

PR #45, opened 2019-05-31
Attempt to use beacon transport when possible.

jfkhoury:
## Description

If the user sets `beacon: true` on the event, we will use the `Beacon` transport only if the event type if not `viewStart`. 

**TODOs to discuss:**

// TODO Validate `beacon`.
// TODO Might have to move somewhere closer to Network Manager.
// TODO Consider adding `beacon` as a config instead of
    // passing it on each command. This way `Alloy` will always
    // use beacon when possible.

## Motivation and Context

With the current implementation, the use can override the `viewStart` event and use `beacon` as transport which results in no response payload.

## How Has This Been Tested?

Sandbox.

## Types of changes

## Checklist:


src/components/DataCollector/index.js
```diff
@@ -29,9 +29,21 @@ const createDataCollector = () => {
...
+    // If the `beacon` config is set, we use it ONLY if not `viewStart`
+    // because `viewStart` events are expected to return a response.
+
+    // TODO Validate `beacon`.
+    // TODO Might have to move somewhere closer to Network Manager.
+    // TODO Consider adding `beacon` as a config instead of
+    // passing it on each command. This way `Alloy` will always
+    // use beacon when possible.
+
+    const shouldUseBeacon = isViewStart || !beacon ? false : beacon;
```
jonsnyder: One alternative would be to have something on the payload which determines if we are expecting a response or not.  Is this what the query section is for on the payload?  For example, if the query section is blank then use collect, else interact.  This gives the flexibility for components like identity and personalization that are listening to onBeforeSend to mark the initial payload as expecting a response.  We'd also need to support the exit-link use case where we want to mark the payload as not accepting any queries thereby forcing a collect call.

sandbox/src/index.js
```diff
@@ -19,8 +19,6 @@ import * as serviceWorker from './serviceWorker';
 
 ReactDOM.render(<App />, document.getElementById('root'));
 
-const promise = new Promise(resolve => resolve("done"));
-
```
jonsnyder: Thanks for removing this.  I added it to test if IE was working correctly with promises.

src/components/DataCollector/index.js
Aaronius: Sorry if I'm just being dumb, but is there a reason why `!beacon ? false : beacon` is better than just `Boolean(beacon)`?

Regarding:

```
// TODO Consider adding `beacon` as a config instead of
// passing it on each command. This way `Alloy` will always
// use beacon when possible.
```

I think that would be good. Also, if it's on the default config as `true`, and we always validate that a user-provided `beacon` value is a boolean, then we should be able to scrap `!beacon ? false : beacon` and `Boolean(beacon)` and just say `beacon`.

I think if the user specifies `beacon=false` and it's a `viewStart`, we should probably throw an error rather than assume `beacon=true`, otherwise we'll probably end up answering the question like "why is content re-rendering when I said `beacon=true`" or something like that.

jfkhoury:
@Aaronius @jonsnyder Moved comments to JIRA ticket to be worked on next week. Will merge this PR. Thank you for the comments!

src/components/DataCollector/index.js
Aaronius: Sorry, my comment should have said "Also, if it's on the default config as `false`". `false` makes more sense as a default than `true`. I just meant if there's a boolean on the default config and we validate that any user-provided value is a boolean, then we can assume that the value is a boolean by the time it gets to the DataCollector.

PR #47, opened 2019-06-03
CORE-30400 Clean up beacon logic.

Aaronius:
## Description

Cleaned up the logic and comments a little. In our check-in today, we decided that the option to force a beacon will only be used internally (not exposed to users). Because Personalization and other components must go through the command exposed by DataCollector, it's technically possible for the consumer to pass in `isBeacon: false` as an option into the command as well, but we would just not document or encourage it.

We do have a couple options to prevent consumers from passing in `isBeacon: false`, should we choose to do so:

1. Add a validator where we can say that no properties other than a specific set of properties can be passed. This would be similar to JSON schema's `additionalProperties: false`. This would require us to set up validation on the options passed to an `event` command (or any other command for that matter).  I don't think validation on event options is set up yet, but it would probably be helpful for use cases beyond just this one? It would also require us to list all the expected options to be passed in, in order to say that anything outside that list should be blocked. Again, having this list of supported options might be a good thing. Finally, it would also require that event commands from the user go through the validation process whereas event commands from other components do not. That might be a little weird.
2. Allow components to expose non-command methods to other components. This would allow DataCollector to expose a command to users that _doesn't_ support `isBeacon`, and expose a different method to other components that _does_ support `isBeacon`. This comes with some more questions about how components would access these methods and how to properly set up error boundaries.

Thoughts?

## Related Issue

https://jira.corp.adobe.com/browse/CORE-30400

## How Has This Been Tested?

Manual

## Types of changes

## Checklist:

- [X] I have updated the documentation accordingly.

### DIFF
--- a/src/components/DataCollector/createEvent.js
+++ b/src/components/DataCollector/createEvent.js
@@ -29,6 +29,9 @@ export default () => {
     mergeDevice: createMerger(content, "device"),
     mergeEnvironment: createMerger(content, "environment"),
     mergePlaceContext: createMerger(content, "placeContext"),
+    expectsResponse() {
+      return Boolean(content.query);
+    },
     toJSON() {
       return content;
     }
--- a/src/components/DataCollector/index.js
+++ b/src/components/DataCollector/index.js
@@ -18,26 +18,20 @@ const createDataCollector = () => {
   let lifecycle;
   let network;
 
-  const makeServerCall = (events, isBeacon) => {
-    const { payload, send } = network.newRequest(isBeacon);
-    events.forEach(event => payload.addEvent(event));
+  const makeServerCall = event => {
+    const expectsResponse = event.expectsResponse();
+    const { payload, send } = network.newRequest(expectsResponse);
+    payload.addEvent(event);
     send();
   };
 
   const createEventHandler = options => {
-    // isBeacon is not intended for consumer usage, but only
-    // for other components to use to force the request to
-    // use beacon transport.
-    const { isBeacon } = options;
     const event = createEvent();
     const isViewStart = options.type === VIEW_START_EVENT;
 
-    // viewStart events are always expected to return a response.
-    const shouldUseBeacon = !isViewStart && Boolean(isBeacon);
-
     event.mergeData(options.data);
     lifecycle.onBeforeEvent(event, isViewStart).then(() => {
-      makeServerCall([event], shouldUseBeacon);
+      makeServerCall(event);
     });
   };
 
--- a/src/components/DataCollector/index.js
+++ b/src/components/DataCollector/index.js
@@ -18,28 +18,22 @@ const createDataCollector = () => {
   let lifecycle;
   let network;
 
-  const makeServerCall = (events, beacon) => {
-    const { payload, send } = network.newRequest(beacon);
+  const makeServerCall = (events, isBeacon) => {
+    const { payload, send } = network.newRequest(isBeacon);
     events.forEach(event => payload.addEvent(event));
     send();
   };
 
   const createEventHandler = options => {
-    const { beacon } = options;
+    // isBeacon is not intended for consumer usage, but only
+    // for other components to use to force the request to
+    // use beacon transport.
+    const { isBeacon } = options;
     const event = createEvent();
     const isViewStart = options.type === VIEW_START_EVENT;
 
-    // Attempt to use Beacon transport.
-    // If the `beacon` config is set, we use it ONLY if not `viewStart`
-    // because `viewStart` events are expected to return a response.
-
-    // TODO Validate `beacon`.
-    // TODO Might have to move somewhere closer to Network Manager.
-    // TODO Consider adding `beacon` as a config instead of
-    // passing it on each command. This way `Alloy` will always
-    // use beacon when possible.
-
-    const shouldUseBeacon = isViewStart || !beacon ? false : beacon;
+    // viewStart events are always expected to return a response.
+    const shouldUseBeacon = !isViewStart && Boolean(isBeacon);
 
     event.mergeData(options.data);
     lifecycle.onBeforeEvent(event, isViewStart).then(() => {
--- a/src/core/network/createNetwork.js
+++ b/src/core/network/createNetwork.js
@@ -29,33 +29,44 @@ export default (config, logger, lifecycle, networkStrategy) => {
      * @property {function} send - call this function when you are ready to send the payload
      * @property {Object} payload - payload object that will be sent
      * @property {Promise} responsePromise - promise that will yield the raw response body
-     * @property {boolean} isBeacon - whether or not this request is a beacon request
      */
     /**
      * Create a new request.  Once "send" on the returned object is called, the lifecycle
-     *  method "onBeforeSend" will be triggered with { payload, responsePromise, isBeacon } as
+     *  method "onBeforeSend" will be triggered with { payload, responsePromise } as
      *  the parameter.  When the response is returned it will call the lifecycle method "onResponse"
      *  with the returned response object.
-     * @param {boolean} isBeacon - true to send a beacon (defaults to false).  If you send
-     *   a beacon, no data will be returned.
-     *
+     * @param {boolean} [expectsResponse=true] The endpoint and request mechanism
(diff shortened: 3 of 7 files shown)


src/components/DataCollector/index.js
```diff
@@ -18,28 +18,22 @@ const createDataCollector = () => {
...
-
-    // TODO Validate `beacon`.
-    // TODO Might have to move somewhere closer to Network Manager.
-    // TODO Consider adding `beacon` as a config instead of
-    // passing it on each command. This way `Alloy` will always
-    // use beacon when possible.
-
-    const shouldUseBeacon = isViewStart || !beacon ? false : beacon;
+    // viewStart events are always expected to return a response.
+    const shouldUseBeacon = !isViewStart && Boolean(isBeacon);
```
Aaronius: I noticed the logic used to be:

```
const shouldUseBeacon = isViewStart || ...
```

Which seems backward. If it's a view start, we want to ensure we do _not_ use a beacon.

jonsnyder:
Why do we have the requirement that "Personalization and other components must go through the command exposed by DataCollector"?

If this is not a requirement, a third option is to have the personalization component call the network gateway directly, and add logic to prevent `isBeacon: false` from being passed into the data collector.

jonsnyder:
Actually, why should anyone be able to pass in `isBeacon: false` to the data collector?  The network gateway should be able to decide which network strategy to use based on the payload.

jonsnyder:
Maybe we should rename the option as `forceBeacon`.  So if that is true, then it will send the data as a beacon.  If it is false, the network gateway can decide based on the payload.

Aaronius:
> Why do we have the requirement that "Personalization and other components must go through the command exposed by DataCollector"?

It's not really a requirement, it's just that we'll need to set up a system for accessing methods exposed by other components that properly takes error boundaries into consideration. When I did the work for improved error handling, I set it up so that an error boundary was created for every command (see https://github.com/adobe/alloy/blob/master/src/core/createComponentRegistry.js#L61). In other words, if a user runs a command from Personalization, then Personalization runs a command from Data Collector, then Data Collector throws an error, we now get an error that says something like:

```
[alloy] Error: [Personlization] An error occurred while executing the XYZ command
[DataCollector] An error occurred while executing the event command.
Caused by: foo is undefined.
```

We're able to show this level of detail because we've essentially wrapped each command in a proxy function that catches any error and prepends additional detail to the error object as the error bubbles up through the stack.

Previously, we had the registry set up so that one component could access another component, but there were no error boundaries set up. Some of the work I did removed any existing need for one component to access another component except through commands, so I removed that feature from the registry entirely. If we need it, we'll probably want to set it up in such a way that the component registry can proxy method calls across components and add the proper error boundaries.

> Actually, why should anyone be able to pass in isBeacon: false to the data collector? The network gateway should be able to decide which network strategy to use based on the payload.

I think the suggestion is for the network gateway to, for example, inspect the request payload and look for a specific pieces of personalization data. If so, I don't think it should be looking to see if some piece of personalization data is on the request, because that seems like a concern that should be owned by personalization and not a network gateway. In other words, while authoring the personalization component, the author would need to edit both the Personalization component AND the network gateway and any changes would need to stay in sync, which is probably a Bad Thing™. It makes the network gateway coupled to the personalization component. I could be persuaded otherwise though.

jonsnyder:
> for example, inspect the request payload and look for a specific pieces of personalization data

I was thinking there was a `query` section in the request.  If that is empty, then we can use beacon.
`const useBeacon = forceBeacon || isEmpty(payload.query)`

From your comment though, I'm assuming it is more complicated than that.  Another option is to have a method on the payload to mark it as expecting a response, so then we could have something like this:
`const useBeacon = forceBeacon || payload.isExpectingResponse()`

With this route we lose the error boundary like you said; however, It feels like if Personalization needs to collect data it should call DataCollector.event, but if it needs to request data, it should use the network gateway directly.

Aaronius:
I don't think I know enough about `query` to determine if we can key off of that. Two of the three places where we're setting `query` have a comment the say something like:

```
// TODO: Remove; We won't need to request id syncs explicitely.
// This is just for demo currently.
```
https://github.com/adobe/alloy/blob/f0564872054e36fdcc85e2ea642fad71b8e19450/src/components/Identity/index.js#L37

The [gateway API docs](https://wiki.corp.adobe.com/pages/viewpage.action?spaceKey=DMSArchitecture&title=Gateway+APIs) for `query` say:

> A-Tag SDKs SHALL permit overriding server-side decisioning by allowing to activate/deactivate capabilities or send them explicit request values (value structure TBD):

Which sounds to me like we could get back a response with valuable information even if `query` isn't set. It really isn't clear to me when we would or would not use `query` though.

Let me take some time to play around with your last suggestion.

Aaronius:
@jonsnyder Could you take another look?

As part of this, I ended up removing `isBeacon` from the object returned from `network.newRequest` as it seems redundant from the `expectsResponse` boolean we pass in. Let me know if you disagree.

src/components/DataCollector/createEvent.js
```diff
@@ -29,6 +29,9 @@ export default () => {
     mergeDevice: createMerger(content, "device"),
     mergeEnvironment: createMerger(content, "environment"),
     mergePlaceContext: createMerger(content, "placeContext"),
+    expectsResponse() {
+      return Boolean(content.query);
```
jfkhoury: My only comment:

I think we should we assume that we expect a response if the event type is `viewStart`, even if query is `undefined`. Thoughts?

Other than that, looks great.

Aaronius: I talked this over with @jfkhoury with our mouth-holes. We agree that we should move forward with the way it is here, but we'll be discussing `query` in more depth with Rares on Monday and will go from there.

Aaronius:
I'm going to go ahead and merge this but feel free to bring up anything later and we can make modifications accordingly.

src/core/network/createNetwork.js
```diff
@@ -29,33 +29,44 @@ export default (config, logger, lifecycle, networkStrategy) => {
...
+      const action = expectsResponse ? "interact" : "collect";
       const url = `https://${edgeDomain}/${action}?propertyID=${propertyID}`;
       const deferred = defer();
       const responsePromise = deferred.promise
         .then(() =>
-          lifecycle.onBeforeSend({ payload, responsePromise, isBeacon })
+          lifecycle.onBeforeSend({
+            payload,
+            responsePromise,
+            isBeacon: expectsResponse
```
jonsnyder: This logic is backwards.  It should be `isBeacon: !expectsResponse`.  Apparently, the tests do not catch this, so we should add a test (or fix a test) for this as well.

src/core/network/createNetwork.js
```diff
@@ -29,33 +29,44 @@ export default (config, logger, lifecycle, networkStrategy) => {
...
+    newRequest(expectsResponse = true) {
       const payload = createPayload();
-      const action = isBeacon ? "collect" : "interact";
+      const action = expectsResponse ? "interact" : "collect";
       const url = `https://${edgeDomain}/${action}?propertyID=${propertyID}`;
       const deferred = defer();
       const responsePromise = deferred.promise
         .then(() =>
-          lifecycle.onBeforeSend({ payload, responsePromise, isBeacon })
+          lifecycle.onBeforeSend({
```
jonsnyder: It's possible for a component to handle the onBeforeSend lifecycle method, and add an event with a query even though this request is a beacon.  We could add a guard for this in the payload object's addEvent method.

src/core/network/createNetwork.js
Aaronius: Good find! Fixed in https://github.com/adobe/alloy/pull/65.

src/core/network/createNetwork.js
Aaronius: It's a good point. Let's talk about this in today's check-in. We may actually want components to be able to add events with a query from within `onBeforeSend`. If so, we have an order-of-operations problem because we're calculating whether a response is expected before `onBeforeSend` is called. I think we need to come to an agreement about the purpose of `onBeforeSend` and the rules about what you can/can't do within that lifecycle phase.

PR #48, opened 2019-06-04
Added library info support.

Aaronius:
## Description

Adds a LibraryInfo component that exposes a `getLibraryInfo` command. Currently, it returns an object with the library's version on it. The source code has a version value of `{{version}}`, which is swapped at build time.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-29456

## Motivation and Context
Customers find it useful to figure out the version of the library on their page. We find it useful because the customer can more easily tell us which version of the library is on their page.

## How Has This Been Tested?

Manual and automated.

## Types of changes

## Checklist:

### DIFF
--- a/rollup.prod.config.js
+++ b/rollup.prod.config.js
@@ -11,11 +11,14 @@ governing permissions and limitations under the License.
 */
 
 import path from "path";
+import replace from "rollup-plugin-replace";
 import resolve from "rollup-plugin-node-resolve";
 import commonjs from "rollup-plugin-commonjs";
 import babel from "rollup-plugin-babel";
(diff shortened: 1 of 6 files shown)


rollup.prod.config.js
```diff
@@ -33,6 +36,12 @@ export default {
     resolve(),
     commonjs(),
     babel(),
+    replace({
+      delimiters: ["{{", "}}"],
+      values: {
+        version
+      }
+    }),
```
Aaronius: I decided to only swap `{{version}}` with the real package version when creating a production build. My reasoning was that it will be less code to maintain and it doesn't seem valuable/applicable in non-prod builds.

PR #49, opened 2019-06-05
use debug_enabled query string parameter

jonsnyder:
Use debug_enabled query string parameter to turn debugging on or off.

## Related Issue

https://jira.corp.adobe.com/browse/CORE-28206

## Motivation and Context

Allow customers to turn on debugging without modifying the code of the website.

## How Has This Been Tested?

unit tests for the stringToBoolean util.  Manual sandbox testing to test the createInstance changes.

## Types of changes

## Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -34,7 +34,7 @@
       "integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
       "dev": true,
       "requires": {
-        "@babel/highlight": "^7.0.0"
+        "@babel/highlight": "7.0.0"
       }
     },
     "@babel/core": {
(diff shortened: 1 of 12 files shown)


Aaronius:
@jonsnyder I added a commit with a fix for the build issue.

Aaronius:
BTW, I think this looks good. I wonder if we should call the parameter `alloy_debug_enabled` to distinguish it from other things that might use `debug_enabled`? I don't know......if you've considered it and decided against it then that's all good with me.

From the demo to the alpha customers, it sounds like we might be renaming "debug" to "log" or "logging" or something like that, so the query param may eventually change due to that as well FYI.

jonsnyder:
The downside to using debug_enabled is it may conflict with a param that the customer is using internally.  This would clog up their logs with alloy logs.  You are right that `debug_enabled` should probably be changed to `log_enabled`.  Of course then we could support log levels with this parameter and call it `log_level`

1. `log_enabled` - may conflict with customer param
2. `alloy_log_enabled`
3. `[instanceVariableName]_log_enabled` - this would allow you to see just one instance's logging

@jfkhoury what do you think?

jfkhoury:
@jonsnyder I would prefix it with ‘alloy’. ‘alloy_log’ or ‘alloy_log_enabled’

jfkhoury:
@Aaronius @jonsnyder  I added a commit to rename `debug_enabled` to `alloy_log`.

PR #56, opened 2019-06-10
CORE-30736 Returning request+response data from event command.

Aaronius:
## Description

This allows users to access the request body and response body after sending an event, like so: 

```
alloy("event", { ... }).then(function(data) {
  // data.requestBody
  // data.responseBody
});
```

Likewise, when the request is sent, we log the request body. When the response comes back, we log the response body.

There was some question as to what we should do when we don't expect a response from the server (in other words, it's just a beacon). In this case, the `data` object returned from the command will not have a `responseBody` property. Regarding logging, when the message that gets logged when a request goes out, it will say `Sending network request (response will be ignored): <requestbody>` instead of just `Sending network request: <requestbody>`. The response will not be logged.

## Related Issue
CORE-30736

## Motivation and Context

This is in response to alpha tester feedback where they stated they wanted access to the request and response after executing an `event` command.

## How Has This Been Tested?
Manually and automated, but we need more automated tests around the Data Collector. I think in this case they should probably be functional tests, which I'd like to start getting more involved with.

## Types of changes

## Checklist:

- [ ] I have updated the documentation accordingly. **Not yet...I want to get feedback on this first**

### DIFF
--- a/src/core/network/createNetwork.js
+++ b/src/core/network/createNetwork.js
@@ -12,61 +12,91 @@ governing permissions and limitations under the License.
 
 import createPayload from "./createPayload";
 import createResponse from "./createResponse";
-import defer from "../../utils/defer";
 
 export default (config, logger, lifecycle, networkStrategy) => {
-  const handleResponse = body => {
-    return new Promise(resolve => resolve(createResponse(JSON.parse(body))))
-      .then(response => lifecycle.onResponse(response))
-      .catch(e => logger.warn(e));
+  const handleResponse = responseBody => {
+    let parsedBody;
+
+    try {
+      parsedBody = JSON.parse(responseBody);
+    } catch (e) {
+      throw new Error(
+        `Error parsing server response.\n${e}\nResponse body: ${responseBody}`
+      );
+    }
+
+    logger.log("Received network response:", parsedBody);
+
+    const response = createResponse(parsedBody);
+
+    return lifecycle.onResponse(response).then(() => response);
   };
 
   const { edgeDomain, propertyID } = config;
+
   return {
     /**
      * The object returned from network.newRequest
+     *
      * @typedef {Object} Request
-     * @property {function} send - call this function when you are ready to send the payload
-     * @property {Object} payload - payload object that will be sent
-     * @property {Promise} responsePromise - promise that will yield the raw response body
+     * @property {Function} send Call this function when you are ready to send
+     * the payload. Returns a promise yielding the raw response body.
+     * @property {Object} payload Payload object that will be sent.
      */
     /**
-     * Create a new request.  Once "send" on the returned object is called, the lifecycle
-     *  method "onBeforeSend" will be triggered with { payload, responsePromise } as
-     *  the parameter.  When the response is returned it will call the lifecycle method "onResponse"
-     *  with the returned response object.
+     * Create a new request.  Once "send" on the returned object is called, the
+     * lifecycle method "onBeforeSend" will be triggered with
+     * { payload, responsePromise, isBeacon } as the parameter.  When the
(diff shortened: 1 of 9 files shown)


src/core/network/createResponse.js
```diff
@@ -25,23 +25,25 @@ import { assign, find } from "../../utils";
...
+        handleContent => handleContent.type === type
+      );
+      return fragment ? fragment.payload : null;
+    },
+    toJSON() {
+      return content;
+    }
+    // TODO: Maybe consider the following API as well?
+    // - getPayloadsByAction
   };
```
Aaronius: The way `createResponse.js` was set up was different from `createPayload.js` and it was useful to make them more consistent. That is, both of them have accessors to get underlying data and both of them have `toJSON` defined to get the underlying data for serialization and cloning purposes. In the case of this pull request, this allows to us to clone the response and provide the result to users so that any changes they make to that data will not affect the data we hold internally.

test/unit/core/network/createNetwork.spec.js
```diff
@@ -161,62 +165,44 @@ describe("createNetwork", () => {
...
-      networkStrategy
-    );
-    const { send } = network.newRequest();
-    send();
-    setTimeout(() => {
-      expect(loggerSpy.warn).toHaveBeenCalledTimes(1);
-      done();
-    }, 100);
-  });
-
```
Aaronius: We were logging parse errors but instead it made sense to throw an error. That error will bubble up, get augmented with additional context along the way, and then the user can either handle the error in a `catch` or, if they don't, it will be logged to the console by virtue of it being an unhandled promise exception.

test/unit/core/network/createNetwork.spec.js
```diff
@@ -18,58 +18,44 @@ describe("createNetwork", () => {
...
+      return Promise.resolve();
     };
-    const network = createNetwork(
-      config,
-      logger,
-      nullLifecycle,
-      networkStrategy
-    );
-    const { send } = network.newRequest();
-    send();
```
Aaronius: Some of the changes to the file, like the changes to this test, were for code simplification.

src/core/network/createNetwork.js
```diff
@@ -12,61 +12,91 @@ governing permissions and limitations under the License.
...
+            return handleResponsePromise;
+          });
 
-      if (expectsResponse) {
-        responsePromise.then(handleResponse);
-      }
+        return responsePromise;
+      };
 
-      return { payload, responsePromise, send: deferred.resolve };
```
Aaronius: We no longer return a `responsePromise` when `newRequest` is called. It made more sense to me to return the promise from `send` when it is called.

src/core/network/createNetwork.js
```diff
@@ -12,61 +12,91 @@ governing permissions and limitations under the License.
...
 
 export default (config, logger, lifecycle, networkStrategy) => {
-  const handleResponse = body => {
-    return new Promise(resolve => resolve(createResponse(JSON.parse(body))))
-      .then(response => lifecycle.onResponse(response))
-      .catch(e => logger.warn(e));
+  const handleResponse = responseBody => {
+    let parsedBody;
+
+    try {
```
jfkhoury: Nice!

src/core/network/createNetwork.js
```diff
@@ -12,61 +12,91 @@ governing permissions and limitations under the License.
...
+          });
 
-      if (expectsResponse) {
-        responsePromise.then(handleResponse);
-      }
+        return responsePromise;
+      };
 
-      return { payload, responsePromise, send: deferred.resolve };
+      return { payload, send };
```
jfkhoury: 👍

PR #59, opened 2019-06-12
Build dev and prod libraries.

jfkhoury:
## Description

Build the minified and un-minified libraries at all times.

## How Has This Been Tested?

Manual.

## Types of changes

- [ x ] New feature (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "prebuild": "npm run clean && npm run format && npm run lint",
     "build": "rollup -c",
     "prebuild:prod": "npm run prebuild",
-    "build:prod": "rollup -c rollup.prod.config.js",
+    "build:prod": "rollup -c && rollup -c rollup.prod.config.js",
     "build:watch": "rollup -c -w",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -18,7 +18,7 @@
     "functional": "BROWSERS='chrome,firefox' TEST_FILES='test/functional/specs/**/*.js' npm run testcafe",
     "functional-sl": "TEST_FILES='test/functional/specs/**/*.js' npm run testcafe",
     "prebuild": "npm run clean && npm run format && npm run lint",
-    "build": "rollup -c",
+    "build": "rollup -c && rollup -c rollup.prod.config.js",
```
Aaronius: What's your reasoning for building the minified version on `npm run build`? I was thinking instead you would change `npm run build:prod` to build both the unminified and minified versions.

jfkhoury: Good question I didn't think of it this way. I thought why not build both all the times? I'm good either way ;)

PR #61, opened 2019-06-13
CORE-30737 Log command execution.

Aaronius:
## Description

Log commands as they are executed. I also made an effort to be consistent in how we log messages.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-30737

## Motivation and Context
Users want to know what's going on inside alloy.

## How Has This Been Tested?

Manually in the sandbox.

## Types of changes

## Checklist:

### DIFF
--- a/src/core/initializeComponentsFactory.js
+++ b/src/core/initializeComponentsFactory.js
@@ -50,8 +50,7 @@ export default (
     componentRegistry.register(namespace, component);
   });
 
-  // Output the finalized configuration
-  logger.info("Runtime configuration:\n", JSON.stringify(config, null, 2));
+  logger.log("Computed configuration:", config.toJSON());
 
   const lifecycle = createLifecycle(componentRegistry);
   const network = createNetwork(config, logger, lifecycle);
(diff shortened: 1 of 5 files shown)


src/core/initializeComponentsFactory.js
```diff
@@ -50,8 +50,7 @@ export default (
     componentRegistry.register(namespace, component);
   });
 
-  // Output the finalized configuration
-  logger.info("Runtime configuration:\n", JSON.stringify(config, null, 2));
+  logger.log("Computed configuration:", config.toJSON());
```
jfkhoury: Will this break the format of the JSON in the console?

Aaronius: What do you mean by "break the format"? `toJSON` returns an object so the browser will just log it as an object rather than a string.

jfkhoury: I guess logging an object is good as well; they can expand it.

Screen Shot 2019-06-13 at 5 34 08 PM

Aaronius: Yeah, I feel like it's preferable.

PR #65, opened 2019-06-18
CORE-30400 Fix isBeacon boolean sent to onBeforeSend.

Aaronius:
## Description
Fixes the `isBeacon` boolean that's sent to `lifecycle.onBeforeSend`. It was the opposite of what it should be. This was [found by @jonsnyder](https://github.com/adobe/alloy/pull/47#discussion_r294391542).

## Related Issue
https://jira.corp.adobe.com/browse/CORE-30400

## How Has This Been Tested?
Automated tests.

## Types of changes

## Checklist:

### DIFF
--- a/src/core/network/createNetwork.js
+++ b/src/core/network/createNetwork.js
@@ -63,7 +63,7 @@ export default (config, logger, lifecycle, networkStrategy) => {
             lifecycle.onBeforeSend({
               payload,
               responsePromise,
-              isBeacon: expectsResponse
+              isBeacon: !expectsResponse
             })
           )
           .then(() => {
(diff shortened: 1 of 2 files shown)


src/core/network/createNetwork.js
```diff
@@ -63,7 +63,7 @@ export default (config, logger, lifecycle, networkStrategy) => {
             lifecycle.onBeforeSend({
```
jfkhoury: @Aaronius @jonsnyder What do you guys think about renaming this hook to `onBeforeNetworkCall` ?

It's really cryptic as it is now since you would need to know that the `Network Gateway` dude is the one triggering this hook, to understand what `send` means.

jonsnyder: I like it.

PR #70, opened 2019-06-20
CORE-31332 Log raw payload with raw events.

Aaronius:
## Description

When we log the payload before a network request goes out, we need to make sure the object is the raw data and not any of our wrapper objects. We were logging this, as an example:

image

See how the event is our fancy wrapper object with all the helpers methods and such? Our users don't care about that. They want the raw data. The reason we were outputting the fancy event wrappers is because we were logging the result of `payload.toJSON()`. Payload's `toJSON()` returns the raw payload data, but it doesn't recursively call `toJSON()` on the fancy event wrappers that it holds. However, `JSON.stringify(payload)` _does_ call `toJSON()` recursively. As a result, if we parse the result of `JSON.stringify(payload)`, we arrive at the full raw payload data in the form of a JavaScript object.

There are a few ways of addressing this problem, but I feel this is the best one at the moment. It does have a performance impact on logging (we have to parse the payload JSON), but I think we can address the performance impact as part of https://jira.corp.adobe.com/browse/CORE-31336

## Related Issue
https://jira.corp.adobe.com/browse/CORE-31332

## Motivation and Context
Users don't care about our fancy wrapper objects. They want the data.

## How Has This Been Tested?
Automated & manual.

## Types of changes

## Checklist:

### DIFF
--- a/src/core/network/createNetwork.js
+++ b/src/core/network/createNetwork.js
@@ -72,9 +72,17 @@ export default (config, logger, lifecycle, networkStrategy) => {
             const responseHandlingMessage = expectsResponse
               ? ""
               : " (response will be ignored)";
+            const stringifiedPayload = JSON.stringify(payload);
+
+            // We want to log raw payload and event data rather than
+            // our fancy wrapper objects. Calling payload.toJSON() is
+            // insufficient to get all the nested raw data, because it's
+            // not recursive (it doesn't call toJSON() on the event objects).
+            // Parsing the result of JSON.stringify(), however, gives the
+            // fully recursive raw data.
             logger.log(
               `Sending network request${responseHandlingMessage}:`,
-              payload.toJSON()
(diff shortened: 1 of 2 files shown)


src/core/network/createNetwork.js
```diff
@@ -72,9 +72,17 @@ export default (config, logger, lifecycle, networkStrategy) => {
             const responseHandlingMessage = expectsResponse
               ? ""
               : " (response will be ignored)";
+            const stringifiedPayload = JSON.stringify(payload);
```
jfkhoury: Should we pass `stringifiedPayload` to the `networkStrategy` on line 90 ?

Aaronius: oof. yes!

Aaronius: Done!

PR #73, opened 2019-06-24
CORE-31305 break apart network's newRequest into two methods

jonsnyder:
## Description

The newRequest method of the network gateway proved to be confusing and limiting.  Having to announce whether the request expects a response before creating the payload is one example.  This PR splits the logic into methods `createPayload` and `sendRequest`.

## Related Issue

https://jira.corp.adobe.com/browse/CORE-31304

## Types of changes

## Checklist:

- [x] I have added/removed/modified tests to cover my changes and all tests pass.

### DIFF
--- a/src/core/network/createNetwork.js
+++ b/src/core/network/createNetwork.js
@@ -36,71 +36,65 @@ export default (config, logger, lifecycle, networkStrategy) => {
 
   return {
     /**
-     * The object returned from network.newRequest
-     *
-     * @typedef {Object} Request
-     * @property {Function} send Call this function when you are ready to send
-     * the payload. Returns a promise yielding the raw response body.
-     * @property {Object} payload Payload object that will be sent.
+     * Create a new payload.  Once you have added data to the payload, send it with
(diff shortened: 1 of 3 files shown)


src/core/network/createNetwork.js
```diff
@@ -36,71 +36,65 @@ export default (config, logger, lifecycle, networkStrategy) => {
...
-     * @typedef {Object} Request
-     * @property {Function} send Call this function when you are ready to send
-     * the payload. Returns a promise yielding the raw response body.
-     * @property {Object} payload Payload object that will be sent.
+     * Create a new payload.  Once you have added data to the payload, send it with
+     * the sendRequest method.
      */
+    createPayload() {
+      return createPayload();
+    },
```
Aaronius: I think you could just do:

```
createPayload,
```

instead of:

```
createPayload() {
  return createPayload();
},
```

jonsnyder: Indeed! Old Java habits die hard

Issue #69, opened 2019-06-19
When sending multilple events in succession, alloy only seems to send the first.


stewartschillingsdi:
### Expected Behaviour
All events sent via `alloy("event",{...})` get sent. 

### Actual Behaviour
Only first event is sent - OR - maybe only the 1st is logged? 

### Reproduce Scenario (including but not limited to)
After configuration call, make several event calls.  Watch logs. 

#### Platform and Version
Alloy alpha.3 - MacOS Chrome 
#### Logs taken while reproducing problem

VM3306:8 alloy configure success undefined
VM3306:27 In edgeAlphaEventBridge(alloy). Sending : Product Viewed {type: "Product Viewed", data: {…}}
VM3306:27 In edgeAlphaEventBridge(alloy). Sending : User Detected {type: "User Detected", data: {…}}
VM3306:27 In edgeAlphaEventBridge(alloy). Sending : Page Loaded {type: "Page Loaded", data: {…}}
alloy.js:2258 [alloy] Sending network request (response will be ignored): {events: Array(1), meta: {…}}
VM3306:32

jfkhoury:
@stewartschillingsdi 

Can you please attach the code snippet?

I tried reproducing the same behavior using the Alpha 3 build, but I'm seeing all the events in the console:

```
[alloy] Runtime configuration:
 {
  "propertyID": 9999999,
  "log": true,
  "edgeDomain": "edgegateway.azurewebsites.net",
  "shouldStoreCollectedData": 1,
  "device": "Chrome-Mac",
  "idSyncsEnabled": true,
  "destinationsEnabled": true,
  "context": [
    "web",
    "device",
    "environment",
    "placeContext"
  ]
}

alloy.js:2258 [alloy] Sending network request: {events: Array(1), meta: {…}, identityMap: {…}}
alloy.js:2258 [alloy] Sending network request (response will be ignored): {events: Array(1), meta: {…}, identityMap: {…}}
alloy.js:2258 [alloy] Sending network request (response will be ignored): {events: Array(1), meta: {…}, identityMap: {…}}

alloy.js:2258 [alloy] Received network response: {handle: Array(3), requestId: "764207dc-f225-49a1-aee4-38aab3694951"}
```

Here's my code:

```javascript
      alloy("configure", {
        propertyID: 9999999,
        log: true
      });

      alloy("event", {
        // NOTE: `viewStart` is a special type of events.
        // View docs for more info.
        type: "viewStart",
        data: {
[...]

PR #74, opened 2019-06-24
CORE-31430 Fix issue where identity would only be requested on viewStart

Aaronius:
## Description
This solves the problem brought up in https://github.com/adobe/alloy/issues/69. That is, if the ECID is not available on the client and the website executes `event` commands that do NOT have `type: viewStart`, then the first request would be sent but no ECID requested. The response therefore did not have an ECID and subsequent requests would never be sent (they were stuck in the queue waiting for an ECID). This PR fixes the problem by requesting the ECID on the first event regardless of whether it is `type: viewStart`.

This PR also gives components greater control over whether a response is expected by exposing an `expectResponse` method on the event rather than basing the response expectation on the existence of a `query` object. We were going to need this anyway when gateway becomes less reliant on the `query` object for sending data back from the server.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-31430
https://jira.corp.adobe.com/browse/CORE-31429

## Types of changes

## Checklist:

- [X] (all tests pass, but we need functional tests) I have added/removed/modified tests to cover my changes and all tests pass.

### DIFF
--- a/src/components/Identity/index.js
+++ b/src/components/Identity/index.js
@@ -58,7 +58,9 @@ const createIdentity = ({ config, logger, cookie }) => {
       // We don't have an ECID, but the first request has gone out to
       // fetch it. We must wait for the response to come back with the
       // ECID before we can apply it to this payload.
+      logger.log("Delaying request while retrieving ECID from server.");
       promise = deferredForEcid.promise.then(() => {
+        logger.log("Resuming previously delayed request.");
         addIdsContext(payload, ecid);
       });
     } else {
--- a/src/components/Identity/index.js
+++ b/src/components/Identity/index.js
@@ -26,18 +26,19 @@ const createIdentity = ({ config, logger, cookie }) => {
   const getEcid = () => cookie.get(ECID_NAMESPACE);
 
   let ecid = getEcid();
+  let responseRequested = false;
   let deferredForEcid;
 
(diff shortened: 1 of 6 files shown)


jonsnyder:
It may be a good idea to log a debug message when the identity component is holding up a request.  I could see there being other reasons why this might happen.

src/components/Identity/index.js
```diff
@@ -26,18 +26,19 @@ const createIdentity = ({ config, logger, cookie }) => {
...
-      return;
+  const onBeforeEvent = event => {
+    if (!ecid && !responseRequested) {
+      // TODO: Remove; We won't need to request id syncs explicitely.
+      // This is just for demo currently.
+      event.mergeQuery({
+        idSyncs: true
+      });
+      event.expectResponse();
+      responseRequested = true;
```
jonsnyder: Long term, we need to figure out a way for the identity component to know when the id request fails or times out.

Aaronius:
Yeah, logging a message would be good. I'll do that right now.

src/components/Identity/index.js
Aaronius: Yes...and what we want to do about it when it does fail or time out...

jfkhoury: We have this ticket that might cover that use case: https://jira.corp.adobe.com/browse/CORE-28205

PR #78, opened 2019-06-25
Adding prehidingSelector to configure along with validator

artur-ciocanu:
Start of flicker management functionality

## Description

Adding `prehidingSelector` configuration to manage body hiding to avoid flicker.

## Motivation and Context

To avoid flicker personalization component needs a way to prehide some containers on the page. The CSS selector for containers can be added to `prehidingSelector` and `personalization` will take care of that, selectors should be specified using comma separator like a normal CSS rule.

## Types of changes

## Checklist:


src/utils/config-validators/validPrehidingSelector.js
```diff
@@ -0,0 +1,26 @@
...
+  if (isNil(currentValue)) {
+    return "";
+  }
+
+  if (isNonEmptyString(currentValue)) {
+    return "";
+  }
+
+  return `Invalid value for ${key}: ${currentValue}`;
+};
```
Aaronius: Can/Should this all be combined?

```
return isNil(currentValue) || isNonEmptyString(currentValue) ? "" : `Invalid value for ${key}: ${currentValue}`;
```

Aaronius: Would it make sense to rename this validator to be more generic? It could be used for any config option that should either be nil or a non-empty string.

artur-ciocanu: @Aaronius do you think something like `validEitherNilOrNonEmpty`?

artur-ciocanu: @Aaronius regarding combining, for me it easy to read a couple of `ifs` vs ternary operator.

Aaronius: If I had my druthers, I would make the usage be like this: `string().minLength(1)`

The `string()` part would validate that it's a string and the `minLength()` part would validate that it meets the minimum length (in your case, you would pass `1`).

PR #79, opened 2019-06-25
Tie request response

Aaronius:
## Description
This logs a request ID when when we log requests and responses so that users can logically tie them together.

I played around with color coding the log messages instead, but there were a few reasons I didn't move forward with it:

* I was concerned users would try to attribute meaning to the colors (e.g., does green mean success and orange mean warning?)
* We can't just pick random colors. We have to pick distinguishable colors and colors with proper contrast, for example. Doing this programatically is a bit problematic and introduces a bunch of code if we want to get close to doing it right. The alternative is to pick a static list of colors. At that point we have the cost of the list of colors and we have to rotate through them. I stole colors the design organization had already defined within the CloudViz charting libraries, but even then we only have 10 colors that we'd have to rotate through for each request.
* Supporting styling of log messages would require us to do some browser detection for browsers that don't support it.
* Supporting styling of logs messages would require some changes to our logger that wouldn't be super clean, because we have to work with the fact that only the first argument passed into `console.log` and friends can be styled (browser rules, not mine). The first argument passed into `console.log` right now is the SDK instance name, which isn't what we're trying to color.
* I'm not sure how accessible it would be for the visually impaired.
* At what point do we say, "Let's let the Experience Cloud Debugger espose this data and handle making things visually nice."? We may actually be past that point already...
* Owen agreed with my assessment

Anyway, I ended up going with the request ID approach. I started working on sending the request ID to the gateway, but [there are problems with that](https://git.corp.adobe.com/Activation/edge-gateway-mock/pull/14#issuecomment-1828165) so right now the ID is only used for this logging. It did require us to add a uuid package which adds about 1 KB to the minified file size. I also removed IE 10 support because Justin said we could and because cryptography APIs don't exist in IE 10, which would cause the UUID package to rely on `Math.random()` for randomness, which is sketchy. 

I should also add that the UUID package bundles logic for IE 10 (and below?), which does get built into our library and is unnecessary for us. It basically amounts to this amount of extra code:

```
else {
	    // Math.random()-based (RNG)
	    //
	    // If all else fails, use Math.random().  It's fast, but is of unspecified
	    // quality.
	    var rnds = new Array(16);

	    module.exports = function mathRNG() {
	      for (var i = 0, r; i < 16; i++) {
	        if ((i & 0x03) === 0) r = Math.random() * 0x100000000;
	        rnds[i] = r >>> ((i & 0x03) << 3) & 0xff;
	      }

	      return rnds;
	    };
	  }
```

I think that's probably an acceptable cost when compared to having to maintain a copy of the uuid library. I've considered giving something like https://www.npmjs.com/package/patch-package a whirl, which may allow us to use a patched/trimmed version without having to maintain a full copy of the npm package. Anyway.....

## Related Issue
https://jira.corp.adobe.com/browse/CORE-31390

## Motivation and Context
Users wanted to logically tie together request and response log messages.

image

## Types of changes

## Checklist:

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -21,21 +21,26 @@ export default {
     {
       file: "dist/alloy.js",
       format: "umd",
-      // Allow non-IE browsers and IE10 and IE11
+      // Allow non-IE browsers and IE11
       // document.documentMode was added in IE8, and is specific to IE.
       // IE7 and lower are not ES5 compatible so will get a parse error loading the library.
       intro:
-        "if (document.documentMode && document.documentMode < 10) { console.warn('The Adobe Experience Cloud Web SDK does not support IE 9 and below.'); return; }"
+        "if (document.documentMode && document.documentMode < 11) { console.warn('The Adobe Experience Cloud Web SDK does not support IE 10 and below.'); return; }"
     },
     {
       file: "sandbox/public/alloy.js",
       format: "umd",
       intro:
-        "if (document.documentMode && document.documentMode < 10) { console.warn('The Adobe Experience Cloud Web SDK does not support IE 9 and below.'); return; }"
+        "if (document.documentMode && document.documentMode < 11) { console.warn('The Adobe Experience Cloud Web SDK does not support IE 10 and below.'); return; }"
     }
   ],
   plugins: [
-    resolve({ preferBuiltins: false }),
+    resolve({
+      preferBuiltins: false,
+      // Support the browser field in dependencies' package.json.
+      // Useful for the uuid package.
+      mainFields: ["module", "main", "browser"]
+    }),
     commonjs(),
     babel(),
     replaceVersion()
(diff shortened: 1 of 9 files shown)


Aaronius:
Looks like the build is failing. 😢  I'll look at it tomorrow.

rollup.config.js
```diff
@@ -21,21 +21,26 @@ export default {
     {
       file: "dist/alloy.js",
       format: "umd",
-      // Allow non-IE browsers and IE10 and IE11
+      // Allow non-IE browsers and IE11
       // document.documentMode was added in IE8, and is specific to IE.
       // IE7 and lower are not ES5 compatible so will get a parse error loading the library.
       intro:
-        "if (document.documentMode && document.documentMode < 10) { console.warn('The Adobe Experience Cloud Web SDK does not support IE 9 and below.'); return; }"
+        "if (document.documentMode && document.documentMode < 11) { console.warn('The Adobe Experience Cloud Web SDK does not support IE 10 and below.'); return; }"
```
jfkhoury: 🎉

Aaronius:
It looks like the build failure was due to the sonar-scanner command not being found for some reason. Anyway, I restarted the build today and it passed. I also merged master into this branch because there was a conflict. Merging now.

PR #80, opened 2019-06-26
CORE-31304 move onBeforeSend into DataCollector as onBeforeDataCollection

jonsnyder:
## Description

Move onBeforeSend into DataCollector and rename it to onBeforeDataCollection.

## Related Issue

https://jira.corp.adobe.com/browse/CORE-31304

## Motivation and Context

This gives the data collector more control over when the request is actually sent out.

## Types of changes

## Checklist:

### DIFF
--- a/src/components/DataCollector/index.js
+++ b/src/components/DataCollector/index.js
@@ -27,19 +27,18 @@ const createDataCollector = () => {
         return lifecycle.onBeforeDataCollection(payload, responsePromise);
       })
       .then(() => {
-        return network
-          .sendRequest(payload, payload.expectsResponse)
-          .then(response => {
-            const data = {
-              requestBody: clone(payload)
-            };
+        return network.sendRequest(payload, payload.expectsResponse);
+      })
+      .then(response => {
+        const data = {
+          requestBody: clone(payload)
+        };
 
-            if (response) {
-              data.responseBody = clone(response);
-            }
+        if (response) {
+          data.responseBody = clone(response);
+        }
 
-            return data;
-          });
+        return data;
       });
     return responsePromise;
   };
--- a/src/components/DataCollector/index.js
+++ b/src/components/DataCollector/index.js
@@ -20,20 +20,28 @@ const createDataCollector = () => {
   let network;
 
   const makeServerCall = event => {
-    const { expectsResponse } = event;
     const payload = network.createPayload();
     payload.addEvent(event);
-    return network.sendRequest(payload, expectsResponse).then(response => {
-      const data = {
-        requestBody: clone(payload)
-      };
+    const responsePromise = Promise.resolve()
+      .then(() => {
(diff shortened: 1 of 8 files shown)


jonsnyder:
Tests are passing now in IE11

src/components/DataCollector/index.js
```diff
@@ -20,20 +20,28 @@ const createDataCollector = () => {
...
-      }
+            if (response) {
+              data.responseBody = clone(response);
+            }
 
-      return data;
-    });
+            return data;
+          });
+      });
```
Aaronius: I recommend keeping promise chaining as shallow when possible rather than nesting chains. For example:

```
    const responsePromise = Promise.resolve()
      .then(() => {
        return lifecycle.onBeforeDataCollection(payload, responsePromise);
      })
      .then(() => {
        return network.sendRequest(payload, payload.expectsResponse);
      })
      .then(response => {
          const data = {
            requestBody: clone(payload)
          };

           if (response) {
            data.responseBody = clone(response);
          }

          return data;
      });
```

src/core/network/createPayload.js
```diff
@@ -30,6 +30,12 @@ export default () => {
       content.events.push(event);
     },
     mergeMeta: createMerger(content, "meta"),
+    get expectsResponse() {
+      return (
+        isNonEmptyArray(content.events) &&
+        find(content.events, event => event.expectsResponse) !== undefined
```
Aaronius: It might be worth implementing a `some` utility that basically does this and matches the [Array.some](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) method. Whatever you'd like to do is fine with me though.

test/unit/components/DataCollector/index.spec.js
```diff
@@ -0,0 +1,170 @@
...
+    });
+  });
+  it("Calls onBeforeEvent with a matching event", () => {
+    eventCommand({ data: { a: 1 }, meta: { b: 2 } }).then(() => {
+      expect(lifecycle.onBeforeEvent.calls.argsFor(0)[0].toJSON().data).toEqual(
+        { a: 1 }
+      );
+      expect(lifecycle.onBeforeEvent.calls.argsFor(0)[0].toJSON().meta).toEqual(
+        { b: 2 }
+      );
```
Aaronius: It might be better to do:

```
expect(lifecycle.onBeforeEvent).toHaveBeenCalledWith({
  data: {
    a: 1
  },
  meta: {
    b: 2
  }
});
```

src/core/network/createPayload.js
Aaronius: Actually, Array.some is supported in all our target browsers. You can just use it.

src/components/DataCollector/index.js
jonsnyder: Yes, that will make it less confusing.

PR #81, opened 2019-06-26
Tnt 32590

artur-ciocanu:
Adding personalization component support for page/containers prehiding

## Description

Personalization component modifies the DOM nodes, so we need a way to manage flicker. The solution was to use `prehidingStyle` config option to be able customize the prehiding behavior. Since our current sandbox uses async loading I have added a prehiding snippet, so we at least have an example how that might work.

Personalization component had to be adjust such that for `personalization:page` handle we have to check all the `elementExists` events and prehide the elements using `prehidingSelector`, this is again necessary to avoid flicker. In the nutshell the workflow is like this:

1. Hide the page/containers
2. Fire a request
3. `onResponse` get the `personalization:page` fragments
4. Prehide all the events that have `moduleType` === `elementExists`
5. Show the page/containers
6. Execute the rules

## Related Issue

None

## Motivation and Context

Flicker management

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Personalization/flicker/index.js
+++ b/src/components/Personalization/flicker/index.js
@@ -10,8 +10,15 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import { createNode, appendNode, removeNode } from "../../../utils/dom";
+import {
+  createNode,
+  appendNode,
+  removeNode,
+  findById
+} from "../../../utils/dom";
 
+const CONTAINERS_HIDING_ID = "alloy-prehiding";
+const HIDING_STYLE_DEFINITION = "{ visibility: hidden }";
 const STYLE_TAG = "style";
 const CACHE = {};
 
@@ -23,8 +30,9 @@ export const hideElements = prehidingSelector => {
     return;
   }
 
-  const node = createNode(STYLE_TAG);
-  node.innerText = `${prehidingSelector} { visibility: hidden }`;
+  const node = createNode(STYLE_TAG, {
+    text: `${prehidingSelector} ${HIDING_STYLE_DEFINITION}`
+  });
 
   appendNode(document.head, node);
 
@@ -37,4 +45,38 @@ export const showElements = prehidingSelector => {
   if (node) {
     removeNode(node);
   }
+};
+
+export const hideContainers = prehidingStyle => {
+  if (!prehidingStyle) {
+    return;
+  }
+
+  // If containers prehiding style has been added
+  // by customer's prehiding snippet we don't
+  // want to add the same node
+  const node = findById(CONTAINERS_HIDING_ID);
+
+  if (node) {
+    return;
+  }
+
+  const styleNode = createNode(STYLE_TAG, {
+    id: CONTAINERS_HIDING_ID,
+    text: prehidingStyle
+  });
+
+  appendNode(document.head, styleNode);
+};
+
+export const showContainers = () => {
+  // If containers prehiding style exists
+  // we will remove it
+  const node = findById(CONTAINERS_HIDING_ID);
+
+  if (!node) {
+    return;
+  }
+
+  removeNode(node);
 };
--- a/src/components/Personalization/flicker/index.js
+++ b/src/components/Personalization/flicker/index.js
@@ -17,37 +17,39 @@ import {
   findById
 } from "../../../utils/dom";
 
-const CONTAINERS_HIDING_ID = "alloy-prehiding";
 const HIDING_STYLE_DEFINITION = "{ visibility: hidden }";
 const STYLE_TAG = "style";
-const CACHE = {};
+const styleNodes = {};
 
 export const hideElements = prehidingSelector => {
   // if we have different events with the same
   // prehiding selector we don't want to recreate
   // the style tag
-  if (CACHE[prehidingSelector]) {
+  if (styleNodes[prehidingSelector]) {
     return;
   }
 
-  const node = createNode(STYLE_TAG, {
-    text: `${prehidingSelector} ${HIDING_STYLE_DEFINITION}`
-  });
+  const attrs = {};
+  const props = {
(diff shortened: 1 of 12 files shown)


artur-ciocanu:
@jonsnyder or @Aaronius could you please have a look! Thank you.

src/components/Personalization/flicker/index.js
```diff
@@ -10,8 +10,15 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
 
-import { createNode, appendNode, removeNode } from "../../../utils/dom";
+import {
+  createNode,
+  appendNode,
+  removeNode,
+  findById
+} from "../../../utils/dom";
 
+const CONTAINERS_HIDING_ID = "alloy-prehiding";
```
jonsnyder: Shouldn't this be configurable? With two alloy instances, you may want to have two different hiding ids.

src/components/Personalization/flicker/index.js
```diff
@@ -38,3 +46,37 @@ export const showElements = prehidingSelector => {
     removeNode(node);
   }
```
jonsnyder: Should you delete the element from the CACHE too?  If you don't do this, calling hideElements with the same selector will not work.

src/components/Personalization/flicker/index.js
```diff
@@ -10,8 +10,15 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+  createNode,
+  appendNode,
+  removeNode,
+  findById
+} from "../../../utils/dom";
 
+const CONTAINERS_HIDING_ID = "alloy-prehiding";
+const HIDING_STYLE_DEFINITION = "{ visibility: hidden }";
 const STYLE_TAG = "style";
 const CACHE = {};
```
jonsnyder: This global state scared me at first.  From a functionality perspective I understand why it is necessary even with multiple instances on a page; namely, the DOM is global so the cache should be global too.  Maybe a comment here could explain why the global state is necessary.

Additionally, I think renaming it as `styleNodes` would make it more clear as to what it is doing.  My understanding is that variables named in all caps should stay constant.

src/components/Personalization/index.js
```diff
@@ -12,43 +12,83 @@ governing permissions and limitations under the License.
...
         event.expectResponse();
       },
       onResponse(response) {
         const fragments = response.getPayloadByType(PAGE_HANDLE) || [];
 
-        fragments.forEach(fragment => {
-          const { rules = [] } = fragment;
+        // On response we first hide all the elements for
+        // personalization:page handle
+        hideElementsForPage(fragments);
```
jfkhoury: Do we need to think about what happens if the request fails, or times out?

If so, do you think we need an `onError` lifecycle hooks so you can decide what to do in case of an error?

src/utils/dom/createNode.js
```diff
@@ -23,12 +24,18 @@ export default function createNode(
   Object.keys(attrs).forEach(key => {
     const value = attrs[key];
 
-    if (typeof value === "function") {
+    if (key === "text") {
+      result.innerText = value;
```
Aaronius: I feel like this is a bit misleading, since `text` comes through as an attribute but sets the `innerText` property. I would rather see `createNode` accept both `attributes` and `properties` options and then the consuming module would pass `{ innerText: "some value" }` for the `properties` option. I would expect any values that are functions to likewise be passed on the `properties` option instead of having the `isFunction` condition below.

src/components/Personalization/flicker/index.js
artur-ciocanu: good catch, added `delete CACHE[prehidingSelector]`

artur-ciocanu:
> Forgive me for commenting on things you haven't changed. It's my first time looking at this code.
> 
> I like how you put the DOM manipulation functions as their own utils. This will be helpful when testing the personalization component logic.
> 
> What is the difference between a container and an element in this code? (hideContainer vs hideElement)

Before applying the DOM changes which are very specific, we need to prehide the page/containers where those DOM changes will be applied. To make sure these are not confused I have added different functions.

src/components/Personalization/flicker/index.js
artur-ciocanu: Fixed

src/components/Personalization/index.js
artur-ciocanu: @jfkhoury good call, do we have an `onResponseError` hook? That would help to ensure we un-hide the containers in case of network request failure.

jfkhoury: @artur-ciocanu, we currently have this TODO in the lifecycle module: `// TODO: We might need an `onError(error)` hook.`. We were waiting for the first use case before we implement it; and this seems like the perfect candidate ;)

jfkhoury:
@artur-ciocanu, I'm gonna create a Documentation ticket and assign it to you, so we can start documenting this stuff, especially the hiding bit.

src/components/Personalization/index.js
jonsnyder: You can use the onBeforeDataCollection event for this.
```suggestion
const onBeforeDataCollection = (payload, promise) => {
  promise.catch(() => {
    ...
  });
};
```

But it's not easy to know which request it is for.

src/utils/dom/createNode.js
artur-ciocanu: I have added `props` to `createNode()` and removed the `isFunction` altogether. The `fireImage` module has been adjusted to pass in `onload`, `onerror` and `onabort` as `props` instead of `attrs`.

src/components/Personalization/flicker/index.js
artur-ciocanu: Added `prehidingId` so we can have multiple instances with their own configurations.

src/components/Personalization/index.js
artur-ciocanu: @jfkhoury and @Aaronius as discussed I'll add a separate `onResponseError` so all this cleanup can be handled properly.

PR #84, opened 2019-06-28
CORE-28205 Retry failed requests.

Aaronius:
## Description
When a request fails, retry it up to a maximum of 3 retries (4 requests total).

We do have a case to consider where request 1 goes out, request 2 goes out, request 1 fails, request 2 succeeds, then the request 1 _retry_ goes out and succeeds. Will the server think that event 1 occurred after event 2? Should we be sending a timestamp so the server can sort it out correctly? I've brought this up on the alloy Slack channel if you'd like to join the conversation.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-28205

## Motivation and Context
Trying hard to make requests succeed.

See the video in the alloy Slack channel.

## Types of changes

## Checklist:

### DIFF
--- a/src/core/network/createNetwork.js
+++ b/src/core/network/createNetwork.js
@@ -12,7 +12,7 @@ governing permissions and limitations under the License.
 
 import createPayload from "./createPayload";
 import createResponse from "./createResponse";
-import { uuid } from "../../utils";
+import { executeWithRetry, stackError, uuid } from "../../utils";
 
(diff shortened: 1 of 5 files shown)


jfkhoury:
I discussed this PR with Aaron. 2 things:

1. If the server can resolve events with unordered timestamps, we will retry all failed events.
2. Otherwise, we will only retry events that are supposed to bring back an ECID.

Waiting on Product to respond.

PR #86, opened 2019-07-09
CORE-31388 use XDM keys and values for event types

jonsnyder:
## Description

Using the [xdm spec](https://github.com/adobe/xdm/blob/master/docs/reference/data/time-series.schema.md) for event type designations.  Key off of this for determining if the page is a viewStart or not.

## Related Issue

[CORE-31388](https://jira.corp.adobe.com/browse/CORE-31388)

## Motivation and Context

The event type was not being passed on and only used internally.  This data should be passed to Konductor so that it can be reported on.

## Types of changes

## Checklist:

### DIFF
--- a/sandbox/src/Home.js
+++ b/sandbox/src/Home.js
@@ -8,7 +8,7 @@ function HomeWithHistory({ history }) {
     if (loc.pathname !== previousPath) {
       const instanceName = loc.pathname.includes("orgTwo") ? "organizationTwo" : "alloy";
       window[instanceName]("event", {
-        type: "viewStart",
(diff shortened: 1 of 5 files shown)


jfkhoury:
@jonsnyder Since we decided to remove the `type: viewStart` property and replace it with `isViewStart: true|false`, we can close this PR now. I will create a ticket for Konductor to account for this new prop.

cc: @Aaronius

jonsnyder:
Sounds good.  We should probably add `eventType: "web.webpagedetails.pageViews",` to the payload for the events in the sandbox.

PR #88, opened 2019-07-09
CORE-31840 Updates to match latest gateway specs.

Aaronius:
## Description
This updates Alloy to match the latest gateway specs. Namely, we no longer have to pass `true` for the query options, as the gateway assumes `true` by default. Also, the gateway endpoint paths are versioned now.

This change is NOT compatible with the current mock gateway, but it IS compatible with this mock gateway PR: https://git.corp.adobe.com/Activation/edge-gateway-mock/pull/16 As such, we should merge the mock gateway PR at the same time-ish as this PR.

This change is NOT compatible with the real gateway, only because the real gateway does not support the plain text MIME type. It only supports the JSON MIME type, which forces pre-flight requests. Therefore, gateway will need to fix the MIME type issue before we can start using Alloy with it.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-31840

## Motivation and Context
Allow alpha users to test the real gateway.

## Types of changes

## Checklist:

- [x] Against the PR - I have run the Sandbox successfully.

### DIFF
--- a/sandbox/public/index.html
+++ b/sandbox/public/index.html
@@ -10,8 +10,8 @@
           content="default-src 'self';
             script-src 'self' 'unsafe-inline';
             style-src 'self' 'unsafe-inline';
-            img-src 'self' http://edgegateway.azurewebsites.net/dests https://edgegateway.azurewebsites.net/dests;
-            connect-src 'self' https://edgegateway.azurewebsites.net http://ex-edge.stable-stage.aam-npe.adobeinternal.net http://localhost:8080"/>
+            img-src *;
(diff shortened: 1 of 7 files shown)


sandbox/public/index.html
```diff
@@ -10,8 +10,8 @@
           content="default-src 'self';
             script-src 'self' 'unsafe-inline';
             style-src 'self' 'unsafe-inline';
-            img-src 'self' http://edgegateway.azurewebsites.net/dests https://edgegateway.azurewebsites.net/dests;
-            connect-src 'self' https://edgegateway.azurewebsites.net http://ex-edge.stable-stage.aam-npe.adobeinternal.net http://localhost:8080"/>
+            img-src *;
```
Aaronius: The property configured in the konductor has destinations that load images from a bunch of different URLs. For this reason, I've changed img-src to a wildcard.

Aaronius:
🤷‍♂ I will make it so, though.

PR #95, opened 2019-07-16
Conditional logging for expensive functions

owenleej:
## Description

a couple calls to logger.log contain expensive calls to JSON.parse or JSON.toString etc. I exposed whether or not logging is enabled within the logger object and wrapped expensive calls to logger.log in a conditional statement.

## Related Issue

CORE-31336

## Motivation and Context

This makes it so logging doesn't cause a performance hit when it is disabled.

## Types of changes

## Checklist:

### DIFF
--- a/src/core/createLogger.js
+++ b/src/core/createLogger.js
@@ -18,6 +18,7 @@ const createLogger = (window, logController, prefix) => {
   };
 
   return {
+    enabled: logController.logEnabled,
     /**
      * Outputs a message to the web console.
      * @param {...*} arg Any argument to be logged.
--- a/src/core/createLogger.js
+++ b/src/core/createLogger.js
@@ -18,7 +18,9 @@ const createLogger = (window, logController, prefix) => {
   };
 
   return {
-    enabled: logController.logEnabled,
+    get enabled() {
+      return logController.logEnabled;
+    },
     /**
      * Outputs a message to the web console.
      * @param {...*} arg Any argument to be logged.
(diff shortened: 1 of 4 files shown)


src/core/createLogger.js
```diff
@@ -18,6 +18,7 @@ const createLogger = (window, logController, prefix) => {
   };
 
   return {
+    enabled: logController.logEnabled,
```
Aaronius: When `logController.logEnabled` changes, `logger.enabled` won't change. To make `logger.enabled` reflect the latest value, you'll need to change this to:

```
get enabled() {
  return logController.logEnabled;
},
```

I recommend adding a test for this scenario inside `createLogger.spec.js`.

test/unit/core/network/createNetwork.spec.js
```diff
@@ -19,6 +19,7 @@ describe("createNetwork", () => {
   };
 
   const logger = console;
+  logger.enabled = true;
```
Aaronius: Because `logger` is literally `window.console`, this is setting `window.console.enabled` to `true`, which could end up tainting other tests since `window.console` sticks around for other tests. I think at this point we need to make logger not be `window.console`. Instead, let's declare the logger here:

```
let logger;
```

then inside `beforeEach` add the following:

```
logger = jasmine.createSpyObj(logger, ["log"]);
logger.enabled = true;
```

then in the tests we can remove the lines that say:

```
spyOn(logger, "log");
```

because `logger.log` will already be a spy.

owenleej: See new commit for changes, thanks for the recommendations.

PR #96, opened 2019-07-17
CORE-32120 Added opt-in support.

Aaronius:
## Description

There's some verbiage that's a little weird here. The user can opt into all purposes or no purposes (later, we plan to support opting into specific purposes). Another way you could say that the user is opted into no purposes is that they've "opted out", but using the term "opted out" can be confusing because "opt out" is a separate effort/task/flag that is separate from opt-in. So, yeah.

Opt-in can currently be in one of three statuses:

* The user's opt-in preferences are pending (we're waiting for the user to tell us what they're okay with).
* The user has opted into all purposes.
* The user has opted into no purposes.

Opt-in can be enabled by settings `optInEnabled: true` through the `configure` command. If opt-in is not enabled, then we assume that the user has opted into all purposes.

Our customers can set whether the user has opted into all purposes or no purposes as follows:

```
alloy("optIn", {
  purposes: "all"
});
```

or

```
alloy("optIn", {
  purposes: "none"
});
```

If the `optIn` command has not been called for the user, then whenever Alloy needs to do something that depends on the user having opted in (e.g., send data to the server), it will wait for the `optIn` command to be called before proceeding. If `optIn` is called with `purposes: "all"`, then Alloy will proceed doing to the work that was previously queued. If `optIn` is called with `purposes: "none"`, then Alloy will not proceed with the work that was previously queued and the promises returned from the prior commands will be rejected.

Once the `optIn` command has been called, the user's preferences will be stored in a cookie and used in perpetuity until the cookie expires or the `optIn` command is called again to change the user's preferences.

As far as architecture goes, I've added an `optIn` object that gets passed to all components. The components can use the `whenOptedIn` and `isOptedIn` methods on the `optIn` object to decision based on whether the user has opted into all purposes. See the documentation above each of the methods for more information and how they might change when we start supporting opting into individual purposes.

I've added a new component called Privacy. It is in charge of setting up the optIn object that handed to all the components. For example, it's in charge of telling the optIn object whether opt-in is enabled. It's also in charge of providing the optIn object with a cookie object that manages the privacy namespace in the alloy cookie. Because this setup needs to happen before components start calling `whenOptedIn` and `isOptedIn`, a function called `enableOptIn` is first passed to all component instances when they are created. The privacy component calls this `enableOptIn` function right away and should be the only component to call that function. Then, all components will receive the configured optIn object during `onComponentsRegistered` and therefore can start calling `whenOptedIn` and `isOptedIn`.

I think that covers the general design.

I did not add tests at the component level. There's a discussion that needs to happen as a team regarding (1) how we flush promises during tests and (2) what we want to unit test vs functional test. I'd like to see where that conversation goes before implementing such tests.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-32120

## Motivation and Context
Users like privacy.

## Types of changes

- [] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

### DIFF
--- a/src/components/Privacy/index.js
+++ b/src/components/Privacy/index.js
@@ -1,0 +1,71 @@
+/*
+Copyright 2019 Adobe. All rights reserved.
+This file is licensed to you under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License. You may obtain a copy
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+import { boolean } from "../../utils/config-validators";
+import { isString } from "../../utils";
+
+const throwInvalidPurposesError = purposes => {
+  throw new Error(
+    `Opt-in purposes must be "all" or "none". Received: ${purposes}`
+  );
+};
+
+const createPrivacy = ({ config, logger, enableOptIn, cookie }) => {
+  let optIn;
+
+  if (config.optInEnabled) {
+    enableOptIn(cookie);
+  }
+
+  return {
+    lifecycle: {
+      onComponentsRegistered(tools) {
+        ({ optIn } = tools);
+      }
+    },
+    commands: {
+      optIn({ purposes }) {
+        if (config.optInEnabled) {
+          if (isString(purposes)) {
+            const lowerCasePurposes = purposes.toLowerCase();
+
+            if (
+              lowerCasePurposes === optIn.ALL ||
+              lowerCasePurposes === optIn.NONE
+            ) {
+              optIn.setPurposes(purposes);
+            } else {
+              throwInvalidPurposesError(purposes);
+            }
+          } else {
+            throwInvalidPurposesError(purposes);
+          }
+        } else {
+          logger.warn(
+            "optInEnabled must be set to true before using the optIn command."
+          );
+        }
+      }
+    }
+  };
+};
+
+createPrivacy.namespace = "Privacy";
+
+createPrivacy.configValidators = {
+  optInEnabled: {
+    defaultValue: false,
+    validate: boolean
+  }
+};
+
+export default createPrivacy;
--- a/src/core/createCookie.js
+++ b/src/core/createCookie.js
@@ -23,23 +23,57 @@ const safeJSONParse = (object, cookieName) => {
   }
 };
 
+/**
+ * The purpose of this proxy is to cache the cookie so we don't have to
+ * read and deserialize it every time a piece of it is accessed.
+ */
+const createCookieProxy = (propertyID, cookieDomain = "") => {
+  const cookieName = `${ALLOY_COOKIE_NAME}_${propertyID}`;
+
+  let deserializedCookie;
+  let cookieHasBeenRead = false;
+
+  return {
+    get() {
+      // We don't read the cookie right when the cookie proxy is created
+      // because we don't know if the user has opted in. If the user
+      // hasn't opted in, we're legally obligated to not read the cookie.
+      // If a component tries to read something off the cookie though,
+      // we assume that the component has received word that the user
+      // has opted-in. The responsibility is on the component.
+      if (!cookieHasBeenRead) {
+        const serializedCookie = cookie.get(cookieName);
+        deserializedCookie =
+          serializedCookie && safeJSONParse(serializedCookie, cookieName);
+        cookieHasBeenRead = true;
+      }
+      return deserializedCookie;
+    },
+    set(updatedCookie) {
+      deserializedCookie = updatedCookie;
+      cookie.set(cookieName, updatedCookie, {
+        expires: ALLOY_COOKIE_TTL,
+        domain: cookieDomain || getTopLevelCookieDomain(window, cookie)
+      });
+    }
+  };
+};
+
 // TODO: Support passing a configurable expiry in the config when creating this cookie.
-const createCookie = (prefix, id, cookieDomain = "") => {
+const createCookie = (componentNamespace, propertyID, cookieDomain = "") => {
+  const cookieProxy = createCookieProxy(propertyID, cookieDomain);
+
   return {
     /**
      * Returns the value from AlloyCookie for a prefix and a key.
      * @param {...*} arg String key stored in alloy cookie under a component prefix.
      */
     get(name) {
-      const cookieName = `${ALLOY_COOKIE_NAME}_${id}`;
-      const currentCookie = cookie.get(cookieName);
-      const currentCookieParsed =
-        currentCookie && safeJSONParse(currentCookie, cookieName);
-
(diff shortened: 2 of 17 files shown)


src/core/createCookie.js
```diff
@@ -23,58 +23,85 @@ const safeJSONParse = (object, cookieName) => {
...
      */
     set(key, value) {
-      const cookieName = `${ALLOY_COOKIE_NAME}_${id}`;
-      const currentCookie = cookie.get(cookieName)
-        ? safeJSONParse(cookie.get(cookieName))
-        : {};
-      const updatedCookie = {
-        ...currentCookie,
-        [prefix]: { ...currentCookie[prefix], [key]: value }
-      };
```
Aaronius: Previously we were cloning the cookie object to prevent mutation of `currentCookie`, but we weren't doing the same thing in `remove()` when deleting and entry from the cookie object. For consistency, we either needed to do it in `remove()` as well, or not do it here. I thought that because any side effects of mutating the object were contained to this module, that it would be reasonable to just mutate the object. I open to disagreement here.

src/core/initializeComponentsFactory.js
```diff
@@ -30,16 +34,15 @@ export default (
...
     // TO-DOCUMENT: Helpers that we inject into factories.
     let component;
     try {
       component = createComponent({
         logger: logger.spawn(`[${namespace}]`),
         cookie: cookie(namespace, propertyID, cookieDomain),
         config,
-        storage
+        storage,
+        enableOptIn: optIn.enable
```
Aaronius: This is kind of a key point. This is how Privacy is able to configure the `optIn` object without any component getting access to the rest of the `optIn` object first.

test/unit/flushPromises.js
```diff
@@ -0,0 +1,16 @@
...
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+/**
+ * Returns a promise that will be resolved after all currently resolved promises.
+ * @returns {Promise}
+ */
+export default () => new Promise(resolve => setTimeout(resolve));
```
Aaronius: If we wanted to be consistent with what's being done in the `test/functional` directory, we would have:

`test/unit/helpers/flushPromises.js`
`test/unit/specs/...` <-- all specs would be moved down to this directory.

Thoughts?

src/core/createCookie.js
```diff
@@ -23,58 +23,85 @@ const safeJSONParse = (object, cookieName) => {
...
+    },
+    set(updatedCookie) {
+      deserializedCookie = updatedCookie;
+      cookie.set(cookieName, updatedCookie, {
+        expires: ALLOY_COOKIE_TTL,
+        domain: cookieDomain || getTopLevelCookieDomain(window, cookie)
+      });
+    }
+  };
+};
```
Aaronius: This is an optimization so that we're not reading+deserializing the cookie every time `cookie.get()` is called, which makes it more amenable for the Identity component to call it every time within its `onBeforeEvent` and `onBeforeDataCollection`.

src/components/Privacy/index.js
```diff
@@ -0,0 +1,71 @@
...
+        }
+      }
+    }
+  };
+};
+
+createPrivacy.namespace = "Privacy";
+
+createPrivacy.configValidators = {
+  optInEnabled: {
```
jonsnyder: This setting name confused me at first.  optInEnabled: true could mean "I want the default to be optedIn".  I wonder if "requireOptIn" would make more sense.  I understand you are trying to follow the logEnabled pattern though.

test/unit/flushPromises.js
jonsnyder: I like it.  I think there are more helpers we could extract.

test/unit/core/createOptIn.spec.js
```diff
@@ -0,0 +1,122 @@
...
+      optIn.whenOptedIn().catch(optedOutSpy);
+
+      expect(cookie.set).toHaveBeenCalledWith("optIn", "none");
+      expect(optIn.isOptedIn()).toBe(false);
+      return flushPromises().then(() => {
+        expect(optedOutSpy).toHaveBeenCalledWith(jasmine.any(Error));
+      });
+    });
+  });
+});
```
jonsnyder: Another test case would be to do a nested whenOptedIn() call.  Although, looking at the code, I think it would pass.

src/core/createCookie.js
```diff
@@ -23,58 +23,85 @@ const safeJSONParse = (object, cookieName) => {
   }
 };
 
+/**
+ * The purpose of this proxy is to cache the cookie so we don't have to
+ * read and deserialize it every time a piece of it is accessed.
+ */
+const createCookieProxy = (propertyID, cookieDomain = "") => {
```
jfkhoury: Nice!

src/core/createOptIn.js
```diff
@@ -0,0 +1,95 @@
...
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+import { defer } from "../utils";
+
+const COOKIE_NAMESPACE = "optIn";
+
+// The user has opted into all behaviors.
```
jfkhoury: into all `purposes` instead of `behaviors`?

src/core/createOptIn.js
```diff
@@ -0,0 +1,95 @@
...
+
+// The user has yet to provide opt-in behaviors.
+const PENDING = "pending";
+
+export default () => {
+  const deferredsAwaitingResolution = [];
+  let cookie;
+  let purposes = ALL;
+
+  const processDeferreds = () => {
```
jfkhoury: Sweet

src/core/createOptIn.js
```diff
@@ -0,0 +1,95 @@
+/*
```
jfkhoury: @Aaronius, I wonder if this module belongs inside the `/Privacy` component?

test/unit/flushPromises.js
Aaronius: Cool, I'll propose this change in a separate PR.

src/core/createOptIn.js
Aaronius: Yes! Thanks for catching that.

src/core/createOptIn.js
Aaronius: Hmm. It would be the first time anything outside the components directory required a module located inside the components directory, aside from the component registration. I think I would prefer to keep it where it is since it more or less keeps component directories as black boxes, but if you feel like it should be in the privacy component directory, I can do it. I don't feel really strongly either way.

test/unit/core/createOptIn.spec.js
Aaronius: I added another test that I think addresses what you're talking about.

Aaronius:
I'm going to merge this. I brought up the `optInEnabled` thing on Slack to see what others think. I'll propose the shift in the test file organization in a separate PR.

test/unit/core/createOptIn.spec.js
jonsnyder: Looks good.  Thanks.

PR #97, opened 2019-07-18
Add the ability to interpret selectors with "eq(...)"

artur-ciocanu:
Add the ability to interpret selectors with "eq(...)"

## Description

If the personalization engine that runs on the Experience Edge is powered by Adobe Target we might get selectors that are not 100% CSS selectors, these might contain `eq(...)` pseudo selectors. This PR add a new DOM related function named `selectNodesWithEq` that should be able to interpret these selectors and find the right DOM nodes.

## Related Issue

None

## Motivation and Context

Allow Adobe Target decisions and personalization content to be used without any issues.

## Types of changes

## Checklist:

### DIFF
--- a/src/utils/dom/selectNodesWithEq.js
+++ b/src/utils/dom/selectNodesWithEq.js
@@ -14,6 +14,7 @@ import isNonEmptyString from "../isNonEmptyString";
 import selectNodes from "./selectNodes";
 
 // ID and CSS classes can not start with digits
+// Pleas check https://www.w3.org/TR/css-syntax-3/#escaping
 const DIGIT_IN_SELECTOR_PATTERN = /(#|\.)(-)?(\d{1})(.*)/g;
 // This is required to remove leading " > " from parsed pieces
 const SIBLING_PATTERN = /^\s*>?\s*/;
--- a/src/utils/dom/selectNodesWithEq.js
+++ b/src/utils/dom/selectNodesWithEq.js
@@ -10,12 +10,12 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
+import escape from "css.escape";
 import isNonEmptyString from "../isNonEmptyString";
 import selectNodes from "./selectNodes";
 
-// ID and CSS classes can not start with digits
-// Pleas check https://www.w3.org/TR/css-syntax-3/#escaping
-const DIGIT_IN_SELECTOR_PATTERN = /(#|\.)(-)?(\d{1})(.*)/g;
+// Trying to match ID or CSS class
+const CSS_IDENTIFIER_PATTERN = /(#|\.)(-?\w+)/g;
 // This is required to remove leading " > " from parsed pieces
 const SIBLING_PATTERN = /^\s*>?\s*/;
 const EQ_START = ":eq(";
@@ -24,20 +24,19 @@ const EQ_PATTERN = /:eq\((\d+)\)/g;
 const cleanUp = str => str.replace(SIBLING_PATTERN, "").trim();
 const isNotEqSelector = str => str.indexOf(EQ_START) === -1;
 
-export const escapeDigitsInSelector = selector => {
-  const replace = value =>
-    value.replace(DIGIT_IN_SELECTOR_PATTERN, `$1$2\\3$3 $4`);
+// Here we use CSS.escape() to make sure we get
+// correct values for ID and CSS class
+// Please check:  https://www.w3.org/TR/css-syntax-3/#escaping
+// CSS.escape() polyfill can be found here: https://github.com/mathiasbynens/CSS.escape
+const replaceIdentifier = (_, $1, $2) => `${$1}${escape($2)}`;
 
-  return selector
-    .split(" ")
-    .filter(isNonEmptyString)
-    .map(replace)
-    .join(" ");
+export const escapeIdentifiersInSelector = selector => {
+  return selector.replace(CSS_IDENTIFIER_PATTERN, replaceIdentifier);
 };
 
 export const parseSelector = rawSelector => {
   const result = [];
-  const selector = escapeDigitsInSelector(rawSelector.trim());
+  const selector = escapeIdentifiersInSelector(rawSelector.trim());
   const parts = selector.split(EQ_PATTERN).filter(isNonEmptyString);
   const { length } = parts;
   let i = 0;
--- a/src/utils/dom/selectNodesWithEq.js
+++ b/src/utils/dom/selectNodesWithEq.js
@@ -1,0 +1,121 @@
+/*
+Copyright 2019 Adobe. All rights reserved.
+This file is licensed to you under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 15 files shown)


src/utils/dom/selectNodesWithEq.js
```diff
@@ -0,0 +1,121 @@
...
+  if (second === "-") {
+    result.val = `${first}${second}\\3${third} `;
+  } else {
+    result.val = `${first}\\3${second} `;
+  }
+
+  return result;
+};
+
+export const escapeDigitsInSelector = selector => {
```
jonsnyder: Why not use replace?  Something like this:
```selector.replace(/(#|\.)(-)?(\d)/g, `$1$2\\$3`);```

File: src/utils/dom/selectNodesWithEq.js [hunk 2]
```diff
@@ -0,0 +1,121 @@
...
+
+  while (i < length) {
+    const { sel, eq } = parts[i];
+    const nodes = selectNodes(sel, context);
+
+    if (i < length - 1) {
+      if (eq == null) {
+        [context] = nodes;
+      } else {
+        context = nodes[eq];
```
jonsnyder: What if eq >= nodes.length?

src/utils/dom/selectNodesWithEq.js
```diff
@@ -0,0 +1,121 @@
...
+  let context = doc;
+  let i = 0;
+
+  while (i < length) {
+    const { sel, eq } = parts[i];
+    const nodes = selectNodes(sel, context);
+
+    if (i < length - 1) {
+      if (eq == null) {
+        [context] = nodes;
```
jonsnyder: What if nodes.length === 0

src/utils/dom/selectNodesWithEq.js
artur-ciocanu: @jonsnyder thanks for suggestions, I have modified the code and added replace with capture groups.

src/utils/dom/selectNodesWithEq.js
artur-ciocanu: Added guarding conditions like:
```
    if (nodesCount === 0) {
      break;
    }
```
And
```
    if (eq != null && eq > nodesCount - 1) {
      break;
    }
```

src/utils/dom/selectNodesWithEq.js
artur-ciocanu: Added guarding condition like:
```
    if (nodesCount === 0) {
      break;
    }
```

test/unit/utils/dom/selectNodesWithEq.spec.js
```diff
@@ -0,0 +1,135 @@
...
+    const result = selectNodesWithEq("#abc:eq(0) > div.b:eq(0) > div.c");
+
+    expect(result[0].tagName).toEqual("DIV");
+    expect(result[0].textContent).toEqual("first");
+    expect(result[1].tagName).toEqual("DIV");
+    expect(result[1].textContent).toEqual("second");
+    expect(result[2].tagName).toEqual("DIV");
+    expect(result[2].textContent).toEqual("third");
+  });
+});
```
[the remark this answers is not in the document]
artur-ciocanu: Added negative tests, that should thrown exception in case the selector is incorrect.

Added a separate tests that shows the difference between `eq()` and `nth-child()`, the same applies to `eq()` to `nth-of-type()`.

artur-ciocanu:
@jonsnyder I have addressed all your comments. Could you please take another look. Thanks a lot!

src/utils/dom/selectNodesWithEq.js
```diff
@@ -24,34 +23,15 @@ const EQ_PATTERN = /:eq\((\d+)\)/g;
...
-
-    return pairs.reduce(
-      (acc, pair) => acc.replace(pair.key, pair.val),
-      selector
-    );
-  }
-
-  return selector;
+  const replace = value =>
+    value.replace(DIGIT_IN_SELECTOR_PATTERN, `$1$2\\3$3 $4`);
```
jonsnyder: Can you add a comment to explain how the css escaping is interpreted? i.e. "\31 " means ascii code point 0x31 = "1", maybe a link to here: https://www.w3.org/TR/css-syntax-3/#escaping

src/utils/dom/selectNodesWithEq.js
```diff
@@ -24,34 +23,15 @@ const EQ_PATTERN = /:eq\((\d+)\)/g;
...
+
+  return selector
+    .split(" ")
```
jonsnyder: Why split on space and then join?

artur-ciocanu: @jonsnyder I've tried without `split(" ")` and the capture groups would replace ONLY the first element, that's I why I decided to split modify and the re-combine.

I am not sure if there is a better way or more elegant way.

artur-ciocanu:
@jonsnyder after looking at this for a while, I am not sure if this is the best approach, since I am just addressing ONLY two issues with digits in identifiers and `eq()` pseudo selector, however generally speaking the identifiers might contain other characters that have to be escaped.

So for now we can leave this as is or we should add a more robust selector parsing, so we would escape not just identifiers that start with digits, but other characters as well.

CC: @jfkhoury @Aaronius

src/utils/dom/selectNodesWithEq.js
jonsnyder: I see.  It's because you have (.*) at the end of the pattern.  Regex's will match the longest string possible.  If you removed that, I think it would work. i.e. ```value.replace(/(#|\.)(-)?(\d{1})/g, `$1$2\\3$3 `);```

Aaronius:
Does Sizzle have a portion of code that handles escaping that we could copy? I don't want the whole Sizzle library, but I would like to mimic the escaping if it's in a reusable piece. If not, then I would be okay with moving forward with what you have, unless there's a simple to identify all characters that need to be escaped (character code ranges or something like that).

I think @jonsnyder is asking good questions and providing good feedback. Aside from what he's mentioned, I approve.

src/utils/dom/selectNodesWithEq.js
artur-ciocanu: Good call, I have added a comment.

src/utils/dom/selectNodesWithEq.js
artur-ciocanu: @jonsnyder thanks for your help, I have removed the capture all group and it works as expected.

artur-ciocanu:
@Aaronius and @jonsnyder after some more thought I have decided to use `CSS.escape()` to escape CSS selector identifiers. I have used: https://github.com/mathiasbynens/CSS.escape.

Basically the new `selectNodesWithEq()` works like this:
1. Parse the selector via a RegExp to find all the CSS identifiers (@jonsnyder thanks for your help with RegExp)
2. Escape identifiers using `CSS.escape()`
3. Put everything together, all this is done via `String.replace()`
4. Check for `eq(0)` and execute the selection logic.

@jonsnyder again thanks a lot for your help, now everything looks much, much cleaner.

PR #102, opened 2019-07-19
TNT-32914 - add setStyle VEC action support

artur-ciocanu:
Adding Target VEC setStyle action support

## Description

Adding Target VEC setStyle action support

## Related Issue

NONE

## Motivation and Context

Adding Target VEC setStyle action support

## Types of changes

## Checklist:

### DIFF
--- a/test/unit/components/Personalization/actions/setStyle.spec.js
+++ b/test/unit/components/Personalization/actions/setStyle.spec.js
@@ -32,14 +32,14 @@ describe("Presonalization::actions::setStyle", () => {
 
     appendNode(document.body, element);
 
-    const settings = { content: { "background-color": "red" }, meta: { a: 1 } };
+    const settings = { content: { "font-weight": "100" }, meta: { a: 1 } };
     const event = { elements, prehidingSelector: "#setStyle" };
 
     setStyle(settings, event);
 
-    expect(elements[0].style.getPropertyValue("background-color")).toEqual(
-      "red"
-    );
+    const result = window.getComputedStyle(element, null);
+
+    expect(result.getPropertyValue("font-weight")).toEqual("100");
     expect(collect).toHaveBeenCalledWith({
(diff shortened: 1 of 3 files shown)


test/unit/components/Personalization/actions/setStyle.spec.js
```diff
@@ -0,0 +1,50 @@
...
+  it("should set personalized content", () => {
+    const collect = jasmine.createSpy();
+    const setStyle = createSetStyle(collect);
+    const element = createNode("div", { id: "setStyle" });
+    const elements = [element];
+
+    appendNode(document.body, element);
+
+    const settings = {
+      content: { "font-size": "33px", priority: "important" },
```
Aaronius: I noticed that `priority` applies to all styles instead of individual styles, which seems like it might be a bit restrictive. It's also a little odd that `priority` is stored on the same level of the object as the styles. Have you considered something like:

```
content: {
  styles: [
    {
      name: "font-size",
      value: "33px",
      priority: "important"
    }
  ]
}
```

It's more verbose but more flexible and more appropriately structured in my opinion.

artur-ciocanu: @Aaronius the way Target VEC stores styles is `key` and `value` or `key`, `value` and `priority`. If we want to go the route of having an array of styles, then I'll have to adjust the edge code, so it returns the data in the proposed format.

artur-ciocanu: @Aaronius because of conflicts I had to open a new PR: https://github.com/adobe/alloy/pull/108. Please let me know if I should change the format or the current approach is OK.

Aaronius: As long as you feel like you understand what I'm trying to communicate and have taken it into consideration, I'm good with whatever you choose. 👍

PR #109, opened 2019-07-23
TNT-32915 - adding Target VEC move action support

artur-ciocanu:
Adding Target VEC move action support

## Description

Adding Target VEC move action support

## Related Issue

NONE

## Motivation and Context

Adding Target VEC move action support

## Types of changes

## Checklist:

### DIFF
--- a/test/unit/components/Personalization/actions/move.spec.js
+++ b/test/unit/components/Personalization/actions/move.spec.js
@@ -4,18 +4,18 @@ import {
   appendNode,
   createNode
 } from "../../../../../src/utils/dom";
-import createSetAttribute from "../../../../../src/components/Personalization/actions/setAttribute";
+import createMove from "../../../../../src/components/Personalization/actions/move";
 
 const cleanUp = () => {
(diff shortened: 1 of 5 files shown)


test/unit/components/Personalization/actions/move.spec.js
```diff
@@ -0,0 +1,49 @@
...
+    const settings = {
+      content: { left: "100px", top: "100px" },
+      meta: { a: 1 }
+    };
+    const event = { elements, prehidingSelector: "#setAttribute" };
+
+    move(settings, event);
+
+    expect(elements[0].style.left).toEqual("100px");
+    expect(elements[0].style.top).toEqual("100px");
```
Aaronius: It seems like you could do `element` here instead of `elements[0]` if you wanted.

artur-ciocanu: Yeah, I kind of wanted to simulate that we get a list in every action, but you are right, I could stick to `element`.

PR #110, opened 2019-07-23
TNT-32916 - adding resize Target VEC action support

artur-ciocanu:
Adding resize Target VEC action support

## Description

Adding resize Target VEC action support

## Related Issue

NONE

## Motivation and Context

Adding resize Target VEC action support

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Personalization/turbine/initRuleComponentModules.js
+++ b/src/components/Personalization/turbine/initRuleComponentModules.js
@@ -17,6 +17,7 @@ import createSetAttribute from "../actions/setAttribute";
 import createSetImageSource from "../actions/setImageSource";
 import createSetStyle from "../actions/setStyle";
 import createMove from "../actions/move";
+import createResize from "../actions/resize";
 
 export default collect => {
   const setHtml = createSetHtml(collect);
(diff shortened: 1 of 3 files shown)


src/components/Personalization/turbine/initRuleComponentModules.js
```diff
@@ -15,18 +15,21 @@ import createSetHtml from "../actions/setHtml";
...
+  const resize = createResize(collect);
 
   return {
     elementExists,
     setHtml,
     setText,
     setAttribute,
-    setImageSource
+    setImageSource,
+    resize
```
Aaronius: It seems like the `move`, `resize`, and `setStyle` actions are basically identical (except that `setStyle` supports `priority` whereas the other two don't). Rather than create an action module for each, you could map to the same module here if you wanted to. Something like:

```
return {
  ...,
  setStyle,
  resize: setStyle,
  move: setStyle
}
```

If you want each of them to have their own module though just so we retain that one-to-one mapping, that's fine with me too.

artur-ciocanu: Yeah, I would like to retain the one-to-one mapping so we could add logging and tracing and other details later.

PR #114, opened 2019-07-23
setCustomerIds - Feature/core 26064

jileeshs:
Adding setCustomer IDs command

## Description

Exposing new command which allows customers to send declared IDs to Adobe Experience platform

## Related Issue

https://jira.corp.adobe.com/browse/CORE-26064

## Types of changes

## Checklist:

- [x ] All tests pass and I've made any necessary test changes.


src/components/Identity/index.js
```diff
@@ -29,11 +36,50 @@ const createIdentity = ({ config, logger, cookie }) => {
...
+    const event = createEvent();
+    validateCustomerIDs(options.customerIDs);
+    Object.keys(options.customerIDs).forEach(key => {
+      customerIDs[key] = options.customerIDs[key];
+    });
+    const normalizedCustomerIDs = normalizeCustomerIDs(customerIDs);
+    const customerIDsHash = createHashFromString(
+      serializeCustomerIDs(normalizedCustomerIDs)
+    );
+    const hasSynced = customerIDsHash === cookie.get("CIDH");
```
jonsnyder: Use the constant you created for CIDH (and on line 72)

src/components/Identity/index.js
```diff
@@ -29,11 +36,50 @@ const createIdentity = ({ config, logger, cookie }) => {
...
+      });
+
+    return responsePromise;
+  };
+
+  const setCustomerIDs = options => {
+    const event = createEvent();
+    validateCustomerIDs(options.customerIDs);
+    Object.keys(options.customerIDs).forEach(key => {
+      customerIDs[key] = options.customerIDs[key];
```
jonsnyder: What do you think about cloning the object?

src/components/Identity/index.js
```diff
@@ -29,11 +36,50 @@ const createIdentity = ({ config, logger, cookie }) => {
...
+    });
+    const normalizedCustomerIDs = normalizeCustomerIDs(customerIDs);
+    const customerIDsHash = createHashFromString(
+      serializeCustomerIDs(normalizedCustomerIDs)
+    );
+    const hasSynced = customerIDsHash === cookie.get("CIDH");
+    event.mergeMeta({ identity: { customerIDs, hasSynced } });
+
+    if (!hasSynced) {
+      cookie.set("CIDH", customerIDsHash);
```
jonsnyder: Does this mean you cannot change the customer IDs in the cookie?

src/components/Identity/index.js
```diff
@@ -100,6 +146,12 @@ const createIdentity = ({ config, logger, cookie }) => {
     commands: {
       getEcid() {
         return optIn.whenOptedIn().then(getEcid);
+      },
+      setCustomerIDs(options) {
```
jonsnyder: Have we considered allowing customers to set the customer ids in the config?  Then the setCustomerIDs data exchange would happen on the first request.

jonsnyder: Should this be `setCustomerIds` ?

src/components/Identity/index.js
jileeshs: Done 👍

src/components/Identity/index.js
jileeshs: Didn't find an advantage to clone this object here.

src/components/Identity/index.js
jileeshs: This means we have the same info in the cookie, and hence it was send to identity backend already, there is no need to perform this operation.

src/components/Identity/index.js
jileeshs: Yes, I think I will do that in a separate task.
Created a subtask here : https://jira.corp.adobe.com/browse/CORE-32507

jileeshs: I just used the existing API name in Visitor JS today.

src/components/Identity/index.js
```diff
@@ -12,7 +12,17 @@ governing permissions and limitations under the License.
...
+  serializeCustomerIDs,
+  createHashFromString,
+  validateCustomerIDs,
+  normalizeCustomerIDs
+} from "./identityUtil";
 
+import { COOKIE_NAMES } from "./constants";
+
+const { CUSTOMER_ID_HASH, EXPERIENCE_CLOUD_ID } = COOKIE_NAMES;
 const ECID_NAMESPACE = "ECID";
```
jfkhoury: Should we remove this guy?

src/components/Identity/index.js
```diff
@@ -25,15 +35,54 @@ const addIdsContext = (payload, ecid) => {
...
+    validateCustomerIDs(options.customerIDs);
+    Object.keys(options.customerIDs).forEach(key => {
+      customerIDs[key] = options.customerIDs[key];
+    });
+    const normalizedCustomerIDs = normalizeCustomerIDs(customerIDs);
+    const customerIDsHash = createHashFromString(
+      serializeCustomerIDs(normalizedCustomerIDs)
+    );
+    const hasSynced = customerIDsHash === cookie.get(CUSTOMER_ID_HASH);
+    event.mergeMeta({ identity: { customerIDs, hasSynced } });
```
jfkhoury: Is that how Konductor expects those? Just double checking that we discussed them with Rares' team.

src/components/Identity/index.js
```diff
@@ -100,6 +149,12 @@ const createIdentity = ({ config, logger, cookie }) => {
     commands: {
       getEcid() {
         return optIn.whenOptedIn().then(getEcid);
+      },
+      setCustomerIDs(options) {
```
jfkhoury: Let's rename to `setCustomerIds` since we had this conversation last week about using `d` for configs and commands; for everything basically :)

jfkhoury: Should we simplify this command to `setCustomerIds(ids)`?

This way customers will do:

```
alloy("setCustomerIds", {
    email: {
    },
    crm: {
    }
});
```
instead of:

```
alloy("setCustomerIds", {
    customerIds: {
        email: { },
        crm: {}
    }
});
```
@jonsnyder  @Aaronius

src/components/Identity/constants.js
```diff
@@ -0,0 +1,12 @@
+const AUTH_STATES = {
```
jfkhoury: Will those be used anywhere other Identity? If so, we should move them with the other top level constants.

src/components/Identity/identityUtil.js
```diff
@@ -0,0 +1,61 @@
+import { AUTH_STATES } from "./constants";
```
jfkhoury: Should we rename this module to simply `utils.js` since it lives in `Identity` already?

src/components/Identity/identityUtil.js
```diff
@@ -0,0 +1,61 @@
...
+import isObject from "../../utils/isObject";
+
+const validateCustomerIDs = customerIDs => {
+  if (!isObject(customerIDs)) {
+    throw new Error(
+      "Invalid customer ID format. Each namespace should be an object"
```
jfkhoury: Let's extract those messages into a single variable to save those precious bits.

src/components/Identity/identityUtil.js
```diff
@@ -0,0 +1,61 @@
+import { AUTH_STATES } from "./constants";
+import isObject from "../../utils/isObject";
+
+const validateCustomerIDs = customerIDs => {
```
jfkhoury: Let's rename all the `D` to `d`.

src/components/Identity/identityUtil.js
```diff
@@ -0,0 +1,61 @@
...
+        "Invalid customer ID format. Each namespace object should have an id"
+      );
+    }
+  });
+};
+const normalizeCustomerIDs = customerIDs => {
+  const normalized = {};
+  Object.keys(customerIDs).forEach(customerID => {
+    const currentID = customerIDs[customerID];
+    normalized[customerID] = { ...currentID };
```
jfkhoury: Does `{ ...currentID }` throw a `Symbol` exception in `IE 11`?

src/components/Identity/identityUtil.js
```diff
@@ -0,0 +1,61 @@
...
+    if (
+      currentID.authState !== AUTH_STATES.AUTHENTICATED ||
+      currentID.authState !== AUTH_STATES.LOGGED_OUT
+    ) {
+      normalized[customerID].authState = AUTH_STATES.UNKNOWN;
+    }
+  });
+  return normalized;
+};
+const serializeCustomerIDs = customerIDs => {
```
jfkhoury: Do we need this? Can we simply stringify the JSON instead, then hash that?

src/components/Identity/identityUtil.js
```diff
@@ -0,0 +1,61 @@
...
+  return Object.keys(customerIDs)
+    .reduce((serialized, customerID) => {
+      return `${serialized}${customerID}|${customerIDs[customerID].id}${
+        customerIDs[customerID].authState
+      }|`;
+    }, "")
+    .slice(0, -1);
+};
+
+const createHashFromString = string => {
```
jfkhoury: Can we use the crypto APIs here, instead of this helper?

@Aaronius

src/components/Identity/index.js
```diff
@@ -25,15 +35,54 @@ const addIdsContext = (payload, ecid) => {
...
   let deferredForEcid;
+  let network;
+  let lifecycle;
+  const customerIDs = {};
+  const getCustomerIDs = () => customerIDs;
+
+  const makeServerCall = event => {
+    const payload = network.createPayload();
+    payload.addEvent(event);
+    const responsePromise = Promise.resolve()
```
jfkhoury: Can't we just return here instead of the `const`?

src/components/Identity/index.js
```diff
@@ -25,15 +35,54 @@ const addIdsContext = (payload, ecid) => {
...
+      })
+      .then(() => {
+        return network.sendRequest(payload, payload.expectsResponse);
+      });
+
+    return responsePromise;
+  };
+
+  const setCustomerIDs = options => {
+    const event = createEvent();
```
jfkhoury: I would validate first.

src/components/Identity/identityUtil.js
jileeshs: Only advantage is size, Stringify adds a `"` character before and after every key and its value.

src/components/Identity/constants.js
jileeshs: I don't know of any other component using these states

src/components/Identity/index.js
jileeshs: I will double check with Rares in the morning.

src/components/Identity/index.js
jileeshs: done

src/components/Identity/identityUtil.js
Aaronius: Here's how to create a sha-256 hash in a cross-browser way:

```
const text = 'An obscure body in the S-K System, your majesty. The inhabitants refer to it as the planet Earth.';

function buffer2hex(buffer) {
  return Array.prototype.map.call(
    new Uint8Array(buffer),
    function (x) { return ('00' + x.toString(16)).slice(-2) }
  ).join('');
}

function textEncode(str) {
  if (window.TextEncoder) {
    return new TextEncoder('utf-8').encode(str);
  } else {
    // IE 11, which doesn't have TextEncoder
    var utf8 = unescape(encodeURIComponent(str));
    var result = new Uint8Array(utf8.length);
    for (var i = 0; i < utf8.length; i++) {
      result[i] = utf8.charCodeAt(i);
    }
    return result;
  }
}

function digestMessage(message) {
  const data = textEncode(message);
  const crypto = window.msCrypto || window.crypto;
  const result = crypto.subtle.digest('SHA-256', data);
  if (result.then) {
    return result;
  } else {
    // IE 11, whose result is a CryptoOperation object instead of a promise
    return new Promise(function(resolve, reject) {
      result.addEventListener("complete", function() {
        resolve(result.result);
      });
      result.addEventListener("error", function() {
        reject();
      })
    })
  }
}

digestMessage(text).then(function(digestValue) {
  // Should result in 6efd383745a964768989b9df420811abc6e5873f874fc22a76fe9258e020c2e1
  console.log("Digest value in hex", buffer2hex(digestValue));
});
```

Aaronius: I don't fully understand what the hash is used for, but are we sure that SHA-256 is what we want to use vs, say, SHA-1 or SHA-512?

jfkhoury: Thanks @Aaronius.

I think it's used today to hash the customer ids and auth states passed in, and store that hash in the cookie, so if the customers calls us again with the same IDs, we don't try to sync them since we've seen those IDs before.

So I don't think it's necessary to use 256

src/components/Identity/util.js
```diff
@@ -0,0 +1,61 @@
+import { AUTH_STATES } from "./constants";
+import isObject from "../../utils/isObject";
+
+const ERROR_MESSAGE = "Invalid customer Id format.";
```
Aaronius: When we're writing in English, let's use `ID`. The capitalization of `*Id` or `id` is just to fit camelcasing rules (as we perceive the rules).

src/components/Identity/util.js
```diff
@@ -0,0 +1,61 @@
...
+const NOT_AN_OBJECT_ERROR = "Each namespace should be an object.";
+const NO_ID_ERROR = "Each namespace object should have an id.";
```
Aaronius: When we're writing in English, let's use `ID`. The capitalization of `*Id` or `id` is just to fit camelcasing rules (as we perceive the rules).

src/components/Identity/identityUtil.js
Aaronius: If we're using SHA to hash the serialized customer IDs, it doesn't matter what size the serialized value is; the SHA hash output will always be the same size regardless. I agree with @jfkhoury that we should use `JSON.stringify()` here, but regardless of what we use, we need to be smarter about how we serialize. Here's why:

```
JSON.stringify({ a: 1, b: 2 }) === JSON.stringify({ b: 2, a: 1 });
```

Conceptually, both objects represent the same thing to the user, but the serialized value, and thus the hashes, will be different. We probably need to break the object down and alphabetize the entries into an array of entries sorted alphanumerically or something like that, such that the output from the two inputs will always be equal. An automated test should be added for this too.

src/components/Identity/index.js
```diff
@@ -12,28 +12,76 @@ governing permissions and limitations under the License.
...
+
+    return responsePromise;
+  };
+
+  const setCustomerIds = ids => {
+    validateCustomerIds(ids);
+    const event = createEvent();
+    Object.keys(ids).forEach(key => {
+      customerIds[key] = ids[key];
+    });
```
Aaronius: If I'm reading this right, it looks like if a user calls `setCustomerIds` twice, the customer IDs passed in the second call will be merged with the customer IDs passed in the first call. Is this intentional/desired? If so, maybe take advantage of the `assign` utility we have:

```
assign(customerIds, ids);
```

src/components/Identity/index.js
```diff
@@ -100,6 +148,12 @@ const createIdentity = ({ config, logger, cookie }) => {
     commands: {
       getEcid() {
         return optIn.whenOptedIn().then(getEcid);
+      },
+      setCustomerIds(options) {
+        return optIn.whenOptedIn().then(setCustomerIds(options));
```
Aaronius: Shouldn't this be:

```
return optIn.whenOptedIn().then(() => setCustomerIds(options));
```

I'm assuming you're wanting to wait for opt in before `setCustomerIds` is called.

src/components/Identity/util.js
```diff
@@ -0,0 +1,61 @@
...
+    .slice(0, -1);
+};
+
+const createHashFromString = string => {
+  return string.split('')
+    .map(char => char.charCodeAt(0))
+    .reduce((hash, charCode) => {
+      const shift = (hash << 5) - hash + charCode; // eslint-disable-line no-bitwise
+      return shift & shift; // eslint-disable-line no-bitwise
+    }, 0);
```
Aaronius: This hashing algorithm is based on Java's hashCode algorithm, which isn't really a standard. It's probably worth using a better, more standard algorithm. CRC32, MD5, and SHA-1, while not cryptographically secure, are fine for comparing two inputs and would all result in a much smaller hash than SHA-256. While there's no native API for generating a CRC32, it would likely be the fastest, produce the smallest hash, and be the most appropriate for what we're trying to do here. It's probably the smallest amount of code too, even when taking into consideration that SHA-1 has a native API. There's a 1 in 4,294,967,296 chance of collision between two inputs with CRC32, which should be sufficient. I believe this is a proper implementation of CRC32: http://crc32.nichabi.com/javascript-function.php

Aaronius: You probably want to run some UTF-8 strings through it and compare the output to some online CRC32 hashing tools.

Aaronius: CRC32 returns a number. Since cookies are stored as strings, each digit in the CRC23 number is represented as one character. You can shorten the number of character needed by converting the number to a string that takes advantage of characters other than 0-9. So if you take the output of CRC32 and call `.toString(36)` you'll get a string using characters 0-9 and A-Z, resulting in a string that's about 6 or 7 characters long instead of a 9 or 10 characters. Technically, we could use an even higher radix by leveraging lowercase characters, etc., but there's no native API for that so it's probably not worth the code it would take to implement it.

src/components/Identity/index.js
```diff
@@ -12,28 +12,76 @@ governing permissions and limitations under the License.
...
+    const event = createEvent();
+    Object.keys(ids).forEach(key => {
+      customerIds[key] = ids[key];
+    });
+    const normalizedCustomerIds = normalizeCustomerIds(customerIds);
+    const customerIdsHash = createHashFromString(
+      serializeCustomerIds(normalizedCustomerIds)
+    );
+    const hasSynced = customerIdsHash === cookie.get(CUSTOMER_ID_HASH);
+    event.mergeMeta({ identity: { customerIds, hasSynced } });
```
Aaronius: This is probably a dumb question because I don't really understand customer ID syncing, but why do we still send the customer IDs to the server if `hasSynced` is true?

jfkhoury:
@jileeshs Should we incorporate all the comments and start over to clean up this PR?

src/components/Identity/identityUtil.js
jileeshs: It is working on IE11, tests are passing

src/components/Identity/identityUtil.js
jileeshs: @Aaronius ,
I updated the normalizeCustomerIds method to sort according to id namespaces and then take only id and authState in the order.

src/components/Identity/identityUtil.js
jileeshs: Thank you @Aaronius for code sample.
I have updated this PR.

src/components/Identity/index.js
jileeshs: Done

src/components/Identity/index.js
jileeshs: Yes, it is the desired behavior. 
I will use the assign utility

src/components/Identity/index.js
raresvla: @jileeshs If I read this correctly, it looks like you're passing both the (declared) customer IDs and a flag to specify that _nothing changed_ in the same object, under `meta.identity`.

_We need to tweak this, as Konductor expects identities to be sent in the `identityMap`._ Identities can be either set at payloads' root level, or at event level. 

What's interesting to notice further is that this part of context data is deeply merged in Konductor: event-level `identityMap` gets merged with the root level `identityMap` entries; this helps you, for example, set the ECID at root level and then have customerIDs sent in at event level, when they become available to Alloy. This is particularly handy when you pass in a batch of events to Konductor -- if only one event is sent, it does not matter where you declare the identities (root or event level).

To sum up:
1. Declared customer IDs need to be sent in `identityMap`, either as root-level or event-level context data
2. To limit the number of calls we send in to the `IdentityService` downstream of Konductor, we need that extra information in `metadata`. It can be either a flag which says "noting changed" (as it looks like it is now) or the opposite, "a new identity has been declared".

jfkhoury: Thank you @raresvla.

Should we go with:

```
metadata: {
    identity: { // Do we need this?
        newDeclaredId: true|false
    }
}
```
@Aaronius @jileeshs

jfkhoury: @jileeshs Once we get an answer on this, let's update the PR, close this one and re-open a new one that includes all the new changes from yesterday. Thanks!

Aaronius: Maybe `customerIdAdded` or `customerIdChanged` (whichever is more technically accurate) instead of `newDeclaredId`? I feel like `customerIdAdded` might suggest that the value should be a boolean more than `newDeclaredId`. The term "declared" seems new too. If it's used elsewhere or we feel like it helps specify what we're talking about, then that might be good to use, but otherwise I would say we stick to the terms we already have (e.g., "customer ID").

jfkhoury: @jileeshs I agree with @Aaronius, let's go with `customerIdChanged` since it's more accurate because an ID is really the ID itself and the Auth State; so the ID can stay the same but the auth state can change.

jfkhoury: Thank you @Aaronius!

jileeshs:
This PR has a lot of back and forth messages. I made most of the changes. Closing this one creating a fresh one.

src/components/Identity/index.js
Aaronius: The ship has probably sailed on this one, but why "customer" and not "user"? The term "customer" seems to suggest an ecommerce context, which doesn't work well with free providers (publishers, social networks, etc). If "customer" is too engrained at this point, just let me know and I'll go on my way...

jonsnyder: I always thought of it as in Adobe's Customer's Custom ID.  i.e. the ID that is owned by Adobe's Customer.

Aaronius: Oh that's interesting. I never thought of it that way.

PR #119, opened 2019-07-24
TNT-32923 - adding Target VEC prependHtml action support

artur-ciocanu:
Adding Target VEC prependHtml action support

## Description

Adding Target VEC prependHtml action support

## Related Issue

NONE

## Motivation and Context

In Target VEC we can insert a piece of HTML as first element inside an HTML element. This action is called `prepend HTML` or `prepend content`, this PR adds support for this VEC action.

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Personalization/actions/prependHtml.js
+++ b/src/components/Personalization/actions/prependHtml.js
@@ -24,10 +24,17 @@ const createFragment = content => {
 const prependHtml = (container, content) => {
   const fragment = createFragment(content);
   const elements = [].slice.call(fragment.children);
+  const { length } = elements;
+  let i = length - 1;
 
-  elements.forEach(element => {
+  // We are inserting elements in reverse order
+  while (i >= 0) {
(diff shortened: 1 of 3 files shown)


src/components/Personalization/actions/prependHtml.js
```diff
@@ -0,0 +1,49 @@
...
+  return result;
+};
+
+const prependHtml = (container, content) => {
+  const fragment = createFragment(content);
+  const elements = [].slice.call(fragment.children);
+
+  elements.forEach(element => {
+    container.insertBefore(element, container.firstElementChild);
+  });
```
Aaronius: If `fragment.children` has more than one child, wouldn't the children end up unexpectedly in reverse order?

artur-ciocanu: Good catch, adjusted and added proper unit test to account for this.

PR #121, opened 2019-07-25
TNT-32988 - move actions Personalization duplicate code to utils/dom …

artur-ciocanu:
Moving actions Personalization duplicate code to utils/dom and utils 

## Description

Moving actions Personalization duplicate code to utils/dom and utils 

## Related Issue

NONE

## Motivation and Context

Currently Personalization component has a lot of duplications. The intent of this patch is to move code to DOM utils utilities or other utils folders if required.

## Types of changes

## Checklist:

### DIFF
--- a/src/utils/dom/append.js
+++ b/src/utils/dom/append.js
@@ -13,8 +13,8 @@ governing permissions and limitations under the License.
 import createFragment from "./createFragment";
 import getChildren from "./getChildren";
 
-export default (container, content) => {
-  const fragment = createFragment(content);
+export default (container, html) => {
+  const fragment = createFragment(html);
   const elements = getChildren(fragment);
 
   elements.forEach(element => {
--- a/src/utils/dom/append.js
+++ b/src/utils/dom/append.js
@@ -10,14 +10,14 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import toArray from "../toArray";
+import createFragment from "./createFragment";
+import getChildren from "./getChildren";
 
-/**
- * Returns an array of matched DOM nodes.
- * @param {String} selector
- * @param {Node} doc, defaults to document
- * @returns {Array} an array of DOM nodes
- */
-export default function selectNodes(selector, doc = document) {
-  return toArray(doc.querySelectorAll(selector));
-}
+export default (container, content) => {
+  const fragment = createFragment(content);
+  const elements = getChildren(fragment);
+
+  elements.forEach(element => {
+    container.appendChild(element);
+  });
+};
--- a/src/utils/dom/getParent.js
+++ b/src/utils/dom/getParent.js
@@ -10,14 +10,12 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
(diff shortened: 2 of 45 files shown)


artur-ciocanu:
@jfkhoury this is the PR that moves all the code duplications to DOM utils and other utils folders. It would be nice if you could take a look.

Thank you!

src/utils/dom/append.js
```diff
@@ -0,0 +1,23 @@
...
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+import createFragment from "./createFragment";
+import getChildren from "./getChildren";
+
+export default (container, content) => {
```
Aaronius: It might be appropriate to rename `content` to `html` or something like that. It took me a while to figure out that `content` was HTML and not DOM nodes, for example.

src/utils/dom/getParent.js
```diff
@@ -0,0 +1,21 @@
...
+*/
+
+export default element => {
+  const result = element.parentNode;
+
+  if (result) {
+    return result;
+  }
+
+  return null;
```
Aaronius: Is this necessary? If the element doesn't have a parent, `element.parentNode` should be `null` as far as I understand, so it seems like you just do `return element.parentNode`. If there is a reason for this extra logic, a comment might help clarify for future developers.

Aaronius:
It seems like there's still an opportunity for code sharing among actions. It looks like all actions are basically composed of:

```
import { insertBefore } from "../../../utils/dom"; // unique
import { showElements } from "../flicker"; // shared

export default collect => { // shared
  return (settings, event) => { // shared
    const { elements, prehidingSelector } = event; // shared
    const { content, meta } = settings; // shared

    // this is a very naive approach, we will expand later
    elements.forEach(element => { // unique
      insertBefore(element, content); // unique
    }); // unique

    // after rendering we should remove the flicker control styles
    showElements(prehidingSelector); // shared

    // make sure we send back the metadata after successful rendering
    collect({ meta: { personalization: meta } }); // shared
  }; // shared
}; // shared
```

Could you do something like:

```
import { insertBefore } from "../../../utils/dom"; 
import runAction from "whatever"; 

runAction((elements, content) => {
  elements.forEach(element => {
      insertBefore(element, content);
    });
});
```

src/utils/dom/append.js
artur-ciocanu: Goog catch, renamed to `html`.

src/utils/dom/getParent.js
artur-ciocanu: Nope, it is not necessary I think I was just carried away. The initial intent was to check if element is `null`, but then I dropped that idea.

Fixed.

artur-ciocanu:
@Aaronius I agree, now that everything has been moved to different utilities, all the actions have basically the same shape, that was the intent.

I would like to address that in a different PR, since this one is already pretty big.

jfkhoury:
Nice @artur-ciocanu!

PR #122, opened 2019-07-26
Click activity collector

liljenback:
Initial commit for click activity collection to start the discussion.

## Description

Added a click handler that collects some basic data and fires a collect call that aligns with existing XDM spec.

## Motivation and Context

Will add optional click activity tracking to alloy.

## Types of changes

## Checklist:

### DIFF
--- a/sandbox/public/index.html
+++ b/sandbox/public/index.html
@@ -74,7 +74,12 @@
         // edgeDomain: "edgegateway.azurewebsites.net",
         propertyId: 8888888,
         imsOrgId: "53A16ACB5CC1D3760A495C99@AdobeOrg",
-        log: true
+        log: true,
+        dataCollector: {
+          clickActivity: {
+            enabled: false
+          }
+        }
       });
     </script>
   </head>
--- a/sandbox/public/index.html
+++ b/sandbox/public/index.html
@@ -75,11 +75,7 @@
         propertyId: 8888888,
         imsOrgId: "53A16ACB5CC1D3760A495C99@AdobeOrg",
         log: true,
-        dataCollector: {
-          clickActivity: {
-            enabled: false
-          }
-        }
+        clickCollectionEnabled: false
       });
     </script>
   </head>
--- a/src/components/DataCollector/activity/click.js
+++ b/src/components/DataCollector/activity/click.js
@@ -81,7 +81,6 @@ const createClickHandler = (config, logger, collect) => {
     if (objectHref && isSupportedHrefObject(clickObject)) {
       // TODO: Update name (link name) and type (exit, other, download) to be collected
       collect({
-        beacon: true,
         data: {
           eventType: "web.webinteraction",
           name: "Link Click",
--- a/src/components/DataCollector/activity/click.js
+++ b/src/components/DataCollector/activity/click.js
@@ -79,15 +79,16 @@ const createClickHandler = (config, logger, collect) => {
       }
     }
     if (objectHref && isSupportedHrefObject(clickObject)) {
+      // TODO: Update name (link name) and type (exit, other, download) to be collected
       collect({
         beacon: true,
         data: {
-          webinteraction: {
-            type: "other",
-            URL: objectHref,
-            linkClicks: {
-              value: 1
-            }
+          eventType: "web.webinteraction",
+          name: "Link Click",
+          type: "other",
+          URL: objectHref,
+          linkClicks: {
+            value: 1
           }
         }
       });
--- a/src/components/DataCollector/activity/click.js
+++ b/src/components/DataCollector/activity/click.js
@@ -96,9 +96,9 @@ const createClickHandler = (config, logger, collect) => {
 };
 
 export default (config, logger, collect) => {
-  const enabled = config.get("dataCollector.clickActivity.enabled", true);
+  const enabled = config.get("clickCollectionEnabled");
   if (!enabled) {
-    logger.log("The click activity collector is disabled");
+    logger.log("Click activity collection is disabled");
     return;
   }
   const clickHandler = createClickHandler(config, logger, collect);
--- a/src/components/DataCollector/activity/click.js
+++ b/src/components/DataCollector/activity/click.js
@@ -1,0 +1,108 @@
+/*
+Copyright 2019 Adobe. All rights reserved.
+This file is licensed to you under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License. You may obtain a copy
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+const isValidObject = clickObj => {
+  if (clickObj) {
+    if (clickObj.tagName || clickObj.parentElement || clickObj.parentNode) {
+      return true;
+    }
+  }
+  return false;
+};
+
+const isSupportedHrefObject = obj => {
+  if (
+    obj.href &&
+    (obj.tagName === "A" || obj.tagName === "AREA") &&
+    (!obj.onclick ||
+      !obj.protocol ||
+      obj.protocol.toLowerCase().indexOf("javascript") < 0)
+  ) {
+    return true;
+  }
+  return false;
+};
+
+const getObjectHref = obj => {
+  const loc = window.location;
+  let href = obj.href ? obj.href : "";
+  let { protocol, host } = obj;
+  let i = href.indexOf(":");
+  const j = href.indexOf("?");
+  const k = href.indexOf("/");
+  if (href && (i < 0 || (j >= 0 && i > j) || (k >= 0 && i > k))) {
+    if (!(protocol && protocol.length > 1)) {
+      protocol = loc.protocol ? loc.protocol : "";
+    }
+    protocol = protocol ? `{$protocol}//` : "";
+    if (!host) {
+      host = loc.host ? loc.host : "";
+    }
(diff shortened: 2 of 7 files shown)


src/components/DataCollector/activity/click.js
```diff
@@ -0,0 +1,108 @@
...
+        : clickObject.parentNode;
+      if (clickObject) {
+        objectHref = getObjectHref(clickObject);
+      }
+    }
+    if (objectHref && isSupportedHrefObject(clickObject)) {
+      collect({
+        beacon: true,
+        data: {
+          webinteraction: {
```
jonsnyder: Reading the xdm is tricky.  You have to look where webinteraction is referenced.  This is the correct path:
```web: { webInteraction: { ... } }```

File: src/components/DataCollector/activity/click.js [hunk 2]
```diff
@@ -0,0 +1,108 @@
...
+        ? clickObject.parentElement
+        : clickObject.parentNode;
+      if (clickObject) {
+        objectHref = getObjectHref(clickObject);
+      }
+    }
+    if (objectHref && isSupportedHrefObject(clickObject)) {
+      collect({
+        beacon: true,
+        data: {
```
jonsnyder: Consider adding `eventType: "web.webInteraction.linkClicks"` as a property under `data`

src/components/DataCollector/activity/click.js
```diff
@@ -0,0 +1,108 @@
...
+      clickObject = clickObject.parentElement
+        ? clickObject.parentElement
+        : clickObject.parentNode;
+      if (clickObject) {
+        objectHref = getObjectHref(clickObject);
+      }
+    }
+    if (objectHref && isSupportedHrefObject(clickObject)) {
+      collect({
+        beacon: true,
```
jonsnyder: The data collector no longer uses the `beacon` property.

src/components/DataCollector/activity/click.js
```diff
@@ -0,0 +1,108 @@
...
+    (!obj.onclick ||
+      !obj.protocol ||
+      obj.protocol.toLowerCase().indexOf("javascript") < 0)
+  ) {
+    return true;
+  }
+  return false;
+};
+
+const getObjectHref = obj => {
```
jonsnyder: Some test cases would be nice for this function, if only to serve as documentation.

src/components/DataCollector/activity/click.js
```diff
@@ -0,0 +1,108 @@
...
+  const loc = window.location;
+  let href = obj.href ? obj.href : "";
+  let { protocol, host } = obj;
+  let i = href.indexOf(":");
+  const j = href.indexOf("?");
+  const k = href.indexOf("/");
+  if (href && (i < 0 || (j >= 0 && i > j) || (k >= 0 && i > k))) {
+    if (!(protocol && protocol.length > 1)) {
```
Aaronius: I believe `if (!protocol) {` would do the same thing.

src/components/DataCollector/activity/click.js
```diff
@@ -0,0 +1,108 @@
...
+      protocol = loc.protocol ? loc.protocol : "";
+    }
+    protocol = protocol ? `{$protocol}//` : "";
+    if (!host) {
+      host = loc.host ? loc.host : "";
```
Aaronius: Maybe simplify:
`host = loc.host || ""`

src/components/DataCollector/activity/click.js
Aaronius: I agree with this ^. I'm having a difficult time understanding what's going on here. My feedback on it is limited to what I understand.

src/components/DataCollector/activity/click.js
```diff
@@ -0,0 +1,108 @@
...
+};
+
+export default (config, logger, collect) => {
+  const enabled = config.get("dataCollector.clickActivity.enabled", true);
+  if (!enabled) {
+    logger.log("The click activity collector is disabled");
+    return;
+  }
+  const clickHandler = createClickHandler(config, logger, collect);
+  if (document && document.addEventListener) {
```
Aaronius: When would `document` or` document.addEventListener` not exist in our target browsers?

src/components/DataCollector/activity/click.js
```diff
@@ -0,0 +1,108 @@
...
+      return;
+    }
+
+    let objectHref = getObjectHref(clickObject);
+    // Iterate through parent objects to obtain valid href
+    // TODO: Replace this with generic DOM tool that fetch all configured object properties
+    while (clickObject && clickObject.tagName !== "BODY" && !objectHref) {
+      clickObject = clickObject.parentElement
+        ? clickObject.parentElement
+        : clickObject.parentNode;
```
Aaronius: Maybe simplify:
`clickObject = clickObject.parentElement || clickObject.parentNode`

src/components/DataCollector/activity/click.js
```diff
@@ -0,0 +1,108 @@
...
+    return true;
+  }
+  return false;
+};
+
+const getObjectHref = obj => {
+  const loc = window.location;
+  let href = obj.href ? obj.href : "";
+  let { protocol, host } = obj;
+  let i = href.indexOf(":");
```
jfkhoury: Let's stick to using clear variable names instead of single chars.

src/components/DataCollector/activity/click.js
jfkhoury: It would be really cool if this helper memoized the parsed hrefs.

src/components/DataCollector/activity/click.js
```diff
@@ -0,0 +1,108 @@
...
+  const j = href.indexOf("?");
+  const k = href.indexOf("/");
+  if (href && (i < 0 || (j >= 0 && i > j) || (k >= 0 && i > k))) {
```
jfkhoury: We need a way to make this `if` clearer somehow: either we place the condition in a variable with a meaningful name or even break it down into multiple variables.

src/components/DataCollector/activity/click.js
liljenback: @jonsnyder are you suggesting that the eventType would represent the XDM path prefix?

src/components/DataCollector/activity/click.js
liljenback: @jonsnyder , looks like it is being set as default so I removed it.

sandbox/public/index.html
```diff
@@ -61,7 +61,12 @@
         // edgeDomain: "edgegateway.azurewebsites.net",
         propertyId: 8888888,
         imsOrgId: "53A16ACB5CC1D3760A495C99@AdobeOrg",
-        log: true
+        log: true,
+        dataCollector: {
+          clickActivity: {
+            enabled: false
+          }
+        }
```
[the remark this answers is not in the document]
liljenback: @Aaronius , I collapsed the config setting to one level for now.

src/components/DataCollector/activity/click.js
```diff
@@ -96,9 +96,9 @@ const createClickHandler = (config, logger, collect) => {
 };
 
 export default (config, logger, collect) => {
-  const enabled = config.get("dataCollector.clickActivity.enabled", true);
+  const enabled = config.get("clickCollectionEnabled");
```
jfkhoury: Let' get rid of `getNestedObject` & `setNestedObject`.

src/components/DataCollector/activity/click.js
jonsnyder: "web.webInteraction.linkClicks" is one of the known values for eventType. See https://github.com/adobe/xdm/blob/master/schemas/data/time-series.schema.json .  It is unclear to me whether this should be the eventType, or whether you increment the value like you are doing here, or both.

The question on what to put inside here brings to light an additional piece of data we are collecting automatically (besides context data).  We are having a discussion on Friday about how to distinguish Alloy generated data versus user generated data and how that gets mapped on the server.

src/components/DataCollector/activity/click.js
jonsnyder: :+1:

src/components/DataCollector/activity/click.js
```diff
@@ -0,0 +1,108 @@
...
+*/
+
+const isValidObject = clickObj => {
+  if (clickObj) {
+    if (clickObj.tagName || clickObj.parentElement || clickObj.parentNode) {
+      return true;
+    }
+  }
+  return false;
+};
```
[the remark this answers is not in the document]
liljenback: I'm removing checks for invalid objects. Couldn't find a valid reason to keep the check. We can add something back if needed.

src/components/DataCollector/activity/click.js
```diff
@@ -0,0 +1,108 @@
...
+    obj.href &&
+    (obj.tagName === "A" || obj.tagName === "AREA") &&
+    (!obj.onclick ||
+      !obj.protocol ||
+      obj.protocol.toLowerCase().indexOf("javascript") < 0)
+  ) {
+    return true;
+  }
+  return false;
+};
```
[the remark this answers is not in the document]
liljenback: IE and Edge do have the protocol property.

src/components/DataCollector/activity/click.js
liljenback: Not sure if that would render any performance increase unless multiple clicks on the same objects. I'll keep it in mind.

liljenback: I reworked the code a bit and added test cases

src/components/DataCollector/activity/click.js
liljenback: I think I agree. However, the greater than one character is probably there for a reason.

src/components/DataCollector/activity/click.js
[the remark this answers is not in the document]
Aaronius: Huh, you're right. For some reason MDN shows it as not supported (https://developer.mozilla.org/en-US/docs/Web/API/HTMLHyperlinkElementUtils/protocol).

PR #124, opened 2019-07-26
CORE-32537 - ensure query -> personalization -> enabled if isViewStart

artur-ciocanu:
Ensure event query contains personalization if event has "view start".

## Description

If event has "view start" we should ensure that request that is passed to backend has `query -> personalization -> enabled` set to `true`. Otherwise backend won't try to run the personalization decision engine.

## Related Issue

NONE

## Motivation and Context

We need to ensure that backend is instructed to return personalization content depending on event "view start".

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Personalization/index.js
+++ b/src/components/Personalization/index.js
@@ -85,14 +85,16 @@ const createPersonalization = ({ config, logger, cookie }) => {
         });
       },
       onBeforeEvent(event, options, isViewStart) {
-        if (isViewStart) {
-          // If isViewStart we enable personalization
-          event.mergeQuery({ personalization: { enabled: true } });
-          event.expectResponse();
-
-          // For viewStart we try to hide the personalization containers
-          hideContainers(prehidingId, prehidingStyle);
+        if (!isViewStart) {
+          // If NOT isViewStart disable personalization
+          event.mergeQuery({ personalization: { enabled: false } });
+          return;
         }
+
+        event.expectResponse();
+
+        // For viewStart we try to hide the personalization containers
+        hideContainers(prehidingId, prehidingStyle);
       },
       onResponse(response) {
         const fragment = response.getPayloadByType(PAGE_HANDLE) || {};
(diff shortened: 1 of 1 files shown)


src/components/Personalization/index.js
```diff
@@ -62,9 +62,12 @@ const createPersonalization = ({ config, logger }) => {
       },
       onBeforeEvent(event, options, isViewStart) {
         if (isViewStart) {
+          // If isViewStart we enable personalization
+          event.mergeQuery({ personalization: { enabled: true } });
```
jonsnyder: From the description in the ticket, we need to do the opposite. i.e:
```
if (!isViewStart) {
  event.mergeQuery({ personalization: { enabled: false } });
} else {
  hideContainers(prehidingId, prehidingStyle);
  event.expectResponse();
}
```

artur-ciocanu: @jonsnyder the way it is currently implemented on our edge is a little bit different. By default edge doesn't process `personalization`, you have to explicitly pass `query -> personalization -> enabled = true`, otherwise it won't work.

If we want to change the edge implementation such that personalization is always enabled and we have to explicitly disabled, then we should probably talk to @jfkhoury and Rares's team.

Let me know your thoughts.

Aaronius: I think all the other queries are `enabled: true` by default, so it's a little sad if personalization is inconsistent. That doesn't mean it's the wrong thing to do, but we should probably scrutinize it a bit more. I do think we should pass this one by Rares's team.

artur-ciocanu: @Aaronius and @jonsnyder I have checked our edge code, by default all features are enabled, so I think it makes sense to be consistent.

I'll adjust the code accordingly and I'll open a PR for our backend code to change the way we handle personalization.

artur-ciocanu:
@jonsnyder I have adjusted the code according to ticket. Could you please take another look. Thank you.

jfkhoury:
@artur-ciocanu @Aaronius @jonsnyder 

Just wanted to clarify one thing: ID Syncs are `disabled` by default in Konductor, so I don't think we can assume that everything is enabled by default.

If you think we should be consistent, let's discuss this issue with the edge team.

It seems like it's a case by case thing, since they service other clients and not just Web.

PR #125, opened 2019-07-26
TNT-32999 - add sessionId tracking for personalization component

artur-ciocanu:
Add user session tracking

## Description

Add user session tracking

## Related Issue

NONE

## Motivation and Context

Personalization component can be used to personalize depending on your in-session activities. Hence we need a way to track user session.

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Personalization/index.js
+++ b/src/components/Personalization/index.js
@@ -10,29 +10,32 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import { isNonEmptyArray, uuid, getTopLevelCookieDomain } from "../../utils";
+import { isNonEmptyArray, uuid } from "../../utils";
 import { initRuleComponentModules, executeRules } from "./turbine";
 import { hideContainers, showContainers, hideElements } from "./flicker";
 
 const PAGE_HANDLE = "personalization:page";
-const SESSION_ID_COOKIE = "alloy-session-id";
+const SESSION_ID_COOKIE = "SID";
 const SESSION_ID_TTL = 31 * 60 * 1000;
 const EVENT_COMMAND = "event";
+
 const isElementExists = event => event.moduleType === "elementExists";
 
 const getOrCreateSessionId = cookie => {
-  let sessionId = cookie.get(SESSION_ID_COOKIE);
-  const domain = getTopLevelCookieDomain(window, cookie);
-  const expires = new Date(Date.now() + SESSION_ID_TTL);
+  let cookieValue = cookie.get(SESSION_ID_COOKIE);
+  const now = Date.now();
+  const expires = now + SESSION_ID_TTL;
 
-  if (!sessionId) {
-    sessionId = uuid();
+  if (!cookieValue || now > cookieValue.expires) {
+    cookieValue = { value: uuid(), expires };
+  } else {
+    cookieValue.expires = expires;
   }
 
-  // We have to extend cookie lifetime
-  cookie.set(SESSION_ID_COOKIE, sessionId, { domain, expires });
+  // We have to extend session ID lifetime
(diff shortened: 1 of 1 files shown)


src/components/Personalization/index.js
```diff
@@ -60,6 +77,11 @@ const createPersonalization = ({ config, logger }) => {
           componentRegistry.getCommand(EVENT_COMMAND)
         );
       },
+      onBeforeDataCollection(payload) {
+        const sessionId = getOrCreateSessionId(cookie);
```
Aaronius: Because `getOrCreateSessionId` uses cookie values and writes to cookies, this probably needs to happen _after_ the user opts in. See https://github.com/adobe/alloy/blob/master/src/components/Identity/index.js#L51 for an example of how this is done.

src/components/Personalization/index.js
```diff
@@ -10,14 +10,31 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
 */
 
-import { isNonEmptyArray } from "../../utils";
+import { isNonEmptyArray, uuid, getTopLevelCookieDomain } from "../../utils";
 import { initRuleComponentModules, executeRules } from "./turbine";
 import { hideContainers, showContainers, hideElements } from "./flicker";
 
 const PAGE_HANDLE = "personalization:page";
+const SESSION_ID_COOKIE = "alloy-session-id";
+const SESSION_ID_TTL = 31 * 60 * 1000;
```
Aaronius: 31 minutes? What the reasoning behind that value? Just curious more than anything.

src/components/Personalization/index.js
```diff
@@ -10,14 +10,31 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+  let sessionId = cookie.get(SESSION_ID_COOKIE);
+  const domain = getTopLevelCookieDomain(window, cookie);
+  const expires = new Date(Date.now() + SESSION_ID_TTL);
+
+  if (!sessionId) {
+    sessionId = uuid();
+  }
+
+  // We have to extend cookie lifetime
+  cookie.set(SESSION_ID_COOKIE, sessionId, { domain, expires });
```
Aaronius: I believe `cookie` is the alloy cookie accessor object (found in `src/core/createCookie.js`), which doesn't accept a `domain` or `expires`. Since the alloy cookie is a plethora of data, it all shares the same expiry (and domain), which is `180` days.

src/components/Personalization/index.js
artur-ciocanu: This is what we currently have in Adobe Target, basically if there is 31 mins of inactivity.

src/components/Personalization/index.js
artur-ciocanu: Thanks, I have adjusted.

src/components/Personalization/index.js
artur-ciocanu: I have removed the domain and expires and I have added expires as part of the cookie value.

artur-ciocanu:
@Aaronius I have addressed all your comments. Could you please take another look.

I have also have a question, may be it's optIn related by I have run into this scenario in sandbox:
1. Clear cache and cookies
2. Reload the page
3. Session ID in personalization is 1111
4. Reload the page again, now session ID is 2222
5. Reload the page again, session ID is 2222

I am not sure either this my implementation or this is related to sandbox or something else.

Aaronius:
Huh. I'll take a look.

Aaronius:
It's a problem with the cookie proxy inside `createCookie.js`. I'll fix it in a new PR.

Aaronius:
Here's the cookie proxy issue: https://jira.corp.adobe.com/browse/CORE-32658

artur-ciocanu:
@Aaronius appreciate! Thanks for review!

PR #126, opened 2019-07-26
CORE-32658 Fixes issue where cookie proxies weren't sharing the same cookie cache.

Aaronius:
## Description

Currently `createCookie`, found in `createCookie.js`, is called once per component. Each time `createCookie` is created, a new cookie proxy is created, which caches the deserialized cookie so that we're not reading the cookie from persistence every time we need a value. Because a cookie proxy is created for each component, we have multiple caches of the same alloy cookie, which means the caches don't stay up-to-date, which eventually results in a persisted cookie that doesn't contain all the latest data.

This fixes that problem by sharing the same cookie proxy for each property ID and therefore each cookie.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-32658

## Motivation and Context
Corrupted cookies make sad users.

## Types of changes

## Checklist:

### DIFF
--- a/src/core/createCookie.js
+++ b/src/core/createCookie.js
@@ -34,38 +34,42 @@ const createCookieProxy = (propertyId, cookieDomain = "") => {
   // caches per cookies would result in caches getting stale.
   let cookieProxy = cookieProxyByPropertyId[propertyId];
 
-  if (!cookieProxy) {
-    const cookieName = `${ALLOY_COOKIE_NAME}_${propertyId}`;
-    let deserializedCookie;
-    let cookieHasBeenRead = false;
-
-    cookieProxy = {
-      get() {
-        // We don't read the cookie right when the cookie proxy is created
-        // because we don't know if the user has opted in. If the user
-        // hasn't opted in, we're legally obligated to not read the cookie.
-        // If a component tries to read something off the cookie though,
-        // we assume that the component has received word that the user
-        // has opted-in. The responsibility is on the component.
-        if (!cookieHasBeenRead) {
-          const serializedCookie = cookie.get(cookieName);
-          deserializedCookie =
-            serializedCookie && safeJSONParse(serializedCookie, cookieName);
-          cookieHasBeenRead = true;
-        }
-        return deserializedCookie;
-      },
-      set(updatedCookie) {
-        deserializedCookie = updatedCookie;
-        cookie.set(cookieName, updatedCookie, {
-          expires: ALLOY_COOKIE_TTL,
-          domain: cookieDomain || getTopLevelCookieDomain(window, cookie)
-        });
-      }
-    };
-
-    cookieProxyByPropertyId[propertyId] = cookieProxy;
+  if (cookieProxy) {
+    return cookieProxy;
   }
+
+  const cookieName = `${ALLOY_COOKIE_NAME}_${propertyId}`;
+  let deserializedCookie;
+  let cookieHasBeenRead = false;
+
+  cookieProxy = {
+    get() {
+      // We don't read the cookie right when the cookie proxy is created
+      // because we don't know if the user has opted in. If the user
+      // hasn't opted in, we're legally obligated to not read the cookie.
+      // If a component tries to read something off the cookie though,
+      // we assume that the component has received word that the user
+      // has opted-in. The responsibility is on the component.
+      if (cookieHasBeenRead) {
+        return deserializedCookie;
+      }
+
+      const serializedCookie = cookie.get(cookieName);
+      deserializedCookie =
+        serializedCookie && safeJSONParse(serializedCookie, cookieName);
+      cookieHasBeenRead = true;
+      return deserializedCookie;
+    },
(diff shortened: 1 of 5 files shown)


src/core/createCookie.js
```diff
@@ -28,35 +29,45 @@ const safeJSONParse = (object, cookieName) => {
...
 const createCookieProxy = (propertyId, cookieDomain = "") => {
-  const cookieName = `${ALLOY_COOKIE_NAME}_${propertyId}`;
+  // Because the cookie proxy caches the deserialized cookie, we want to make
+  // sure there's only one cookie proxy created per cookie because multiple
+  // caches per cookies would result in caches getting stale.
+  let cookieProxy = cookieProxyByPropertyId[propertyId];
 
-  let deserializedCookie;
-  let cookieHasBeenRead = false;
+  if (!cookieProxy) {
```
artur-ciocanu: Could we invert the condition?

src/core/createCookie.js
```diff
@@ -28,35 +29,45 @@ const safeJSONParse = (object, cookieName) => {
...
-          serializedCookie && safeJSONParse(serializedCookie, cookieName);
-        cookieHasBeenRead = true;
+    cookieProxy = {
+      get() {
+        // We don't read the cookie right when the cookie proxy is created
+        // because we don't know if the user has opted in. If the user
+        // hasn't opted in, we're legally obligated to not read the cookie.
+        // If a component tries to read something off the cookie though,
+        // we assume that the component has received word that the user
+        // has opted-in. The responsibility is on the component.
```
artur-ciocanu: May be we can invert the condition, here as well.

artur-ciocanu:
@Aaronius looks good!

src/core/createCookie.js
```diff
@@ -108,3 +119,11 @@ const createCookie = (componentNamespace, propertyId, cookieDomain = "") => {
 };
 
 export default createCookie;
+
+// #if _TEST
```
jonsnyder: You could avoid this by adding this as a parameter to the createCookieProxy function.  Then you would define the cookieProxyByPropertyId object in main.js.

Aaronius: I'm not sure how I feel about this. In some ways, I like that it would remove the test-specific code, but on the other hand, `cookieProxyByPropertyId` feels like an implementation detail that shouldn’t be exposed to or modified by the consumer of the module. I'll bring it up in chat, because I'd like to see how others feel about it.

src/core/createCookie.js
Aaronius: Sure. I added a commit with the change.

src/core/createCookie.js
Aaronius: Sure. I added a commit with the change.

src/core/createCookie.js
jfkhoury: I don't mind the `_TEST`; I see it as just another tool to help us test, similar to the build tools and plugins. And it makes it clear to understand that this portion is for testing purposes only; plus it looks like you are using it in multiple places. (To clear the top level domain cache)

I'd rather do that, than leaking the proxy stuff outside of this module.

jonsnyder: I think of it as the createCookie function having a dependency on a cache.  As it is in this PR, the cache is a hidden global variable.  In general explicit dependencies are better than hidden dependencies, and global state should be avoided. In this case, the test logic is fairly benign, but `#if _TEST` could quickly get out of hand to the point where the system behaves differently when under test.  

I'm good either way, and understand that everyone prefers testing in different ways.  I'm just pointing out that the `#if _TEST` could be avoided in this case.

jfkhoury: @jonsnyder I might have misunderstood your original comment. I absolutely agree that the system should behave the same way whether it's in test mode or not, and we definitely don't want hidden dependencies. 

DI is usually the answer to every similar situation; so the question is, are `proxy` in this case, and `cached topleveldomain` in the Top Level Domain module dependencies? Maybe. And maybe we have a `createMemoizer` or something that we re-use to inject modules in need of caching with a generic memoizer.

jonsnyder: Yes, I would make a similar comment on the getTopLevelCookieDomain function.  Expanding on what you said @jfkhoury, we could remove the caching/lookup logic from these functions, so they compute the values every time.  Then we could add a utility to memoize the function, and pass the memoized function to those that need it.

I misspoke before; these aren't really global variables.  They don't get defined on window, but it is global state.

getTopLevelCookieDomain should return the same value for any instance of alloy, so we need to have global state somewhere.  My preference is to explicitly have global state defined at the highest level (i.e. main.js).  The benefit of avoiding global state in these functions is that then they can be unit tested more easily.  I see main.js as containing glue code to wire up all the dependencies and avoiding logic.  Then main.js is tested using functional/integration tests to make sure everything is wired up properly and works together.

Aaronius: Thanks for the feedback. I'm going to play around with these ideas and open in a new pull request if it shakes out well.

PR #127, opened 2019-07-29
CORE-31591 If enableSyncs is false, set the syncs to false in Query

wgamboa:
## Related Issue
CORE-31591

## Motivation and Context
To turn off fetching ID syncs, pass the following in the Query portion of the Request:

"identity":

{ "exchange": false }
exchange is turned off by default in Konductor. So, Identity should set exchange to true when we need them.

## Types of changes

## Checklist:


src/components/Identity/processIdSyncs.js
```diff
@@ -35,15 +38,15 @@ const setControlObject = obj => {
...
-    expires: 6 * 30 // 6 months
-  });
+  alloyCookie.set(ID_SYNC_CONTROL, arr.join("_"));
 };
 
-export default ({ destinations, config, logger }) => {
+export default ({ destinations, config, logger, cookie }) => {
+  alloyCookie = cookie;
+
   if (config.idSyncsEnabled) {
```
jfkhoury: Just a suggestion: I would `return` if this config is `false`, and take the code out of the `if` statement's block.

src/components/Identity/processIdSyncs.js
```diff
@@ -10,14 +10,17 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import { assign, cookie, fireDestinations } from "../../utils";
-import namespace from "../../constants/namespace";
+import { assign, fireDestinations } from "../../utils";
 
-const millisecondsPerHour = 60 * 60 * 1000;
+const ID_SYNC_CONTROL = "idSyncControl";
+const ID_SYNC_TIMESTAMP = "idSyncTimestamp";
```
jfkhoury: It looks like you are using `ID_SYNC_TIMESTAMP` and `MILLISECONDS_PER_HOUR` here in the `index.js` modules. I would move them to the `constants` module and reuse them.

src/components/Identity/index.js
```diff
@@ -37,6 +39,17 @@ const createIdentity = ({ config, logger, cookie }) => {
       },
       // Waiting for opt-in because we'll be reading the ECID from a cookie
       onBeforeEvent(event) {
+        const nowInHours = Math.round(
```
jfkhoury: I think we should run this logic after we've received the opt-in consent, especially that we are reading the timestamp from the cookie.

jfkhoury: This `nowInHours` guy seems like a good candidate for a util since you are using it in both modules.

src/components/Identity/processIdSyncs.js
jfkhoury: It can be a `constants.js` module within the `Identity` component if no one else uses those consts.

src/components/Identity/index.js
wgamboa: There's only one place where I use it.

src/components/Identity/processIdSyncs.js
```diff
@@ -10,14 +10,17 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
-const millisecondsPerHour = 60 * 60 * 1000;
+let alloyCookie;
 
-// TODO: use alloy cookie once https://github.com/adobe/alloy/pull/26 is merged
 const getControlObject = () => {
-  const val = cookie.get(`${namespace}idSyncControl`) || "";
+  const val = alloyCookie.get(ID_SYNC_CONTROL) || "";
   const arr = val ? val.split("_") : [];
 
   return arr.reduce((obj, pair) => {
```
Aaronius: In this `reduce`, why not use `obj` instead of declaring the new variable `o`?

Aaronius: Can we use a more descriptive name than `ts` as well?

Aaronius: Is there a more descriptive name for `obj` and `pair`?

src/components/Identity/processIdSyncs.js
```diff
@@ -35,54 +38,60 @@ const setControlObject = obj => {
 
   Object.keys(obj).forEach(id => arr.push(`${id}-${obj[id]}`));
```
Aaronius: Use `Object.keys(obj).map` here.

src/components/Identity/processIdSyncs.js
```diff
@@ -35,54 +38,60 @@ const setControlObject = obj => {
...
 
-  cookie.set(`${namespace}idSyncControl`, arr.join("_"), {
-    expires: 6 * 30 // 6 months
-  });
+  alloyCookie.set(ID_SYNC_CONTROL, arr.join("_"));
```
Aaronius: Are we using hyphens and underscores rather than arrays or objects for optimization purposes? I would typically propose using arrays and objects, but I don't know enough about the context. Is there any risk of `id` or `obj[id]` having a hyphen or underscore in them?

src/components/Identity/processIdSyncs.js
```diff
@@ -35,54 +38,60 @@ const setControlObject = obj => {
...
+    }).then(result => {
+      const timeStamp = Math.round(
+        new Date().getTime() / MILLISECONDS_PER_HOUR
+      ); // hours
+
+      result.succeeded.forEach(idSync => {
+        const ttl = (idSync.ttl || 7) * 24; // hours
+
+        if (idSync.id !== undefined) {
+          controlObject[idSync.id] = timeStamp + ttl;
```
Aaronius: In one of the recent standups, we decided that when saving numbers into cookies from our various components, we were going to use a base radix of 36 for optimized storage. In other words, take the result of `timeStamp + ttl` and call `toString(36)` and then when you use the timestamp from the cookie, call `parseInt(myNumber, 36)`.

src/constants/times.js
```diff
@@ -0,0 +1,14 @@
...
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+export const MILLISECONDS_PER_HOUR = 60 * 60 * 1000;
+export const SEVEN_DAYS_IN_HOURS = 7 * 24;
```
Aaronius: I like the general idea here, but I feel like it would be much more flexible/reusable as a utility called `convertTime.js` that looks like this:

```
export const MILLISECOND = 1;
export const SECOND = MILLISECOND * 1000;
export const MINUTE = SECOND * 60;
export const HOUR = MINUTE * 60;
export const DAY = HOUR * 24;
export const WEEK = DAY * 7;
// etc...

export default (fromUnit, toUnit, amount) => (fromUnit * amount) / toUnit;
};
```

src/components/Identity/processIdSyncs.js
wgamboa: Partly for optimization, partly because that's how we had it in the previous libraries. There is no risk.

test/unit/specs/components/Identity/processIdSyncs.spec.js
```diff
@@ -42,7 +42,9 @@ describe("Identity::processIdSyncs", () => {
 
     cookie.set(
       ID_SYNC_CONTROL,
-      `123-${Math.round(new Date().getTime() / 1000 / 60 / 60) - 10}`
+      `123-${(Math.round(new Date().getTime() / 1000 / 60 / 60) - 10).toString(
```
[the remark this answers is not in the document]
Aaronius: I just realized this was a spec. Definitely feel free to do what you want here. I'm okay either way.

wgamboa:
Thank you for all the suggestions, @Aaronius !

src/components/Identity/processIdSyncs.js
```diff
@@ -84,8 +81,10 @@ export default ({ destinations, config, logger, cookie }) => {
 
       cookie.set(
         ID_SYNC_TIMESTAMP,
-        Math.round(new Date().getTime() / MILLISECONDS_PER_HOUR) +
-          SEVEN_DAYS_IN_HOURS
+        (
+          Math.round(convertTimes(MILLISECOND, HOUR, new Date().getTime())) +
+          convertTimes(DAY, HOUR, 7)
```
Aaronius: Good call on this.

wgamboa: Thanks.

PR #128, opened 2019-07-29
CORE-32550 Allow onBeforeEvent run before opt-in.

Aaronius:
## Description
We need to capture context data at the time the customer attempted to run an `event` command, not when the data is sent to the server after opt-in is received. In order to accomplish this, the data collector component needs to run the `onBeforeEvent` lifecycle method _before_ waiting for opt-in.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-32550

## Motivation and Context
Users like data that was captured at the time they attempted to execute the command.

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Identity/index.js
+++ b/src/components/Identity/index.js
@@ -62,7 +62,7 @@ const createIdentity = ({ config, logger, cookie }) => {
             logger.log("Delaying request while retrieving ECID from server.");
             promise = deferredForEcid.promise.then(() => {
               logger.log("Resuming previously delayed request.");
-              addIdsContext(payload, ecid);
+              addIdsContext(payload, getEcid());
             });
           } else {
             // We don't have an ECID and no request has gone out to fetch it.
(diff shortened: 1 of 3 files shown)


src/components/Identity/index.js
```diff
@@ -62,7 +62,7 @@ const createIdentity = ({ config, logger, cookie }) => {
             logger.log("Delaying request while retrieving ECID from server.");
             promise = deferredForEcid.promise.then(() => {
               logger.log("Resuming previously delayed request.");
-              addIdsContext(payload, ecid);
+              addIdsContext(payload, getEcid());
```
Aaronius: We had a bug where this line was referring to the `ecid` variable, which was `undefined` because we didn't have an ECID when that variable was declared. Instead, we need to pull the ECID anew now that we know an ECID has been retrieved. I wrote a functional test case in TestRails that should cover this.

PR #129, opened 2019-07-30
Feature/core 26064

jileeshs:
## Description

Allowing customers to send declared IDs to experience platform.
## Related Issue

https://jira.corp.adobe.com/browse/CORE-26064

## Motivation and Context

Alloy should support this feature in the existing VisitorJS library.

## Types of changes

## Checklist:

### DIFF
--- a/src/components/DataCollector/index.js
+++ b/src/components/DataCollector/index.js
@@ -30,10 +30,8 @@ const createDataCollector = ({ config }) => {
       }
     });
 
-    const responsePromise = Promise.resolve()
-      .then(() => {
-        return lifecycle.onBeforeDataCollection(payload, responsePromise);
-      })
+    return lifecycle
+      .onBeforeDataCollection(payload)
       .then(() => {
         return network.sendRequest(payload, payload.expectsResponse);
       })
@@ -48,7 +46,6 @@ const createDataCollector = ({ config }) => {
 
         return data;
       });
-    return responsePromise;
   };
 
   const createEventHandler = options => {
--- a/src/components/Identity/index.js
+++ b/src/components/Identity/index.js
@@ -10,14 +10,22 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import { defer, flatMap, convertTimes } from "../../utils";
+import { defer, assign, flatMap, crc32, convertTimes } from "../../utils";
 import processIdSyncs from "./processIdSyncs";
+import createEvent from "../DataCollector/createEvent";
+import { validateCustomerIds, normalizeCustomerIds } from "./util";
 import { HOUR, MILLISECOND } from "../../utils/convertTimes";
-import { ECID_NAMESPACE, ID_SYNC_TIMESTAMP } from "./constants";
+import { COOKIE_NAMES } from "./constants";
+
+const {
+  CUSTOMER_ID_HASH,
+  EXPERIENCE_CLOUD_ID,
+  ID_SYNC_TIMESTAMP
+} = COOKIE_NAMES;
 
 const addIdsContext = (payload, ecid) => {
   // TODO: Add customer ids.
-  payload.addIdentity(ECID_NAMESPACE, {
+  payload.addIdentity(EXPERIENCE_CLOUD_ID, {
     id: ecid
   });
 };
@@ -25,14 +33,51 @@ const addIdsContext = (payload, ecid) => {
 const createIdentity = ({ config, logger, cookie }) => {
   // We avoid reading the ECID from the cookie right away, because we
   // need to wait for the user to opt in first.
-  const getEcid = () => cookie.get(ECID_NAMESPACE);
+  const getEcid = () => cookie.get(EXPERIENCE_CLOUD_ID);
   let optIn;
   let deferredForEcid;
+  let network;
+  let lifecycle;
+  const customerIds = {};
+
+  const makeServerCall = payload => {
+    return lifecycle.onBeforeDataCollection(payload).then(() => {
+      return network.sendRequest(payload, payload.expectsResponse);
+    });
+  };
+
+  const setCustomerIds = ids => {
+    validateCustomerIds(ids);
+    const event = createEvent();
+    event.mergeData(ids.data);
+    const payload = network.createPayload();
+    payload.addEvent(event);
+    assign(customerIds, ids);
+
+    const normalizedCustomerIds = normalizeCustomerIds(customerIds);
+
+    Object.keys(normalizedCustomerIds).forEach(idName => {
+      payload.addIdentity(idName, normalizedCustomerIds[idName]);
+    });
+    const customerIdsHash = crc32(
+      JSON.stringify(normalizedCustomerIds)
+    ).toString(36);
+    // TODO: add more tests around this piece
+    const customerIdChanged = customerIdsHash !== cookie.get(CUSTOMER_ID_HASH);
+    payload.mergeMeta({ identityMap: { customerIdChanged } });
+    if (customerIdChanged) {
+      cookie.set(CUSTOMER_ID_HASH, customerIdsHash);
+    }
+    return lifecycle
+      .onBeforeEvent(event, normalizedCustomerIds, customerIdChanged)
+      .then(() => optIn.whenOptedIn())
+      .then(() => makeServerCall(payload));
+  };
 
   return {
     lifecycle: {
       onComponentsRegistered(tools) {
-        ({ optIn } = tools);
+        ({ lifecycle, network, optIn } = tools);
       },
       // Waiting for opt-in because we'll be reading the ECID from a cookie
       onBeforeEvent(event) {
@@ -87,7 +132,7 @@ const createIdentity = ({ config, logger, cookie }) => {
           const ecidPayloads = response.getPayloadsByType("identity:persist");
 
           if (ecidPayloads.length > 0) {
-            cookie.set(ECID_NAMESPACE, ecidPayloads[0].id);
+            cookie.set(EXPERIENCE_CLOUD_ID, ecidPayloads[0].id);
 
             if (deferredForEcid) {
               deferredForEcid.resolve();
@@ -111,6 +156,10 @@ const createIdentity = ({ config, logger, cookie }) => {
     commands: {
       getEcid() {
         return optIn.whenOptedIn().then(getEcid);
+      },
+      // TODO: Discuss renaming of CustomerIds to UserIds
+      setCustomerIds(options) {
+        return optIn.whenOptedIn().then(() => setCustomerIds(options));
       }
     }
   };
--- a/src/components/Identity/util.js
+++ b/src/components/Identity/util.js
@@ -1,0 +1,46 @@
+import { AUTH_STATES } from "./constants";
+import { isObject, values, includes } from "../../utils";
+
+const ERROR_MESSAGE = "Invalid customer ID format.";
+const NOT_AN_OBJECT_ERROR = "Each namespace should be an object.";
+const NO_ID_ERROR = "Each namespace object should have an ID.";
+
+const validateCustomerIds = customerIds => {
+  if (!isObject(customerIds)) {
+    throw new Error(`${ERROR_MESSAGE} ${NOT_AN_OBJECT_ERROR}`);
+  }
+  Object.keys(customerIds).forEach(customerId => {
+    if (!isObject(customerIds[customerId])) {
+      throw new Error(`${ERROR_MESSAGE} ${NOT_AN_OBJECT_ERROR}`);
+    }
+    if (!customerIds[customerId].id) {
+      throw new Error(`${ERROR_MESSAGE} ${NO_ID_ERROR}`);
+    }
+  });
+};
+
+const sortObjectKeyNames = object => {
+  return Object.keys(object)
+    .sort()
+    .reduce((newObject, key) => {
+      newObject[key] = object[key];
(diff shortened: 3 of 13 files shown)


src/components/Identity/index.js
```diff
@@ -10,30 +10,80 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
   // We avoid reading the ECID from the cookie right away, because we
   // need to wait for the user to opt in first.
-  const getEcid = () => cookie.get(ECID_NAMESPACE);
+  const getEcid = () => cookie.get(EXPERIENCE_CLOUD_ID);
   let optIn;
   let responseRequested = false;
   let deferredForEcid;
+  let network;
+  let lifecycle;
+  const customerIds = {};
```
jfkhoury: To make sure we don't lose @Aaronius 's comment about the name:

Let's add a TODO here to discuss renaming `customer` to `user` with Product. It might be too late because it's a known construct within Adobe and our customers, but let's try for sure.

@jileeshs

jfkhoury:
Hey @jileeshs, are you done making changes / force pushing, so we can start reviewing?

src/components/Identity/index.js
jileeshs: > Hey @jileeshs, are you done making changes / force pushing, so we can start reviewing?

@jfkhoury ,
Yes I am done making changes. 
@jfkhoury , @Aaronius , @jonsnyder  please review

jileeshs: > To make sure we don't lose @Aaronius 's comment about the name:
> 
> Let's add a TODO here to discuss renaming `customer` to `user` with Product. It might be too late because it's a known construct within Adobe and our customers, but let's try for sure.
> 
> @jileeshs

Added a TODO commend near the exposed command names

src/components/Identity/util.js
```diff
@@ -0,0 +1,79 @@
...
+    result.addEventListener("complete", () => {
+      resolve(result.result);
+    });
+    result.addEventListener("error", () => {
+      reject();
+    });
+  });
+};
+
+export { validateCustomerIds, normalizeCustomerIds, bufferToHex, createHash };
```
Aaronius: This is still using sha-256. Please see https://github.com/adobe/alloy/pull/114#discussion_r306970665 and https://github.com/adobe/alloy/pull/114#discussion_r307103432

test/unit/components/Identity/util.spec.js
```diff
@@ -0,0 +1,175 @@
...
+      const stringifiedNormalizedTwo = JSON.stringify(
+        normalizeCustomerIds(customerIdTwo)
+      );
+      expect(stringifiedNormalizedOne).toBe(stringifiedNormalizedTwo);
+
+      createHash(stringifiedNormalizedOne).then(resultOne => {
+        createHash(stringifiedNormalizedTwo).then(resultTwo => {
+          expect(bufferToHex(resultOne)).toBe(bufferToHex(resultTwo));
+        });
+      });
```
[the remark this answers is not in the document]
Aaronius: Better yet, use `Promise.all`:

```
Promise.all([
  createHash(stringifiedNormalizedOne),
  createHash(stringifiedNormalizedTwo);
])..then([resultOne, resultTwo] => {
  expect(bufferToHex(resultOne)).toBe(bufferToHex(resultTwo));
});
```

I don't think you'll need to use promises at all once you move to crc32.

src/components/Identity/index.js
```diff
@@ -10,30 +10,80 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+        return network.sendRequest(payload, false);
+      });
+  };
+
+  const setCustomerIds = ids => {
+    validateCustomerIds(ids);
+    const event = createEvent();
+    const payload = network.createPayload();
+
+    assign(customerIds, ids);
```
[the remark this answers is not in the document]
jileeshs: `customerIds` are not stored in the cookie.
We store only a hash of all `customerIds with authState`.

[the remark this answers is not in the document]
Aaronius: Oh yeah, I forgot. Should it retrieve the customer IDs from the server then (I don't even know if this is a thing)? I'm still trying to figure out the use case, especially if it only reflects what was recently passed in via `setCustomerIds`.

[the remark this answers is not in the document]
Aaronius: Sorry..I'm referring to the use case for the `getCustomerIds` command here: ^

src/components/Identity/util.js
```diff
@@ -0,0 +1,79 @@
...
+  }
+  Object.keys(customerIds).forEach(customerId => {
+    if (!isObject(customerIds[customerId])) {
+      throw new Error(`${ERROR_MESSAGE} ${NOT_AN_OBJECT_ERROR}`);
+    }
+    if (!customerIds[customerId].id) {
+      throw new Error(`${ERROR_MESSAGE} ${NO_ID_ERROR}`);
+    }
+  });
+};
```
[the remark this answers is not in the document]
jileeshs: authState is optional field and if it is not from the enumeration values, it has a default.

src/components/Identity/index.js
Aaronius: @jonsnyder made an interesting comment [here](https://github.com/adobe/alloy/pull/114#discussion_r309428045) that was from a different perspective than mine. When approached from his perspective, "customer" seems sufficiently appropriate.

src/components/Identity/index.js
```diff
@@ -10,30 +10,80 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+    const normalizedCustomerIds = normalizeCustomerIds(customerIds);
+
+    Object.keys(normalizedCustomerIds).forEach(idName => {
+      payload.addIdentity(idName, normalizedCustomerIds[idName]);
+    });
+
+    createHash(JSON.stringify(normalizedCustomerIds)).then(hash => {
+      const customerIdsHash = bufferToHex(hash);
+      const customerIdChanged =
+        customerIdsHash === cookie.get(CUSTOMER_ID_HASH);
```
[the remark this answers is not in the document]
jileeshs: You are right.
Previously it was `hasSynced` and the expression was according to that.
I will e adding more tests in next PR

src/components/Identity/index.js
```diff
@@ -10,30 +10,80 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+  const getCustomerIds = () => customerIds;
+
+  const makeServerCall = payload => {
+    return Promise.resolve()
+      .then(() => {
+        return lifecycle.onBeforeDataCollection(payload);
+      })
+      .then(() => {
+        return network.sendRequest(payload, false);
+      });
```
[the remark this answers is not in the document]
jileeshs: Done. Thank You!

src/components/Identity/index.js
```diff
@@ -100,6 +150,13 @@ const createIdentity = ({ config, logger, cookie }) => {
     commands: {
       getEcid() {
         return optIn.whenOptedIn().then(getEcid);
+      },
+      // TODO: Discuss renaming of CustomerIds to UserIds
+      setCustomerIds(options) {
+        return optIn.whenOptedIn().then(() => setCustomerIds(options));
+      },
+      getCustomerIds() {
+        return optIn.whenOptedIn().then(getCustomerIds);
```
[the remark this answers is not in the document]
jileeshs: Removed getCustomerIds.

src/components/Identity/index.js
```diff
@@ -10,30 +10,80 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+    Object.keys(normalizedCustomerIds).forEach(idName => {
+      payload.addIdentity(idName, normalizedCustomerIds[idName]);
+    });
+
+    createHash(JSON.stringify(normalizedCustomerIds)).then(hash => {
+      const customerIdsHash = bufferToHex(hash);
+      const customerIdChanged =
+        customerIdsHash === cookie.get(CUSTOMER_ID_HASH);
+      payload.mergeMeta({ identity: { customerIdChanged } });
+      if (!customerIdChanged) {
```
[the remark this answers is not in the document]
jileeshs: Yeah, again made sense when it was `hasSynced`. Forgot to change it when the flag changed to `customerIdChanged`. Corrected it. Thank You for spotting it.

src/components/Identity/index.js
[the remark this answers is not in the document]
jileeshs: Removed `getCustomerIds` as we discussed

src/components/Identity/util.js
jileeshs: Added a new `crc32` in utils and using it here

test/unit/components/Identity/util.spec.js
```diff
@@ -0,0 +1,175 @@
...
+      expect(() => {
+        validateCustomerIds(objToTest);
+      }).toThrow(
+        new Error(
+          "Invalid customer ID format. Each namespace should be an object."
+        )
+      );
+    });
+
+    it("should throw an error when any of teh namespace doesn't have an id parameter", () => {
```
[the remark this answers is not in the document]
jileeshs: Fixed. Thank You!

Aaronius:
You have crc32 tests failing in IE because you're using `fill`.

src/utils/crc32.js
```diff
@@ -0,0 +1,24 @@
...
+  cleanString
+    .split("")
+    .map(char => char.charCodeAt(0))
+    .forEach(char => {
+      const y = (crc ^ char) & 0xff;
+      crc = (crc >>> 8) ^ buffer[y];
+    });
+  crc = crc ^ -1;
+  return (crc >>> 0).toString(16);
+};
```
Aaronius: Where did you get this code? Is there any reason you went with it instead of http://crc32.nichabi.com/javascript-function.php?

jileeshs: I got this code from there.
I made a few changes because of lint errors.
Added `.toString(16)` because the function was returning a decimal number in JS, and every other crc32 example was returning a `hex`.
I compared the out put of many strings with available crc32 online examples

jileeshs: I removed fill and added a loop as it was before.

test/unit/utils/crc32.spec.js
```diff
@@ -17,15 +17,15 @@ describe("crc32", () => {
     };
     const resultOne = crc32(JSON.stringify(idsTohash));
     const resultTwo = crc32(JSON.stringify(idsTohash));
-    expect(typeof resultOne).toBe("string");
-    expect(resultOne).toBe("bc420422");
-    expect(resultTwo).toBe("bc420422");
+    expect(typeof resultOne).toBe("number");
```
Aaronius: You don't have to change your code, but as an FYI, you can also do:

```
expect(resultOne).toEqual(jasmine.any(Number))
```

The error that shows up if it doesn't pass the assertion might be a little more friendly (I can't recall what it actually says), but other than that I think it's probably subjective which one is better.

src/components/Identity/index.js
```diff
@@ -10,30 +10,67 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
   let responseRequested = false;
   let deferredForEcid;
+  let network;
+  let lifecycle;
+  const customerIds = {};
+
+  const makeServerCall = payload => {
+    return lifecycle
+      .onBeforeDataCollection(payload)
+      .then(() => network.sendRequest(payload, false));
```
jileeshs: @Aaronius ,
This line should be 
```
return Promise.resolve()
    .then(() => {
      return lifecycle.onBeforeDataCollection(payload);
    })
    .then(() => {
      return network.sendRequest(payload, false);
    });
```

Aaronius: Why?

jileeshs: Without this wrapping the the command is not waiting for first interact call to get an ECID

Aaronius: It sounds like we're basically getting lucky with using `Promise.resolve` here then and not being intentional in our strategy of sequencing. We need to think this through in more depth. I'll try to dedicate some time to figuring out how to properly solve the problem.

Aaronius: @jileeshs I think your problem might be this line: https://github.com/adobe/alloy/pull/129/files#diff-c626ca482d730eb64b741a9d718607adR102 and how it's using the `ecid` variable that was initialized `before` the request for the ECID occurred, and therefore it is undefined. Compare this with the `master` branch, which retrieves the `ecid` again instead.

If this doesn't fix the issue, let me know and I might need you to show me the problem.

jileeshs: That didn't work.

src/components/DataCollector/index.js
```diff
@@ -31,10 +31,8 @@ const createDataCollector = () => {
       }
     });
 
-    const responsePromise = Promise.resolve()
-      .then(() => {
-        return lifecycle.onBeforeDataCollection(payload, responsePromise);
-      })
+    return lifecycle
+      .onBeforeDataCollection(payload)
```
Aaronius: @jonsnyder This is a change you should be aware of. It would take my quite a while to write about why it was needed and I probably won't communicate it clearly, but let's talk with our mouth holes when you get back. If we need to add it back in, we'll need to do some additional thinking about how to work out some timing things.

jonsnyder: I put that in there so that components could get the matching response without having to match up requests and responses in onResponse.  We are not using it, so it's okay to remove.

Aaronius:
@mandody are you waiting for me on any of this?

Aaronius:
...or @jileeshs

PR #130, opened 2019-07-30
AN-32319 use isViewStart boolean instead of type

jonsnyder:
## Description

Change `type: "viewStart"` to `isViewStart: true` within the data collector options.

## Related Issue

https://jira.corp.adobe.com/browse/CORE-32319

## Motivation and Context

We made the decision that customer can set event types in their data, instead of top level on the event, especially that XDM already has an `eventType` property, with 20 built-in values.

Since the view start concept is important to Personalization, and probably few other components as well, we will convert the `type: viewStart` to a boolean: `isViewStart: true|false`. (false by default)

The components will still receive this boolean in the lifecycle hooks.

## Types of changes

## Checklist:

### DIFF
--- a/sandbox/src/Home.js
+++ b/sandbox/src/Home.js
@@ -8,8 +8,9 @@ function HomeWithHistory({ history }) {
     if (loc.pathname !== previousPath) {
       const instanceName = loc.pathname.includes("orgTwo") ? "organizationTwo" : "alloy";
       window[instanceName]("event", {
-        isViewStart: true,
+        viewStart: true,
         data: {
+          "eventType": "page-view",
           "url": window.location.href,
           "name": loc.pathname.substring(1)
         }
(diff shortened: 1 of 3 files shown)


sandbox/src/Home.js
```diff
@@ -8,7 +8,7 @@ function HomeWithHistory({ history }) {
     if (loc.pathname !== previousPath) {
       const instanceName = loc.pathname.includes("orgTwo") ? "organizationTwo" : "alloy";
       window[instanceName]("event", {
-        type: "viewStart",
+        isViewStart: true,
```
artur-ciocanu: I was wondering if we should use `viewStart: true`, instead of `isViewStart`. We had a discussion around NOT using options/configs that look like functions or method calls.

jfkhoury:
@jonsnyder Let's hold off on merging this guy, And updating docs until we meet this week to review the Alpha feedback please. Thanks.

Aaronius:
We've decided to move forward with `viewStart` and merge this asap, haven't we?

jonsnyder:
I'm working on it now.

Aaronius:
Excellent. After the name change, I approve, even though I can't officially approve the PR.

Aaronius:
Reminder to update the docs.

jonsnyder:
You're right. I'll add another PR.

PR #131, opened 2019-07-30
CORE-21547 Separated createCookieProxy for dependency injection.

Aaronius:
## Description
As discussed in https://github.com/adobe/alloy/pull/126#discussion_r308275505, I split out `createCookieProxy` from `createCookie`, primarily so we could eliminate the need for test-specific code and because the modules can be more effectively tested in isolation.

I'm planning on opening another pull request that eliminates the test-specific code from `getTopLevelDomain.js`.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-32658

## Motivation and Context
Eliminate test-specific code so that the code we test will exactly match what we see in production.

## Types of changes

## Checklist:

### DIFF
--- a/src/constants/cookieDetails.js
+++ b/src/constants/cookieDetails.js
@@ -12,5 +12,5 @@ governing permissions and limitations under the License.
 
 export default {
   ALLOY_COOKIE_NAME: "adobe_alloy", // TODO: Rename this cookie
-  ALLOY_COOKIE_EXPIRES: 180
+  ALLOY_COOKIE_TTL_IN_DAYS: 180
 };
--- a/src/constants/cookieDetails.js
+++ b/src/constants/cookieDetails.js
@@ -12,5 +12,5 @@ governing permissions and limitations under the License.
 
 export default {
   ALLOY_COOKIE_NAME: "adobe_alloy", // TODO: Rename this cookie
-  ALLOY_COOKIE_TTL: 180
+  ALLOY_COOKIE_EXPIRES: 180
 };
--- a/src/core/createCookie.js
+++ b/src/core/createCookie.js
@@ -9,125 +9,53 @@ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTA
 OF ANY KIND, either express or implied. See the License for the specific language
 governing permissions and limitations under the License.
 */
-import cookie from "../utils/cookie";
-import cookieDetails from "../constants/cookieDetails";
-import getTopLevelCookieDomain from "../utils/getTopLevelCookieDomain";
 
-const { ALLOY_COOKIE_NAME, ALLOY_COOKIE_TTL } = cookieDetails;
-const cookieProxyByPropertyId = {};
-
-const safeJSONParse = (object, cookieName) => {
-  try {
-    return JSON.parse(object);
-  } catch (error) {
-    throw new Error(`Invalid cookie format in ${cookieName} cookie`);
-  }
-};
-
-/**
- * The purpose of this proxy is to cache the cookie so we don't have to
- * read and deserialize it every time a piece of it is accessed.
- */
-const createCookieProxy = (propertyId, cookieDomain = "") => {
-  // Because the cookie proxy caches the deserialized cookie, we want to make
-  // sure there's only one cookie proxy created per cookie because multiple
-  // caches per cookies would result in caches getting stale.
-  let cookieProxy = cookieProxyByPropertyId[propertyId];
-
-  if (cookieProxy) {
-    return cookieProxy;
-  }
-
-  const cookieName = `${ALLOY_COOKIE_NAME}_${propertyId}`;
-  let deserializedCookie;
-  let cookieHasBeenRead = false;
-
-  cookieProxy = {
-    get() {
-      // We don't read the cookie right when the cookie proxy is created
-      // because we don't know if the user has opted in. If the user
-      // hasn't opted in, we're legally obligated to not read the cookie.
-      // If a component tries to read something off the cookie though,
-      // we assume that the component has received word that the user
-      // has opted-in. The responsibility is on the component.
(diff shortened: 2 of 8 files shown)


src/core/initializeComponentsFactory.js
```diff
@@ -10,20 +10,23 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import createLifecycle from "./createLifecycle";
-import createComponentRegistry from "./createComponentRegistry";
-import createNetwork from "./network";
 import { stackError } from "../utils";
-import createOptIn from "./createOptIn";
```
Aaronius: We were importing a bunch of things here, but I thought the general consensus was that we would only import non-utility modules from `main.js` and injected them all down the tree. As such, I moved the imports to `main.js`. This will allow for easier unit testing of `initializeComponentsFactory.js` should we choose to do so.

src/constants/cookieDetails.js
```diff
@@ -12,5 +12,5 @@ governing permissions and limitations under the License.
 
 export default {
   ALLOY_COOKIE_NAME: "adobe_alloy", // TODO: Rename this cookie
-  ALLOY_COOKIE_TTL: 180
+  ALLOY_COOKIE_EXPIRES: 180
```
jfkhoury: Can you please change the name to specify the type? is `180` in days?
I already have a ticket for it, so 2 birds, 1 PR: CORE-32680

src/core/createCookie.js
```diff
@@ -9,125 +9,53 @@ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTA
...
-  cookieProxyByPropertyId[propertyId] = cookieProxy;
-
-  return cookieProxy;
-};
-
-// TODO: Support passing a configurable expiry in the config when creating this cookie.
-const createCookie = (componentNamespace, propertyId, cookieDomain = "") => {
-  const cookieProxy = createCookieProxy(propertyId, cookieDomain);
 
+export default (cookieProxy, componentNamespace) => {
```
jfkhoury: Love DI!

src/core/createCookie.js
```diff
@@ -9,125 +9,53 @@ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTA
...
-      // it's reasonably controlled since the side effects are all contained
-      // within this file.
-      currentCookie[componentNamespace] =
-        currentCookie[componentNamespace] || {};
-      currentCookie[componentNamespace][key] = value;
-      cookieProxy.set(currentCookie);
+      const updatedCookie = {
+        ...currentCookie,
+        [componentNamespace]: {
+          ...currentCookie[componentNamespace],
```
jfkhoury: Will this pass in IE 11?

src/constants/cookieDetails.js
Aaronius: Yeah, this ran across my mind when I was changing it and now I regret not doing it. I renamed it to match the `expires` option of `cookie-js`, but I agree it should have the unit.

src/core/createCookie.js
Aaronius: Yes. Babel's object spreading output doesn't try to use symbols or Array.from in IE.

src/core/initializeComponentsFactory.js
```diff
@@ -10,20 +10,31 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+  createCookie,
+  createLifecycle,
+  createComponentRegistry,
+  createNetwork,
+  createOptIn
 ) => config => {
   const componentRegistry = createComponentRegistry();
   const { imsOrgId, propertyId, cookieDomain } = config;
+  const cookieName = `${ALLOY_COOKIE_NAME}_${propertyId}`;
+  const cookieProxy = createCookieProxy(
```
jonsnyder: If we ever support multiple instances with the same propertyId, we could memoize the createCookieProxy method so that if two cookies are fetched with the same name and domain, it would actually return the same instance.

src/core/createCookie.js
jonsnyder: These names are somewhat confusing.  Having a cookie object and a cookieProxy object, I would assume the cookieProxy wraps the cookie Object.  Perhaps a way to solve this potential confusion is to rename the cookie object to `namespacedCookie`.  Indeed, the main purpose of the createCookie function is to add namespace functionality to the cookieProxy.  This would align nicely with the `createNamespacedStorage` function as well.

test/unit/specs/core/createCookieProxy.spec.js
```diff
@@ -0,0 +1,77 @@
...
+};
+
+const removeAllCookies = () => {
+  const cookies = document.cookie.split(";");
+  cookies.forEach(cookieName => {
+    removeCookie(cookieName.split("=")[0]);
+  });
+};
+
+describe("createCookie", () => {
```
jonsnyder: should describe `createCookieProxy`

src/core/initializeComponentsFactory.js
jonsnyder: Making these dependencies explicit is great.  This highlights that initializeComponentsFactory is probably doing too much when we are passing in 9 dependencies.  This is beyond the scope of this PR, but I have some ideas on how to simplify this.

test/unit/specs/core/createCookieProxy.spec.js
Aaronius: Fixed this is a later commit. Sorry for pushing commits while you were reviewing.

src/core/createCookie.js
Aaronius: Yeah, I like that. I'll make the change.

Aaronius: Actually, I'm going to do this in a separate PR. Thanks for the feedback.

PR #132, opened 2019-07-31
Update getPayload(s)ByType to include all matching payloads

jonsnyder:
## Description

rename `getPayloadByType` to `getPayload*s*ByType`.  Update components that use getPayloadByType to loop through responses.

## Related Issue

https://jira.corp.adobe.com/browse/CORE-32538

## Motivation and Context

Response can have multiple handles with the same type

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Audiences/index.js
+++ b/src/components/Audiences/index.js
@@ -11,6 +11,7 @@ governing permissions and limitations under the License.
 */
 
 import processDestinations from "./processDestinations";
+import { flatMap } from "../../utils";
 
 const createAudiences = ({ config, logger }) => {
   return {
@@ -21,7 +22,10 @@ const createAudiences = ({ config, logger }) => {
         }
       },
       onResponse(response) {
-        const destinations = response.getPayloadByType("activation:push") || [];
+        const destinations = flatMap(
+          response.getPayloadsByType("activation:push"),
+          payloads => payloads
+        );
 
         processDestinations({
           destinations,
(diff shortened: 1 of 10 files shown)


src/utils/flatMap.js
```diff
@@ -0,0 +1,3 @@
+export default (array, mapFunction) => {
+  return Array.prototype.concat.apply([], array.map(mapFunction));
```
Aaronius: Why not `return [].concat(array.map(mapFunction));`?

src/components/Audiences/index.js
```diff
@@ -21,7 +22,10 @@ const createAudiences = ({ config, logger }) => {
         }
       },
       onResponse(response) {
-        const destinations = response.getPayloadByType("activation:push") || [];
+        const destinations = flatMap(
+          response.getPayloadsByType("activation:push"),
+          payloads => payloads
+        );
```
Aaronius: Both of these `flatMap` usages seem like `flat` would be more appropriate, since the mapping is unnecessary. Did you choose not to implement `flat` because you would need to support a depth of greater than `1` to align with the JavaScript spec?

src/utils/flatMap.js
Aaronius: Nevermind, that was a dumb question. We could do `return [].concat(...array.map(mapFunction));`, which would be a little shorter in es6 world, but babel would transpile it to something more complicated and expensive. I like yours.

src/components/Audiences/index.js
jonsnyder: Exactly.  I suppose we could create `flatten` or `flat1`, but I thought flatmap would be more understandable and standard.

Aaronius: Sounds good. I'm not really opinionated on the matter.

PR #133, opened 2019-07-31
AN-32729 use smaller names for cookie namespaces

jonsnyder:
## Description

I added an optional property of the component function called `cookieNamespace`.  These are used as the first level keys for the cookie's JSON.  I considered just shortening namespace values, but long namespaces are useful especially in the logs.  If `cookieNamespace` is undefined, it uses the `namespace` property.  Some of the components never use cookies so I didn't bother defining a `cookieNamespace` property for them.

## Related Issue

https://jira.corp.adobe.com/browse/CORE-32729

## Motivation and Context

Customers are sensitive to the size of cookies, so we should make them smaller if possible.

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Identity/index.js
+++ b/src/components/Identity/index.js
@@ -109,7 +109,7 @@ const createIdentity = ({ config, logger, cookie }) => {
 };
 
 createIdentity.namespace = "Identity";
-createIdentity.cookieNamespace = "I";
+createIdentity.abbreviation = "ID";
 
 createIdentity.configValidators = {
   idSyncsEnabled: {
--- a/src/components/Identity/index.js
+++ b/src/components/Identity/index.js
@@ -109,6 +109,7 @@ const createIdentity = ({ config, logger, cookie }) => {
 };
 
 createIdentity.namespace = "Identity";
+createIdentity.cookieNamespace = "I";
 
 createIdentity.configValidators = {
   idSyncsEnabled: {
--- a/src/components/Privacy/index.js
+++ b/src/components/Privacy/index.js
@@ -60,7 +60,7 @@ const createPrivacy = ({ config, logger, enableOptIn, cookie }) => {
 };
 
 createPrivacy.namespace = "Privacy";
-createPrivacy.cookieNamespace = "C"; // C for consent
+createPrivacy.abbreviation = "PR";
 
 createPrivacy.configValidators = {
   optInEnabled: {
--- a/src/components/Privacy/index.js
+++ b/src/components/Privacy/index.js
@@ -60,6 +60,7 @@ const createPrivacy = ({ config, logger, enableOptIn, cookie }) => {
 };
 
 createPrivacy.namespace = "Privacy";
+createPrivacy.cookieNamespace = "C"; // C for consent
 
 createPrivacy.configValidators = {
   optInEnabled: {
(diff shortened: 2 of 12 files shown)


src/components/Identity/index.js
```diff
@@ -106,6 +106,7 @@ const createIdentity = ({ config, logger, cookie }) => {
 };
 
 createIdentity.namespace = "Identity";
+createIdentity.cookieNamespace = "I";
```
Aaronius: It might be good to rename `cookieNamespace` to `abbreviation` so it could potentially be used in other contexts. 🤷‍♂ We can rename it later if needed.

src/components/Privacy/index.js
```diff
@@ -60,6 +60,7 @@ const createPrivacy = ({ config, logger, enableOptIn, cookie }) => {
 };
 
 createPrivacy.namespace = "Privacy";
+createPrivacy.cookieNamespace = "C"; // C for consent
```
Aaronius: I think I would go with two characters for each of these so they're more identifiable and hopefully avoid duplicates (e.g., `ID`, `PE`, `PR`). I understand this is subjective though.

src/core/initializeComponentsFactory.js
```diff
@@ -45,13 +45,13 @@ export default (
...
   componentCreators.forEach(createComponent => {
-    const { namespace } = createComponent;
+    const { namespace, cookieNamespace } = createComponent;
     // TO-DOCUMENT: Helpers that we inject into factories.
     let component;
     try {
       component = createComponent({
         logger: logger.spawn(`[${namespace}]`),
-        cookie: createCookie(cookieProxy, namespace),
+        cookie: createCookie(cookieProxy, cookieNamespace || namespace),
```
Aaronius: I'm a bit concerned that one of the components that doesn't have `cookieNamespace` defined (or some new component) will start using the cookie utility without the developer realizing they need to define `cookieNamespace` and we won't notice until the library in prod and we'll be locked into using `namespace` for that component forever instead of `cookieNamespace`. We could throw an error here or in createCookie but I think it would also be reasonable to just add `cookieNamespace` to all the components right now, which I believe would greatly reduce the chance of this happening.

jfkhoury:
I agree with all of Aaron’s comments

src/core/createCookie.js
```diff
@@ -10,13 +10,24 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+    }
+  };
+
   return {
     /**
      * Returns a value from the Alloy cookie for a given key.
      * @param {string} key
      */
     get(key) {
+      validateNamespace();
```
Aaronius: Is there a reason why we don't just validate it once before returning the cookie object instead of on every `get`, `set`, and `remove`?

jonsnyder: I didn't want to throw an error for a component that doesn't use the cookie.

Aaronius: In my opinion, it isn't worth it. If `componentNamespace` isn't defined, I'd like to hear about it ASAP regardless of whether the cookie is used. In any case, if you choose to keep it, maybe add a comment for future devs because this is something I would likely change if I didn't understand your intent.

jonsnyder: I don't have a strong opinion either way, so I have made your suggested change.

PR #134, opened 2019-07-31
CORE-32836 Remove test-specific code from getTopLevelCookieDomain.

Aaronius:
## Description
Remove test-specific code from getTopLevelCookieDomain.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-32836

## Motivation and Context
Keep code under test as closely aligned with production code as possible.

## Types of changes

## Checklist:

### DIFF
--- a/src/core/createCookieProxy.js
+++ b/src/core/createCookieProxy.js
@@ -12,6 +12,9 @@ governing permissions and limitations under the License.
 
 import cookie from "../utils/cookie";
 import getTopLevelCookieDomain from "../utils/getTopLevelCookieDomain";
+import memoize from "../utils/memoize";
+
+const memoizedGetTopLevelCookieDomain = memoize(getTopLevelCookieDomain);
 
 const safeJSONParse = (object, cookieName) => {
   try {
@@ -52,7 +55,7 @@ export default (name, expires, domain = "") => {
       deserializedCookie = updatedCookie;
       cookie.set(name, updatedCookie, {
         expires,
-        domain: domain || getTopLevelCookieDomain(window, cookie)
+        domain: domain || memoizedGetTopLevelCookieDomain(window, cookie)
       });
     }
   };
--- a/src/core/createCookieProxy.js
+++ b/src/core/createCookieProxy.js
@@ -11,10 +11,6 @@ governing permissions and limitations under the License.
 */
 
 import cookie from "../utils/cookie";
-import getTopLevelCookieDomain from "../utils/getTopLevelCookieDomain";
-import memoize from "../utils/memoize";
-
-const memoizedGetTopLevelCookieDomain = memoize(getTopLevelCookieDomain);
 
 const safeJSONParse = (object, cookieName) => {
   try {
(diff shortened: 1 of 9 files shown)


src/core/createCookieProxy.js
```diff
@@ -52,7 +55,7 @@ export default (name, expires, domain = "") => {
       deserializedCookie = updatedCookie;
       cookie.set(name, updatedCookie, {
         expires,
-        domain: domain || getTopLevelCookieDomain(window, cookie)
+        domain: domain || memoizedGetTopLevelCookieDomain(window, cookie)
```
Aaronius: This is a simple enough use case that memoize probably isn't necessary, but I thought it would be nice anyway and I imagine we'll end up using the util elsewhere.

src/core/createCookieProxy.js
```diff
@@ -12,6 +12,9 @@ governing permissions and limitations under the License.
 
 import cookie from "../utils/cookie";
 import getTopLevelCookieDomain from "../utils/getTopLevelCookieDomain";
+import memoize from "../utils/memoize";
+
+const memoizedGetTopLevelCookieDomain = memoize(getTopLevelCookieDomain);
```
Aaronius: We could create `memoizedGetTopLevelCookieDomain` in a few different places:

1. Right here like we are now. This is really only useful if we only end up using it in this module.
2. We could create a utility called `memoizedGetTopLevelCookieDomain.js` that would live beside `getTopLevelCookieDomain.js`.
3. We could create `memoizedGetTopLevelCookieDomain` up in `main.js` and pass it into `createCookieProxy.js`, and later pass it down to all components if they end up needing it.

As I think about this again, I think I'm going to add a commit that changes it to number 3.

Aaronius:
Thanks for the review @jonsnyder. Sorry to have you do this, but do you mind taking a look at the commit I just added?

src/utils/memoize.js
```diff
@@ -0,0 +1,37 @@
...
+ * Creates a function that memoizes the result of `fn`. If `keyResolver` is
+ * provided, it determines the cache key for storing the result based on the
+ * arguments provided to the memoized function. By default, the first argument
+ * provided to the memoized function is used as the map cache key.
+ *
+ * @param {Function} fn The function to have its output memoized.
+ * @param {Function} [keyResolver] The function to resolve the cache key.
+ * @returns {Function} The new memoized function.
+ */
+export default (fn, keyResolver) => {
```
jfkhoury: Sweet! We should use this little guy in the activity collector to memoize parsing URLs

jfkhoury:
I'm restarting the build.

src/core/createCookieProxy.js
```diff
@@ -55,7 +51,7 @@ export default (name, expires, domain = "") => {
       deserializedCookie = updatedCookie;
       cookie.set(name, updatedCookie, {
         expires,
-        domain: domain || memoizedGetTopLevelCookieDomain(window, cookie)
+        domain
```
jonsnyder: 👍

PR #142, opened 2019-08-05
CORE-32922 Add suppressErrors option.

Aaronius:
## Description

Adds an option allowing customers to suppress errors. When set to true, the errors will only be logged to the console when logging is enabled.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-32922

## Motivation and Context
Customers (most often?) don't like seeing errors in their console.

## Types of changes

## Checklist:

- [x] https://github.com/AdobeDocs/alloy-docs/pull/26 -> I have updated the documentation accordingly.

### DIFF
--- a/src/core/createInstance.js
+++ b/src/core/createInstance.js
@@ -24,6 +24,7 @@ export default (
 ) => {
   let componentRegistry;
   let configurationFailed = false;
+  let suppressErrors;
 
   const logCommand = ({ enabled }) => {
     // eslint-disable-next-line no-param-reassign
@@ -31,6 +32,7 @@ export default (
   };
 
   const configureCommand = options => {
+    ({ suppressErrors } = options);
     if (options.log !== undefined) {
       logCommand({ enabled: options.log });
     }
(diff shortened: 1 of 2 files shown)


src/core/createInstance.js
```diff
@@ -24,13 +24,15 @@ export default (
...
   let configurationFailed = false;
+  let suppressErrors;
 
   const logCommand = ({ enabled }) => {
     // eslint-disable-next-line no-param-reassign
     logController.logEnabled = enabled;
   };
 
   const configureCommand = options => {
+    ({ suppressErrors } = options);
```
Aaronius: I would really like to add a config validator for this option, along with the `log` option, but can't, because of when the config validation is run. I've logged an issue here: https://jira.corp.adobe.com/browse/CORE-32996

Aaronius: Also, we need to add tests around the core modules generally, which I've also logged here: https://jira.corp.adobe.com/browse/CORE-32995

jfkhoury: Thank you for logging the tickets @Aaronius !

PR #144, opened 2019-08-06
CORE-32956 Fix tests.

Aaronius:
## Description

Fixes tests...hopefully. **Don't merge this yet. I want to run the CI build a bunch of times and make sure it doesn't fail.**

## Related Issue
https://jira.corp.adobe.com/browse/CORE-32956

## Motivation and Context
Failing tests are 👎 .

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] All tests pass and I've made any necessary test changes.
- [x] I have run the Sandbox successfully.

PR #145, opened 2019-08-07
TNT-33101 - ensure sessionId is passed in event for data fetching and data collection

artur-ciocanu:
Personalization sessionId should be passed in event -> meta, to ensure data collection request is properly processed by our backend.

## Description

Session ID used to be passed at the request level and it was working OK for data fetching, but it was failing for data collection. To ensure that both data fetching and data collection requests have the session ID details, the session ID has been moved to `onBeforeEvent` and we make sure that both data fetching and data collection request have a session ID.

As part of this PR I have also added an utility named `deepAssign`. This utility was required since our `event.mergeMeta()` was misbehaving in situations like when we try to merge:
```json
{
  "meta": {
     "personalization": {
         "notification": {"a": 1, "b": 2}
     }
  }
}
```
And
```json
{
  "meta": {
     "personalization": {
         "sessionId": "......."
     }
  }
}
```
 
## Related Issue

NONE

## Motivation and Context

Ensure that data collection requests are not failing on our backend

## Types of changes

## Checklist:

### DIFF
--- a/src/utils/deepAssign.js
+++ b/src/utils/deepAssign.js
@@ -10,13 +10,13 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
+import isNil from "./isNil";
 import isObject from "./isObject";
 
 const deepAssignObject = (target, source) => {
   Object.keys(source).forEach(key => {
     if (isObject(target[key]) && isObject(source[key])) {
       deepAssignObject(target[key], source[key]);
-
       return;
     }
 
@@ -26,29 +26,20 @@ const deepAssignObject = (target, source) => {
 
 /**
  * Recursively copy the values of all enumerable own properties from a source item to a target item if the both items are objects
- * @private
  * @param {Object} target - a target object
- * @param {Object} source - a source object
+ * @param {...Object} source - an array of source objects
  * @example
  * deepAssign({ a: 'a', b: 'b' }, { b: 'B', c: 'c' });
  * // { a: 'a', b: 'B', c: 'c' }
  */
 export default (target, ...sources) => {
-  if (!isObject(target)) {
-    return {};
+  if (isNil(target)) {
+    throw new TypeError('deepAssign "target" cannot be null or undefined');
   }
 
-  const { length } = sources;
+  const result = Object(target);
 
-  for (let i = 0; i < length; i += 1) {
-    const source = sources[i];
+  sources.forEach(source => deepAssignObject(result, Object(source)));
 
-    if (!isObject(source)) {
-      break;
-    }
-
-    deepAssignObject(target, source);
-  }
-
-  return target;
+  return result;
 };
--- a/src/utils/deepAssign.js
+++ b/src/utils/deepAssign.js
@@ -1,0 +1,54 @@
+/*
+Copyright 2019 Adobe. All rights reserved.
+This file is licensed to you under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License. You may obtain a copy
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+import isObject from "./isObject";
+
+const deepAssignObject = (target, source) => {
+  Object.keys(source).forEach(key => {
+    if (isObject(target[key]) && isObject(source[key])) {
+      deepAssignObject(target[key], source[key]);
+
+      return;
+    }
+
+    target[key] = source[key];
+  });
+};
+
+/**
+ * Recursively copy the values of all enumerable own properties from a source item to a target item if the both items are objects
+ * @private
+ * @param {Object} target - a target object
+ * @param {Object} source - a source object
+ * @example
+ * deepAssign({ a: 'a', b: 'b' }, { b: 'B', c: 'c' });
+ * // { a: 'a', b: 'B', c: 'c' }
+ */
+export default (target, ...sources) => {
+  if (!isObject(target)) {
+    return {};
+  }
+
+  const { length } = sources;
+
+  for (let i = 0; i < length; i += 1) {
+    const source = sources[i];
+
+    if (!isObject(source)) {
+      break;
+    }
+
+    deepAssignObject(target, source);
+  }
+
+  return target;
+};
--- a/test/unit/specs/utils/deepAssign.spec.js
+++ b/test/unit/specs/utils/deepAssign.spec.js
@@ -14,13 +14,59 @@ import deepAssign from "../../../../src/utils/deepAssign";
 import assign from "../../../../src/utils/assign";
 
 describe("deepAssign", () => {
-  it("should copy values like assign", () => {
-    expect(deepAssign({}, { a: 1 }, { b: 2 })).toEqual(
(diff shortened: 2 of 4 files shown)


jfkhoury:
@artur-ciocanu It looks like the build failure might be legit. Here's the test that is failing:

`Allows other components to hold up the event`

artur-ciocanu:
@jfkhoury it's interesting since `travis-ci/push` was OK, but `travis/ci-pr` failed. Also as far as I can tell it fails again on IE 11 in data collector and promise polyfill related code.

Aaronius:
FYI I'm working on a fix here: https://github.com/adobe/alloy/pull/144. I wouldn't let it hold up the merge once the PR gets approved.

src/utils/deepAssign.js
```diff
@@ -0,0 +1,54 @@
...
+    return {};
+  }
+
+  const { length } = sources;
+
+  for (let i = 0; i < length; i += 1) {
+    const source = sources[i];
+
+    if (!isObject(source)) {
+      break;
```
jonsnyder: We should throw an error here instead of silently failing.  This would happen for `deepAssign({}, "foo")` or `deepAssign({}, undefined)`.

test/unit/specs/utils/deepAssign.spec.js
```diff
@@ -0,0 +1,33 @@
...
+    const result = deepAssign(
+      {},
+      { a: { c: 1 } },
+      { b: 2 },
+      { a: { c: 2, d: 3 } }
+    );
+
+    expect(result).toEqual({ a: { c: 2, d: 3 }, b: 2 });
+  });
+});
```
jonsnyder: Can you add some more tests?
* make sure it works with strings
* make sure arrays can be assigned
* test undefined as first or second parameter

src/utils/deepAssign.js
```diff
@@ -0,0 +1,54 @@
...
+ * @private
+ * @param {Object} target - a target object
+ * @param {Object} source - a source object
+ * @example
+ * deepAssign({ a: 'a', b: 'b' }, { b: 'B', c: 'c' });
+ * // { a: 'a', b: 'B', c: 'c' }
+ */
+export default (target, ...sources) => {
+  if (!isObject(target)) {
+    return {};
```
jonsnyder: I think this should throw an error as well.

src/utils/deepAssign.js
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  const { length } = sources;
```
Aaronius: This really isn't a big deal in any way, but caching length _generally_ doesn't help with performance anymore and can often hinder it. Here's a little on the issue: https://mrale.ph/blog/2014/12/24/array-length-caching.html In any case, it will make little difference either way.

src/utils/deepAssign.js
Aaronius: Not throwing an error would probably be more native-like, for better or worse. For example:
image

src/utils/deepAssign.js
```diff
@@ -0,0 +1,54 @@
...
+
+    target[key] = source[key];
+  });
+};
+
+/**
+ * Recursively copy the values of all enumerable own properties from a source item to a target item if the both items are objects
+ * @private
+ * @param {Object} target - a target object
+ * @param {Object} source - a source object
```
Aaronius: This should be `{...Object}`

src/utils/deepAssign.js
```diff
@@ -0,0 +1,54 @@
...
+      return;
+    }
+
+    target[key] = source[key];
+  });
+};
+
+/**
+ * Recursively copy the values of all enumerable own properties from a source item to a target item if the both items are objects
+ * @private
```
Aaronius: It's interesting that you marked this as private. Is that because you wouldn't want it to show up in our JSDoc if we ever generated it? It's an interesting thing to think about...whether our produced JSDoc would be intended for Alloy customers or Alloy developers. I would probably think we would ever only produce JSDoc for other developers, if ever. 🤷‍♂

test/unit/specs/utils/deepAssign.spec.js
Aaronius: I think this would be good too.

src/utils/deepAssign.js
Aaronius: Here's an error though:

image

artur-ciocanu: @jonsnyder and @Aaronius thanks for your suggestions I have added the following changes:
1. If `target` is `null` or `undefined` we will throw a `TypeError`
2. To make sure we are compliant with `Object.assign()` we convert both `target` and `source` to object. Which means that all the weird combinations of different types are supported

test/unit/specs/utils/deepAssign.spec.js
artur-ciocanu: Done.

src/utils/deepAssign.js
artur-ciocanu: Fixed.

src/utils/deepAssign.js
artur-ciocanu: @Aaronius it's just a habit, not necessarily for performance reasons.

src/utils/deepAssign.js
artur-ciocanu: Fixed.

src/utils/deepAssign.js
artur-ciocanu: My bad, I just copied it from another place. Right I am not sure we discussed about JSDoc and the way we want to is it. For now I just removed `private` since it doesn't make any sense.

src/utils/deepAssign.js
artur-ciocanu: After some code changes I replaced it with `forEach()`

test/unit/specs/utils/deepAssign.spec.js
```diff
@@ -14,13 +14,59 @@ import deepAssign from "../../../../src/utils/deepAssign";
 import assign from "../../../../src/utils/assign";
 
 describe("deepAssign", () => {
-  it("should copy values like assign", () => {
-    expect(deepAssign({}, { a: 1 }, { b: 2 })).toEqual(
-      assign({}, { a: 1 }, { b: 2 })
-    );
+  it("should throw when target is null or unbdefined", () => {
```
jonsnyder: typo `undefined`

test/unit/specs/utils/deepAssign.spec.js
```diff
@@ -14,13 +14,59 @@ import deepAssign from "../../../../src/utils/deepAssign";
...
+    expect(() => {
+      deepAssign(null, { a: 1 });
+    }).toThrow();
+
+    expect(() => {
+      deepAssign(undefined, { a: 1 });
+    }).toThrow();
+  });
+
+  it("should assing when target is string", () => {
```
jonsnyder: typo `assign` (and on the next 5)

artur-ciocanu: Fixed.

test/unit/specs/utils/deepAssign.spec.js
artur-ciocanu: FIxed.

test/unit/specs/utils/deepAssign.spec.js
```diff
@@ -14,13 +14,59 @@ import deepAssign from "../../../../src/utils/deepAssign";
...
+  });
+
+  it("should assing when target is object and source is number", () => {
+    const result1 = deepAssign({}, 1);
+    const result2 = assign({}, 1);
+
+    expect(result1).toEqual(result2);
+  });
+
+  it("should assing when target is object and source is array", () => {
```
Aaronius: Assing. Lol.

PR #147, opened 2019-08-08
CORE-30186 Support an ID Sync container config

wgamboa:
## Description

Here's the config:

"query": { "identity":
{ "containerId": 1 }}

## Related Issue

CORE-30186

## Types of changes

## Checklist:


src/components/Identity/index.js
```diff
@@ -88,11 +88,19 @@ const createIdentity = ({ config, logger, cookie }) => {
...
+            const identityQuery = {
               identity: {
                 exchange: true
               }
-            });
+            };
+
+            const containerId = parseInt(config.idSyncContainerId, 10);
+
+            if (!Number.isNaN(containerId)) {
```
jonsnyder: IE does not support `Number.isNaN` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN

.travis.yml
```diff
@@ -41,7 +39,6 @@ script:
 - npm run allure:generate
 - npm run functional
 - rollup -c --environment BUILD:production && rollup -c --environment MINIFY,BUILD:production && bundlesize
-- sonar-scanner
```
jonsnyder: how does this relate to this PR?

sandbox/public/index.html
```diff
@@ -42,7 +42,8 @@
         imsOrgId: "53A16ACB5CC1D3760A495C99@AdobeOrg",
         log: true,
         prehidingId: "alloy-prehiding",
-        prehidingStyle: "section:nth-of-type(1) { opacity: 0 !important }"
+        prehidingStyle: "section:nth-of-type(1) { opacity: 0 !important }",
+        idSyncContainerId: 81
```
jonsnyder: Is this a required config parameter if idSyncsEnabled?  If so you should add config validation for it.

Aaronius: I agree. Even if it isn't required, we should add a config validator to restrict it to a number (assuming it's always a number).

src/components/Identity/index.js
```diff
@@ -88,11 +88,19 @@ const createIdentity = ({ config, logger, cookie }) => {
...
+              assign(identityQuery.identity, { containerId });
```
Aaronius: I don't think `assign` provides anything particularly useful here. How about just `identityQuery.identity.containerId = containerId`?

src/components/Identity/index.js
```diff
@@ -88,11 +88,19 @@ const createIdentity = ({ config, logger, cookie }) => {
...
           if (config.idSyncsEnabled && nowInHours > timestamp) {
-            event.mergeQuery({
+            const identityQuery = {
               identity: {
                 exchange: true
               }
-            });
+            };
+
+            const containerId = parseInt(config.idSyncContainerId, 10);
```
Aaronius: I feel it would be better to add config validator that rejects the user's value if it's not a number.

src/components/Identity/index.js
wgamboa: Thanks.

.travis.yml
wgamboa: It allows the build to pass as suggested by @owenleej .

src/components/Identity/index.js
wgamboa: I'm getting
error  Unexpected use of 'isNaN'  no-restricted-globals
Should I edit eslint? It only accepts Number.isNaN, which doesn't exist in IE

Aaronius: That sounds appropriate to me.

src/components/Identity/index.js
```diff
@@ -179,6 +178,10 @@ createIdentity.abbreviation = "ID";
 createIdentity.configValidators = {
   idSyncsEnabled: {
     defaultValue: true
+  },
+  idSyncContainerId: {
+    validate: nonNegativeInteger,
```
Aaronius: We need to start making our validators more composable. I wrote a JIRA issue for this a while back: https://jira.corp.adobe.com/browse/CORE-32266 In this case, `nonNegativeInteger` would be more flexible if it didn't try to do multiple things. So instead of:

`validate: nonNegativeInteger`

you would do something like:

`validate: number.integer().min(0)`

and later on some other component could do something like:

`validate: number.min(-15).max(250)` or whatever it needed.

If you wanted to do that work now or in a new PR, that would be excellent. Otherwise, add a comment to https://jira.corp.adobe.com/browse/CORE-32266 describing that we need to rework `nonNegativeInteger` later.

src/components/Identity/index.js
```diff
@@ -179,6 +178,10 @@ createIdentity.abbreviation = "ID";
...
+    defaultValue: null
```
Aaronius: Is this line necessary?

test/unit/specs/utils/isNumber.spec.js
```diff
@@ -0,0 +1,27 @@
...
+*/
+
+import isNumber from "../../../../src/utils/isNumber";
+
+describe("isNumber", () => {
+  it("tests", () => {
+    expect(isNumber(null)).toBe(false);
+    expect(isNumber(undefined)).toBe(false);
+    expect(isNumber("abc")).toBe(false);
+    expect(isNumber(parseInt("abc", 10))).toBe(false);
```
Aaronius: This one is a bit odd. Since you're really just testing if `isNumber` returns `false` for `NaN`, how about just passing `NaN` into `isNumber` (ignore eslint if needed)?

src/components/Identity/index.js
```diff
@@ -90,7 +91,13 @@ const createIdentity = ({ config, logger, cookieJar }) => {
               identity: {
                 exchange: true
               }
-            });
+            };
+
+            if (config.idSyncContainerId !== null) {
+              identityQuery.identity.containerId = config.idSyncContainerId;
```
Aaronius: Should I be seeing `identityQuery` being initialized or passed in somewhere?

Aaronius:
Be sure to update https://github.com/AdobeDocs/alloy-docs/blob/master/fundamentals/configuring-the-sdk.md

test/unit/specs/utils/isNumber.spec.js
```diff
@@ -0,0 +1,27 @@
...
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+import isNumber from "../../../../src/utils/isNumber";
+
+describe("isNumber", () => {
+  it("tests", () => {
```
Aaronius: If you could be more specific in the description, that would be nice. Maybe even split the tests into true tests and false tests. Something like:

`it("returns true if the value is a number"...`
`it("returns false if the value is not a number"...`

src/components/Identity/index.js
```diff
@@ -90,7 +91,13 @@ const createIdentity = ({ config, logger, cookieJar }) => {
               identity: {
                 exchange: true
               }
-            });
+            };
```
Aaronius: It looks like you're missing a close parenthesis here.

src/components/Identity/index.js
wgamboa: That would be interesting, but there are other things I need to work on right now.

Regarding IdentityQuery, I did a dangerous merge conflict resolution through the pull request page and messed it up. I won't do that again since it bypasses the commit and push tests.

src/components/Identity/index.js
wgamboa: I guess not.

test/unit/specs/utils/isNumber.spec.js
wgamboa: Will do about the docs.

sandbox/public/index.html
wgamboa: It's not required.

.travis.yml
wgamboa: I took this out but put it back in through a rebase.

PR #149, opened 2019-08-09
TNT-33100 - ensure Alloy is compatible with Target Visual Experience Editor

artur-ciocanu:
Ensure Alloy can be used with Adobe Target Visual Experience Composer.

## Description

To make sure that Adobe Target can edit and create activities on the pages that have Alloy I have added a few changes:
- a new config has been added named `authoringMode` it's a boolean that indicates if we are in authoring mode or not. Currently it is initialized in the sandbox using this code:
```js
  authoringMode: document.location.href.indexOf("mboxEdit") !== -1
```
When loading a site in VEC, VEC opens the site in an IFRAME and it appends `mboxEdit` to the URL. I didn't want to hard code the `mboxEdit` checking logic into personalization component hence we have `authoringMode` config.
- personalization component has been made `authoringMode` aware. If we are in authoring mode personalization component bails out and does nothing. This is to allow VEC to do it's business so Alloy doesn't interfere with VEC authoring functionality.

## Related Issue

NONE

## Motivation and Context

Allow anyone to use Alloy and Target VEC

## Types of changes

## Checklist:

### DIFF
--- a/sandbox/public/index.html
+++ b/sandbox/public/index.html
@@ -42,7 +42,9 @@
         propertyId: 9999999,
         imsOrgId: "53A16ACB5CC1D3760A495C99@AdobeOrg",
         log: true,
-        authoringMode: document.location.href.indexOf("mboxEdit") !== -1,
+        // Target VEC appends "mobxEdit" to URL when loading the page.
+        // By checking this params we know we are in authoring mode
+        authoringMode: document.location.href.indexOf("mboxEdit") !== -1, 
         prehidingId: "alloy-prehiding",
         prehidingStyle: "section:nth-of-type(1) { opacity: 0 !important }"
       });
--- a/sandbox/public/index.html
+++ b/sandbox/public/index.html
@@ -8,19 +8,20 @@
     <script src="https://cdn.jsdelivr.net/npm/promise-polyfill@8/dist/polyfill.min.js"></script>
     <meta http-equiv="Content-Security-Policy"
           content="default-src 'self';
-            script-src 'self' 'unsafe-inline';
-            style-src 'self' 'unsafe-inline';
-            img-src *;
-            connect-src 'self' https://alpha.konductor.adobedc.net https://edgegateway.azurewebsites.net http://ex-edge.stable-stage.aam-npe.adobeinternal.net http://localhost:8080"/>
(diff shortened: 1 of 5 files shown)


src/components/Personalization/index.js
```diff
@@ -93,6 +93,18 @@ const createPersonalization = ({ config, logger, cookieJar }) => {
...
+        if (authoringMode) {
+          logger.warn("Rendering is disabled, authoring mode.");
+
+          event.mergeQuery({
+            personalization: {
+              enabled: false
+            }
+          });
+
+          return Promise.resolve();
```
Aaronius: Feel free to leave this the way that it is, but it's also fine if you just want to `return` instead of returning a promise.

sandbox/public/index.html
```diff
@@ -41,6 +42,7 @@
         propertyId: 9999999,
         imsOrgId: "53A16ACB5CC1D3760A495C99@AdobeOrg",
         log: true,
+        authoringMode: document.location.href.indexOf("mboxEdit") !== -1,
```
jfkhoury: Can you pls add a comment here to explain the reason why we need the `mboxEdit` voodoo?

src/components/Personalization/index.js
artur-ciocanu: @Aaronius I've tried that, but ESLint is complaining.

sandbox/public/index.html
artur-ciocanu: Done

artur-ciocanu:
@Aaronius I'll add a ticket to update the docs.

test/unit/specs/components/Identity/createIdSyncs.spec.js
```diff
@@ -32,7 +32,7 @@ describe("Identity::createIdSyncs", () => {
     }, {});
   };
 
-  it("tracks id syncs", done => {
+  it("tracks id syncs", () => {
```
Aaronius: Are you sure this change should be made? I don't think this will run the `expect` without the `done`. If `idSyncs.process` returned a promise, we wouldn't have to do this `setTimeout` business at all.

It looks like you left an `fdescribe` in as well.

Aaronius:
Don't forget to delete the branch if you no longer need it.

PR #150, opened 2019-08-09
Fixes issue where setCustomerIds was hitting interact endpoint.

Aaronius:
## Description
Fixes issue where setCustomerIds was hitting interact endpoint.

## Related Issue
None

## Motivation and Context
We shouldn't be hitting `interact` for setCustomerIds.

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Identity/index.js
+++ b/src/components/Identity/index.js
@@ -48,10 +48,10 @@ const createIdentity = ({ config, logger, cookieJar }) => {
 
   const setCustomerIds = ids => {
     validateCustomerIds(ids);
-    const event = createEvent();
-    event.mergeData(ids.data);
+    const event = createEvent(); // FIXME: We shouldn't need an event.
+    event.mergeData({}); // FIXME: We shouldn't need an event.
     const payload = network.createPayload();
-    payload.addEvent(event);
+    payload.addEvent(event); // FIXME: We shouldn't need an event.
     assign(customerIds, ids);
 
     const normalizedCustomerIds = normalizeCustomerIds(customerIds);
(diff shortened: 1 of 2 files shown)


src/components/Identity/index.js
```diff
@@ -70,7 +70,7 @@ const createIdentity = ({ config, logger, cookieJar }) => {
       cookieJar.set(CUSTOMER_ID_HASH, customerIdsHash);
     }
     return lifecycle
-      .onBeforeEvent(event, normalizedCustomerIds, customerIdChanged)
+      .onBeforeEvent(event, {}, false) // FIXME: We shouldn't need an event.
```
Aaronius: Though we shouldn't need an event on payload, Konductor currently requires one. It also has to have a `data` property on it. :/

Aaronius: BTW, the reason this was forcing a request to the `interact` endpoint was because the third parameter of `onBeforeEvent` is supposed to indicate whether the event was due to a view start. Because we were passing `customerIdChanged`, which was `true`, other components like Personalization though that the event was due to a view start, and therefore requested a response.

jileeshs: Yes, I tried it without event, kondcutor was throwing an error. That's why I added event and data.

jileeshs: Yeah, probably I didn't notice the issue because I might have tested in an older session where `customerIdChanged` returned false :(

jileeshs:
Can we comment out the line
https://github.com/adobe/alloy/blob/bcb529c0228ee48d88091c82e1c8110e4eb174f9/.travis.yml#L25
until we resolve issues with sonar connection from travis?

Aaronius:
@jileeshs It was a different line causing problems, but I got it done.

PR #151, opened 2019-08-09
Id sync query fix & refactor processIdSyncs

jfkhoury:
## Description

Stop querying for ID Syncs on every event, and multiple times per lifecycle hook.

Refactor processIdSyncs to make it an object:

Add a process and hasExpired APIs; this way we move all the logic inside that object instead of make Identity be aware of those details. Tell don't ask! (Great catch @Aaronius)

## Related Issue

https://jira.corp.adobe.com/browse/CORE-33129

## Motivation and Context

Querying for ID Syncs multiple times was causing lots of `interact` requests.

## Types of changes

## Checklist:

- [ x ] All tests pass and I've made any necessary test changes.

### DIFF
--- a/src/components/Identity/createIdSyncs.js
+++ b/src/components/Identity/createIdSyncs.js
@@ -36,7 +36,7 @@ const setControlObject = (controlObject, cookieJar) => {
   cookieJar.set(ID_SYNC_CONTROL, arr.join("_"));
 };
 
-export default ({ destinations, config, logger, cookieJar }) => {
+const createProcessor = (config, logger, cookieJar) => destinations => {
   if (!config.idSyncsEnabled) {
     return;
   }
@@ -89,4 +89,21 @@ export default ({ destinations, config, logger, cookieJar }) => {
       );
     });
   }
+};
+
+const createExpiryChecker = cookieJar => () => {
+  const nowInHours = Math.round(
+    convertTimes(MILLISECOND, HOUR, new Date().getTime())
+  );
+  const timestamp = parseInt(cookieJar.get(ID_SYNC_TIMESTAMP) || 0, 36);
+
+  return nowInHours > timestamp;
+};
+
+export default (config, logger, cookieJar) => {
+
+  return {
+    process: createProcessor(config, logger, cookieJar),
+    hasExpired: createExpiryChecker(cookieJar)
+  };
 };
--- a/src/components/Identity/index.js
+++ b/src/components/Identity/index.js
@@ -10,18 +10,13 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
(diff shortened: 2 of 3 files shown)


src/components/Identity/createIdSyncs.js
```diff
@@ -90,3 +92,21 @@ export default ({ destinations, config, logger, cookieJar }) => {
...
+  const nowInHours = Math.round(
+    convertTimes(MILLISECOND, HOUR, new Date().getTime())
+  );
+  const timestamp = parseInt(cookieJar.get(ID_SYNC_TIMESTAMP) || 0, 36);
+
+  return nowInHours > timestamp;
+};
+
+export default (config, logger, _cookieJar) => {
+  cookieJar = _cookieJar;
```
Aaronius: I think we should pass cookieJar down through `createProcessor` and `hasExpired`. @wgamboa did what you did in his PR and I suggested he change it. Here's the reasoning: https://github.com/adobe/alloy/pull/127#discussion_r310269642

jfkhoury: I went back on forth on this, but I should go forth because you're right :) thanks.

src/components/Identity/index.js
```diff
@@ -83,12 +80,12 @@ const createIdentity = ({ config, logger, cookieJar }) => {
...
-          );
-          const timestamp = parseInt(cookieJar.get(ID_SYNC_TIMESTAMP) || 0, 36);
-
-          if (config.idSyncsEnabled && nowInHours > timestamp) {
+          if (
+            !alreadyQueriedForIdSyncs &&
+            config.idSyncsEnabled &&
+            idSyncs.hasExpired()
+          ) {
+            alreadyQueriedForIdSyncs = true;
```
jonsnyder: This logic means that you only query for IdSyncs once per page load.  Is this what we want for long running pages?  You could rename this variable to `idSyncsInFlight`. and then reset it when the query sync comes back.

src/components/Identity/createIdSyncs.js
```diff
@@ -90,3 +90,20 @@ export default ({ destinations, config, logger, cookieJar }) => {
     });
   }
 };
+
+const createExpiryChecker = cookieJar => () => {
+  const nowInHours = Math.round(
```
jonsnyder: I'd like to see some unit tests on this logic.  From looking at it, it will take 90 minutes for a 1 hour expiration to expire.  You could replace `Math.round` with `Math.ceil`

src/components/Identity/index.js
jfkhoury: Jon, we sync once every 7 days.

src/components/Identity/createIdSyncs.js
jfkhoury: Good catch, I will create a testing ticket.

PR #154, opened 2019-08-13
Rename log and suppressErrors to be more consistent.

jfkhoury:
## Description

Rename `log` to `logEnabled`.
Rename `suppressErrors` to `errorsEnabled`.

TODO: I think we should suppress errors in lifecycle hooks as well, not only commands. Thoughts?

## Motivation and Context

All our boolean configs end with `enabled` except for these 2.

## Types of changes

## Checklist:

- [ X ] My change requires a change to the documentation.

### DIFF
--- a/sandbox/public/index.html
+++ b/sandbox/public/index.html
@@ -44,7 +44,7 @@
         logEnabled: true,
         // Target VEC appends "mobxEdit" to URL when loading the page.
         // By checking this params we know we are in authoring mode
-        authoringMode: document.location.href.indexOf("mboxEdit") !== -1, 
+        authoringModeEnabled: document.location.href.indexOf("mboxEdit") !== -1, 
         prehidingId: "alloy-prehiding",
         prehidingStyle: "section:nth-of-type(1) { opacity: 0 !important }"
       });
--- a/sandbox/public/index.html
+++ b/sandbox/public/index.html
@@ -38,10 +38,10 @@
       alloy("configure", {
         // edgeDomain: "edgegateway.azurewebsites.net",
         // optInEnabled: true,
-        // suppressErrors: true,
+        // errorsEnabled: false,
         propertyId: 9999999,
         imsOrgId: "53A16ACB5CC1D3760A495C99@AdobeOrg",
-        log: true,
+        logEnabled: true,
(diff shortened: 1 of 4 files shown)


sandbox/public/index.html
```diff
@@ -38,10 +38,10 @@
...
         // optInEnabled: true,
-        // suppressErrors: true,
+        // errorsEnabled: false,
         propertyId: 9999999,
         imsOrgId: "53A16ACB5CC1D3760A495C99@AdobeOrg",
-        log: true,
+        logEnabled: true,
         // Target VEC appends "mobxEdit" to URL when loading the page.
         // By checking this params we know we are in authoring mode
         authoringMode: document.location.href.indexOf("mboxEdit") !== -1,
```
Aaronius: Should `authoringMode` be `authoringModeEnabled` then? I think for consistency it should.

Aaronius:
@jfkhoury Regarding this:

> TODO: I think we should suppress errors in lifecycle hooks as well, not only commands. Thoughts?

What your reasoning behind this thought? Errors in lifecycle hooks should also bubble up and be handled by this logic, unless I'm not seeing something you are.

Aaronius:
I can't approve or request changes on this PR either. I think that might happen when you request 2 other people to review or something?

sandbox/public/index.html
jfkhoury: Good catch, I'll add that as well.

jfkhoury: @Aaronius Added one more commit to address this comment. Thanks!

Aaronius: I think you need to update `authoringMode` in index.html still.

sandbox/public/index.html
```diff
@@ -44,7 +44,7 @@
         logEnabled: true,
         // Target VEC appends "mobxEdit" to URL when loading the page.
         // By checking this params we know we are in authoring mode
-        authoringMode: document.location.href.indexOf("mboxEdit") !== -1, 
+        authoringModeEnabled: document.location.href.indexOf("mboxEdit") !== -1,
```
jfkhoury: Good catch @Aaronius

PR #157, opened 2019-08-14
CORE-32734 - move not-shared DOM utils to personalization

artur-ciocanu:
Move DOM utils that are not shared to Personalization component helpers

## Description

Move DOM utils that are not shared to Personalization component helpers

## Related Issue

NONE

## Motivation and Context

This is required to improve components cohesion, there is no need to have functionality specific to a single component being part of an utility folder.

## Types of changes

## Checklist:


src/components/Personalization/helper/dom/getElementById.js
```diff
@@ -13,9 +13,9 @@ governing permissions and limitations under the License.
 /**
  * Returns an array of matched DOM nodes.
  * @param {String} id
- * @param {Node} doc, defaults to document
+ * @param {Node} context, defaults to document
```
Aaronius: It's really not a big deal, but if you're wanting to stick to JSDoc, the way you would show an optional param with a default value would be like this: `@param {Node} [context=document]`

artur-ciocanu: Will change, thanks!

PR #159, opened 2019-08-14
Fix nonNegativeInteger - clean up config validators.

jfkhoury:
## Description

- The optional `idSyncContainer` config is causing an error because the `nonNegativeInteger` validator is was not allowing `undefined` values.
- Moved config validators from `core` to `Data Collector`.
- The `config` arg was never used in the validators.
- Came-cased `config-validators` to match the rest of the folders.

## Related Issue

https://jira.corp.adobe.com/browse/CORE-33402

## Types of changes

- [ X ] Bug fix (non-breaking change which fixes an issue)

## Checklist:


src/components/DataCollector/index.js
```diff
@@ -73,4 +74,18 @@ const createDataCollector = ({ config }) => {
 
 createDataCollector.namespace = "DataCollector";
 createDataCollector.abbreviation = "DC";
+
```
liljenback: I thought we already had this before?

src/utils/configValidators/required.js
```diff
@@ -10,7 +10,7 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-export default (config, key, currentValue) => {
+export default (key, currentValue) => {
   let err = "";
   if (currentValue == null) {
```
Aaronius: I'd like to see this as `if (currentValue === undefined)`. I think validation should fail f the user passes in `null` for a required option.

src/utils/configValidators/nonNegativeInteger.js
```diff
@@ -12,8 +12,9 @@ governing permissions and limitations under the License.
 
 import isInteger from "../isInteger";
 
-export default (config, key, currentValue) => {
-  return isInteger(currentValue) && currentValue >= 0
+export default (key, currentValue) => {
+  return currentValue === undefined ||
```
Aaronius: I like that you used `=== undefined` here.

src/utils/configValidators/eitherNilOrNonEmpty.js
```diff
@@ -13,7 +13,7 @@ governing permissions and limitations under the License.
 import isNil from "../isNil";
 import isNonEmptyString from "../isNonEmptyString";
 
-export default (config, key, currentValue) => {
+export default (key, currentValue) => {
   if (isNil(currentValue)) {
```
Aaronius: I think this should be `if (currentValue === undefined){`. I think validation should fail if the user passes in `null`.

src/utils/configValidators/boolean.js
```diff
@@ -12,7 +12,7 @@ governing permissions and limitations under the License.
 
 import isBoolean from "../isBoolean";
 
-export default (config, key, currentValue) => {
+export default (key, currentValue) => {
   return isBoolean(currentValue)
```
Aaronius: I believe this should return an empty string if the value is `undefined`. I'm not sure if you wanted to do that as part of this same PR.

src/utils/configValidators/eitherNilOrNonEmpty.js
Aaronius: Oh wait.....nevermind on this one.

src/utils/configValidators/required.js
Aaronius: OMG...I'm not thinking straight. Ignore this comment too.

src/components/DataCollector/index.js
liljenback: Looks like it was removed by mistake, but this shiny new one looks better :)

PR #162, opened 2019-08-15
Add large payload testing page to the sandbox.

jfkhoury:
## Description

We need a page where we can test beacon calls and large payloads being send to the edge.

## Types of changes

## Checklist:

- [ ] All tests pass and I've made any necessary test changes.

### DIFF
--- a/sandbox/src/LargePayload.js
+++ b/sandbox/src/LargePayload.js
@@ -1,0 +1,30 @@
+import React from "react";
+
+export default function LargePayload() {
+
+  const makePayload = size => () => {
+    const payload = new Uint8Array(size * 1024);
+    window.alloy("event", {
+      beacon: true,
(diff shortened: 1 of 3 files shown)


sandbox/src/LargePayload.js
```diff
@@ -0,0 +1,30 @@
+import React from "react";
+
+export default function LargePayload() {
+
+  const makePayload = size => () => {
+    const payload = new Uint8Array(size * 1024);
+    window.alloy("event", {
+      beacon: true,
```
jonsnyder: `beacon: true` doesn't actually do anything.

jfkhoury: :) That's what happens when I copy/paste :)

I think from the Links page. I'll clean it up.

PR #165, opened 2019-08-19
CORE-33469 Suppress errors that occur within lifecycle methods.

Aaronius:
## Description
There were two lifecycle methods where, if an error was thrown, the error was not being suppressed if the customer had configured `errorsEnabled: false` in the configuration. Those two lifecycle methods were `onComponentsRegistered` on `onResponseError`. This PR fixes those holes so the error gets properly suppressed when `errorsEnabled: false`.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-33469

## Motivation and Context
Inconsistent and inaccurate behavior.

## Types of changes

## Checklist:

### DIFF
--- a/src/core/createInstance.js
+++ b/src/core/createInstance.js
@@ -21,9 +21,8 @@ export default (
   logger,
   window
 ) => {
-  let componentRegistry;
-  let configurationFailed = false;
   let errorsEnabled;
+  let configurePromise;
 
   const logCommand = ({ enabled }) => {
     // eslint-disable-next-line no-param-reassign
@@ -43,48 +42,59 @@ export default (
     }
     const config = createConfig(options);
 
-    try {
-      componentRegistry = initializeComponents(config);
-    } catch (e) {
-      configurationFailed = true;
-      throw e;
-    }
+    return initializeComponents(config);
   };
 
   const executeCommand = (commandName, options) => {
-    let command;
+    let execute;
 
-    if (configurationFailed) {
-      // We've decided if configuration fails we'll return
-      // never-resolved promises for all subsequent commands rather than
-      // throwing an error for each of them.
-      command = () => new Promise(() => {});
-    } else if (commandName === "configure") {
-      if (componentRegistry) {
+    if (commandName === "configure") {
+      if (configurePromise) {
         throw new Error(
           "The library has already been configured and may only be configured once."
         );
       }
-      command = configureCommand;
-    } else if (commandName === "log") {
-      command = logCommand;
+
+      execute = () => {
+        configurePromise = configureCommand(options);
+        return configurePromise;
+      };
     } else {
-      if (!componentRegistry) {
+      if (!configurePromise) {
         throw new Error(
           `The library must be configured first. Please do so by calling ${namespace}("configure", {...}).`
         );
       }
 
-      command = componentRegistry.getCommand(commandName);
-
-      if (!isFunction(command)) {
-        throw new Error(`The ${commandName} command does not exist.`);
+      if (commandName === "log") {
+        execute = () => logCommand(options);
+      } else {
+        execute = () => {
+          return configurePromise.then(
+            componentRegistry => {
+              const command = componentRegistry.getCommand(commandName);
+              if (!isFunction(command)) {
+                throw new Error(`The ${commandName} command does not exist.`);
+              }
+              return command(options);
+            },
+            () => {
+              // If configuration failed, we prevent the configuration
+              // error from bubbling here because we don't want the
+              // configuration error to be reported in the console every
+              // time any command is executed. Only having it bubble
+              // once when the configure command runs is sufficient.
+              // Instead, for this command, we'll just return a promise
+              // that never gets resolved.
+              return new Promise(() => {});
+            }
+          );
+        };
       }
     }
 
     logger.log(`Executing "${commandName}" command.`, "Options:", options);
-
-    return command(options);
+    return execute();
   };
 
   return args => {
--- a/src/core/initializeComponentsFactory.js
+++ b/src/core/initializeComponentsFactory.js
@@ -80,12 +80,12 @@ export default (
 
   const lifecycle = createLifecycle(componentRegistry);
   const network = createNetwork(config, logger, lifecycle);
-  lifecycle.onComponentsRegistered({
-    componentRegistry,
-    lifecycle,
-    network,
-    optIn
-  });
-
-  return componentRegistry;
+  return lifecycle
+    .onComponentsRegistered({
+      componentRegistry,
+      lifecycle,
+      network,
+      optIn
+    })
+    .then(() => componentRegistry);
 };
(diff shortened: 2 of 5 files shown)


src/core/initializeComponentsFactory.js
```diff
@@ -80,12 +80,12 @@ export default (
 
   const lifecycle = createLifecycle(componentRegistry);
   const network = createNetwork(config, logger, lifecycle);
-  lifecycle.onComponentsRegistered({
-    componentRegistry,
-    lifecycle,
-    network,
-    optIn
-  });
```
Aaronius: This is why errors in `onComponentsRegistered` weren't being suppressed. Nothing was handling errors from this promise.

src/core/createInstance.js
```diff
@@ -43,48 +42,59 @@ export default (
     }
     const config = createConfig(options);
 
-    try {
-      componentRegistry = initializeComponents(config);
-    } catch (e) {
-      configurationFailed = true;
-      throw e;
-    }
+    return initializeComponents(config);
```
Aaronius: Notice how the configure command now returns the promise from `initializeComponents`? That's a key piece to the error handling.

src/core/network/createNetwork.js
```diff
@@ -106,8 +106,20 @@ export default (config, logger, lifecycle, networkStrategy) => {
           return handleResponsePromise;
         })
         .catch(error => {
-          lifecycle.onResponseError(error);
```
Aaronius: This is why errors from `onResponseError` were not being suppressed. Nothing was handling errors from this promise.

src/core/createInstance.js
```diff
@@ -43,48 +42,59 @@ export default (
...
 
-      command = componentRegistry.getCommand(commandName);
-
-      if (!isFunction(command)) {
-        throw new Error(`The ${commandName} command does not exist.`);
+      if (commandName === "log") {
+        execute = () => logCommand(options);
+      } else {
+        execute = () => {
+          return configurePromise.then(
```
Aaronius: By waiting for `configurePromise` to complete before executing any other commands, I was able to remove the crufty workaround that was in `createLifecycle.js`.

test/unit/specs/core/network/createNetwork.spec.js
```diff
@@ -101,35 +101,28 @@ describe("createNetwork", () => {
     });
   });
 
-  it("runs onResponseError hook and rejects the returned promise", () => {
-    networkStrategy.and.returnValue(Promise.reject(new Error("myerror")));
-    return network
-      .sendRequest({}, true)
-      .then(() => {
-        throw Error("Expecting sendRequest to fail.");
```
Aaronius: I'm not sure why we were throwing an error here...

test/unit/specs/core/network/createNetwork.spec.js
```diff
@@ -101,35 +101,28 @@ describe("createNetwork", () => {
...
-        expect(error.message).toContain("Error parsing server response.\n");
-        expect(error.message).toContain("\nResponse body: badbody");
-      });
+    return network.sendRequest({}, true).catch(error => {
+      expect(lifecycle.onResponseError).toHaveBeenCalledWith(error);
+      // The native parse error message is different based on the browser
+      // so we'll just check to parts we control.
+      expect(error.message).toContain("Error parsing server response.\n");
+      expect(error.message).toContain("\nResponse body: badbody");
+    });
```
Aaronius: Technically, there is no test that checks to see whether the promise that is returned by `sendRequest` is chained off the promise that is returned from `onResponseError`. I was on the fence about whether to add a test for that. If anyone feels like I should, let me know and I'll do so.

src/core/createInstance.js
```diff
@@ -43,48 +42,59 @@ export default (
...
+            },
+            () => {
+              // If configuration failed, we prevent the configuration
+              // error from bubbling here because we don't want the
+              // configuration error to be reported in the console every
+              // time any command is executed. Only having it bubble
+              // once when the configure command runs is sufficient.
+              // Instead, for this command, we'll just return a promise
+              // that never gets resolved.
+              return new Promise(() => {});
```
jonsnyder: Should we log something here like, "An error during configure is preventing this command from executing"

src/core/createInstance.js
```diff
@@ -43,48 +42,59 @@ export default (
...
         throw new Error(
           `The library must be configured first. Please do so by calling ${namespace}("configure", {...}).`
         );
       }
 
-      command = componentRegistry.getCommand(commandName);
-
-      if (!isFunction(command)) {
-        throw new Error(`The ${commandName} command does not exist.`);
+      if (commandName === "log") {
```
jonsnyder: This change requires that configure be called before "log".  Do we want that?

test/unit/specs/core/network/createNetwork.spec.js
jonsnyder: If sendRequest resolves without error, then this test will pass.  Said another way, the expects only run when sendRequest fails.

src/core/createInstance.js
Aaronius: I think so. I like the blanket statement of saying you have to run `configure` before anything else. I think I like this because it doesn't run the risk of being a slippery slope of what may or may not be supported before configure and it makes it pretty clear that `configure` lays the foundation for settings and then other commands apply on top of that. For example, if I were to have:

```
alloy("log", {
  enabled: true
});

alloy("configure", {
  logEnabled: false
});
```

It's just a little odd to me and might result in confusion as to whether logging will ultimately be enabled. Because `configure` to me sounds like an initialization/foundational step, this seems like a more understandable approach:

```
alloy("configure", {
  logEnabled: false
});

alloy("log", {
  enabled: true
});
```

Anyway, that's my thought behind it. Thanks for bringing it up.

test/unit/specs/core/network/createNetwork.spec.js
Aaronius: Ah, that makes a lot of sense. Thanks for explaining. I've added it back.

src/core/createInstance.js
Aaronius: I think it's a good idea. @jfkhoury and I didn't want it to be too naggy for each command attempted to be executed after configuration fails, but I think a warning would be good, so I've added logging for that. I'm still not sure if this is the right approach. In some ways, I think it would be the most technically correct to reject the promise for each command. But it does get quite naggy. I'm curious how you feel about it.

jonsnyder: Yes, there is already an error when the configure command fails.  If we had log levels these decisions would be easier.  I could see us logging every command call at a debug or trace level.

Aaronius: We have the ability to log at a debug level. There's a `console.debug` command whose messages will only show up if you have "verbose" checked in the developer tools console (although I think that behavior varies by browser).

PR #166, opened 2019-08-19
And sendBeacon fallback, and 'exitLink: true' event option

jonsnyder:
## Description

 This PR is to align alloy with `sendBeacon`'s intended purpose.  sendBeacon will only be used for exit links.  It adds an option to the event command of `exitLink: true`.  This option determines whether or not sendBeacon is used.  If `exitLink: true` is set and the payload expects a response, the exitLink flag takes precedence and the collect endpoint is used.  Additionally if during an exitLink's sendBeacon request, the browser rejects it, Alloy will fallback to using fetch.

## Related Issue

https://jira.corp.adobe.com/browse/CORE-33392

## Motivation and Context

From the research I've done, I come to the conclusion that `sendBeacon` was designed to be used while the page is unloading so analytics can be recorded without holding up the navigation to the next page.  Browsers impose a size limit on the payload and refuse to send the hit if it is too large. 

## Types of changes

## Checklist:

### DIFF
--- a/src/components/DataCollector/index.js
+++ b/src/components/DataCollector/index.js
@@ -21,7 +21,7 @@ const createDataCollector = ({ config, logger }) => {
   let network;
   let optIn;
 
-  const makeServerCall = (event, exitLink) => {
+  const makeServerCall = (event, documentUnloading) => {
     const payload = network.createPayload();
     payload.addEvent(event);
     payload.mergeMeta({
@@ -33,7 +33,11 @@ const createDataCollector = ({ config, logger }) => {
     return lifecycle
       .onBeforeDataCollection(payload)
       .then(() => {
-        return network.sendRequest(payload, payload.expectsResponse, exitLink);
+        return network.sendRequest(
+          payload,
+          payload.expectsResponse,
+          documentUnloading
+        );
       })
       .then(response => {
         const data = {
@@ -50,15 +54,20 @@ const createDataCollector = ({ config, logger }) => {
 
   const createEventHandler = options => {
     const event = createEvent();
-    const { viewStart = false, exitLink = false, data, meta } = options;
+    const {
+      viewStart = false,
+      documentUnloading = false,
+      data,
+      meta
+    } = options;
 
     event.mergeData(data);
     event.mergeMeta(meta);
 
     return lifecycle
-      .onBeforeEvent(event, options, viewStart, exitLink)
+      .onBeforeEvent(event, options, viewStart, documentUnloading)
       .then(() => optIn.whenOptedIn())
-      .then(() => makeServerCall(event, exitLink));
+      .then(() => makeServerCall(event, documentUnloading));
   };
 
   createClickActivityCollector(config, logger, createEventHandler);
--- a/src/components/DataCollector/index.js
+++ b/src/components/DataCollector/index.js
@@ -21,7 +21,7 @@ const createDataCollector = ({ config, logger }) => {
   let network;
   let optIn;
 
-  const makeServerCall = event => {
+  const makeServerCall = (event, exitLink) => {
     const payload = network.createPayload();
     payload.addEvent(event);
     payload.mergeMeta({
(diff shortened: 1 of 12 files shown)


src/components/DataCollector/index.js
```diff
@@ -49,15 +49,15 @@ const createDataCollector = ({ config }) => {
...
     const event = createEvent();
-    const { viewStart = false, data, meta } = options;
+    const { viewStart = false, exitLink = false, data, meta } = options;
 
     event.mergeData(data);
     event.mergeMeta(meta);
 
     return lifecycle
-      .onBeforeEvent(event, options, viewStart)
+      .onBeforeEvent(event, options, viewStart, exitLink)
```
jfkhoury: I wonder it it's time to change this signature to `{ event, options, viewStart, exitLink }`

src/components/Personalization/index.js
```diff
@@ -76,7 +76,8 @@ const createCollect = collect => {
     const notification = Object.assign({}, payload, { id, timestamp });
     const personalization = { notification };
 
-    collect({ meta: { personalization } });
+    const data = new Uint8Array(94 * 1024);
```
jfkhoury: This stuff can go away now; it was just for testing.

src/components/DataCollector/index.js
jfkhoury: @Aaronius @artur-ciocanu

Aaronius: I'm good either way. It's a little odd that `onBeforeDataCollection` would still just be `payload`. These decisions give me nightmares. Make it stop.

jfkhoury: I think moving to `({...})` now, for all hooks, can set us up for the future when we want to pass more params.

src/core/network/createNetwork.js
```diff
@@ -50,15 +50,23 @@ export default (config, logger, lifecycle, networkStrategy) => {
...
      * @returns {Promise} a promise resolved with the response object once the response is
      * completely processed.  If expectsResponse==false, the promise will be resolved
      * with undefined.
      */
-    sendRequest(payload, expectsResponse = true) {
+    sendRequest(payload, expectsResponse = true, exitLink = false) {
       const requestId = uuid();
+      if (expectsResponse && exitLink) {
+        logger.log(
+          `Attempt to get response during exit link.  Assuming no response is expected.`
```
Aaronius: I'm not sure if this would make sense or be helpful to a customer. I think they may wonder why we're telling them to "attempt to get response during exit link". Maybe "No response requested due to exit link." and always log it whenever `exitLink: true` regardless of the `expectsResponse` value? Just thinking out loud.

test/unit/specs/core/network/createNetworkStrategy.spec.js
```diff
@@ -186,6 +190,43 @@ describe("networkStrategyFactory", () => {
...
+                      body: largeRequestBody
+                    },
+                    1,
+                    1
+                  ) // verify atLeast 1, atMost 1 times called
+                  .then(done, done.fail);
+              })
+              .catch(done.fail);
+          });
+      });
```
Aaronius: I don't think there's any need to use `done` in these tests. Just return the resulting promise from the promise chain. If the promise gets resolved, Jasmine will consider it passed. If the promise gets rejected, Jasmine will consider it failed.

test/unit/specs/core/network/createNetworkStrategy.spec.js
```diff
@@ -186,6 +190,43 @@ describe("networkStrategyFactory", () => {
...
+              statusCode: 204
+            }
+          })
+          .then(() => {
+            networkStrategy(
+              "http://localhost:1080/myapi",
+              largeRequestBody,
+              true
+            )
+              .then(body => {
```
Aaronius: Can we flatten the promise chain?

test/unit/specs/core/network/createNetworkStrategy.spec.js
```diff
@@ -186,6 +190,43 @@ describe("networkStrategyFactory", () => {
...
+                  .verify(
+                    {
+                      method: "POST",
+                      path: "/myapi",
+                      body: largeRequestBody
+                    },
+                    1,
+                    1
+                  ) // verify atLeast 1, atMost 1 times called
+                  .then(done, done.fail);
```
Aaronius: Can we flatten the promise chain here too?

src/components/DataCollector/index.js
Aaronius: Upon further review, I agree with this. ☝️

PR #167, opened 2019-08-20
Feature/core 32301

jileeshs:
Allowing sha256 hashing of customer personal identifiers

## Description
Customers should be given an option to hash personal identifiers before sending.
We are allowing to pass customer ids like 
```
setCustomerIds: {
   namespace: {
      id,
      authState,
      hash
   }
}
```

## Related Issue

https://jira.corp.adobe.com/browse/CORE-32301

## Types of changes

## Checklist:


src/components/Identity/syncDeclaredIds.js
```diff
@@ -0,0 +1,52 @@
...
+  optIn
+) => {
+  const normalizedCustomerIds = normalizeCustomerIds(customerIds);
+  const idNames = Object.keys(normalizedCustomerIds);
+  const idsToHash = idNames.filter(idName => customerIds[idName].hash);
+  const hashedIds = idsToHash.map(id =>
+    sha256Buffer(normalizedCustomerIds[id].id)
+  );
+
+  Promise.all(hashedIds).then(values => {
```
jfkhoury: Don't we need a `return` statement here?

sandbox/public/index.html
```diff
@@ -61,17 +61,18 @@
       });
 
       alloy("setCustomerIds", {
-        "email" : {
+        "emailHash" : {
           "id": "me@gmail.com",
-          "authState": 0
+          "authState": 0,
+          "hash": true
```
jfkhoury: Let's make sure we document this change.

src/components/Identity/index.js
```diff
@@ -38,39 +39,22 @@ const createIdentity = ({ config, logger, cookieJar }) => {
...
-    payload.mergeMeta({ identityMap: { customerIdChanged } });
-    if (customerIdChanged) {
-      cookieJar.set(CUSTOMER_ID_HASH, customerIdsHash);
-    }
-    return lifecycle
-      .onBeforeEvent(event, {}, false) // FIXME: We shouldn't need an event.
-      .then(() => optIn.whenOptedIn())
-      .then(() => makeServerCall(payload));
+    syncDeclaredIds(
+      customerIds,
```
jfkhoury: How about we extract ALL the `setCustomerIds` logic in this module, instead of splitting the logic between here and `syncDeclaredIds`? Tell don't Ask right? logic in 2 places, even if it's only preparation can lead to bugs and confusion.

src/components/Identity/syncDeclaredIds.js
```diff
@@ -0,0 +1,52 @@
...
+    payload.mergeMeta({ identity: { customerIdChanged } });
+    if (!customerIdChanged) {
+      cookie.set(CUSTOMER_ID_HASH, customerIdsHash);
+    }
+    return lifecycle
+      .onBeforeEvent(event, {}, false) // FIXME: We shouldn't need an event.
+      .then(() => optIn.whenOptedIn())
+      .then(() => makeServerCall(payload, lifecycle, network));
+  });
+};
```
jonsnyder: You need to add automated tests for this logic.

There are a lot of concerns here which make it hard to unit test.  I wonder if creating a customerIds class would help.  I could see methods like `getChecksum`, `getNormalizedAndHashedIds`, This class could more easily be unit tested.

src/components/Identity/syncDeclaredIds.js
```diff
@@ -0,0 +1,52 @@
...
+  payload,
+  cookie,
+  lifecycle,
+  network,
+  optIn
+) => {
+  const normalizedCustomerIds = normalizeCustomerIds(customerIds);
+  const idNames = Object.keys(normalizedCustomerIds);
+  const idsToHash = idNames.filter(idName => customerIds[idName].hash);
+  const hashedIds = idsToHash.map(id =>
```
Aaronius: Maybe call this `idHashPromises` or something to indicate that they are promises and not the actual hashed Ids...

src/components/Identity/syncDeclaredIds.js
Aaronius: (continued from https://github.com/adobe/alloy/pull/167/files#r315935018) ...then rename `values` to `hashedIds` to be more specific.

src/components/Identity/syncDeclaredIds.js
```diff
@@ -0,0 +1,52 @@
...
+    sha256Buffer(normalizedCustomerIds[id].id)
+  );
+
+  Promise.all(hashedIds).then(values => {
+    const normalizedAndHashedIds = values.reduce((valueObj, value, index) => {
```
Aaronius: Then rename `value` here to be `hashedId` to be more specific.

src/components/Identity/syncDeclaredIds.js
Aaronius: I would prefer no es6 classes or prototypal "classes", please, though you might have meant something else. I would agree that this should be broken down into smaller chunks though. I would advocate separate modules for some of this logic. Like @jonsnyder said, `getChecksum` and `getNormalizedAndHashedIds` would probably be good modules, along with something like `detectCustomerIdChange` that would handle this part:

```
    idNames.forEach(idName => {
      payload.addIdentity(idName, normalizedAndHashedIds[idName]);
    });
    const customerIdsHash = crc32(
      JSON.stringify(normalizedCustomerIds)
    ).toString(36);

     const customerIdChanged = customerIdsHash === cookie.get(CUSTOMER_ID_HASH);
    payload.mergeMeta({ identity: { customerIdChanged } });
    if (!customerIdChanged) {
      cookie.set(CUSTOMER_ID_HASH, customerIdsHash);
    }
```

src/components/Identity/util.js
```diff
@@ -43,4 +43,37 @@ const normalizeCustomerIds = customerIds => {
...
+  // IE 11, whose result is a CryptoOperation object instead of a promise
+  return new Promise((resolve, reject) => {
+    result.addEventListener("complete", () => {
+      resolve(result.result);
+    });
+    result.addEventListener("error", () => {
+      reject();
+    });
+  });
+};
```
Aaronius: I feel like `sha256Buffer` and `bufferToHex` could eventually be useful to other components which, in my mind, qualifies it to be placed up in `core/utils`. I think other people might disagree though, so I've started a thread in the alloy-engineering Slack channel.

src/components/Identity/syncDeclaredIds.js
jonsnyder: I meant a class just like Payload or Event that we have now.

PR #169, opened 2019-08-21
Reworked click handler

liljenback:
## Description

Alloy click handler

## Motivation and Context

Initial click handler that collects against existing XDM schema

## Types of changes

## Checklist:

### DIFF
--- a/src/components/DataCollector/activity/click.js
+++ b/src/components/DataCollector/activity/click.js
@@ -15,7 +15,7 @@ import {
   isSupportedAnchorElement
 } from "../utils";
 
-const createClickHandler = (logger, collect) => {
+const createClickHandler = (window, logger, collect) => {
   return event => {
     // TODO: Consider safeguarding from the same object being clicked multiple times in rapid succession?
     let clickedObj = event.target;
@@ -50,6 +50,6 @@ export default (config, logger, collect) => {
   if (!enabled) {
     return;
   }
-  const clickHandler = createClickHandler(logger, collect);
+  const clickHandler = createClickHandler(window, logger, collect);
   document.addEventListener("click", clickHandler, true);
 };
--- a/src/components/DataCollector/activity/click.js
+++ b/src/components/DataCollector/activity/click.js
@@ -32,12 +32,16 @@ const createClickHandler = (window, logger, collect) => {
       // TODO: Update name (link name) and support exit, other, and download link types
       collect({
         data: {
-          eventType: "web.webinteraction",
-          name: "Link Click",
-          type: "other",
-          URL: linkUrl,
-          linkClicks: {
-            value: 1
+          eventType: "web.webinteraction.linkClicks",
+          web: {
+            webinteraction: {
+              name: "Link Click",
+              type: "other",
+              URL: linkUrl,
+              linkClicks: {
+                value: 1
+              }
+            }
           }
         }
       });
--- a/src/components/DataCollector/activity/click.js
+++ b/src/components/DataCollector/activity/click.js
@@ -48,7 +48,6 @@ const createClickHandler = (logger, collect) => {
 export default (config, logger, collect) => {
   const enabled = config.get("clickCollectionEnabled");
   if (!enabled) {
-    logger.log("Click activity collection is disabled");
     return;
   }
   const clickHandler = createClickHandler(logger, collect);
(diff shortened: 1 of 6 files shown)


src/components/DataCollector/activity/click.js
```diff
@@ -0,0 +1,56 @@
...
+        }
+      });
+    }
+  };
+};
+
+export default (config, logger, collect) => {
+  const enabled = config.get("clickCollectionEnabled");
+  if (!enabled) {
+    logger.log("Click activity collection is disabled");
```
jfkhoury: I don't think this log is necessary, especially that it's the customer that configured the `clickCollectionEnabled` in the first place.

src/components/DataCollector/activity/click.js
```diff
@@ -0,0 +1,56 @@
...
+    return;
+  }
+  const clickHandler = createClickHandler(logger, collect);
+  document.addEventListener("click", clickHandler, true);
```
jfkhoury: Should we inject the `document` or `window` instead of referring to the global ones?

src/components/DataCollector/index.js
```diff
@@ -60,6 +61,8 @@ const createDataCollector = ({ config }) => {
       .then(() => makeServerCall(event));
   };
 
+  createClickActivityCollector(config, logger, createEventHandler);
```
jfkhoury: Maybe we should create the `Click Collector` if it's not enabled in the config? Which means you won't need to pass `config`?

src/components/DataCollector/activity/click.js
liljenback: I do this in some of the utilities. I'm not sure if it is needed here. What would be the use case?

src/components/DataCollector/index.js
liljenback: I don't think we want to attach an event handler if it's not enabled.

src/components/DataCollector/activity/click.js
liljenback: Good point, removing ...

src/components/DataCollector/activity/click.js
jfkhoury: easier to test

src/components/DataCollector/index.js
jfkhoury: :) Sorry I meant maybe we should NOT create.... not that important though.

src/components/DataCollector/activity/click.js
liljenback: ok, adding it to the signature. I think this part will need TestCafe tests though.

src/components/DataCollector/activity/click.js
```diff
@@ -0,0 +1,55 @@
...
+    while (clickedObj && clickedObj !== document.body && !linkUrl) {
+      clickedObj = clickedObj.parentElement || clickedObj.parentNode;
+      if (clickedObj) {
+        linkUrl = getAbsoluteUrlFromAnchorElement(window, clickedObj);
+      }
+    }
+    if (linkUrl && isSupportedAnchorElement(clickedObj)) {
+      // TODO: Update name (link name) and support exit, other, and download link types
+      collect({
+        data: {
```
jonsnyder:
```
data: {
  eventType: "web.webinteraction.linkClicks",
  web: {
    webinteraction: {
      name: "Link Click",
      type: "other",
      URL: linkUrl,
      linkClicks: {
        value: 1
      }
    }
  }
};
```

src/components/DataCollector/activity/click.js
```diff
@@ -0,0 +1,55 @@
...
+          eventType: "web.webinteraction",
+          name: "Link Click",
```
jonsnyder: Should we extract the text of the link and put it here as the name?

src/components/DataCollector/index.js
```diff
@@ -85,6 +87,9 @@ createDataCollector.configValidators = {
   },
   imsOrgId: {
     validate: required
+  },
+  clickCollectionEnabled: {
```
jonsnyder: this is `linkClickCollection` right, or will it expand into clickCollection?

liljenback: It is link-click-collection. I didn't want to reduce the scope and kept it as click. But I also don't want to make it misleading. Once we know the other properties that it needs to collect we can revisit the naming convention.

src/components/DataCollector/activity/click.js
liljenback: I'll do that in a separate PR once this is merged. There are other features that needs to be added as well.

PR #179, opened 2019-08-29
allowing to pass declared ids from config

jileeshs:
## Description

Some customers might have users id info during config itself,
This change will allow to pass these IDs while config
```
alloy("configure", {
        propertyId: 9999999,
        customerIds: {
          email: {
            id: "you@gmail.com",
              "authState": 0,
              "hash": false 
            }
         }
      });
```
## Related Issue

https://jira.corp.adobe.com/browse/CORE-32507

## Types of changes

## Checklist:

### DIFF
--- a/sandbox/public/index.html
+++ b/sandbox/public/index.html
@@ -46,14 +46,7 @@
         // By checking this params we know we are in authoring mode
         authoringModeEnabled: document.location.href.indexOf("mboxEdit") !== -1,
         prehidingStyle: "section:nth-of-type(1) { opacity: 0 !important }",
-        idSyncContainerId: 81,
-        "customerIds": {
-          "email" : {
-          "id": "you@gmail.com",
-            "authState": 0,
-            "hash": false 
-          }
-         }
+        idSyncContainerId: 81, 
       });
 
       alloy("event", {
@@ -72,29 +65,21 @@
           "id": "me@gmail.com",
           "authState": 0,
           "hash": true //TODO: document customer ID hashing syntax
-        },
-        "crm" : {
+        },"crm" : {
           "id": "1234",
           "authState": 0
         }
       }).then(function() {
         console.log("Sandbox: Set customer IDs event has completed.");
       });
-
+      
       // For Testing multiple instances
       organizationTwo("configure", {
         // edgeDomain: "edgegateway.azurewebsites.net",
         propertyId: 8888888,
         imsOrgId: "53A16ACB5CC1D3760A495C99@AdobeOrg",
         logEnabled: true,
-  clickCollectionEnabled: false,
-  "customerIds": {
-          "email" : {
-          "id": "youtoo@gmail.com",
-            "authState": 0,
-            "hash": false 
-          }
-         }
+        clickCollectionEnabled: false, 
       });
     </script>
   </head>
--- a/sandbox/public/index.html
+++ b/sandbox/public/index.html
@@ -46,7 +46,14 @@
         // By checking this params we know we are in authoring mode
         authoringModeEnabled: document.location.href.indexOf("mboxEdit") !== -1,
         prehidingStyle: "section:nth-of-type(1) { opacity: 0 !important }",
-        idSyncContainerId: 81
+        idSyncContainerId: 81,
+        "customerIds": {
+          "email" : {
+          "id": "you@gmail.com",
+            "authState": 0,
+            "hash": false 
+          }
+         }
       });
 
       alloy("event", {
@@ -61,7 +68,7 @@
       });
 
       alloy("setCustomerIds", {
-        "emailHash" : {
+        "Email_LC_SHA256" : {
           "id": "me@gmail.com",
           "authState": 0,
           "hash": true //TODO: document customer ID hashing syntax
@@ -80,7 +87,14 @@
         propertyId: 8888888,
         imsOrgId: "53A16ACB5CC1D3760A495C99@AdobeOrg",
         logEnabled: true,
-	clickCollectionEnabled: false
+  clickCollectionEnabled: false,
+  "customerIds": {
+          "email" : {
+          "id": "youtoo@gmail.com",
+            "authState": 0,
+            "hash": false 
+          }
+         }
       });
     </script>
   </head>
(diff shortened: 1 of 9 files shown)


sandbox/public/index.html
```diff
@@ -80,7 +87,14 @@
         propertyId: 8888888,
         imsOrgId: "53A16ACB5CC1D3760A495C99@AdobeOrg",
         logEnabled: true,
-	clickCollectionEnabled: false
+  clickCollectionEnabled: false,
+  "customerIds": {
```
Aaronius: Looks like there's some weird indentation going on here.

src/components/Identity/addCustomerIdsToPayload.js
```diff
@@ -0,0 +1,24 @@
+import { assign } from "../../utils";
+import { validateCustomerIds } from "./util";
+import processCustomerIds from "./processCustomerIds";
+
+export default (ids, cookieJar, payload) => {
+  validateCustomerIds(ids);
+  const customerIds = assign({}, ids);
```
Aaronius: Maybe just use es6 for this:

```
const customerIds = { ...ids };
```

src/components/Identity/index.js
```diff
@@ -90,7 +89,7 @@ const createIdentity = ({ config, logger, cookieJar }) => {
             logger.log("Delaying request while retrieving ECID from server.");
             promise = deferredForEcid.promise.then(() => {
               logger.log("Resuming previously delayed request.");
-              addIdsContext(payload, getEcid());
+              addIdsContext(payload, getEcid(), config);
```
Aaronius: It doesn't look like passing `config` here is doing anything useful. Old code?

src/components/Identity/setCustomerIds.js
```diff
@@ -10,30 +8,14 @@ const makeServerCall = (payload, lifecycle, network) => {
...
-      payload.mergeMeta({ identity: { customerIdChanged } });
-      if (customerIdChanged) {
-        customerIdsProcess.updateChecksum(cookieJar);
-      }
-      return lifecycle
-        .onBeforeEvent(event, {}, false) // FIXME: We shouldn't need an event.
-        .then(() => optIn.whenOptedIn())
-        .then(() => makeServerCall(payload, lifecycle, network));
-    });
+  addCustomerIdsToPayload(ids, cookieJar, payload).then(() => {
```
Aaronius: I believe we should be returning this promise chain for a couple reasons: (1) When a customer executes `setCustomerIds`, I think it would be appropriate to resolve the promise returned to the customer _after_ the event has been sent to the server and (2) if an error is thrown anywhere in this promise chain, we want to reject the promise returned to the customer.

jfkhoury:
@jileeshs Let's hold off on commenting or making changes to this PR until we talk please. Thanks!

jfkhoury:
Latest update/decisions:

1. We won't be supporting `customerIds` as config, because the current thinking is that we should send customer IDs to the server as soon as we receive them; which means, the config has to either trigger a server call automatically (which is something we don't want to support in Alloy), or behave differently than the `setCustomerIds`, which can be confusing and error prone.

2. As long as we have customer IDs in a session, we should be sending them on every event, and use the meta flag to declare any changes in the customer IDs, mostly for optimization.

Thank you @jileeshs & @Aaronius !

src/components/Identity/addCustomerIdsToPayload.js
```diff
@@ -0,0 +1,24 @@
+import { assign } from "../../utils";
```
jonsnyder: Consider adding a method to `processCustomerIds` called `addToPayload(cookieJar, payload)`

test/unit/specs/components/Identity/addCustomerIdsToPayload.spec.js
```diff
@@ -0,0 +1,91 @@
...
+      expect(payload.addIdentity.calls.count()).toBe(2);
+      expect(payload.addIdentity).toHaveBeenCalledWith("Email_LC_SHA256", {
+        id: "me@gmail.com",
+        authState: 0
+      });
+      expect(payload.addIdentity).toHaveBeenCalledWith("crm", {
+        id: "1234",
+        authState: 0
+      });
+      expect(payload.mergeMeta).toHaveBeenCalled();
```
jonsnyder: `toHaveBeenCalledWith` makes sure the function was called, so this line is not needed.

test/unit/specs/components/Identity/addCustomerIdsToPayload.spec.js
```diff
@@ -0,0 +1,91 @@
...
+      get: jasmine.createSpy().and.returnValue("807ct1"),
+      set: jasmine.createSpy()
+    };
+
+    const test = addCustomerIdsToPayload(idsWithoutHash, cookieJar, payload);
+    test.then(() => {
+      expect(cookieJar.get).toHaveBeenCalledWith(CUSTOMER_ID_HASH);
+      expect(cookieJar.set).not.toHaveBeenCalled();
+      expect(payload.addIdentity).toHaveBeenCalled();
+      expect(payload.addIdentity.calls.count()).toBe(2);
```
jonsnyder: Likewise, line 75 and 76 are not needed.

test/unit/specs/components/Identity/addCustomerIdsToPayload.spec.js
```diff
@@ -0,0 +1,91 @@
...
+      addIdentity: jasmine.createSpy(),
+      mergeMeta: jasmine.createSpy()
+    };
+    const cookieJar = {
+      get: jasmine.createSpy().and.returnValue("807ct1"),
+      set: jasmine.createSpy()
+    };
+
+    const test = addCustomerIdsToPayload(idsWithoutHash, cookieJar, payload);
+    test.then(() => {
```
jonsnyder: Your test should return this promise so that Jasmine knows to wait for the promise to complete before marking the test as good.

test/unit/specs/components/Identity/addCustomerIdsToPayload.spec.js
```diff
@@ -0,0 +1,91 @@
+import addCustomerIdsToPayload from "../../../../../src/components/Identity/addCustomerIdsToPayload";
+import { COOKIE_NAMES } from "../../../../../src/components/Identity/constants";
+
+const { CUSTOMER_ID_HASH } = COOKIE_NAMES;
+
+describe("Identity::addCustomerIdsToPayload", () => {
+  it("should call validate customerIds", () => {
```
jonsnyder: See my comments on the other test case as they apply here as well.

jileeshs:
Closing as we are not having this feature

PR #182, opened 2019-09-03
CORE-32819 Enforce unique propertyId and imsOrgId across instances

jonsnyder:
## Description

Add config validators to make sure propertyId and imsOrgId are unique across instances

## Related Issue

https://jira.corp.adobe.com/browse/CORE-32819

## Motivation and Context

Due to problems with overlapping cookies, Alloy doesn't support duplicated propertyIds or imsOrgIds across instances.

## Types of changes

## Checklist:


jonsnyder:
@jfkhoury The sandbox has two instances configured with the same imsOrgId.  Do we have another orgId we can test with?

src/components/DataCollector/createConfigValidators.js
```diff
@@ -0,0 +1,36 @@
...
+*/
+
+import {
+  required,
+  validDomain,
+  allValidate,
+  unique
+} from "../../utils/configValidators";
+
+export default (propertyIdValues, orgIdValues) => {
```
Aaronius: Is there any benefit to these parameters rather than just creating the arrays inside the function?

jonsnyder: It makes testing easier.

Aaronius: How? It seems we're passing in `[], []` everywhere.

jonsnyder: Correct, but each test is isolated because they all use different array instances.

Aaronius: That would also be the case if we just created the arrays inside this function, no?

jonsnyder: Oh, I see we could just have empty arrays in the function.  Yes that would work.

src/utils/configValidators/allValidate.js
```diff
@@ -0,0 +1,29 @@
...
+export default (...validators) => {
+  if (validators.length < 2) {
+    throw Error("allValidate requires at least 2 validators");
+  }
+
+  return (...args) => {
+    let error;
+    let i;
+    for (i = 0; i < validators.length; i += 1) {
+      error = validators[i](...args);
```
Aaronius: Consider doing `const error = ...` here instead of declaring `error` away from its usage. You could do the same thing with `for (let i`. I used to do what you've done here with `var`, but stopped with `let`/`const`. I suspect you may have a different perspective on these things though, which is fine.

src/utils/configValidators/createUnique.js
```diff
@@ -0,0 +1,22 @@
...
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+export default () => {
+  const values = [];
```
Aaronius: Even better!

src/components/DataCollector/createConfigValidators.js
```diff
@@ -0,0 +1,36 @@
...
+  return {
+    propertyId: {
+      validate: allValidate(required, createUnique())
+    },
+    edgeDomain: {
+      validate: validDomain,
+      defaultValue: "alpha.konductor.adobedc.net"
+    },
+    imsOrgId: {
+      validate: allValidate(required, createUnique())
```
Aaronius: I'd rather see this type of format: `validate: string().required().unique()`, but there's a JIRA issue to make validators more composable that could cover that work if we choose to follow that format: https://jira.corp.adobe.com/browse/CORE-32266. One of the advantages of that format is that it naturally scopes certain validation functionality to its respective value type. In other words, as a developer I can do this:

`validate: number().minimum(500)`

but not this:

`validate: string().minimum(50)`

`minimum` doesn't really apply to a string value, but it does a number value. If one of us tried to do `string().minimum(50)`, we'd get an error saying that `minimum` is undefined, which is what I would want. Whereas, if we did `allValidate(minimum(50), string)` our `minimum` validator would require type checking (or worse, implicitly cast the value to a number), which should be the `string` validator's job instead.

jfkhoury: @Aaronius / @jonsnyder, do you want to implement this change in this PR and get it over with?

Aaronius: The amount of refactoring does seem to be mounting. I'd like to see it done sooner rather than later, whether @jonsnyder does it or I do it.

src/utils/configValidators/allValidate.js
jonsnyder: This is a for loop not a forEach, so I can't do `const error = ...`.  I'm using a for loop because I want to short circuit the return.  I will put the let inside the for.

Aaronius: You can use const in a `for` loop since the `for` iteration creates a new scope for `const`.

src/components/DataCollector/createConfigValidators.js
jonsnyder: I'll take a stab at it now.

liljenback: There is an old commit in the validation logic that accepts an array of validators. We could resurrect that one to support what allValidate does.

src/utils/configValidators/createUnique.js
liljenback: Will this be tied to a specific key?

jonsnyder: Yes, each key has its own unique set.  So if you had a property id that matched a ims org id, it would still validate.  When you call `createUnique()`, it creates a new values array.  So when this is called as part of the data collector validator initialization it will create one instance for propertyId, and one for imsOrgId.

src/utils/configValidators/boolean.js
```diff
@@ -11,9 +11,12 @@ governing permissions and limitations under the License.
...
 import isBoolean from "../isBoolean";
+import createExpected from "./createExpected";
+
+const expected = createExpected("true or false");
 
 export default (key, currentValue) => {
   return currentValue === undefined || isBoolean(currentValue)
     ? ""
-    : `Value for ${key} is not a boolean: ${currentValue}`;
+    : expected(key, currentValue);
```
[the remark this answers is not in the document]
jonsnyder: Yes, I don't particularly like the empty string meaning valid either.  I have `expected` use the same interface as the other validators; it just always returns an error.

src/utils/configValidators/createMinimum.js
```diff
@@ -0,0 +1,20 @@
...
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+import createExpected from "./createExpected";
+
+export default minimum => (key, currentValue) => {
+  const expected = createExpected(
+    `an integer greater than or equal to ${minimum}`
```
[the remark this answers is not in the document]
jonsnyder: I have integer because this is only available on the integer validator, but if we had `number().minimum(3)` then we would need to change it.

test/unit/specs/utils/configValidators/integer.spec.js
```diff
@@ -0,0 +1,36 @@
...
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+import { integer } from "../../../../../src/utils/configValidators";
+
+describe("configValidator::integer", () => {
+  [true, "", "foo", [1], {}].forEach(value => {
```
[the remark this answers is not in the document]
jonsnyder: 👍

src/utils/configValidators/createMinimum.js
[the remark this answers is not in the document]
Aaronius: Actually, that brings up a different problem I didn't notice before. In my opinion `integer` shouldn't be a top-level validator; `number` should be. So the usage should be `number.integer().minimum(...)`. The top-level validators should correspond to JavaScript types: number(), boolean(), string(), array(), object(), etc. As such, `unique` shouldn't be a top-level validator either; it should just be a method attached to all the top-level validators.

PR #184, opened 2019-09-04
CORE-30233 Implement manual ID sync

wgamboa:
## Related Issue
CORE-30233

## Types of changes

## Checklist:


sandbox/public/index.html
```diff
@@ -74,6 +74,23 @@
         console.log("Sandbox: Set customer IDs event has completed.");
       });
 
+      alloy("syncIdByUrl", {
```
Aaronius: Should this be named `syncIdsByUrl` (plural)?

Aaronius: Are we going to be providing other commands that are like `syncIdsBySomething`? If so, why are we making them separate commands? If not, then should this one be named something more generic, like `syncIds` or `manuallySyncIds` or something like that? Are URL-based IDs the only identities that can be synced "manually"? I'm having a hard time understanding the overall/long-term vision here.

src/components/Identity/createManualIdSyncs.js
```diff
@@ -0,0 +1,76 @@
...
+  return Promise.resolve({
+    message: "[syncIdByUrl] No ID syncs were in queue"
+  });
+};
+
+export default (config, logger, cookieJar) => {
+  const idSyncsProcessor = createIdSyncs(config, logger, cookieJar);
+
+  return {
+    syncIdByUrl: createSyncIdByUrl(config, logger, cookieJar, idSyncsProcessor)
```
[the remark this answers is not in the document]
wgamboa: Yes. syncIdsByDataSource. There's no endpoint for it yet.

src/components/Identity/createIdSyncs.js
```diff
@@ -91,10 +93,14 @@ const createProcessor = (config, logger, cookieJar) => destinations => {
           convertTimes(DAY, HOUR, 7)
         ).toString(36)
       );
+
+      return Promise.resolve(result);
```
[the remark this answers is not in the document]
wgamboa: Initially just for testing, but it would also be useful for customers to see if the command returns an error.

src/components/Identity/createIdSyncs.js
```diff
@@ -38,7 +38,9 @@ const setControlObject = (controlObject, cookieJar) => {
 
 const createProcessor = (config, logger, cookieJar) => destinations => {
   if (!config.idSyncsEnabled) {
-    return Promise.resolve();
+    return Promise.resolve({
+      message: "ID syncs not enabled"
+    });
```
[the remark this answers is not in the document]
wgamboa: I did `return Promise.reject("ID syncs not enabled");`
And I got an alloy log error as well as an uncaught exception. Is that normal? I don't remember our previous discussion clearly.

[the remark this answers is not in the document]
wgamboa: @Aaronius Please see the previous comment.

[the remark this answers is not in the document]
Aaronius: Reject with an error instead. `return Promise.reject(new Error("ID syncs not enabled"));`

src/components/Identity/createManualIdSyncs.js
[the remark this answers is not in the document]
Aaronius: Can't we use the same command (the command currently named `syncIdByUrl`) for those IDs as well, considering each ID sync object has a `type` property on it?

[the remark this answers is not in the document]
wgamboa: The type is always URL. The second command takes an additional field called data source unique user ID. Together with the ID, the URL is constructed from this. So the validation will also be different. This is how it currently works in the ID service anyway.

src/components/Identity/createManualIdSyncs.js
```diff
@@ -0,0 +1,76 @@
...
+const createSyncIdByUrl = (
+  config,
+  logger,
+  cookieJar,
+  idSyncsProcessor
+) => data => {
+  let hasInvalidIdSyncs = false;
+
+  const idSyncs = (data.idSyncsArray || []).filter((idSync = {}) => {
+    if (
```
jonsnyder: I wonder if we could re-use the config validation code to perform this validation.  I'm working on the config stuff right now so it is top of mind, but I think this is something we could do in the future.

test/unit/specs/components/Identity/createManualIdSyncs.spec.js
```diff
@@ -0,0 +1,94 @@
+import createCookieProxy from "../../../../../src/core/createCookieProxy";
```
jonsnyder: In general, I'd like to avoid making http requests to actual servers in tests.  This means the test will fail if the rlcdn.com domain goes down or stops accepting requests.  If you pulled out the processor creation to Identity/index.js, you could pass in a mock idSyncProcessor for this test.

wgamboa: @jonsnyder I wouldn't feel it's a true test though. In the case with this code, if the domain stopped responding, It would just return the failed id sync in the failed array instead of the succeeded array. This doesn't affect my test.

wgamboa: @jonsnyder Sorry, that will make the test fail I just realized. I'll work on a mock.

wgamboa: @jonsnyder @Aaronius @jfkhoury from Aaron's Explanation of What a unit test is, It sounds like I shouldn't be testing idSyncs.process at all in createManualIdSyncs.spec.js and createIdSyncs.spec.js Because it's a subprocess in these files. So I guess I'll reserve idSyncs.process For future functional testing. I'll leave in the unit tests that actually apply.

src/components/Identity/createManualIdSyncs.js
```diff
@@ -0,0 +1,60 @@
...
+    return Promise.reject(
+      new Error("syncIdsByUrl was passed one or more invalid id syncs.")
+    );
+  }
+
+  return idSyncsProcessor.process(idSyncs);
+};
+
+export default (config, logger, cookieJar) => {
+  const idSyncsProcessor = createIdSyncs(config, logger, cookieJar);
```
Aaronius: Like @jonsnyder was saying this morning, it would be best of you created `idSyncsProcessor` up in `Identity/index.js` and then passed it into this function. That way, in your unit test you can mock `idSyncsProcessor` and test that a call to `syncIdsByUrl` with valid IDs results in a call to `idSyncsProcessor.process`.

src/components/Identity/createManualIdSyncs.js
```diff
@@ -0,0 +1,60 @@
...
+      !isObject(idSync.spec) ||
+      !isNonEmptyString(idSync.spec.url) ||
+      (idSync.spec.hideReferrer !== 0 && idSync.spec.hideReferrer !== 1) ||
+      (idSync.spec.ttlMinutes !== undefined &&
+        !isNumber(idSync.spec.ttlMinutes))
+    ) {
+      return false;
+    }
+
+    return true;
```
Aaronius: There's no need for the `if` statement here. Just do:

```
return isObject(idSync) && idSync.type === "url" && isNumber(idSync.id)...
```

test/unit/specs/components/Identity/createIdSyncs.spec.js
```diff
@@ -57,11 +46,7 @@ describe("Identity::createIdSyncs", () => {
...
     const idSyncs = createIdSyncs(config, logger, cookieJar);
     return idSyncs.process(idsToSync).then(() => {
-      const now = Math.round(new Date().getTime() / 1000 / 60 / 60); // hours
-
       obj = getControlObject();
-      expect(obj[411]).toBeDefined();
-      expect(obj[411] - now).toEqual(2);
       expect(obj[123]).toBeUndefined();
     });
   });
```
Aaronius: Let's talk about how to add more tests for `createIdSyncs` sometime. You can probably add some more useful unit tests but I need to dig into the code some more in order to offer guidance. I wouldn't let it prevent this PR from being merged though.

src/components/Identity/createIdSyncs.js
```diff
@@ -61,40 +61,38 @@ const createProcessor = (config, logger, cookieJar) => destinations => {
...
+
+    cookieJar.set(
+      ID_SYNC_TIMESTAMP,
+      (
+        Math.round(convertTimes(MILLISECOND, HOUR, new Date().getTime())) +
+        convertTimes(DAY, HOUR, 7)
+      ).toString(36)
+    );
+
+    return Promise.resolve(result);
```
Aaronius: You're already inside a `.then`, so there's not need to return a promise. Just do `return result;`

src/components/Identity/createManualIdSyncs.js
```diff
@@ -0,0 +1,60 @@
...
+  });
+
+  if (!areAllIdSyncsValid) {
+    return Promise.reject(
+      new Error("syncIdsByUrl was passed one or more invalid id syncs.")
+    );
+  }
```
Aaronius: We should probably give more information about which ID sync(s) is problematic. Here's what I would do. I would break your logic that validates a single ID sync into a function called something like `isIdSyncValid()`. I would then use the `find` utility to find the first ID sync that's invalid. I would then report the ID of the invalid ID sync in the error. Something like this:

```
const invalidIdSync = find(idSyncs, idSync => !isIdSyncValid(idSync));

if (invalidIdSync) {
  return Promise.reject(
      new Error("An invalid ID sync with the ID of ${invalidIdSync.id} was passed to syncIdsByUrl.")
    );
}
```

src/components/Identity/createIdSyncs.js
Aaronius: If the result has failures, shouldn't we reject the promise provided to the customer since at least part of the ID syncing actually failed? I would think if all ID syncs were successful, we should resolve the promise with no resolution value. If any of the ID syncs failed, we should reject the promise with an error that probably contains some details about which ID syncs failed (maybe list out the IDs in the error message).

src/components/Identity/createManualIdSyncs.js
wgamboa: Thanks. No need for return either.

src/components/Identity/createIdSyncs.js
wgamboa: I'm testing to see if all id syncs were processed, whether successful or failed, in order to know when to remove the iframe.

wgamboa:
Merging with two approvals.

PR #185, opened 2019-09-05
Feature/core 32507

jileeshs:
Send customer Ids in every request if it is available in a session

## Description

The current implementation of setCustomerIds would send the declared Ids only once to the server.
The backend team mentioned that it is beneficial to have this data with evrey payload

## Types of changes

## Checklist:


src/components/Identity/index.js
```diff
@@ -35,6 +36,9 @@ const createIdentity = ({ config, logger, cookieJar }) => {
   let lifecycle;
   const idSyncs = createIdSyncs(config, logger, cookieJar);
   let alreadyQueriedForIdSyncs = false;
+  // Defining initial state of customerIds
+  let customerIdState = {};
```
jfkhoury: Do we need both? isn't a non-empty state an indication that we have customer IDs?

src/components/Identity/index.js
```diff
@@ -74,10 +78,18 @@ const createIdentity = ({ config, logger, cookieJar }) => {
       },
       // Waiting for opt-in because we'll be reading the ECID from a cookie
       // TO-DOCUMENT: We wait for ECID before trigger any events.
-      onBeforeDataCollection(payload) {
+      onBeforeDataCollection(payload, processedIds) {
```
jfkhoury: It feels incorrect to pass the ids into this hook, since this hook is app wide; Hooks' signatures are predefined so that components know how to implement them.

src/components/Identity/setCustomerIds.js
```diff
@@ -1,10 +1,9 @@
-import { assign } from "../../utils";
 import createEvent from "../DataCollector/createEvent";
 import { validateCustomerIds } from "./util";
 import processCustomerIds from "./processCustomerIds";
 
-const makeServerCall = (payload, lifecycle, network) => {
-  return lifecycle.onBeforeDataCollection(payload).then(() => {
+const makeServerCall = (payload, processedIds, lifecycle, network) => {
```
jonsnyder: I agree with Joe's comment above about not passing processedIds into the onBeforeDataCollection.  Could you add the customerIds to the payload here?

test/unit/specs/components/Identity/processCustomerIds.spec.js
```diff
@@ -120,7 +120,7 @@ describe("Identity::processCustomerIds", () => {
 
       const normalizedIds = normalizeCustomerIds(ids);
       const idSyncProcess = processCustomerIds(ids);
-      idSyncProcess.getNormalizedAndHashedIds().then(normalized => {
+      return idSyncProcess.getNormalizedAndHashedIds().then(normalized => {
```
jonsnyder: 👍

src/components/Identity/index.js
jileeshs: Yes, i was trying to avoid the expensive operation to determine if the object is empty - and it would happen on every request.

src/components/Identity/setCustomerIds.js
jileeshs: I could, do you suggest I add from here and use payload.toJSON to get customerIds inside the onbeforeHooke?

jonsnyder: Why do you need the customerIds in the onBeforeHook?

src/components/Identity/customerIds/createCustomerIds.js
```diff
@@ -0,0 +1,86 @@
...
+
+        lifecycle
+          .onBeforeEvent(event, {}, false) // FIXME: We shouldn't need an event.
+          .then(() => optIn.whenOptedIn())
+          .then(() => makeServerCall(payload));
+      });
+    }
+  };
+
+  return customerIds;
```
jfkhoury: One object to rule them all?

src/components/Identity/customerIds/createCustomerIds.js
```diff
@@ -34,7 +34,9 @@ export default (cookieJar, lifecycle, network, optIn) => {
   };
 
   const makeServerCall = payload => {
-    return network.sendRequest(payload, payload.expectsResponse);
+    return lifecycle.onBeforeDataCollection(payload).then(() => {
```
jfkhoury: This whole flow might change because setCustomerIDs might not be considered an event to start with. So the createCustomerIds might get injected with the ECID getter, and it adds it instead of calling the lifecycle.

jfkhoury:
@jileeshs I added 2 commits. Let me know what you think.

@Aaronius @jonsnyder

src/components/Identity/index.js
```diff
@@ -110,12 +99,8 @@ const createIdentity = ({ config, logger, cookieJar }) => {
             deferredForEcid = defer();
             payload.expectResponse();
           }
-          if (haveCustomerIds) {
-            addCustomerIdsToPayload(
-              customerIdState,
-              customerIdChanged,
-              payload
-            );
+          if (customerIds.state.hasIds) {
```
jonsnyder: This feels like we are violating encapsulation of the customerIds' state.  I'd recommend putting this check inside the `addToPayload` method.

src/components/Identity/customerIds/createCustomerIds.js
jonsnyder: Love it!

src/components/Identity/customerIds/createCustomerIds.js
```diff
@@ -0,0 +1,88 @@
...
+  };
+
+  const makeServerCall = payload => {
+    return lifecycle.onBeforeDataCollection(payload).then(() => {
+      return network.sendRequest(payload, payload.expectsResponse);
+    });
+  };
+
+  const customerIds = {
+    state: {
```
jonsnyder: It seems like `state` and `setState` could be encapsulated here (i.e. not returned as part of the public interface for this object)

src/components/Identity/customerIds/createCustomerIds.js
```diff
@@ -0,0 +1,88 @@
+import { normalizeCustomerIds, validateCustomerIds } from "../util";
+import {
+  crc32,
+  convertBufferToHex,
+  convertStringToSha256Buffer
+} from "../../../utils";
+import { COOKIE_NAMES } from "../constants";
+import createEvent from "../../DataCollector/createEvent";
```
jonsnyder: I don't think we should have dependencies across components.  Perhaps we should move createEvent into core?

src/components/Identity/customerIds/createCustomerIds.js
```diff
@@ -0,0 +1,88 @@
+import { normalizeCustomerIds, validateCustomerIds } from "../util";
```
jonsnyder: Can we move all things customerId into the same directory?

src/components/Identity/index.js
```diff
@@ -78,18 +76,9 @@ const createIdentity = ({ config, logger, cookieJar }) => {
       },
       // Waiting for opt-in because we'll be reading the ECID from a cookie
       // TO-DOCUMENT: We wait for ECID before trigger any events.
-      onBeforeDataCollection(payload, processedIds) {
+      onBeforeDataCollection(payload) {
```
jonsnyder: 👍

src/components/Identity/customerIds/createCustomerIds.js
Aaronius: I agree with this.

jfkhoury: I agree!

jfkhoury:
Thank you @Aaronius @jonsnyder.

@jileeshs I agree with all the new comments. I will leave it to you to incorporate them, or comment on them first. Thanks.

Aaronius:
I still plan on giving a full review, but @jonsnyder's comments sounds pretty good so feel free to start making those changes and I'll review ASAP.

src/components/Identity/customerIds/createCustomerIds.js
Aaronius: Considering we are planning on not using an event at all once the new APIs open up (that's my understanding anyway), I don't mind keeping this as it is for now if you want. IMO it would be nice if we can later keep DataCollector the only component that creates events.

src/components/Identity/index.js
```diff
@@ -124,13 +127,8 @@ const createIdentity = ({ config, logger, cookieJar }) => {
       getEcid() {
         return optIn.whenOptedIn().then(getEcid);
       },
-      // TODO: Discuss renaming of CustomerIds to UserIds
       setCustomerIds(options) {
-        return optIn
-          .whenOptedIn()
```
Aaronius: Where did the `whenOptedIn()` go? It looks like we're no longer waiting for opt-in before reading the CUSTOMER_ID_HASH cookie within `customerIds.sync`.

test/unit/specs/components/Identity/addCustomerIdsToPayload.spec.js
```diff
@@ -0,0 +1,65 @@
...
+//     });
+//     expect(payload.addIdentity).toHaveBeenCalledWith("crm", {
+//       id: "1234",
+//       authState: 0
+//     });
+//     expect(payload.mergeMeta).toHaveBeenCalledWith({
+//       identity: { customerIdChanged: false }
+//     });
+//   });
+// });
```
Aaronius: I assume you're planning on cleaning up these tests, but just in case, be sure to clean up these tests.

src/components/Identity/util.js
```diff
@@ -43,4 +43,11 @@ const normalizeCustomerIds = customerIds => {
   }, {});
 };
 
-export { validateCustomerIds, normalizeCustomerIds };
+const updateCustomerIdState = (currentState, newState) => {
+  return {
+    ...currentState,
+    ...newState
+  };
+};
```
Aaronius: I don't see this being used anywhere.

Aaronius: I probably would advise removing it anwyay as it's simple enough and generic enough that I don't think it warrants being a utility.

test/unit/specs/components/Identity/util.spec.js
```diff
@@ -100,4 +111,49 @@ describe("Identity::identityUtil", () => {
...
+          id: "me@gmail.com",
+          authState: 0
+        }
+      };
+      const newState = updateCustomerIdState({}, state);
+      expect(newState).not.toBe(state);
+      expect(newState).toEqual(state);
+    });
+  });
+  it("should append values to an existing state", () => {
```
Aaronius: Would "merge" be more appropriate than "append" here?

PR #186, opened 2019-09-06
CORE-33593 Restructured payload to consolidate XDM

Aaronius:
## Description
Conslidates XDM data (both customer XDM data and automatically collected data) inso a single `xdm` key on the event object. XDM data can be provided by the customer through the `xdm` option, while non-XDM data can be provided through the `data` option. XDM data provided by the customer is deeply merged on top of automatically collected data.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-33593
https://jira.corp.adobe.com/browse/EXEG-215

## Motivation and Context
* Users can more easily understand how the payload data maps to the schema they've created in AEP.
* Customers easily override or add to automatically collected data.
* Customers can more easily indicate which data is XDM vs non-XDM.

## Types of changes

## Checklist:

- [ ] (I will pronto!) I have updated the documentation accordingly.

### DIFF
--- a/sandbox/public/index.html
+++ b/sandbox/public/index.html
@@ -11,7 +11,7 @@
             script-src 'self' 'unsafe-inline' assets.adobedtm.com cdn.tt.omtrdc.net;
             style-src 'self' 'unsafe-inline' cdn.tt.omtrdc.net;
             img-src * data:;
-            connect-src 'self' unifiedjslab.tt.omtrdc.net *.demdex.net http://localhost:8080 https://alpha.konductor.adobedc.net https://edgegateway.azurewebsites.net "/>
+            connect-src 'self' unifiedjslab.tt.omtrdc.net *.demdex.net http://localhost:8080 https://alpha.konductor.adobedc.net https://edgegateway.azurewebsites.net konductor-qe.apps-exp-edge-npe-va6.experience-edge.adobeinternal.net"/>
 
     <title>Mock website hosting Alloy</title>
 
@@ -37,6 +37,7 @@
       // https://git.corp.adobe.com/Activation/edge-gateway-mock
       alloy("configure", {
         // edgeDomain: "edgegateway.azurewebsites.net",
+        // edgeDomain: "konductor-qe.apps-exp-edge-npe-va6.experience-edge.adobeinternal.net",
         // optInEnabled: true,
         // errorsEnabled: false,
         propertyId: 9999999,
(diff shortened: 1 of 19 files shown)


src/components/DataCollector/index.js
```diff
@@ -54,19 +54,19 @@ const createDataCollector = ({ config, logger }) => {
...
     return lifecycle
       .onBeforeEvent(event, options, viewStart, documentUnloading)
-      .then(() => optIn.whenOptedIn())
+      .then(() => {
+        // We merge the user's data after onBeforeEvent so that
+        // it overlays on top of any data Alloy automatically
+        // provides. This allows the user to override the
+        // automatically collected data.
+        event.mergeXdm(xdm);
+        event.data = data;
```
Aaronius: A consequence of setting `xdm` and `data` onto the event after `onBeforeEvent` is that none of the components will have access to the customer's `xdm` and `data` in `onBeforeEvent`. I think this is appropriate, since customer-provided `xdm` and `data` are probably not useful to components anyhow, as their structure is/can be rather customer-specific.

sandbox/public/index.html
```diff
@@ -51,10 +52,19 @@
 
       alloy("event", {
         viewStart: true,
+        xdm: {
+          "_myorg": {
```
jfkhoury: So this is required right? :(

src/components/DataCollector/index.js
jfkhoury: I  like/agree.

sandbox/public/index.html
Aaronius: For data related to customer-defined mixins, yeah. :/

sandbox/src/LargePayload.js
```diff
@@ -8,7 +8,7 @@ export default function LargePayload() {
       const payload = new Uint8Array(size * 1024);
       window.alloy("event", {
         documentUnloading: true,
-        data: {
+        xdm: {
           payload
```
jonsnyder: This isn't xdm.

src/components/DataCollector/index.js
jonsnyder: 👍

sandbox/src/LargePayload.js
Aaronius: Thanks @jonsnyder. I've switched it back to `data`.

PR #189, opened 2019-09-09
CORE-3698: 

shammowla:


test/functional/localConfig.json
```diff
@@ -4,6 +4,12 @@
       "browser": [
         "all"
       ],
+      "saucelabs": [
+        "saucelabs:Internet Explorer@10.0:Windows 8",
```
owenleej: We decided we are not supporting IE10, but we are supporting IE11.

.gitignore
```diff
@@ -67,3 +67,5 @@ typings
 .rpt2_cache/
 
 TMP_DIR/
+
+bin/
```
shammowla: The bin directory is being used by Saucelabs locally to tunnel.

.travis.yml
```diff
@@ -6,8 +6,12 @@ language:
 node_js:
 - '10'
 
+cache:
+  directories:
+    - ~/.npm
+    - node_modules
+
 install:
-- npm i -g npm
```
shammowla: Caching the NPM directories instead of installing npm globally improves Travis run time.

.travis.yml
```diff
@@ -33,11 +37,8 @@ before_script:
 
 script:
 - npm install
-- npm run build
-- npm run test
-- npm run test:saucelabs
-- npm run test:report
-- npm run functional
+- npm run test:slreport
+- npm run functional:ci
```
shammowla: The "functional:ci" script builds and unit tests runs during the build step. The script also runs http-serve and the the Testcafe in parallel.

test/functional/localConfig.json
```diff
@@ -4,6 +4,12 @@
       "browser": [
         "all"
       ],
+      "saucelabs": [
```
shammowla: Saucelabs environment is read from this configuration file by the Testcafe testrunner.js

test/functional/sandbox/html/bogusCommand.dev.html
```diff
@@ -0,0 +1,22 @@
...
+        <script>
+            !function(n,o){o.forEach(function(o){n[o]||((n.__alloyNS=n.__alloyNS||
+            []).push(o),n[o]=function(){var u=arguments;return new Promise(
+            function(i,l){n[o].q.push([i,l,u])})},n[o].q=[])})}
+            (window,["alloy"]);
+        </script>
+        <script src="/dist/standalone/alloy.js" async></script>
+
+        <script>
+            alloy("configure", {
```
shammowla: If logging isn't enabled using the logEnabled configuration, Testcafe is unable to fetch the browser console logs.

test/functional/specs/desktop/C2560.js
```diff
@@ -1,27 +1,22 @@
...
   .page(gatewayDomain)
   .requestHooks(
     networkLogger.gatewayEndpointLogs,
     networkLogger.sandboxEndpointLogs
   );
 
-test.before(() => console.log("Test Started"));
-test.after(() => console.log("Test Completed"));
-
-const testDescription =
```
shammowla: The Allure test report doesn't require the testDescription separately when the test type is mentioned. For example: "Regression: Test description here."

package.json
```diff
@@ -111,9 +115,8 @@
     "rollup-plugin-node-resolve": "^4.0.0",
     "rollup-plugin-replace": "^2.2.0",
     "rollup-plugin-terser": "^4.0.4",
-    "testcafe": "^1.3.3",
+    "testcafe": "^1.4.3",
```
shammowla: New version of testcafe supports script injection and getting console messages using getBrowserConsoleMessages. Testcafe live comes bundles with the new version.

test/functional/sandbox/html/bogusCommand.dev.html
owenleej: If you add "?alloy_log=true" to the test urls it will ensure logging is enabled before the configure command is called so you get log output for the config command being called.

owenleej: Also I like that you used the word bogus. 👍 👍 👍

shammowla: Adding a separate test spec with the url flag to see if getBrowserConsoleMessages is able to retrieve the logs.

test/functional/testrunner.js
```diff
@@ -51,29 +51,115 @@ createReport = () => {
...
-    const runOptions = {
-      skipJsErrors: config.desktop.skipCriticalConsoleJsErrors,
-      quarantineMode: config.desktop.quarantineMode,
-      speed: config.desktop.speed,
-      debugMode: false,
-      selectorTimeout: config.desktop.selectorTimeOut,
-      assertionTimeout: config.desktop.assertionTimeout
+const isSL = process.argv.includes("--sl");
+if (environment === "dev" && isSL === false) {
+  createTestCafe().then(tc => {
```
jonsnyder: I don't understand what is different in each environment.  Why the 4 cases when they all do the same thing?

test/functional/specs/desktop/C2588.js
```diff
@@ -1,21 +1,22 @@
...
+});
+
+test("Regression: Throw error when configure is executed multiple times.", async () => {
+  const message = await t.getBrowserConsoleMessages();
   await t
-    .expect(error)
-    .ok([
-      "[alloy] The library has already been configured and may only be configured once."
-    ]);
+    .expect(message.error[0])
```
jonsnyder: In general I try to be as lenient as possible when checking for error messages.  Changing an error message format slightly should not fail a test.  You could verify that there is one error message, or you could check to see that the error has the word 'already'.  Also could you verify that the second configure command returned a rejected promise?

test/functional/sandbox/html/bogusCommand.prod.html
```diff
@@ -0,0 +1,22 @@
...
+
+        <script>
+            alloy("configure", {
+                errorsEnabled: false,
+                logEnabled: true
+            });
+            alloy("boguscommand");
+        </script>
+    </head>
+</html>
```
Aaronius: From what I can tell just by looking here, it looks like bogusCommand.dev.html is different from bogusCommand.prod.html only in that it loads `alloy.js` vs `alloy.min.js`. Are you planning on reproducing this 
pattern for every test? If so, can we use a querystring parameter and templating engine to produce the script tag that loads the respective alloy file? Or...if we've decided we're going to test the minified build, then I don't think there's any need to test the unminified build. The chance that the minified build would behave correctly while the unminified build would not behave correctly is incredibly low. I like the idea of having the unminified build available for easier debugging, but if that would just require us to come into the html file and manually remove `.min` while we're debugging, that might be sufficient.

sandbox/public/index.html
```diff
@@ -33,12 +33,11 @@
...
 
     <script>
-      // Set localEdge to true to use a locally-running edge gateway mock
-      // https://git.corp.adobe.com/Activation/edge-gateway-mock
+
       alloy("configure", {
         // edgeDomain: "edgegateway.azurewebsites.net",
         // optInEnabled: true,
-        // errorsEnabled: false,
+        errorsEnabled: false,
```
wgamboa: Why is this needed?

wgamboa: Never mind.

PR #192, opened 2019-09-09
Change lifecycle calls to use an object parameter

jonsnyder:
## Description

This changes all the lifecycle calls and implementations to use an object as a parameter.

## Motivation and Context

On the PR to add documentUnloading to the onBeforeEvent, we decided it would be best to use an object parameter in all lifecycle calls.

## Types of changes

- [x] Code Refactor
## Checklist:


src/components/DataCollector/index.js
```diff
@@ -57,7 +57,7 @@ const createDataCollector = ({ config, logger }) => {
     const { viewStart = false, documentUnloading = false, xdm, data } = options;
 
     return lifecycle
-      .onBeforeEvent(event, options, viewStart, documentUnloading)
+      .onBeforeEvent({ event, options, viewStart, documentUnloading })
```
Aaronius: It looks like the name `viewStart` is inconsistent with `isViewStart` that's being referenced in, for example, `Audiences/index.js`

PR #193, opened 2019-09-09
Include `data` property on event while Konductor still requires it.

Aaronius:
## Description
Temporarily resolves an issue where Konductor was throwing errors because we weren't including `data` on `event`.

## Related Issue
None.

## Motivation and Context
Server errors are bad.

## Types of changes

## Checklist:

### DIFF
--- a/src/components/DataCollector/index.js
+++ b/src/components/DataCollector/index.js
@@ -63,8 +63,12 @@ const createDataCollector = ({ config, logger }) => {
         // it overlays on top of any data Alloy automatically
         // provides. This allows the user to override the
         // automatically collected data.
-        event.mergeXdm(xdm);
-        event.data = data;
+        if (xdm) {
+          event.mergeXdm(xdm);
+        }
+        if (data) {
+          event.data = data;
+        }
         return optIn.whenOptedIn();
       })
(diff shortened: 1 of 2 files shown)


src/components/DataCollector/index.js
```diff
@@ -63,8 +63,12 @@ const createDataCollector = ({ config, logger }) => {
...
         // provides. This allows the user to override the
         // automatically collected data.
-        event.mergeXdm(xdm);
-        event.data = data;
+        if (xdm) {
+          event.mergeXdm(xdm);
+        }
+        if (data) {
+          event.data = data;
+        }
```
Aaronius: I think it's probably a good idea to only set these properties on `event` if we actually have values for them, but it's also important for this temporary fix to work, since we would be overwriting the default `data` object with `undefined` if the customer didn't pass in `data` as an option.

Aaronius:
Sorry, false alartm. This isn't the real issue. The real issue is that I shouldn't have merged my `xdm`/`data` restructuring PR to master until Konductor releases their respective changes. Here's the PR that reverts my restructuring PR: https://github.com/adobe/alloy/pull/194

PR #196, opened 2019-09-10
CORE-33885 add beforeSend hook to network config

jonsnyder:
## Description

The spec for this feature is very thin.  This PR is a starting point for discussion on how this feature should be implemented.  I'm starting here with the most simple and most broadly applicable implementation.

### beforeSend hook implemented in the network component (in this PR)
Pros
* User is able to edit the entire payload.
* Simple for a user to understand how the passed in JSON maps to the new payload.
* This hook will be called for all network ops even idSyncs.

Cons
* User can mess up the payload to the point where Konductor rejects it
* Can be difficult to use if only a part of the payload needs to be modified
* If Konductor changes its expected format and Alloy releases a new version, customers' beforeSend hooks would need to be modified.

### Other ideas:
* Provide a hook for just the event object
* Provide a hook for just the automatically collected values (i.e. the xdm part of the event before it is merged with the xdm part of the user's event)
* Provide all of these hooks

## Related Issue

https://jira.corp.adobe.com/browse/CORE-33885

## Motivation and Context

* We want the user to be able to change the data that is automatically collected in Alloy.
* We want the user to be able to modify the sent data in one place.

## Types of changes

## Checklist:

### DIFF
--- a/src/core/network/createNetwork.js
+++ b/src/core/network/createNetwork.js
@@ -34,7 +34,8 @@ export default (config, logger, lifecycle, networkStrategy) => {
     return lifecycle.onResponse(response).then(() => response);
   };
 
-  const { edgeDomain, propertyId } = config;
+  // TODO: add config validation to beforeSend
+  const { edgeDomain, propertyId, beforeSend = payload => payload } = config;
 
   return {
     /**
@@ -62,8 +63,9 @@ export default (config, logger, lifecycle, networkStrategy) => {
         logger.log(`No response requested due to document unloading.`);
       }
       const reallyExpectsResponse = documentUnloading ? false : expectsResponse;
-      return Promise.resolve()
-        .then(() => {
+      return Promise.resolve(payload)
+        .then(beforeSend)
+        .then(newPayload => {
           const action = reallyExpectsResponse ? "interact" : "collect";
 
           let baseUrl = `https://${edgeDomain}`;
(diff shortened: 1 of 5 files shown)


Aaronius:
Thanks for writing this up and for the good description. My take is we should wait to hear the use cases before implementing this. I think use cases will provide some important insight into how it's built and I would hate to maintain something that's unnecessary. Some related thoughts:

* Once https://github.com/adobe/alloy/pull/186 is merged, customers will be able to overwrite, add to, and delete (by setting values to `undefined`) the auto-collected data without a `beforeSend` option. This ability only applies to auto-collected data that gets placed on the event's `xdm` object, however (in other words, they wouldn't be able modify the `identityMap` data because that sits at the payload level instead of the event level), so there might still be a need for something like `beforeSend`.
* It might be a bit difficult to do something useful with the payload if `beforeSend` were called for event data _and_ ID syncs and potentially other types of requests. The plan is that ID syncs will soon be using a different endpoint with a different payload structure. If I, as a customer, wanted to modify event data on every request, I would probably need to add code in my `onBeforeSend` function to basically duck-type-check the payload to determine if it's related to events or if it's related ID syncs or something else.

I don't know that the approach in this PR is the wrong thing, and I think what you've proposed in your "other ideas" section could also be valid approaches, but I don't think we have the right information to make a good decision.

dancingcactus:
The more I think about this the more I worry that it could come back to bit us. We will have breaking changes to Konductor in the future. At least for the next year or so as we go through the ramp in Adoption. 

The biggest fear is that customers implement some logic that only fires on certain requests then they upgrade the library or we upgrade Konductor and only parts of their implementation fail. Those types of errors are extremely difficult to track down. 

I also don't want them to have to become experts on the entire structure of the request. Rather I would like to let them use an interface that looks more like what they use with Event 
```json
{
  "viewStart":True,
  "data":{blah},
}
```

My preference would be two fold. 1- Allow customers to modify the event before it is sent out but before it is merged (I don't want them editing the raw request). 2- In that hook don't provide the raw request but provide the customer variables (e.g. view start, data, xdm, etc) and the auto collected stuff to modify.

PR #197, opened 2019-09-10
Event merge id to xdm

Aaronius:
## Description
Supports promises being passed in options and moved `eventMergeId` to be part of the `xdm` block rather than a top-level option.

For a little background, we supported a promise being passed as an `eventMergeId` value. When a promise is passed, we were waiting for the promise to be resolved and then using the resolved value as the `eventMergeId`. Now that `eventMergeId` is no longer a top-level option, we didn't really want components to be digging through `xdm` blocks looking for `eventMergeId`. Preferably, components could be unaware of specific things inside `xdm` blocks.

I talked with @jfkhoury about this and he mentioned that he's heard customers sometimes want to pass promises for values in some of the things he's worked on. In other words, instead of just supporting promises for the `eventMergeId` value, why not support promises for any value inside `xdm` or `data` options? I kind of took it a step further, for better or worse, and proposed we support promises for any value in any of the options. Maybe this is a great idea, or maybe this is a really dumb idea.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-33593

## Motivation and Context
`eventMergeId` would be part of the scheme the user created, so it makes sense to pass it as part of the `xdm` option and not as a top-level option.

## Types of changes

## Checklist:


src/utils/promiseAllObject.js
```diff
@@ -0,0 +1,47 @@
...
+ * @returns {Promise}
+ */
+const promiseAllObject = object => {
+  const keys = Object.keys(object);
+  return Promise.all(
+    keys.map(key => {
+      const value = object[key];
+      if (
+        Array.isArray(value) ||
+        (isPlainObject(value) && !isFunction(value.then))
```
Aaronius: `isPlainObject` is important here because we don't want to be messing with non-plain-objects like dates, regexes, etc. The spec covers this.

src/core/createInstance.js
```diff
@@ -121,7 +126,19 @@ export default (
...
+      logger.log(`Executing ${commandName} command.`, "Options:", options);
+      _resolve(
+        // Note that promiseAllObject also clones the options object.
+        // This is an important aspect because it serves as taking a snapshot
+        // of options content (particularly important with xdm and data
+        // options where the original objects are more likely to be modified
+        // by the customer over time).
+        // TODO: Add a test for createInstance that ensures options are
+        // being cloned.
+        promiseAllObject(options).then(resolvedOptions => {
```
Aaronius: Notice that of `options` has a promise somewhere in it, we're not logging that we're waiting for that promise(s) to be resolved. I think it would probably be helpful, but I wanted to see what you think first because it's going to make `promiseAllObject` a little gross because it has to return both the new promise and whether promises were found inside the object or something like that.

test/unit/specs/utils/promiseAllObject.spec.js
```diff
@@ -0,0 +1,105 @@
...
+        name: "Avocado",
+        calories: Promise.resolve(234)
+      },
+      {
+        name: "Banana",
+        calories: 133
+      },
+      Promise.resolve("Cherry")
+    ],
+    vegetables: Promise.resolve([
```
owenleej: What happens if the promise is rejected? Can you add a test for that scenario?
Can you add a test for a promise that isn't resolved instantly? Maybe that doesn't matter... idk.

src/utils/isPlainObject.js
```diff
@@ -0,0 +1,42 @@
...
+ * constructor.
+ * @param value
+ * @returns {boolean}
+ */
+export default value => {
+  if (isObjectObject(value) === false) return false;
+
+  // If has modified constructor
+  const ctor = value.constructor;
+  if (typeof ctor !== "function") return false;
```
owenleej: Would it be better to check that "ctor === Object" ?
image

owenleej: And then you wouldn't have to do the other two check I think

src/core/createInstance.js
owenleej: Can you pass the logger to promiseAllObject and put the logging statements for that inside the promiseAllObject code?

test/unit/specs/utils/promiseAllObject.spec.js
Aaronius: The promise returned from `promiseAllObject` should be rejected. It's a good idea to add a test for that though. I will do that.

src/utils/isPlainObject.js
Aaronius: This code is copied from the [`is-plain-object` package](https://github.com/jonschlinkert/is-plain-object) (see the header at the top of the file) but modified only to re-use our own `isObject` utility. I'd rather not modify the code because what's here is very well battle-tested (depended on by 3+ million github projects, installed from npm 11+ million times weekly).

src/core/createInstance.js
Aaronius: We could, but that would probably make `promiseAllObject` aware of and coupled to the context in which it is running. In other words, if inside `promiseAllObject` we had something like:

```
logger.log("Promises found within provided options. Waiting for promises to be resolved before continuing command execution.");
```

It would restrict usage of `promiseAllObject` to this one use case. Making the error more generic might make it too generic. We could do something like this:

```
promiseAllObject(options, logger, "Promises found within provided options. Waiting for promises to be resolved before continuing command execution.");
```

But I think that's worse than just returning an object with both the promise and a boolean indicating whether promises were found (and, probably ideally, where those promises were found in the object structure).

Aaronius:
BTW @owenleej, I fully expect this to be a hit to performance that will show up in your reports. I think it's a reasonable benefit/cost tradeoff though.

src/core/createInstance.js
owenleej: That makes sense. I think promiseAllObject(option,logger,"Message") and returning a boolean/nestedPromises are both acceptable options.

test/unit/specs/utils/promiseAllObject.spec.js
Aaronius: Regarding this:

> Can you add a test for a promise that isn't resolved instantly?

I think it helps to understand that promises are ALWAYS asynchronous, regardless of whether they're resolved immediately or not. In other words, even though we resolve the promise here immediately, any callbacks passed through a `.then()` on the promise will always be called _asynchronously_. So it doesn't really matter whether it takes .001 milliseconds for the callback to be called or 1000 seconds, it's guaranteed to be asynchronous, which is the important part, and we can reasonably assume that the difference in the amount of time won't cause a difference in behavior within `promiseAllObject`.

Aaronius:
Thanks @owenleej. I've added a test for the rejected promise.

src/core/createInstance.js
```diff
@@ -121,7 +126,19 @@ export default (
     // underlying function call.
     // Also note that executeCommand may or may not return a promise.
     new Promise(_resolve => {
-      _resolve(executeCommand(commandName, options));
+      logger.log(`Executing ${commandName} command.`, "Options:", options);
+      _resolve(
+        // Note that promiseAllObject also clones the options object.
```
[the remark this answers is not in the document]
Aaronius: I think it's sufficient to put it in the util's JSdoc, which I'll add another commit for. I think I feel that way because of other functions that do things similar to cloning where the cloning part isn't called out specifically in the name (e.g., `Array.prototype.concat`). I just want to make sure that we (I?) add a test for cloned options when we add tests for the core modules. I think cloning the options object is a good thing and something we should be doing regardless of `resolvePromisesAndClone`, but more importantly I don't want to accidentally go from cloning options to not cloning options later, since it could result in a negative change in behavior for the user.

src/utils/promiseAllObject.js
```diff
@@ -0,0 +1,47 @@
...
+      ) {
+        return promiseAllObject(value);
+      }
+      return value;
```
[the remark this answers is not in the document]
Aaronius: > In the case where an object is passed that was created with a custom constructor, the values of the object are not cloned.

I don't think we want to support custom constructor objects for this function. I don't think it's worth adding a programmatic check, but I do think it's worth adding to the JSDoc. I'll do that in a new commit.

> Additionally, the resolved promise is not cloned. For example, the promise could resolve to an object, and then the object could change before the network call is made.

It's a good point. Let me think about this one. I might make `promiseAllObject` recursively apply to the resolution object as well. I was trying to avoid it because of the extra code it would require, but it might be worth it to also address what you've brought up.

src/core/createInstance.js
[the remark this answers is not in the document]
Aaronius: I think it would be helpful. I will make it so.

Aaronius: @jonsnyder I like the approach you've outlined, but one concern I might have is that if someone executed an `event` command twice, and the first one has a promise that gets resolved in, say, 20 milliseconds, and the second `event command doesn't have any promises in its data, the data from the second event will likely get sent to the server before the data from the first event and there won't be any log messages that suggest why that may have happened.

src/core/createInstance.js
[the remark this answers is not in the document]
Aaronius: I've added a commit intended to improve the jsdoc in this regard.

src/utils/promiseAllObject.js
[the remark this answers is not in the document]
Aaronius: I've added a commit intended to improve the jsdoc in this regard. Also, I made `promiseAllObject` recursively search resolved values if they are objects or arrays (which includes cloning them). IMO it actually made the code simpler, so I like the change. Depending on how we decide to log that we're waiting for promises, it might get dirtier again (because we'll need logic to determine whether an object is a promise), but it's good for now.

src/core/createInstance.js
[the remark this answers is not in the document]
Aaronius: Done.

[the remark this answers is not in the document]
Aaronius: Sorry again for not having tests for any of the core modules. I understand how important it is that we get tests for that stuff. It's still on the priority list in JIRA, but is having a hard time reaching the top. :(

Aaronius:
For the logging that tells users we're waiting for promises, I'm going to push off the work to a separate PR.

PR #206, opened 2019-09-13
Add name to data to get audiences.

jfkhoury:
## Description

Remove QE edge domain.

## Types of changes

## Checklist:

### DIFF
--- a/sandbox/public/index.html
+++ b/sandbox/public/index.html
@@ -42,8 +42,8 @@
     <script>
       alloy("configure", {
         // edgeDomain: "edgegateway.azurewebsites.net",
-        edgeDomain:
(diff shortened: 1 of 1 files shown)


sandbox/public/index.html
```diff
@@ -71,7 +71,8 @@
           }
         },
         data: {
-          nonXdmKey: "nonXdmValue"
+          nonXdmKey: "nonXdmValue",
+          name: "home"
```
Aaronius: It might be good to add a comment here as to the importance of `name`.

jonsnyder: Can you add a comment that this key/value is for Audience Manager?

jfkhoury: Done @jonsnyder

jonsnyder: Jinx!

PR #208, opened 2019-09-16
CORE-34857: Toggle logging through configuration

shammowla:
1) Created a test page that configures the SDK with click events. 
2) Updated C2585 to use the new test page. 

C2585 is currently failing due to logging configurations bug. 
https://jira.corp.adobe.com/browse/CORE-34934


test/functional/specs/desktop/C2585.js
```diff
@@ -11,10 +11,15 @@ test.meta({
...
+
   const message = await t.getBrowserConsoleMessages();
+
   await t
-    .expect(message.error)
-    .contains(
-      '[alloy] Error: [alloy] The library must be configured first. Please do so by calling alloy("configure", {...}).'
+    .expect(message.log)
+    .match(
+      '[alloy] The library must be configured first. Please do so by calling alloy("configure", {...}).'
```
jonsnyder: Be careful not to make these tests too fragile.

test/functional/specs/desktop/C2583.js
```diff
@@ -0,0 +1,87 @@
...
+  ID: "C2583",
+  SEVERITY: "P0",
+  TEST_RUN: "Regression"
+});
+
+test("Regression: Set the log option to true. Load the page. Execute an event command.", async () => {
+  await t
+    .click(
+      Selector(
+        "#body > section > div.left-nav > div > ul > li:nth-child(1) > a"
```
Aaronius: Rather than having these complex selectors that might break if you restructure the HTML, I recommend adding a data attribute to each of the links and then selecting off the data attribute.

https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes
https://css-tricks.com/almanac/selectors/a/attribute/

PR #209, opened 2019-09-16
CORE-31911 add requestId to queryString

jonsnyder:
## Description

Add the requestId to the queryString of the request.  Also, I added requestId to the onResponse, and onResponseError lifecycle calls.

## Related Issue

https://jira.corp.adobe.com/browse/CORE-31911

## Motivation and Context

The requestId can be used by Konductor and other downstream systems so we can trace the processing of the data.

## Types of changes

## Checklist:

### DIFF
--- a/src/core/network/createNetwork.js
+++ b/src/core/network/createNetwork.js
@@ -31,7 +31,7 @@ export default (config, logger, lifecycle, networkStrategy) => {
 
     const response = createResponse(parsedBody);
 
-    return lifecycle.onResponse({ response }).then(() => response);
+    return lifecycle.onResponse({ response, requestId }).then(() => response);
   };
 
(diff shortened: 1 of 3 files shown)


src/core/network/createNetwork.js
```diff
@@ -31,7 +31,7 @@ export default (config, logger, lifecycle, networkStrategy) => {
 
     const response = createResponse(parsedBody);
 
-    return lifecycle.onResponse({ response }).then(() => response);
+    return lifecycle.onResponse({ response, requestId }).then(() => response);
```
jfkhoury: Jon, the response already contains the request ID that was provided, or the one that was created by Konductor if none was provided. Just an FYI, I don’t think it hurts to have it to match the onResponseError.

PR #210, opened 2019-09-17
Fix validator for optInEnabled, add better error message

jonsnyder:
## Description

Aaron found a bug with the optInEnabled validation while working on the launch extension.  The bug was the missing parentheses in the validation.  This fixes the bug and provides better logging for when it will happen in the future.

## Types of changes

## Checklist:

### DIFF
--- a/src/core/createConfig.js
+++ b/src/core/createConfig.js
@@ -10,7 +10,7 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import { assign, isString } from "../utils";
+import { assign } from "../utils";
 
 const CONFIG_DOC_URI =
   "https://launch.gitbook.io/adobe-experience-platform-web-sdk/fundamentals/configuring-the-sdk";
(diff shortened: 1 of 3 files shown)


src/core/createConfig.js
```diff
@@ -76,7 +76,11 @@ const createConfig = config => {
             validator.defaultValue
           );
           if (errorMessage) {
-            ac.push(errorMessage);
+            if (isString(errorMessage)) {
+              ac.push(errorMessage);
+            } else {
+              ac.push(`An internal error occurred while validating '${key}'`);
```
liljenback: Should you log the faulty errorMessage here just so we could debug the issue if needed?

Aaronius:
You might want to add a test in `createConfig.spec.js`.

src/core/createConfig.js
Aaronius: @jonsnyder thinking through this again, would be better to just add tests for our config validators (e.g, validate that `optInEnabled` accepts a boolean and only a boolean)?

PR #211, opened 2019-09-17
CORE-34934 Assume errors enabled until configuration says otherwise.

Aaronius:
## Description
With a clean local storage and attempting to execute a command before executing the `configure` command, the promise returned to the customer should be rejected. If the promise goes unhandled, the error should show up in the console. 

## Related Issue
https://jira.corp.adobe.com/browse/CORE-34934

## Motivation and Context
Customers need to know why the command didn't execute.

## Types of changes

## Checklist:

### DIFF
--- a/src/core/createInstance.js
+++ b/src/core/createInstance.js
@@ -29,7 +29,8 @@ export default (
   logger,
   window
 ) => {
-  let errorsEnabled;
+  // Assume errors are enabled until configuration says otherwise.
+  let errorsEnabled = true;
   let configurePromise;
 
   const logCommand = ({ enabled }) => {


src/core/createInstance.js
```diff
@@ -29,7 +29,8 @@ export default (
   logger,
   window
 ) => {
-  let errorsEnabled;
+  // Assume errors are enabled until configuration says otherwise.
+  let errorsEnabled = true;
```
Aaronius: Previously, it was `undefined` until the configure command executed, so the logic [here](https://github.com/adobe/alloy/blob/1e24d20e2e0c97496bf3b6a9489b537cbbc17d7c/src/core/createInstance.js#L177-L181) decided not to reject the promise.

PR #212, opened 2019-09-18
CORE-35071 Reject promise if not all id syncs are successful

wgamboa:
## Related Issue
CORE-35071

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Identity/createIdSyncs.js
+++ b/src/components/Identity/createIdSyncs.js
@@ -94,9 +94,7 @@ const createProcessor = (config, logger, cookieJar) => destinations => {
     if (result.failed.length) {
       const failedIds = result.failed.map(idSync => idSync.id);
 
-      return Promise.reject(
-        new Error(`The following id sync ids failed: ${failedIds.join(", ")}.`)
-      );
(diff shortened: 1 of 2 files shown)


src/components/Identity/createIdSyncs.js
```diff
@@ -91,6 +91,14 @@ const createProcessor = (config, logger, cookieJar) => destinations => {
       ).toString(36)
     );
 
+    if (result.failed.length) {
+      const failedIds = result.failed.map(idSync => idSync.id);
+
+      return Promise.reject(
+        new Error(`The following id sync ids failed: ${failedIds.join(", ")}.`)
+      );
```
Aaronius: You can just throw the error rather than returning a rejected promise.
I would recommend capitalizing ID here: `The following ID sync IDs failed: ${failedIds.join(", ")}.`

wgamboa:
@jonsnyder the successful id syncs (manual and server-based) and failed server-based id syncs won't be fired, while the failed manual id syncs will be re-attempted. Failed server-based id syncs will be re-attempted after 7 days.

PR #213, opened 2019-09-20
CORE-33793 Rename idSyncsEnabled to idSyncEnabled

wgamboa:
## Types of changes

## Checklist:

### DIFF
--- a/src/components/Identity/createIdSyncs.js
+++ b/src/components/Identity/createIdSyncs.js
@@ -37,7 +37,7 @@ const setControlObject = (controlObject, cookieJar) => {
 };
 
 const createProcessor = (config, logger, cookieJar) => destinations => {
-  if (!config.idSyncsEnabled) {
+  if (!config.idSyncEnabled) {
     return Promise.reject(new Error("ID syncs not enabled"));
   }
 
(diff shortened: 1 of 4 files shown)


src/components/Identity/createIdSyncs.js
```diff
@@ -37,7 +37,7 @@ const setControlObject = (controlObject, cookieJar) => {
 };
 
 const createProcessor = (config, logger, cookieJar) => destinations => {
-  if (!config.idSyncsEnabled) {
+  if (!config.idSyncEnabled) {
     return Promise.reject(new Error("ID syncs not enabled"));
```
Aaronius: You might want to change this error message to "ID sync not enabled" or "ID synchronization not enabled".

src/components/Identity/index.js
```diff
@@ -58,7 +58,7 @@ const createIdentity = ({ config, logger, cookieJar }) => {
         return optIn.whenOptedIn().then(() => {
           if (
             !alreadyQueriedForIdSyncs &&
```
Aaronius: Do you want to change some of these variable names, or does it make sense to leave them as "syncs"?

src/components/Identity/createIdSyncs.js
wgamboa: Thanks, I'll use the second one.

src/components/Identity/index.js
wgamboa: I think it makes sense to leave them.

PR #214, opened 2019-09-23
Adding unit-tests for identity:onResponse

jileeshs:
Added unit tests

## Description

https://jira.corp.adobe.com/browse/CORE-34087

## Related Issue

https://jira.corp.adobe.com/browse/CORE-34087

## Types of changes

## Checklist:

### DIFF
--- a/test/unit/specs/components/Identity/index.spec.js
+++ b/test/unit/specs/components/Identity/index.spec.js
@@ -11,12 +11,14 @@ governing permissions and limitations under the License.
 */
 
 import createIdentity from "../../../../../src/components/Identity";
+import flushPromiseChains from "../../../helpers/flushPromiseChains";
+import { COOKIE_NAMES } from "../../../../../src/components/Identity/constants";
 
+const { EXPERIENCE_CLOUD_ID } = COOKIE_NAMES;
 describe("Identity", () => {
   describe("reactor-specific functionality", () => {
     let identity;
     let reactorRegisterGetEcid;
-
     beforeEach(() => {
       reactorRegisterGetEcid = jasmine.createSpy();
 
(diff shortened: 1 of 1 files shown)


test/unit/specs/components/Identity/index.spec.js
```diff
@@ -59,5 +61,61 @@ describe("Identity", () => {
...
+            EXPERIENCE_CLOUD_ID,
+            "ABC"
+          );
+          expect(response.getPayloadsByType).toHaveBeenCalledWith(
+            "identity:exchange"
+          );
+        });
+      });
+      it("should not do anything if the wrong object is passed", () => {
+        identity.lifecycle.onResponse(response);
```
jonsnyder: Did you mean to add curly braces around response?

test/unit/specs/components/Identity/index.spec.js
```diff
@@ -59,5 +61,61 @@ describe("Identity", () => {
...
+        return flushPromiseChains().then(() => {
+          expect(response.getPayloadsByType).not.toHaveBeenCalled();
+          expect(cookieJar.set).not.toHaveBeenCalled();
+          expect(response.getPayloadsByType).not.toHaveBeenCalled();
```
jonsnyder: This line is here twice.  You can do `...not.toHaveBeenCalledWith("identity:exchange")`

test/unit/specs/components/Identity/index.spec.js
```diff
@@ -59,5 +61,61 @@ describe("Identity", () => {
...
+          expect(cookieJar.set).toHaveBeenCalledWith(
+            EXPERIENCE_CLOUD_ID,
+            "ABC"
+          );
+          expect(response.getPayloadsByType).toHaveBeenCalledWith(
+            "identity:exchange"
+          );
+        });
+      });
+      it("should not do anything if the wrong object is passed", () => {
```
jonsnyder: A better test might be for when getPayloadsByType("identity:persist") returns [].

test/unit/specs/components/Identity/index.spec.js
```diff
@@ -59,5 +61,61 @@ describe("Identity", () => {
...
+        return flushPromiseChains().then(() => {
+          expect(response.getPayloadsByType).toHaveBeenCalledWith(
+            "identity:persist"
+          );
+          expect(cookieJar.set).toHaveBeenCalledWith(
+            EXPERIENCE_CLOUD_ID,
+            "ABC"
+          );
+          expect(response.getPayloadsByType).toHaveBeenCalledWith(
+            "identity:exchange"
```
jonsnyder: This is where having dependency injection makes things easy to test.  Because `idSyncs` is created inside the createIdentity function, there is no way to pass in a mock for it.  Therefore we cannot test that the result of response.getPayloadsByType("identity:exchange") is actually passed to idSyncs.process.  Ideally we would return a known value when the response is queried for "identity:exhange" and then just have verification here that idSyncs.merge was called with that value.

To follow dependency injection principles, you want to separate the code that creates objects with the code that does the business logic.  The code that creates the objects should be really simple, and will be tested by functional tests, but the code with the business logic should have unit tests.  In index.js, create all the objects you need, but then pass them into other functions or objects.  Then you can test the other functions or objects with mocks passed in.  You can look at Context/index.js for an example.

You are welcome to merge this PR without making this change, as this would be a fairly large refactor.

test/unit/specs/components/Identity/index.spec.js
jileeshs: No, the test was to see if the response object doesn't have a property named `response`, it shouldn't do any of the  actions.

jonsnyder: Ah, I'd recommend passing empty object in the test so that it is more clear.

test/unit/specs/components/Identity/index.spec.js
Aaronius: I agree with this. Components shouldn't have to code for, or test, the possible case where `onResponse` is called with an invalid object because we own all the code that could potentially call `lifecycle.onResponse`, and that code should have its own tests to ensure it's passing the right object.

test/unit/specs/components/Identity/index.spec.js
Aaronius: I agree with this. Without restructuring the code, this test isn't nearly as valuable as it could it. By testing that `getPayloadsByType` is being called, we're getting a hint that something expected is happening, but we can't be confident that the information is being used correctly.

test/unit/specs/components/Identity/index.spec.js
```diff
@@ -59,5 +61,61 @@ describe("Identity", () => {
...
+          get(key) {
+            return key === "ECID" ? "ABC" : null;
+          },
+          set: jasmine.createSpy()
+        };
+        response = {
+          getPayloadsByType: jasmine
+            .createSpy()
+            .and.returnValue([{ id: "ABC" }]),
+          toJSON: jasmine.createSpy()
```
Aaronius: Don't feel like you need to stub all methods on your mock objects. If `toJSON` isn't used in the code under test, I feel like it's fine to not fill it in here.

jfkhoury:
@jileeshs I think we can merge this PR at this point.

Thanks.

PR #215, opened 2019-09-24
CORE-31590 Support a config to turn off 3rd party cookie

wgamboa:
## Types of changes

## Checklist:

### DIFF
--- a/src/components/Identity/index.js
+++ b/src/components/Identity/index.js
@@ -72,6 +72,11 @@ const createIdentity = ({ config, logger, cookieJar }) => {
               identityQuery.identity.containerId = config.idSyncContainerId;
             }
 
+            if (config.thirdPartyCookiesEnabled !== undefined) {
+              identityQuery.identity.thirdPartyCookiesEnabled =
+                config.thirdPartyCookiesEnabled;
+            }
+
             event.mergeQuery(identityQuery);
           }
         });
@@ -153,6 +158,10 @@ createIdentity.configValidators = {
       .integer()
       .minimum(0)
       .expected("an integer greater than or equal to 0")
+  },
+  thirdPartyCookiesEnabled: {
+    defaultValue: undefined,
+    validate: boolean()
   }
 };
 


src/components/Identity/index.js
```diff
@@ -153,6 +158,10 @@ createIdentity.configValidators = {
       .integer()
       .minimum(0)
       .expected("an integer greater than or equal to 0")
+  },
+  thirdPartyCookiesEnabled: {
+    defaultValue: undefined,
```
Aaronius: Why isn't the `defaultValue` `true` or `false`?

wgamboa: If it's true or false, then it will be sent in the request. If it's undefined, it won't be.

Aaronius: What's the default in Konductor?

wgamboa: Probably true. But @jfkhoury specified that It shouldn't be sent if it's not set.

Aaronius: Here's my take. Our default value here should match Konductor's default value. If the Alloy user doesn't set `thirdPartyCookiesEnabled` to a value other than the default, then we shouldn't send the value to the server. I think it's important to have a default value of `true` or `false` defined here though, because that default value is both in our documentation and reported at runtime in the console as part of the "computed configuration". As an Alloy user, I want to see how it's going to behave by default, and `undefined` isn't explicit about that.

jfkhoury: I agree with @Aaronius. Yes, `true` is the default.

wgamboa: I'll make another pull request.

Aaronius: Regarding `idSyncContainerId`: I was told that `idSyncContainerId` essentially has a default value of `0`. If that's true, does it make sense for the `idSyncContainerId` option to have a default value of `0` instead of `undefined` or does a container ID of `0` not really have any significance to a user? In other words, do users know that there's such a thing as a container ID of 0 (in which case it would probably be useful to set `defaultValue` to `0`) or is that an internal-only concept?

wgamboa: Konductor has the default value as null.

PR #216, opened 2019-09-25
Core unit tests

Aaronius:
## Description
Adds unit tests for a bunch of the core modules. It also removes support for supporting promises as option values (discussed previously with some of the team, including Justin). I've incorporated feedback from @jonsnyder into this PR as well. Thanks @jonsnyder!

## Related Issue
https://jira.corp.adobe.com/browse/CORE-32995

## Motivation and Context
Some level of certainty that our code is working the way we expect.

## Types of changes

## Checklist:

- [x] I have updated the documentation accordingly. (https://github.com/AdobeDocs/alloy-docs/pull/50)


Aaronius:
I'm good with squashing my commits during the merge.

src/core/initializeComponentsFactory.js
```diff
@@ -22,35 +22,25 @@ const { ALLOY_COOKIE_NAME, ALLOY_COOKIE_TTL_IN_DAYS } = cookieDetails;
...
-  createOptIn
-) => config => {
-  const componentRegistry = createComponentRegistry();
-  const { imsOrgId, propertyId, cookieDomain } = config;
-  const cookieName = `${ALLOY_COOKIE_NAME}_${propertyId}`;
+  optIn
+}) => config => {
+  const network = createNetwork(config, logger, lifecycle);
+  const { imsOrgId, cookieDomain } = config;
+  const cookieName = `${ALLOY_COOKIE_NAME}_${imsOrgId}`;
```
jfkhoury: Nice! We will discuss this some more to decide if we need to support multiple property IDs inside the same cookie.

src/core/main.js
```diff
@@ -84,14 +85,19 @@ if (namespaces) {
...
+    });
+
+    const executeCommand = executeCommandFactory({
       logger,
       configureCommand,
       logCommand,
-      getErrorsEnabled() {
-        return errorsEnabled;
-      }
+      handleError
```
jfkhoury: Nice!

src/core/main.js
```diff
@@ -64,7 +78,17 @@ if (namespaces) {
...
+      tools: {
+        config: configToolFactory(),
+        cookieJar: cookieJarToolFactory(
+          createCookieProxy,
+          createComponentNamespacedCookieJar,
+          getTopLevelDomain
+        ),
+        enableOptIn: enableOptInToolFactory(optIn),
+        logger: loggerToolFactory(logger),
+        network: networkToolFactory(createNetwork, lifecycle, logger)
```
jfkhoury: Great idea to move this here.

src/core/main.js
```diff
@@ -28,11 +33,17 @@ import initializeComponentsFactory from "./initializeComponentsFactory";
 import createConfig from "./createConfig";
 import configValidators from "./configValidators";
 import handleErrorFactory from "./handleErrorFactory";
+import networkToolFactory from "./tools/networkToolFactory";
```
jfkhoury: Good idea @jonsnyder

sandbox/src/EventMerge.js
```diff
@@ -1,28 +1,30 @@
 import React, { useRef, useEffect } from "react";
 
 export default function EventMerge() {
-  const eventMergeId = useRef(window.alloy("createEventMergeId"));
+  const eventMergeIdPromise = useRef(window.alloy("createEventMergeId"));
```
jfkhoury: Need to update the docs?

Aaronius: Yeah, right here: https://github.com/AdobeDocs/alloy-docs/pull/50

src/core/configureCommandFactory.js
```diff
@@ -0,0 +1,44 @@
...
+  });
+  config.validate();
+  setErrorsEnabled(config.errorsEnabled);
+  const parsedQueryString = queryString.parse(window.location.search);
+  logCommand({
+    enabled:
+      parsedQueryString[logQueryParam] !== undefined
+        ? stringToBoolean(parsedQueryString[logQueryParam])
+        : config.logEnabled
+  });
```
jonsnyder: It looks like lines 33-40 could be put in the code for the loggerToolFactory, but I see that we are doing this so that the computed configuration can be logged correctly.  If we put the log line in initializeComponentsFactory after the tools are configured, then these lines can go in the loggerToolFactory.

src/core/configureCommandFactory.js
```diff
@@ -0,0 +1,44 @@
...
+  // toJson is expensive so we short circuit if logging is disabled
+  if (logger.enabled) logger.log("Computed configuration:", config.toJSON());
+  return initializeComponents(config);
```
jonsnyder: Seems weird to pass this function in to just call it once at the end.  It's basically a callback for a sync function.

I think it'd be better to do this inside of core/main.js:
```
const configureCommand = options => initializeComponents(buildAndValidateConfig(options))
```

src/core/tools/networkToolFactory.js
```diff
@@ -0,0 +1,18 @@
...
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+/**
+ * Tool-specific dependencies => config => componentCreator => result
+ */
+export default (createNetwork, lifecycle, logger) => config => () => {
+  return createNetwork(config, logger, lifecycle);
```
jonsnyder: This creates a new network instance for each component even though it doesn't depend on the componentCreator.  What is your reasoning?

test/unit/specs/core/componentCreators.spec.js
```diff
@@ -0,0 +1,28 @@
...
+describe("componentCreators", () => {
+  it("is an array of component creators", () => {
+    expect(componentCreators).toEqual(jasmine.any(Array));
+    componentCreators.forEach(componentCreator => {
+      expect(componentCreator).toEqual(jasmine.any(Function));
+      expect(componentCreator.namespace).toEqual(jasmine.any(String));
+      expect(componentCreator.abbreviation).toEqual(jasmine.any(String));
+
+      if (componentCreator.configValidators) {
+        expect(componentCreator.configValidators).toEqual(jasmine.any(Object));
```
jonsnyder: We could do some more checks here on the configValidators
* check for only keys "defaultValue" & "validator"
* check that the validator function is of the correct name (We'd have to change newValidator in utils/configValidators/chain.js to be a named function)
* check that leaf nodes have at least "validator".
* check that the defaultValue validates (if it isn't null or undefined).
* make sure there aren't any overlapping fields (i.e. two components validating the same field)

Maybe put these in a separate spec file that would run it against the coreConfigValidators too.

src/core/configureCommandFactory.js
Aaronius: I think it's preferable to enable/disable logging as soon as possible after the config is validated so that any log statements (current or future) from anything in Alloy appropriately hits the console or not without the need for the author to understand whether the logger has yet to be appropriately configured. For example, if an error were to be thrown in `initializeComponentsFactory.js` during tool configuration for whatever reason, we would want to make sure the logging were appropriately configured beforehand because it would come into play when the error gets reported to the customer: https://github.com/adobe/alloy/pull/216/files/4de6267c353ca4f9791cb8951a310e931a3bf70f#diff-6ac18022080e60b6d96019ed637bae25R21

This might already be understood, but the configuration of the logger affects both the original logger that the core modules use and the loggers that are spawned for the components. It seems like `loggerToolFactory.js` should stick to dealing with managing the spawned loggers only and not the parent logger.

src/core/configureCommandFactory.js
Aaronius: Yeah, I debated doing this. Here was my mindset: I wasn't sure what the appetite was for logic inside `main.js`. Even though there's very little here, it would be the first function declared in main.js that wasn't just getting or setting (I went back and forth on those too).  If I did do this in main.js, I would want to move the `logCommandFactory.js` logic inside main.js as well so it's not a loner. I'm totally cool with this though. Still prefer it?

src/core/tools/networkToolFactory.js
Aaronius: Ooh, good catch. No reasoning. Just a mistake. I'll fix it.

src/core/configureCommandFactory.js
jonsnyder: I'm not sure what the guideline should be between object-creation-and-wiring-code and testable-code.  To me it feels like function composition is the same as passing in or wiring and doesn't need to be testable.  We could just as easily make a util called `compose` to do this.

Your comment about logging brings up something that has bothered me too. We have logging concerns in a bunch of different places.
* createLogController
* createLogger
* createLogCommand
* loggerToolFactory

I'm wondering if we could move all of this inside of createLogController
```
const createLogController = (instanceNamespace, createNamespacedStorage, ...) => {
  ...
  return {
    get logEnabled() {
      ...
    },
    set logEnabled(value) {
      ...
    },
    readQueryString(window) {
    },
    readConfig(config) {
    },
    logCommand(options) {
    },
    logTool() {
    },
    getCoreLogger() {
    }
}
```
After doing this exercise, I realized that we could probably put the check here to grab the query string parameter even before the configure command.  I see it is here because the query string parameter should override the config, but if we had all the logging logic together we could easily implement this functionality while still having the query string parameter take precedence.  We may even be able to incapsulate the get and set of logEnabled.

test/unit/specs/core/componentCreators.spec.js
Aaronius: Created a JIRA issue: https://jira.corp.adobe.com/browse/CORE-35633

Aaronius:
Created a JIRA ticket here: https://jira.corp.adobe.com/browse/CORE-35634

src/core/configureCommandFactory.js
Aaronius: Regarding `const configureCommand = options => initializeComponents(buildAndValidateConfig(options))`. At that point, there's not really a need for `initializeComponents` and `buildAndValidateConfig` to come from factories. For example, main.js could have this instead:

```
const configureCommand = options => {
  const config = buildAndVailidateConfig({
    config: options,
    componentCreators,
    createConfig,
    configValidators,
    logCommand,
    logger,
    initializeComponents,
    setErrorsEnabled,
    window
  });
  return initializeComponents({
    config,
    componentCreators,
    logger,
    createCookieProxy,
    createComponentNamespacedCookieJar,
    lifecycle,
    componentRegistry,
    tools: {
      config: configToolFactory(),
      cookieJar: cookieJarToolFactory(
        createCookieProxy,
        createComponentNamespacedCookieJar,
        getTopLevelDomain
      ),
      enableOptIn: enableOptInToolFactory(optIn),
      logger: loggerToolFactory(logger),
      network: networkToolFactory(createNetwork, lifecycle, logger)
    },
    optIn
  });
}
```

I'm going to create a separate ticket for this.

PR #217, opened 2019-09-25
CORE-32995 Network-related tests.

Aaronius:
## Description
Unit tests related to network core modules.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-32995

## Motivation and Context
Some level of certainty that our code is working the way we expect.

## Types of changes

## Checklist:


test/unit/specs/core/network/sendBeaconFactory.spec.js
```diff
@@ -0,0 +1,50 @@
...
+    };
+    const fetchPromise = Promise.resolve();
+    const fetch = jasmine.createSpy().and.returnValue(fetchPromise);
+    const logger = {
+      log: jasmine.createSpy()
+    };
+    const sendBeacon = sendBeaconFactory(navigator, fetch, logger);
+    const body = { a: "b" };
+    const result = sendBeacon("https://example.com/endpoint", body);
+    expect(navigator.sendBeacon).toHaveBeenCalledWith(
```
jonsnyder: It may be worth it to add a check for the context of the call.  When I was developing sendBeacon I wanted to just give it navigator.sendBeacon, but the context matters.
```
expect(navigator.sendBeacon.calls.first().object).toBe(navigator)
```

Aaronius:
@jonsnyder Yeah, that's how I see it.  Do you feel like it's too much overlap?

jonsnyder:
> @jonsnyder Yeah, that's how I see it. Do you feel like it's too much overlap?

No, I don't think it's too much.

Aaronius:
I'm going to wait until https://github.com/adobe/alloy/pull/220 gets merged, then clean up the conflicts here.

test/unit/specs/core/network/sendBeaconFactory.spec.js
```diff
@@ -0,0 +1,58 @@
...
+// When running these tests in IE 11, they fail because IE doesn't like the
+// way the blob is constructed (see
+// https://github.com/bpampuch/pdfmake/pull/297/files for a workaround).
+// Fortunately, if navigator.sendBeacon doesn't exist (IE 11), sendBeaconFactory
+// should never be used (see createNetworkStrategy.js), so we can skip
+// these tests altogether.
+if (window.navigator.sendBeacon) {
+  describe("sendBeaconFactory", () => {
+    it("falls back to fetch if sendBeacon fails", () => {
+      const navigator = {
```
jonsnyder: I think it would be better to mark the test as pending so we can see which tests were not run for ie in the reports.

something like this:
```
describe("sendBeaconFactory", () => {
  it("falls back to fetch if sendBeacon fails", () => {
    if (!window.navigator.sendBeacon) {
      pending("No sendBeacon API available");
      return;
    }
   const navigator = {
   ...
```

Aaronius: K, I'll make the change.

PR #219, opened 2019-09-27
Spec updates and additional test cases

shammowla:
1) Testcafe was unable to retrieve the console log error messages with getBrowserConsoleMessages() method, added a catch sandbox helper.js to catch messages and output it with console.log.

2) Added data attributes and updated the test cases. 

3) Added Allure to the gitignore, the the report has its own git config and package for hosting.

4) Testcase C2586 - Toggle logging through the querystring parameter, this spec is broken. Passing false still logs error messages to console. 

5) Adding additional specs to this branch


test/functional/specs/desktop/C2583.js
```diff
@@ -12,74 +12,34 @@ test.meta({
...
   const message2 = await t.getBrowserConsoleMessages();
 
-  await t
-    .expect(message2.log)
-    .match(/Config nologconfig initiated./)
-    .expect(message2.log)
-    .match(/Config event initiated./)
-    .expect(message2.log)
-    .match(/\[alloy] Executing event command./);
+  await t.expect(message2.log).match(/\[alloy] Executing event command./);
```
Aaronius: FYI, this behavior will change in the future: https://jira.corp.adobe.com/browse/CORE-33754

test/functional/specs/desktop/C2587.js
```diff
@@ -13,8 +13,6 @@ test.meta({
 test("Regression: Throw error when executing command that doesn't exist", async () => {
   const message = await t.getBrowserConsoleMessages();
   await t
-    .expect(message.warn)
-    .contains(
-      "[alloy] An error during configuration is preventing the boguscommand command from executing."
-    );
+    .expect(message.error)
+    .match(/\[alloy] Resolve these configuration problems:/);
```
Aaronius: This doesn't seem to be testing what the spec description is describing. If you're getting a message about configuration problems, you're not even getting past the configure command.

test/functional/specs/desktop/C2585.js
```diff
@@ -11,15 +11,15 @@ test.meta({
...
+  await t.click(Selector("#logger-button"));
+  await t.click(Selector("#event-button"));
   const message = await t.getBrowserConsoleMessages();
 
   await t
-    .expect(message.log)
+    .expect(message.warn)
     .match(
-      '[alloy] The library must be configured first. Please do so by calling alloy("configure", {...}).'
+      /\[alloy] An error during configuration is preventing the event command from executing./
```
Aaronius: This doesn't seem to be testing what the spec description is describing. You should be getting a message about needing to run the configure command first like your old code shows. Is something broken?

PR #224, opened 2019-10-02
TNT-33359 - adjust personalization to latest Konductor JSON response

artur-ciocanu:
Adjust personalization code to properly process Konductor personalization JSON response

## Description

The patch adjusts the rendering actions such that those are grouped by rules and each action is responsible for sending data collection details upon success or failure

## Related Issue

NONE

## Motivation and Context

This is required to be aligned with recent Konductor changes for personalization

## Types of changes

## Checklist:


src/components/Personalization/helper/index.js
```diff
@@ -24,23 +26,34 @@ export { default as setAttributes } from "./setAttributes";
...
+      .then(
+        () => {
+          // if everything is OK, notify and show elements
+          collect(meta);
+          showElements(prehidingSelector);
+        },
+        () => {
+          // in case of awaiting timing or error, we need to remove the style tag
+          // hence showing the pre-hidden elements and notify
+          collect(meta);
```
jfkhoury: Will this `reject` callback catch all the exceptions in this flow, or should keep those in a `catch`?

src/components/Personalization/turbine/executeRules.js
```diff
@@ -10,26 +10,28 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
```
jfkhoury: A little unit test for this module would be great.

src/components/DataCollector/index.js
```diff
@@ -68,6 +74,7 @@ const createDataCollector = ({ config, logger, network }) => {
         // it overlays on top of any data Alloy automatically
         // provides. This allows the user to override the
         // automatically collected data.
+        event.mergeMeta(meta);
```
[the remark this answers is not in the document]
artur-ciocanu: @jonsnyder I am not against having a different mechanism for appending `meta` or any other data to the outgoing request. However this code used to  work and probably when `event` command was switched from ordinary `data` to `xdm` and `data`, this functionality has been broken.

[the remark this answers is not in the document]
artur-ciocanu: @jfkhoury and @jonsnyder if we don't want to expose `meta`, it will still be part of the public API exposed by Konductor and we will have to document that anyway.

src/components/Personalization/helper/index.js
artur-ciocanu: @jfkhoury I have tried different combinations and the `rejected` callback catches any error that might be thrown.

An alternative approach would be to guard `collect()` and `showElements()` using a `try/catch` and just have `then().catch()`.

We don't want to accidentally invoke collect twice, although I think it would be good to augment the `error` collect call with an error object, so we could report back to conductor that something is broken.

src/components/DataCollector/index.js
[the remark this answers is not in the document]
Aaronius: I'd rather not expose `meta` to the user until we understand the use case. I understand it will be a part of Konductor's API and will need to be documented there, but that doesn't mean it needs to be a part of Alloy's and I'm all about reducing the cognitive load for the Alloy user where possible.

I think this is pretty easily solvable by changing:

```
commands: {
   event: createEventHandler
}
```

to something like:

```
commands: {
 event(options) {
    const { viewStart, documentUnloading, xdm, data }) = options;
    // We don't want to allow customers to pass a meta option.
    return createEventHandler({
      viewStart,
      documentUnloading,
      xdm,
      data
    });
  }
}
```

I would also change `createEventHandler` to not pass `options` into `lifecycle.onBeforeEvent`. I don't think it provides value and we don't seem to be using it anywhere.

Another option, which is part of a larger validation effort that we should do anyway, is we should be validating options for all commands, similar to how we're validating options passed into the `configure` command. If we did that, and we failed validation on extraneous options, then we could set up validation for the `event` command to not accept a `meta` option.

src/components/Personalization/helper/index.js
Aaronius: This looks good to me.  I like the idea of augmenting the error collect call.

src/components/Personalization/helper/index.js
```diff
@@ -24,23 +26,34 @@ export { default as setAttributes } from "./setAttributes";
 export { default as swapImage } from "./swapImage";
 export { default as rearrangeChildren } from "./rearrangeChildren";
 
+const renderContent = (elements, content, renderFunc) => {
+  const executions = elements.map(element =>
+    Promise.resolve(renderFunc(element, content))
```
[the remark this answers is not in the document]
artur-ciocanu: Removed.

src/components/Personalization/helper/index.js
artur-ciocanu: I'll leave as is and I'll add a separate ticket to capture errors in `collect` call.

src/components/Personalization/index.js
```diff
@@ -44,16 +42,17 @@ const hideElementsForPage = fragments => {
...
+        const { settings = {} } = action;
         const { prehidingSelector } = settings;
 
-        if (prehidingSelector) {
-          hideElements(prehidingSelector);
+        if (!prehidingSelector) {
+          return;
         }
+
+        hideElements(prehidingSelector);
```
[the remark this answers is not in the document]
artur-ciocanu: Removed.

src/components/Personalization/turbine/executeRules.js
artur-ciocanu: Done.

src/components/Personalization/turbine/executeRules.js
```diff
@@ -10,26 +10,28 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
 */
 
+const DEFAULT_EVENT = { moduleType: "libraryLoaded" };
+
+const getEvents = rule => {
+  const { events = [] } = rule;
+
+  return events.length === 0 ? [DEFAULT_EVENT] : events;
```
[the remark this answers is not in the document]
artur-ciocanu: I'll adjust Konductor to always add `libraryLoaded`

src/components/DataCollector/index.js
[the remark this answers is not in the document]
artur-ciocanu: Leaving as is for now. Hopefully will come with a better approach in the next commit.

src/components/Personalization/index.js
```diff
@@ -128,15 +103,9 @@ const createPersonalization = ({ config, logger, cookieJar }) => {
 
         const fragments = response.getPayloadsByType(DECISIONS_HANDLE);
 
-        // On response we first hide all the elements for
-        // personalization:page handle
-        hideElementsForPage(fragments);
```
Aaronius: Sorry if I'm being dumb, but what functionality did we lose by removing this? I'm not sure I follow the hiding/showing process.

artur-ciocanu: Removed

artur-ciocanu: Initially the `prehidingSelector` was part of the `event` now since I moved it to `action` this is no longer required. Each `action` handles this internally.

So thanks for pointing that out, now it's much, much cleaner.

src/components/Personalization/turbine/executeRules.js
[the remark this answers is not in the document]
artur-ciocanu: @Aaronius since JSON response from Konductor should be consumed by Mobile as well, I am wondering if we should force Konductor to add `libraryLoaded` or should we leave it to Alloy to decide.

PR #225, opened 2019-10-03
Add implementationDetails to context

jonsnyder:
## Description

Send the alloy version as part of the xdm.  See https://github.com/adobe/xdm/pull/769

## Related Issue
https://jira.corp.adobe.com/browse/CORE-31964

## Motivation and Context

This will enable us to see which version customers are running.

## Types of changes

## Checklist:

### DIFF
--- a/test/unit/specs/components/Context/implementationDetailsFactory.spec.js
+++ b/test/unit/specs/components/Context/implementationDetailsFactory.spec.js
@@ -1,0 +1,20 @@
+import versionFactory from "../../../../../src/components/Context/implementationDetailsFactory";
+
+describe("Context::versionFactory", () => {
+  const version = "1.2.3";
+  let event;
+  beforeEach(() => {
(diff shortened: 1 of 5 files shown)


Aaronius:
Reminder to add this option to the extension and docs as well.

Aaronius:
Actually, should `implementationDetails` be required?

jfkhoury:
I think it should be required, but let’s bring it up with Justin and Rares

test/unit/specs/components/Context/implementationDetailsFactory.spec.js
```diff
@@ -0,0 +1,20 @@
+import versionFactory from "../../../../../src/components/Context/implementationDetailsFactory";
```
Aaronius: I think you probably meant to name this `implementationDetailsFactory`.

PR #227, opened 2019-10-07
TNT-33684 - adding click action handling

artur-ciocanu:
Adding click action handling

## Description

Adding click action handling, whenever we get a click action from our server side we would save the `selector` and the `meta` to an in-memory data structure.

The assumption here is that whenever some will click on an element, the element will be passed as an argument to the click tracking lifecycle hook and the personalization component would try to find if there is a matching selector. If we would find a matching selector then we would make sure that the associated `meta` is added to the outgoing request.

## Related Issue

NONE

## Motivation and Context

First step for adding click tracking support.

## Types of changes

## Checklist:


src/utils/storageFactory.js
```diff
@@ -50,6 +64,7 @@ export default context => additionalNamespace => {
   const finalNamespace = baseNamespace + (additionalNamespace || "");
   return {
     session: getStorageByType(context, "sessionStorage", finalNamespace),
-    persistent: getStorageByType(context, "localStorage", finalNamespace)
+    persistent: getStorageByType(context, "localStorage", finalNamespace),
+    memory: createMemoryStorage(finalNamespace)
```
liljenback: Volatile keyword here instead of memory, thoughts?

artur-ciocanu: @liljenback let's ask the team :)

src/components/Personalization/helper/click.js
```diff
@@ -0,0 +1,30 @@
...
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+import { uuid } from "../../../utils";
+import { awaitSelector } from "../../../utils/dom";
+import { selectNodesWithEq, setAttribute } from "./dom";
+
+const ALLOY_KEY = "data-alloy-key";
```
liljenback: I think we already discussed if this needed to be configurable and that it won't be required (until asked for), correct?

artur-ciocanu: Yeah, I don't think has to be configurable, but I do think it should probably be part of `constants` so others can access it, thoughts?

src/components/Personalization/helper/click.js
```diff
@@ -0,0 +1,29 @@
...
+governing permissions and limitations under the License.
+*/
+
+import { uuid } from "../../../utils";
+import { awaitSelector } from "../../../utils/dom";
+import ALLOY_KEY from "../../../constants/clickTrackingAttribute";
+import { selectNodesWithEq, setAttribute } from "./dom";
+
+export default (settings, store) => {
+  const key = uuid();
```
Aaronius: This isn't a big deal, but it might be good to just move this line down inside the `.then` so that we only generate a UUID if elements matching the selector are found.

src/components/Personalization/turbine/initRuleComponentModules.js
```diff
@@ -24,10 +24,11 @@ import {
   appendHtml,
   prependHtml,
   insertHtmlAfter,
-  insertHtmlBefore
+  insertHtmlBefore,
+  click
 } from "../helper";
 
-export default collect => {
+export default (collect, storage) => {
```
Aaronius: This parameter should probably be named `store` to be consistent.

src/components/Personalization/helper/click.js
```diff
@@ -0,0 +1,29 @@
...
+  const { selector, meta } = settings;
+
+  return awaitSelector(selector, selectNodesWithEq).then(elements => {
+    store(key, meta);
+
+    elements.forEach(element => {
+      setAttribute(element, ALLOY_KEY, key);
+    });
```
Aaronius: Do we ever have elements that are related to multiple `settings` objects? If so, the `data-alloy-key` UUID that was stored on the element from the first `settings` object will get overwritten by the UUID created for the second `settings` object, which I assume would be a bad thing.

src/constants/clickTrackingAttribute.js
```diff
@@ -0,0 +1 @@
+export default "data-alloy-key";
```
Aaronius: Do you think this will be used for components other than personalization? If not, then this file should probably be stored inside the personalization directory. If it will be used for components other than personalization, then I think how `data-alloy-key` values are assigned and how storage is managed will need some tweaking (either now or later). Just thinking out loud a bit here, but right now the UUID assigned to the elements are specific to the "click" `settings` object. In other words, multiple elements will share the same UUID. This works for the personalization case, but doesn't work for the generic case of wanting to store some information for a specific element. You may have thought about this already and are thinking we resolve it at the point in time that we need to share the storage across components. Anyway, let me know if you want to talk it through.

Aaronius:
After thinking about this some more, I feel like this is a great use case for WeakMap. Rather than creating a UUID and storing it on the element `data-alloy-key` attribute and then tying that to a storage object key, I think we should just store the element itself as the key inside a WeakMap. This should reduce the complexity and also cleans up the stored data automatically when the element is freed from memory.

src/components/Personalization/helper/click.js
artur-ciocanu: @Aaronius the `settings` will contain a `selector` and we try to decorate ALL the elements matched by that `selector`. I am not sure what you mean by "multiple settings object".

src/components/Personalization/helper/click.js
artur-ciocanu: Done.

src/components/Personalization/turbine/initRuleComponentModules.js
artur-ciocanu: Done

src/constants/clickTrackingAttribute.js
artur-ciocanu: I have moved this to `constants` assuming that data collector click tracking functionality would need to know the name of the attribute when it starts to scan the DOM elements. Once the `data-alloy-key` attribute is found, it's value will be passed into `click` life-cycle hook and personalization will have to subscribe to this hook, similar to `onResponse`.

Probably something will change when @liljenback will finish his work, but I don't think we will have a lot of changes.

src/components/Personalization/helper/click.js
Aaronius: Sorry I wasn't clear. Let's say there's a rule with an action with a `moduleType` of `click` and the value of its `settings.selector` is `.foo`. Then let's say there's another rule with an action with a `moduleType` of `click` and the value of its `settings.selector` is `.bar`. Then let's say there's an element on the page with `class="foo bar"`. In this case, if I'm following the code right (which I very well may not), when the first rule executes, a UUID will be generated and applied to the element's `data-alloy-key` attribute and meta will be stored using that UUID. Then, when the second rule executes, a new UUID will be generated for the exact same element. That UUID would be used to override the `data-alloy-key` attribute that had previously been applied to that element. I'm guessing that would foul up the tracking part of the process.

src/constants/clickTrackingAttribute.js
Aaronius: I thought we had discussed having DataCollector either (1) pass the element that was clicked (instead of searching for elements with `data-alloy-key`) or (2) pass an array of elements, where each one is an element in the hierarchy of the element that was clicked. In each case, Personalization would be the only one aware of `data-alloy-key`.  I think it makes more sense to get rid of `data-alloy-key` and instead have DataCollector pass the element that was clicked and then personalization would iterate over the hierarchy and look up each element in the WeakMap to see if there's an entry (e.g., `metaByElement.has(element)`)

jfkhoury: @Aaronius is correct. This way other Components in the future can implement this lifecycle hook without it being specific to Personalization.

jonsnyder:
> After thinking about this some more, I feel like this is a great use case for WeakMap. Rather than creating a UUID and storing it on the element `data-alloy-key` attribute and then tying that to a storage object key, I think we should just store the element itself as the key inside a WeakMap. This should reduce the complexity and also cleans up the stored data automatically when the element is freed from memory.

Would this approach work with the myriad of javascript frameworks that may replace or modify the content dynamically?  For example what would happen if the content is some HTML that then is modified by a javascript framework which would have copied over the data element, but would not preserve the identity of the DOM node?

src/components/Personalization/helper/click.js
jonsnyder: To solve this you could put multiple values in the key data element. (maybe separated by space).

test/unit/specs/components/Personalization/actions/click.spec.js
```diff
@@ -0,0 +1,36 @@
...
+    appendNode(document.body, element);
+
+    const meta = { a: 1 };
+    const settings = {
+      selector: "#click",
+      meta
+    };
+
+    return click(settings, store).then(() => {
+      const key = elements[0].getAttribute("data-alloy-key");
```
jonsnyder: This should reference "constants/clickCollectionAttribute"

Aaronius:
Why are we trying to find all the matching elements up front in the first place? Can we just test the elements against the selectors as the user clicks on them?

1. The user clicks on an element.
2. DataCollector passes the clicked-on element to Personalization through the lifecycle hook.
3. For each selector that has come from the server, Personalization calls `element.matches(selector)` to see if the element matches.
4. If the element matches, pass the corresponding meta or whatever to DataCollector so it can be sent to the server. If the element doesn't match, continue traversing the ancestry of the element and repeating step 3 & 4 for each.

This is what we do for many of the event types in the Launch Core extension. It's great because you don't have to worry about DOM changes that might occur between the time you search for elements and when the click occurs.

artur-ciocanu:
@Aaronius and @jonsnyder thanks a lot for your help. I have made a few changes. I will try to apply the idea of matching selectors when we'll have the click lifecycle hooks.

src/components/Personalization/index.js
```diff
@@ -37,16 +38,25 @@ const createCollect = collect => {
...
 
+const createStore = storage => {
+  return (selector, meta) => {
+    storage.push({ selector, meta });
+  };
+};
+
 const createPersonalization = ({ config, logger }) => {
   const { authoringModeEnabled, prehidingStyle } = config;
+  const storage = [];
```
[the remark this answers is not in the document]
artur-ciocanu: Will do

src/components/Personalization/turbine/initRuleComponentModules.js
```diff
@@ -45,7 +46,6 @@ export default collect => {
     replaceHtml: createAction(collect, replaceHtml),
     prependHtml: createAction(collect, prependHtml),
     appendHtml: createAction(collect, appendHtml),
-    // TODO: NOT IMPLEMENTED
-    click: () => {}
+    click: settings => click(settings, store)
   };
 };
```
jonsnyder: I'm wondering if there is a better word than click here.  All of these are verbs or verb phrases, as is click, but for the other commands they are telling the client code what to do.  In the case of click, we are telling the client code what to collect when something is clicked.  What if we changed this to `collectClicks`?

artur-ciocanu: @jonsnyder it is already part of Konductor response, so to change it I would have to change Konductor, not a big deal, but I would like to leave it as is fir now

PR #228, opened 2019-10-08
Rename authState - Support setting a primary ID

jfkhoury:
### 1st change:

We have a new requirement for the integration with AEP: Set a primary ID in your payload/identityMap.

This will be supported in the Config Service as well.

Identity Item schema in XDM already supports this property: [DOC](https://github.com/adobe/xdm/blob/master/docs/reference/context/identity.schema.md#identity-properties)

### 2nd change:

Rename `authState` to `authenticatedState` since that's how XDM defines it: [DOC](https://github.com/adobe/xdm/blob/master/docs/reference/context/identityitem.schema.md#xdmauthenticatedstate)

Per the XDM doc, we also had to change the values from `0, 1, 2` to `ambiguous, authenticated, loggedOut`. This change might require we expose the enum to make it easier for customers to set those values. Thoughts??

## Types of changes

- [ x ] New feature (non-breaking change which adds functionality)

## Checklist:

- [ x ] All tests pass and I've made any necessary test changes.

### DIFF
--- a/src/components/Identity/customerIds/util.js
+++ b/src/components/Identity/customerIds/util.js
@@ -1,20 +1,33 @@
 import { AUTH_STATES } from "../constants";
-import { isObject, includes, keys, values } from "../../../utils";
+import {
+  isObject,
+  includes,
+  keys,
+  values,
+  isBoolean,
+  isNil
+} from "../../../utils";
 
 const ERROR_MESSAGE = "Invalid customer ID format.";
 const NOT_AN_OBJECT_ERROR = "Each namespace should be an object.";
 const NO_ID_ERROR = "Each namespace object should have an ID.";
+const PRIMARY_NOT_BOOLEAN = "The `primary` property should be true or false.";
 
 const validateCustomerIds = customerIds => {
   if (!isObject(customerIds)) {
     throw new Error(`${ERROR_MESSAGE} ${NOT_AN_OBJECT_ERROR}`);
   }
   keys(customerIds).forEach(customerId => {
+    const { primary } = customerIds[customerId];
+
     if (!isObject(customerIds[customerId])) {
       throw new Error(`${ERROR_MESSAGE} ${NOT_AN_OBJECT_ERROR}`);
     }
     if (!customerIds[customerId].id) {
       throw new Error(`${ERROR_MESSAGE} ${NO_ID_ERROR}`);
+    }
+    if (!isNil(primary) && !isBoolean(primary)) {
+      throw new Error(`${ERROR_MESSAGE} ${PRIMARY_NOT_BOOLEAN}`);
     }
   });
 };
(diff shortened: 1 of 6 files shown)


src/components/Identity/customerIds/util.js
```diff
@@ -30,15 +43,21 @@ const sortObjectKeyNames = object => {
...
+
     normalizedIds[customerId] = {
       id,
-      authState: includes(authStates, authState)
-        ? authState
-        : AUTH_STATES.UNKNOWN
+      authenticatedState: includes(authStates, authenticatedState)
+        ? authenticatedState // Set the auth state to the string value like `authenticated`.
+        : AUTH_STATES.AMBIGUOUS,
+      ...(primary !== undefined && { primary })
```
Aaronius: Fancy.

src/components/Identity/customerIds/util.js
```diff
@@ -1,26 +1,39 @@
...
       throw new Error(`${ERROR_MESSAGE} ${NOT_AN_OBJECT_ERROR}`);
     }
     if (!customerIds[customerId].id) {
       throw new Error(`${ERROR_MESSAGE} ${NO_ID_ERROR}`);
     }
+    if (!isNil(primary) && !isBoolean(primary)) {
+      throw new Error(`${ERROR_MESSAGE} ${PRIMARY_NOT_BOOLEAN}`);
+    }
   });
 };
```
jonsnyder: Do you also need to validate that only one customerId has primary set?

jfkhoury:
> My understanding of the primary customer id is that it is used as the seed for the marketing cloud visitor id. i.e. it is hashed and xor-ed with some value and that becomes the visitor id. At least inside analytics collection, the primary id is configured server side (and must match the ismcseed query param). So why can't the primary id be configured server-side?

@jonsnyder, this is more for AEP than Identity Feature in Konductor. If you send data to the Unified Profile, you are required to set a primary ID. There will be logic in Konductor to fallback to device IDs if no declared IDs are provided. And even if no declared IDs were set as primary, Konductor will set one randomly, unless the customer has set it in the Config Service already.

src/components/Identity/customerIds/util.js
```diff
@@ -1,26 +1,39 @@
...
+  keys,
+  values,
+  isBoolean,
+  isNil
+} from "../../../utils";
 
 const ERROR_MESSAGE = "Invalid customer ID format.";
 const NOT_AN_OBJECT_ERROR = "Each namespace should be an object.";
 const NO_ID_ERROR = "Each namespace object should have an ID.";
+const PRIMARY_NOT_BOOLEAN = "The `primary` property should be true or false.";
```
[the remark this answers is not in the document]
jfkhoury: I thought it makes it clearer, but maybe it's too much :)

PR #230, opened 2019-10-09
CORE-33754 CORE-35634 Improved persistence of log enablement.

Aaronius:
## Description
Right now, there are three ways of turning on logging:

(1) Through the logEnabled option during configuration.
(2) Using the log command.
(3) Setting the alloy_log querystring parameter.

For all scenarios, if logging is enabled, it stays enabled until you explicitly set it to false. I've found behavior to be awkward in scenario #1. If I set logEnabled to true in configuration, and then I remove the logEnabled option, I expect logging to turn off, but right now it stays on until I set logEnabled: false.

I brought this up with the alpha testers and I agree with the feedback we received, that is, if logEnabled: true in configuration, logging should be enabled. If logEnabled is set to false or removed entirely from configuration, logging should be turned off.

If logging is enabled by using the log command or the alloy_log querystring parameter, logging should stay on for the session (right now we're using local storage, so we should switch to session storage) or until the user explicitly runs the log command with enabled: false or sets the alloy_log querystring parameter value to false.

The log command and the alloy_log querystring parameter should always trump the logEnabled configuration option.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-33754
https://jira.corp.adobe.com/browse/CORE-35634

## Motivation and Context
More predictable behavior when toggling logging.

## Types of changes

## Checklist:

- [ ] I'll queue up a PR right now -> I have updated the documentation accordingly.

### DIFF
--- a/src/core/buildAndValidateConfig.js
+++ b/src/core/buildAndValidateConfig.js
@@ -27,7 +27,7 @@ export default ({
   });
   config.validate();
   setErrorsEnabled(config.errorsEnabled);
-  setLogEnabled(config.logEnabled, { persist: false, highPriority: false });
+  setLogEnabled(config.logEnabled, { fromConfig: true });
   // toJson is expensive so we short circuit if logging is disabled
   if (logger.enabled) {
     logger.log("Computed configuration:", config.toJSON());
--- a/src/core/buildAndValidateConfig.js
+++ b/src/core/buildAndValidateConfig.js
@@ -10,18 +10,14 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import { queryString, stringToBoolean } from "../utils";
-import logQueryParam from "../constants/logQueryParam";
-
 export default ({
   options,
   componentCreators,
   createConfig,
   coreConfigValidators,
-  logCommand,
   logger,
-  setErrorsEnabled,
-  window
+  setLogEnabled,
+  setErrorsEnabled
 }) => {
   const config = createConfig(options);
   config.addValidators(coreConfigValidators);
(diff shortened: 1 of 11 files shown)


src/core/buildAndValidateConfig.js
```diff
@@ -31,14 +27,10 @@ export default ({
...
   config.validate();
   setErrorsEnabled(config.errorsEnabled);
-  const parsedQueryString = queryString.parse(window.location.search);
-  logCommand({
-    enabled:
-      parsedQueryString[logQueryParam] !== undefined
-        ? stringToBoolean(parsedQueryString[logQueryParam])
-        : config.logEnabled
-  });
+  setLogEnabled(config.logEnabled, { persist: false, highPriority: false });
```
jonsnyder: Setting persist and highPriority feels like the LogController is a leaky abstraction.  What if we added a source parameter?  So the sources could be: `"config", "queryString", or "command"` Then the log controller can determine what do to with each source, and how they overwrite each other.

Aaronius: Yeah, I think you're right. I'll make the change.

Aaronius:
I decided to go with a `fromConfig` boolean instead of a source enumeration. I think anything that can or may change persistence can probably be lumped into "config" or "everything else" and whatever comes last in the "everything else" bucket trumps whatever came before it. We can move to an enum later if we find we need it.

jfkhoury:
Nice work @Aaronius. Do you still need to update the docs? Need a ticket?

Aaronius:
I'll update the docs today (it looks like I lied in my PR when I said I'd queue up a PR 😢). No need for a ticket. I triple promise I'll get it done today.

src/core/createLogger.js
```diff
@@ -42,19 +42,10 @@ const createLogger = (console, logController, prefix) => {
      * Outputs an error message to the web console.
      * @param {...*} arg Any argument to be logged.
      */
-    error: process.bind(null, "error"),
+    error: process.bind(null, "error")
     /**
      * Creates a new logger with an additional prefix.
      * @param {String} additionalPrefix
      */
```
enortonAtAdobe: should this comment be removed since the subsequent code has been?

Aaronius: Yes, thanks! I'm submit a new PR for that.

PR #231, opened 2019-10-11
Feature/support more link types

liljenback:
Adding support for download links and exit links

CORE-35830
CORE-36239

## Motivation and Context

Feature is part of the XDM specification we opted to support.

## Types of changes

## Checklist:


src/components/DataCollector/activity/click.js
```diff
@@ -29,14 +41,16 @@ const createClickHandler = (window, logger, collect) => {
       }
     }
     if (linkUrl && isSupportedAnchorElement(clickedObj)) {
+      const linkType = determineLinkType(window, config, linkUrl, clickedObj);
+      const linkName = "Link Click";
       // TODO: Update name (link name) and support exit, other, and download link types
```
jonsnyder: Part of this TODO is now complete.

src/components/DataCollector/utils.js
```diff
@@ -50,8 +50,23 @@ const isSupportedAnchorElement = element => {
   return false;
 };
 
+const isDownloadLink = (downloadLinkQualifier, linkUrl, clickedObj) => {
```
jonsnyder: Instead of passing the downloadLinkQualifier as a String and then creating a RegExp every time, you could create the RegExp once in createClickHandler, and then pass the RegExp object.

src/components/DataCollector/utils.js
```diff
@@ -50,8 +50,23 @@ const isSupportedAnchorElement = element => {
...
+  const re = new RegExp(downloadLinkQualifier);
+  return clickedObj.download ? true : re.test(linkUrl.toLowerCase());
+};
+
+const isExitLink = (window, linkUrl) => {
+  const currentHostname = window.location.hostname;
```
jonsnyder: Is `window.location.hostname` guaranteed to be lowercase already?

src/components/DataCollector/utils.js
```diff
@@ -50,8 +50,23 @@ const isSupportedAnchorElement = element => {
...
+  if (linkUrl.toLowerCase().indexOf(currentHostname) >= 0) {
```
jonsnyder: Wouldn't this match?
hostname = example.com
linkUrl = http://example.com.foo.com/bar

sandbox/src/Links.js
```diff
@@ -14,13 +14,38 @@ export default function Links() {
...
+        records and transmits an exit-link web interaction event.
+      </p>
+      <p>
+        When clicking on this{" "}
+        <a href="download.zip" target="_blank">
+          download link
+        </a>
+        , Alloy records and transmits a download-link web interaction event.
+      </p>
+      <p>
```
jonsnyder: Can you add an example where the download property is set on a link?

src/components/DataCollector/createConfigValidators.js
```diff
@@ -27,6 +27,11 @@ export default () => {
     clickCollectionEnabled: {
       defaultValue: true,
       validate: boolean()
+    },
+    downloadLinkQualifier: {
+      defaultValue:
+        "\\.(exe|zip|wav|mp3|mov|mpg|avi|wmv|pdf|doc|docx|xls|xlsx|ppt|pptx)$",
+      validate: string().regexp()
     }
```
jonsnyder: Do we have plans to support linkInternalFilters like in AppMeasurement?

src/components/DataCollector/activity/click.js
```diff
@@ -29,14 +41,16 @@ const createClickHandler = (window, logger, collect) => {
...
       collect({
```
jonsnyder: Do we want to automatically set `documentUnloading` here based on the link type?

src/components/DataCollector/utils.js
liljenback: I read that it was but I can probably throw in a toLowerCase to be safe

src/components/DataCollector/activity/click.js
jfkhoury: I like that!

src/components/DataCollector/utils.js
liljenback: It would. If the domain name is repeated or included in an outbound exit link it would trigger as an internal link. Although unlikely we can enhance this later and parse the link-url for the domain.

src/components/DataCollector/createConfigValidators.js
liljenback: I think we can leave that one out until we have a valid use-case.

src/components/DataCollector/activity/click.js
liljenback: Yes, good point

src/components/DataCollector/utils.js
liljenback: I was thinking about doing that but added the extra code in the function instead. The performance impact will be minimal.

sandbox/src/Links.js
liljenback: I will add this to the documentation instead.

PR #238, opened 2019-10-16
Moving identityMap under xdm  in payload

jileeshs:
Updated `payload.addIdentity` to match Konductor change

## Description
Konductor is expecting `identityMap` under `xdm` in request payload

## Related Issue
https://jira.corp.adobe.com/browse/CORE-36460

## Types of changes

## Checklist:

- [ ] All tests pass and I've made any necessary test changes.


Aaronius:
From the conversation in standup today I came away with the impression that `identityMap` was going to be added to the `xdm` property under each event rather than under payload. It looks like I misunderstood?

jfkhoury:
> From the conversation in standup today I came away with the impression that `identityMap` was going to be added to the `xdm` property under each event rather than under payload. It looks like I misunderstood?

Sorry I should have left a comment @Aaronius. I misunderstood as well and thought it's going under each event. This is much better!

PR #239, opened 2019-10-16
CORE-34189 Added note to all errors about how to suppress errors.

Aaronius:
## Description
CORE-34189 Added note to all errors about how to suppress errors.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-34189

## Motivation and Context
Help users understand how they can modify error behavior.

## Types of changes

## Checklist:

### DIFF
--- a/src/core/handleErrorFactory.js
+++ b/src/core/handleErrorFactory.js
@@ -12,9 +12,12 @@ governing permissions and limitations under the License.
 
 import { toError } from "../utils";
 
+const suppressionNote =
+  "Note: Errors can be suppressed via the errorsEnabled configuration option.";
+
 export default ({ instanceNamespace, getErrorsEnabled, logger }) => error => {
   const err = toError(error);
-  err.message = `[${instanceNamespace}] ${err.message}`;
+  err.message = `[${instanceNamespace}] ${err.message}\n${suppressionNote}`;
   if (getErrorsEnabled()) {
     throw err;
   } else {
--- a/src/core/handleErrorFactory.js
+++ b/src/core/handleErrorFactory.js
@@ -13,12 +13,13 @@ governing permissions and limitations under the License.
 import { toError } from "../utils";
 
 const suppressionNote =
-  "Note: Errors can be suppressed via the errorsEnabled configuration option.";
+  "Note: Errors can be suppressed by setting the errorsEnabled configuration option to false.";
 
 export default ({ instanceNamespace, getErrorsEnabled, logger }) => error => {
   const err = toError(error);
-  err.message = `[${instanceNamespace}] ${err.message}\n${suppressionNote}`;
+  err.message = `[${instanceNamespace}] ${err.message}`;
   if (getErrorsEnabled()) {
+    err.message += `\n${suppressionNote}`;
     throw err;
   } else {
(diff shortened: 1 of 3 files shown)


Aaronius:
I'm not sure if we actually want to do this, but I'm still trying to figure out how to create URL aliases so we can include a link to the docs. We'll probably need to use Go URLs: https://wiki.corp.adobe.com/pages/viewpage.action?pageId=1039074007

src/core/network/createPayload.js
```diff
@@ -20,7 +20,8 @@ export default () => {
     addIdentity: (namespaceCode, identity) => {
       content.xdm = content.xdm || {};
       content.xdm.identityMap = content.xdm.identityMap || {};
-      content.xdm.identityMap[namespaceCode] = content.xdm.identityMap[namespaceCode] || [];
+      content.xdm.identityMap[namespaceCode] =
+        content.xdm.identityMap[namespaceCode] || [];
```
Aaronius: For some reason people are able to commit+push when code doesn't match prettier rules. I have a note to figure out why this is happening.

jfkhoury:
> I'm not sure if we actually want to do this, but I'm still trying to figure out how to create URL aliases so we can include a link to the docs. We'll probably need to use Go URLs: https://wiki.corp.adobe.com/pages/viewpage.action?pageId=1039074007

Most complicated WIKI I've read so far.

src/core/handleErrorFactory.js
```diff
@@ -12,9 +12,12 @@ governing permissions and limitations under the License.
 
 import { toError } from "../utils";
 
+const suppressionNote =
+  "Note: Errors can be suppressed via the errorsEnabled configuration option.";
+
```
jonsnyder: It may not be clear what you have to set the errorsEnabled configuration option in order to suppress errors.

`Note: Errors can be suppressed by setting errorsEnabled to false as an option to configure`

src/core/handleErrorFactory.js
```diff
@@ -12,9 +12,12 @@ governing permissions and limitations under the License.
...
 export default ({ instanceNamespace, getErrorsEnabled, logger }) => error => {
   const err = toError(error);
-  err.message = `[${instanceNamespace}] ${err.message}`;
+  err.message = `[${instanceNamespace}] ${err.message}\n${suppressionNote}`;
   if (getErrorsEnabled()) {
     throw err;
   } else {
```
jonsnyder: Assuming logging is enabled, the suppressionNote will be logged even when errors are not enabled.

src/core/handleErrorFactory.js
Aaronius: I was hoping it would subtly prompt them to go read the documentation before just setting it to `false` so they could understand the ramifications a little more. I'll change it though.

src/core/handleErrorFactory.js
Aaronius: That's a good point. I should only add the note if errorsEnabled is false. I'll make the change.

Aaronius: I mean....true.

Aaronius:
@jonsnyder Fixed.

PR #240, opened 2019-10-16
Fix setCustomerIds promise chain and tests.

Aaronius:
## Description
When working on https://github.com/adobe/alloy/pull/239, I was testing whether the error message would be successfully appended with the message about how to suppress errors. I happened to add `throw new Error("foo")` right before [this line](https://github.com/adobe/alloy/blob/1666324ac0790190f080efe0c948890c06be5b9a/src/components/Identity/index.js#L90) inside `onBeforeDataCollection` in the Identity component. What I discovered is that when using the `setCustomerIds` command, the promise returned from `sync()` inside `createCustomerIds.js` wasn't incorporating the lifecycle promise chain. Because of this, any errors thrown during those asynchronous processes (e.g., `lifecycle.onBeforeEvent`, `lifecycle.onBeforeDataCollection`, `network.sendRequest`) weren't being handled by our global handler. This meant that such errors weren't being properly suppressed by `errorsEnabled: false` or being given additional context like other errors in Alloy are. It also meant the promise being returned from the command was being resolved before the customer IDs were fully synced with the server.

While I was fixing that, I noticed that some of the tests weren't returning promises when they were supposed to, which meant Jasmine was considering the tests as passed before the promises resolved. Tests were succeeding when really they should have been failing. I fixed those as well.

## Related Issue
None

## Motivation and Context
Predictable and consistent behavior.

## Types of changes

## Checklist:

### DIFF
--- a/test/unit/specs/components/Identity/customerIds/createCustomerIds.spec.js
+++ b/test/unit/specs/components/Identity/customerIds/createCustomerIds.spec.js
@@ -27,21 +27,20 @@ describe("Identity::createCustomerIds", () => {
       set: jasmine.createSpy()
     };
     lifecycle = {
-      onBeforeEvent: jasmine
-        .createSpy()
-        .and.returnValue(() => Promise.resolve()),
+      onBeforeEvent: jasmine.createSpy().and.returnValue(Promise.resolve()),
       onBeforeDataCollection: jasmine
         .createSpy()
-        .and.returnValue(() => Promise.resolve())
+        .and.returnValue(Promise.resolve())
     };
     payload = {
       addEvent: jasmine.createSpy(),
       addIdentity: jasmine.createSpy(),
-      mergeMeta: jasmine.createSpy()
+      mergeMeta: jasmine.createSpy(),
+      expectsResponse: false
     };
     network = {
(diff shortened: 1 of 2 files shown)


jfkhoury:
Awesome PR! Thank you for noticing and fixing.

test/unit/specs/components/Identity/customerIds/createCustomerIds.spec.js
```diff
@@ -83,10 +82,18 @@ describe("Identity::createCustomerIds", () => {
...
-      const test = customerIds.sync(idsWithHash);
-      test.then(() => {
-        expect(lifecycle.onBeforeEvent).toHaveBeenCalledWith();
-        expect(lifecycle.onBeforeDataCollection).toHaveBeenCalledWith(payload);
+      return customerIds.sync(idsWithHash).then(() => {
+        // Normally we would test that the event passed to onBeforeEvent
+        // was the event created by createEvent(), but createEvent() is
+        // imported directly from the DataCollector component so we can't
+        // mock it. This should probably all change once Konductor opens
+        // up an endpoint specific to setting customer IDs.
```
jfkhoury: 👌🏻👌🏻

PR #241, opened 2019-10-17
Added an onClick lifecycle

liljenback:
## Description

Created a new onClick lifecycle and separated out the click handler into a generic click handler under the new ActivityCollector component.

## Related Issue

CORE-33897

## Motivation and Context

The main reason is to integrate the click handler used for link clicks to also support personalization.

## Types of changes

## Checklist:


src/components/ActivityCollector/createLinkClick.js
```diff
@@ -0,0 +1,73 @@
...
+  let linkType = "other";
+  if (isDownloadLink(config.downloadLinkQualifier, linkUrl, clickedObj)) {
+    linkType = "download";
+  } else if (isExitLink(window, linkUrl)) {
+    linkType = "exit";
+  }
+  return linkType;
+};
+
+export default (window, config, clickedObject) => {
```
jonsnyder: I think this would be better encapsulated if you returned the lifecycle function instead of an object with two methods.

```
export default (window, config) => {
  return (event, clickedObject) => {
    ...
  };
};
```

src/components/ActivityCollector/createLinkClick.js
```diff
@@ -0,0 +1,73 @@
...
+  return {
+    isValid() {
+      return isValidLink;
+    },
+    populateEvent(event) {
+      if (linkType === "exit") {
+        event.documentUnloading();
+      }
+      event.mergeXdm({
+        eventType: "web.webinteraction.linkClicks",
```
jonsnyder: What happens if the ActivityCollector determines the click is not a valid link, and the Personalization component determines that the clicked on object has attached data that needs to be sent?  How would the personalization component know to set an eventType?  There are a few options to solve for this:

1. Leave things how they are in this PR.  Require the Personalization component to set `eventType: "web.webinteraction.linkClicks"` when adding event data.  This way the order of the lifecycle resolution would not matter.
2. Use multiple events - pass a payload object to onClick and have the lifecycle method implementations add events as needed.
3. Set the event type in the clickHandler.

src/components/DataCollector/index.js
```diff
@@ -52,9 +50,10 @@ const createDataCollector = ({ config, logger, network }) => {
       });
   };
 
-  const createEventHandler = options => {
-    const event = createEvent();
-    const { viewStart = false, documentUnloading = false, xdm, data } = options;
+  const createEventHandler = (options, event = createEvent()) => {
+    const { viewStart = false, xdm, data } = options;
+    const documentUnloading =
+      options.documentUnloading || event.isDocumentUnloading();
```
jonsnyder: I like the addition of isDocumentUnloading() on the event, but for consistency we should probably set it on the event before calling onBeforeEvent, change the onBeforeEvent handlers to use that, and use it in makeServerCall.

```
if (options.documentUnloading) {
  event.documentUnloading();
}
return lifecycle
  .onBeforeEvent({ event, options, isViewStart: viewStart }) // notice I removed isDocumentUnloading
  .then(() => {
    ...
  })
  .then(() => makeServerCall(event)) // use the event.isDocumentUnloading() inside makeServerCall
};
```

src/components/DataCollector/createEvent.js
```diff
@@ -24,12 +26,21 @@ export default () => {
...
     mergeMeta: createMerger(content, "meta"),
     mergeQuery: createMerger(content, "query"),
+    documentUnloading() {
+      documentUnloading = true;
+    },
+    isDocumentUnloading() {
+      return documentUnloading;
+    },
     expectResponse() {
       expectsResponse = true;
```
jonsnyder: We could do some validation here and within documentUnloading() so that you can't expect a response when the document is unloading and vice versa.  I'm not sure if we would want to throw an error, log a warning, not allow you to, or have some kind of priority.  I haven't thought it through.

src/components/ActivityCollector/createClickActivityCollector.js
```diff
@@ -0,0 +1,35 @@
...
+
+import createEvent from "../DataCollector/createEvent";
+
+const createClickHandler = (collect, lifecycle) => {
+  return clickEvent => {
+    // TODO: Consider safeguarding from the same object being clicked multiple times in rapid succession?
+    const clickedObj = clickEvent.target;
+    const event = createEvent();
+    lifecycle.onClick(event, clickedObj).then(() => {
+      if (!event.isEmpty()) {
```
jonsnyder: It feels weird to check `isEmpty` when what you want to accomplish is to see if the event should be sent.  I'm wondering if there is a use case for sending nothing (knowing the context would update it).  Conversely, I'm wondering if there is a use case where a component could augment the event data from the clickedObject, but use a different component to determine if the event should be sent.  Does it make sense to have a `shouldSend` and `isShouldSend` methods?  Conversely, the isEmpty works for Personalization and ActivityCollector so maybe I'm overthinking this.

src/components/ActivityCollector/index.js
```diff
@@ -0,0 +1,42 @@
...
+const createActivityCollector = ({ config }) => {
+  return {
+    lifecycle: {
+      onComponentsRegistered(tools) {
+        const { lifecycle, componentRegistry } = tools;
+        const collect = componentRegistry.getCommand(EVENT_COMMAND);
+        createClickActivityCollector(config, collect, lifecycle);
+        // TODO: createScrollActivityCollector ...
+      },
+      onClick(event, clickedObject) {
```
jonsnyder: After submitting my first review I saw this problem.  All the lifecycle events so far pass objects with named arguments, so it should be onClick({ event, clickedObject })

src/components/ActivityCollector/createClickActivityCollector.js
```diff
@@ -0,0 +1,35 @@
+/*
```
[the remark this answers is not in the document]
jfkhoury: I agree, this is more of an action than a factory.

src/components/ActivityCollector/createClickActivityCollector.js
liljenback: Right, so the component creating the event in this scenario will need to know if it should be sent or not. In the current case if it has been modified it needs to get sent. Going forward this might not be sufficient to determine whether to send it or not. If that is the case we can move the responsibility to any logic modifying the event. We could change the naming of the function.

src/components/ActivityCollector/createLinkClick.js
Aaronius: Yeah, this is interesting. Does Personalization only intend to send data for link clicks? Or could it potentially send data for non-link clicks as well? If it's the latter, then `eventType` probably shouldn't be set to `web.webinteraction.linkClicks` in that case, right?

Aaronius:
I'd like to talk/type more about https://github.com/adobe/alloy/pull/241/files#r336211963 as I think it may affect the overall design.

src/components/ActivityCollector/createLinkClick.js
liljenback: We touched on this earlier. The eventType doesn't carry much valuable information when the data it directly correlates to is part of the payload.

Aaronius: Sorry, I don't follow. Sorry if I missed the discussion.

liljenback: I will remove eventType for linkClicks for now. If there is a need to put it back we'll have to decide how to change the request.

src/components/ActivityCollector/createLinkClick.js
liljenback: Agreed 👍

src/components/ActivityCollector/createLinkClick.js
```diff
@@ -0,0 +1,73 @@
...
+export default (window, config, clickedObject) => {
+  let linkName;
+  let linkType;
+  let clickedObj = clickedObject;
+  let linkUrl = getAbsoluteUrlFromAnchorElement(window, clickedObj);
+  let isValidLink = false;
+  // Search parent elements for an anchor element
+  // TODO: Replace with generic DOM tool that can fetch configured properties
+  while (clickedObj && clickedObj !== document.body && !linkUrl) {
+    clickedObj = clickedObj.parentElement || clickedObj.parentNode;
```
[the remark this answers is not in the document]
liljenback: This is from legacy code and I believe this is the reason: https://developer.mozilla.org/en-US/docs/Web/API/Node/parentElement#Browser_compatibility. That said I don't know (and have not reproduced) a scenario where parentElement has failed (I did try this). I just don't want to resurface a previous bug.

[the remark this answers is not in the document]
Aaronius: Can we please remove it until we find and understand a case? I would prefer not to have code in the codebase that's not understood or for which a test could not be written.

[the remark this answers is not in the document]
Aaronius: Sorry, I may have been unclear. I think `clickedObj.parentNode` should stay and we should get rid of `clickedObj.parentElement`. I understand `clickedObj.parentElement` may fail, but if it can fail but `clickedObj.parentNode` can't, I don't understand why we try to use `clickedObj.parentElement` at all.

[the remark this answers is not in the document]
liljenback: It does look like parentNode will always be available.

PR #242, opened 2019-10-18
CORE-36571 Prevent developers from committing code not matching standards

Aaronius:
## Description
If a developer has some code that doesn't match prettier formatting rules, then commits and pushes the code, prettier will fix the code to match the rules, but the changes will be stored as working changes rather than being included in the commit. 

This PR runs prettier and eslint, both in fix/write mode, on staged changes during a commit and includes the fixes on the commit itself. If prettier or eslint fail, the commit will be aborted.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-36571

## Motivation and Context
Code on the github repo was not matching standards.

## Types of changes

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,10 +4,8 @@
   "description": "Client-Side SDK for Unified Data Collection",
   "main": "src/core/main.js",
   "scripts": {
-    "prepush-msg": "echo 'Running pre-push scripts...' && exit 0",
-    "precommit-msg": "echo 'Running pre-commit scripts...' && exit 0",
     "clean": "rimraf dist",
-    "lint": "eslint \"*.js\" \"src/**/*.js\" \"test/**/*.js\"",
+    "lint": "eslint --fix \"*.js\" \"{src,test}/**/*.js\"",
     "format": "prettier --write \"*.{html,js}\" \"{sandbox,src,test}/**/*.{html,js}\"",
     "test": "karma start --single-run",
     "test:report": "karma start --single-run --reporters allure",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -37,21 +35,36 @@
...
+    ],
+    "./*.{html,js}": [
+      "prettier --write",
+      "git add"
+    ],
+    "./{sandbox,src,test}/**/*.{html,js}": [
+      "prettier --write",
+      "git add"
+    ]
+  },
```
Aaronius: Yeah, it kind of sucks that these blob matching patterns are duplicated in the `scripts` block and here, but thems the breaks. It would also be nice to be able to combine multiple glob patterns for a particular set of tasks, but it [hasn't been implemented in lint-staged yet](https://github.com/okonet/lint-staged/issues/201).

Aaronius: Apparently `./` is important in these glob patterns here because, without it, the file `/Users/aahardy/dev/alloy/test/functional/sandbox/html/linkPageWithoutClickHandler.html` matches both `*.{html,js}` and `{sandbox,src,test}/**/*.{html,js}` and prettier would run twice on the (I found this out by enabling debugging on lint-staged).

However....for the `format` npm script (line 9), we _shouldn't_ use `./` because for whatever reason that forces prettier to stop ignoring `node_modules` so it tries to format `node_modules` files as well (I also found this out by enabling debugging because the process was taking a long time).

I admit I don't really understand the details of why these things are the way they are or if I should be doing something different. It may be intentional by the libraries or simply bugs.

package.json
```diff
@@ -37,21 +35,36 @@
...
+  "husky": {
+    "hooks": {
+      "pre-commit": "echo 'Running pre-commit scripts...' && lint-staged",
+      "pre-push": "echo 'Running pre-push scripts...' && npm run test"
+    }
+  },
+  "lint-staged": {
+    "./*.js": [
+      "eslint --fix",
+      "git add"
```
jonsnyder: Is this going to add files that aren't in the original commit?

Aaronius: If I understand what you mean by "original commit", then no. With lint-staged, what we're basically saying is: of the staged files that are attempting to be committed, find the files that match `./*.js`, then run that subset of files through `eslint --fix`, and then run that same subset through `git add`. In other words, unstaged files won't be affected.

PR #244, opened 2019-10-21
CORE-36609 Upgraded dependencies to use latest node to fix coverage report.

Aaronius:
## Description

When running a coverage report (`npm run coverage`), errors would be thrown due to out-of-memory errors. Using a recent version of node (I'm now running 12.12.0) fixes the issue. In order to use the latest version of node, we needed to upgrade several dependencies, which is what this PR does.

I'll add some comments to this PR to provide some detail.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-36609

## Motivation and Context
Be able to generate a coverage report.

## Types of changes

## Checklist:


package.json
```diff
@@ -5,7 +5,7 @@
   "main": "src/core/main.js",
   "scripts": {
     "clean": "rimraf dist",
-    "lint": "eslint --fix \"*.js\" \"{src,test}/**/*.js\"",
+    "lint": "eslint --fix \"*.js\" \"src/**/*.js\" \"test/**/*.js\"",
```
Aaronius: This one is fairly interesting. We were leveraging [brace expansion](https://www.linuxjournal.com/content/bash-brace-expansion) (`{src,test}/**/*.js`) in order to get both `src/**/*.js` and `test/**/*.js`. Brace expansion, when performed by the `sh` shell at least, works because the shell performs the expansion. However, we've added quotes around the pattern here, which means it's no longer the shell performing the expansion, but instead eslint. It seems that eslint doesn't actually support brace expansion. I can't find evidence in the docs or otherwise that brace expansion is supported and it's not behaving as I would expect if it were supported. What happens is that estlint starts digging through the `node_modules` directories and chokes as follows:

```
aarons-mbp-2:alloy aahardy$ npm run lint

> @adobe/alloy@0.0.1-alpha.8 lint /Users/aahardy/dev/alloy
> eslint --fix "*.js" "{src,test}/**/*.js"


Oops! Something went wrong! :(

ESLint: 6.5.1.

ESLint couldn't find the config "dev" to extend from. Please check that the name of the config is correct.

The config "dev" was referenced from the config file in "/Users/aahardy/dev/alloy/sandbox/node_modules/@csstools/convert-colors/package.json".

If you still have problems, please stop by https://gitter.im/eslint/eslint to chat with the team.
```

It shouldn't even be looking in `node_modules`. Instead, we just need to not use brace expansion, as I've done here, which fixes the issue. While we could leverage brace expansion by removing the quotes around the pattern and letting the shell perform the expansion, that's typically bad practice because it assumes the default shell supports brace expansion (which it very well may not).

package.json
```diff
@@ -85,26 +85,22 @@
...
     "@babel/preset-env": "^7.4.5",
-    "babel-loader": "^8.0.5",
-    "bundlesize": "^0.17.2",
+    "bundlesize": "^0.18.0",
     "dotenv": "^8.1.0",
-    "eslint": "^5.16.0",
-    "eslint-config-airbnb": "^17.1.0",
-    "eslint-config-prettier": "^4.0.0",
+    "eslint": "^6.5.1",
+    "eslint-config-airbnb-base": "^14.0.0",
```
Aaronius: You'll see here we were using `eslint-config-airbnb` which includes a bunch of React stuff and requires us to also install `eslint-plugin-react`, `eslint-plugin-jsx-a11y`, and, most recently, `eslint-plugin-react-hooks`. We're not actually linting react files (even the ones in sandbox), so we can use `eslint-config-airbnb-base` which doesn't require any of the react linting stuff.

PR #246, opened 2019-10-23
CORE-34953 Created a spec for every source file. Added pre-push check.

Aaronius:
## Description
I created a spec for every source file. I also added a pre-push check that will validate that a spec file exists for every src file (with exceptions for `constants` directories, `index.js` files, the base code, and files beginning with `.`). It does this by matching up file names. For example, `thing.js` needs a `thing.spec.js` within the associated test directory.

We had some constants lumped together in a constants.js file. I've broken them out into separate, more specifically named files and put them inside `constants` directories like we were doing for some of our other constants. Aside from maybe being better organizationally, this also allows us to have a single exception pattern for constants where we can ignore any directory name `constants`.

After talking to @artur-ciocanu, I moved  some of the Personalization specs to a more accurate location (the location they were in was due to old personalization rule structure). I noticed the Personalization "helper" specs are a bit unique in that they aren't so much unit tests as integration tests. For example, `click.spec.js` doesn't import `click.js`, but instead imports `initRuleComponentModules` and tests the click module indirectly through `initRuleComponentModules`. Because of this, some of the specs were previously named after the property name that `initRuleComponentModules` exposes rather than the underlying helper file being tested. For example, `initRuleComponentModules` contains a property called `insertAfter` which uses the `insertHtmlAfter` module. Therefore, there was a spec named `insertAfter.spec.js`. To comply with the new rule of needing a spec that matches the name of every source file, I've renamed some of the spec files. For example, `insertAfter.spec.js` has been renamed to `insertHtmlAfter.spec.js`, since it's primarily testing the `insertHtmlAfter.js` helper file. Anyway, @artur-ciocanu, you'll probably want to take a look and chime in on how you'd like to organize this. We could make the helper specs actually unit tests, in which case they would import the helper module directly and only test its unit. On the other hand, I see how using `initRuleComponentModules` more accurately tests whether the helpers are properly using native APIs.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-34953

## Types of changes

## Checklist:


Aaronius:
I added a commit. I realized the `baseCode/**` entry was technically unnecessary because it's covered by `**/index.js`. If we put other files in `baseCode`, maybe we'll need to add back the `baseCode/**`, but I find that doubtful.

Aaronius:
I sneaked in one more commit. Some helpers within `test/unit` were showing up in the coverage report, so I expanded the exclusion pattern so they would be excluded from the coverage report.

PR #249, opened 2019-10-23
CORE-36261 - ensure JS code that is part of an offer is properly executed

artur-ciocanu:
Fixing JS code execution in the personalization component rendering

## Description

Fixing JS code execution in the personalization component rendering

## Motivation and Context

Fixing JS code execution in the personalization component rendering

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Personalization/helper/scripts.js
+++ b/src/components/Personalization/helper/scripts.js
@@ -21,7 +21,7 @@ const isInlineScript = element =>
 const isRemoteScript = element =>
   is(element, SCRIPT) && getAttribute(element, SRC);
 
-const getInlineScripts = fragment => {
+export const getInlineScripts = fragment => {
   const scripts = selectNodes(SCRIPT, fragment);
   const result = [];
   const { length } = scripts;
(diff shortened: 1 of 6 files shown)


src/components/Personalization/helper/scripts.js
```diff
@@ -21,7 +21,7 @@ const isInlineScript = element =>
 const isRemoteScript = element =>
   is(element, SCRIPT) && getAttribute(element, SRC);
 
-const getInlineScripts = fragment => {
+export const getInlineScripts = fragment => {
   const scripts = selectNodes(SCRIPT, fragment);
```
jonsnyder: Will this execute scripts at the top level twice? Should the selector be "* script"?

artur-ciocanu: @jonsnyder I know this is counter intuitive, but SCRIPT tags are very weird. I'll try to explain what happens and why scripts tags are not executed twice.

For the sake of the argument lets say that we have an HTML string like this:
```
<script>console.log('Top level')</script>
<div id=\"foo\"><h1>Awesomeness</h1>
  <script>console.log('Inner')</script>
</div>
```

This is what happens during HTML rendering:
1. We create an in-memory DOM node that will be our fragment
2. We user `innerHTML` to add HTML string to the fragment
3. We get a list of nodes so we could append them to real DOM
4. We append the nodes

The tricky part with SCRIPT nodes is that these are not executed, even if we can list and append them to real DOM. We have to explicitly create SCRIPT nodes and append them to real DOM. We can't execute SCRIPTs that have transitively created via `innerHTML`.

PR #251, opened 2019-10-24
CORE-35805 - adding personalization click tracking support

artur-ciocanu:
Adding click tracking capabilities to personalization component.

## Description

Adding click tracking capabilities to personalization component. The personalization component uses the `onClick` life cycle method to make sure it is notified and it can inspect it's local storage to make sure that the cached click related metadata is sent to the backend.

## Related Issue

NONE

## Motivation and Context

Adding click tracking capabilities to personalization component.

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Personalization/helper/dom/matchesSelector.js
+++ b/src/components/Personalization/helper/dom/matchesSelector.js
@@ -10,8 +10,22 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-export default (settings, store) => {
-  const { selector, meta } = settings;
+import { selectNodesWithEq } from "./selectNodesWithEq";
 
-  store({ selector, meta });
+// Using node selection vs matches selector, because of :eq()
+export default (selector, element) => {
+  // Find all nodes using document as context
+  const nodes = selectNodesWithEq(selector, document);
+  let result = false;
+
+  // Iterate through all the identified elements
+  // and reference compare with element
+  for (let i = 0; i < nodes.length; i += 1) {
+    if (nodes[i] === element) {
+      result = true;
+      break;
+    }
+  }
+
+  return result;
 };
--- a/src/components/Personalization/index.js
+++ b/src/components/Personalization/index.js
@@ -11,14 +11,18 @@ governing permissions and limitations under the License.
 */
 
 import { isNonEmptyArray } from "../../utils";
+import { string } from "../../utils/configValidators";
 import { initRuleComponentModules, executeRules } from "./turbine";
 import { hideContainers, showContainers } from "./flicker";
-import { string } from "../../utils/configValidators";
+import collectClicks from "./helper/clicks/collectClicks";
 
 const DECISIONS_HANDLE = "personalization:decisions";
 
 // This is used for Target VEC integration
 const isAuthoringMode = () => document.location.href.indexOf("mboxEdit") !== -1;
+const mergeMeta = (event, meta) => {
+  event.mergeMeta({ personalization: { ...meta } });
+};
 
 const executeFragments = (fragments, modules, logger) => {
   fragments.forEach(fragment => {
@@ -31,20 +35,12 @@ const executeFragments = (fragments, modules, logger) => {
 };
 
 const createCollect = eventManager => {
-  return payload => {
+  return meta => {
     const event = eventManager.createEvent();
-    event.mergeMeta({
-      personalization: { ...payload }
-    });
+
+    mergeMeta(event, meta);
+
     eventManager.sendEvent(event);
-  };
-};
-
-const createStore = () => {
-  const storage = [];
-
-  return (selector, meta) => {
-    storage.push({ selector, meta });
   };
 };
 
@@ -52,7 +48,8 @@ const createPersonalization = ({ config, logger, eventManager }) => {
   const { prehidingStyle } = config;
   const authoringModeEnabled = isAuthoringMode();
   const collect = createCollect(eventManager);
-  const store = createStore();
+  const storage = [];
+  const store = value => storage.push(value);
   const ruleComponentModules = initRuleComponentModules(collect, store);
 
   return {
@@ -87,6 +84,11 @@ const createPersonalization = ({ config, logger, eventManager }) => {
       },
       onResponseError() {
         showContainers();
+      },
+      onClick({ event, clickedElement }) {
+        const merger = meta => mergeMeta(event, meta);
+
+        collectClicks(merger, clickedElement, storage);
       }
     }
   };
--- a/src/components/ActivityCollector/attachClickActivityCollector.js
+++ b/src/components/ActivityCollector/attachClickActivityCollector.js
@@ -13,21 +13,26 @@ governing permissions and limitations under the License.
 const createClickHandler = (eventManager, lifecycle) => {
   return clickEvent => {
     // TODO: Consider safeguarding from the same object being clicked multiple times in rapid succession?
-    const clickedObject = clickEvent.target;
+    const clickedElement = clickEvent.target;
     const event = eventManager.createEvent();
-    lifecycle.onClick({ event, clickedObject }).then(() => {
-      if (!event.isEmpty()) {
-        eventManager.sendEvent(event);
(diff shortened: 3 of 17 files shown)


src/components/Personalization/index.js
```diff
@@ -40,26 +41,66 @@ const createCollect = collect => {
...
+  const nodes = selectNodesWithEq(selector, node);
+
+  return nodes.length > 0;
+};
+
+const collectIfMatches = (collect, element, value) => {
+  const { selector, meta } = value;
+
+  if (matches(selector, element)) {
+    collect(meta);
```
liljenback: What is meta in this case?

liljenback: The reason I ask is that I don't think meta will be captured in the event command options.

src/components/Personalization/index.js
```diff
@@ -40,26 +41,66 @@ const createCollect = collect => {
...
+    },
+    getAll: () => {
+      return storage;
+    }
   };
 };
 
+// Using node selection vs matches selector, because of :eq()
+const matches = (selector, node) => {
+  const nodes = selectNodesWithEq(selector, node);
```
Aaronius: Doesn't `selectNodesWithEq` search `node` and all its children for an element matching the selector? If so, and if I understand correctly, it seems like that might not be what you want to use here since you're just trying to see if the current node (not its children) matches the selector. Right?

src/components/Personalization/index.js
artur-ciocanu: @liljenback true, it won't work today, but once @Aaronius adds his patch related to `sendEvent` I'll replace this with appropriate invocation.

Today even without click tracking, personalization functionality is half-broken because I can't properly send a notification back to the server since the `event` command doesn't allow `meta` manipulation.

artur-ciocanu: @liljenback one more thing, by design Alloy shouldn't really care or know what is in `meta`. The assumption is that `meta` is echoed back to the server without any client side understanding or manipulation.

src/components/Personalization/index.js
artur-ciocanu: @Aaronius `selectNodesWitEq` is basically similar to `node.querySelectorAll(selector)`. I was going to create a more optimized version something like `matchesSelectorWithEq`, that would use `node.matches(...)`, but the code was way more complex.

Aaronius: It seems like using something like `node.querySelectorAll` would result in incorrect behavior here though, unless I'm just misunderstanding something. For example, let's say you have `A` contains `B` contains `C` contains `D`. Let's say `D` is the element that has a class of `foo` and your personalization rule is targeting the selector `.foo`.

The user clicks on `D`. At this point, the code runs `selectNodesWithEq(".foo", D)`, which would return no results, because `D` doesn't contain an element with `.foo` (`D` _is_ the element matching `.foo`). I think you would want to collect at this point and not move on to inspecting `C`.

Anyway, the code continues and runs `selectNodesWithEq(".foo", C)`. It finds that D is a descendant of `C` and matches `.foo`, so data is collected.

The code continues and runs `selectNodesWithEq(".foo", B)`. It finds that D is a descendant of `B` and matches `.foo`, so data is collected.

The code continues and runs `selectNodesWithEq(".foo", A)`. It finds that D is a descendant of `A` and matches `.foo`, so data is collected.

This seems wrong. It seems like you would want to collect on the first iteration (when inspecting `D`) and not on the subsequent iterations (when inspecting `C`, `B`, and `A`).

There's another problem I believe exists. Let's say you have `X` which contains both `Y` and `Z`. Let's say `Y` has a class of `foo` and `Z` does not. The user clicks element `Z`. In this scenario, data would erroneously be collected, because `selectNodesWithEq(".foo", X)` would return `Y` as a result. But the user never clicked Y so nothing should be collected.

artur-ciocanu: @Aaronius I have adjusted the matching logic. The way I have implemented is the following:
1. Using the selector find all the elements
2. Iterate through elements and do a reference check against the clicked element
3. If the reference checks passes collect the metadata

I wish it would have been simpler, but because of `eq()` it is really, really non-trivial.

src/components/Personalization/helper/dom/matchesSelector.js
```diff
@@ -0,0 +1,31 @@
...
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+import { selectNodesWithEq } from "./selectNodesWithEq";
+
+// Using node selection vs matches selector, because of :eq()
+export default (selector, element) => {
+  // Find all nodes using document as context
+  const nodes = selectNodesWithEq(selector, document);
```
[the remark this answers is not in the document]
artur-ciocanu: @jonsnyder yeah, I saw that this might not be optimal, but I wanted to make sure that it is OK functional wise.

If this will become a HUGE issue we can review and come up with a better algo. The first that comes to mind is that instead of invoking all that selector parsing we could memoize or do something smarter.

Thanks for you suggestion, I'll add it to my next commit.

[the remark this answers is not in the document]
artur-ciocanu: @Aaronius I have a check similar to `selectNodesWithEq`, so if the selector has `eq()` we will our own parsing etc, if it's an ordinary selector we would use `element.matches()`.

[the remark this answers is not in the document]
Aaronius: Hey @artur-ciocanu, I don't think you've really resolved what @jonsnyder and I were pointing out. You may be intentionally putting it off until later, but I just want to point in out in case it was an oversight.

Considering 20 elements in the hierarchy of the clicked element and 10 items in storage (all using `:eq`), your later commit changed the logic from:

```
for each element in clicked element hierarchy: 20 elements
  for each item in storage: 10 items
    for each found element that matches selectNodesWithEq: 200 selectNodesWithEq queries
      check to to see if found element == current element in hierarchy
```

to

```
for each item in storage: 10 items
  for each element in clicked element hierarchy: 20 elements
    for each found element that matches selectNodesWithEq: 200 selectNodesWithEq queries
      check to to see if found element == current element in hierarchy
```

In both cases, there are 200 calls to `selectNodesWithEq`, which I think is the concern because it's likely to be the bottleneck.

What was being suggested, I think, was that you re-work the logic to only call `selectNodesWithEq` 10 times (one for each item in storage), like this:

```
for each item in storage, call selectNodesWithEq: 10 selectNodesWithEq queries
  for each element in clicked element hierarchy: 20 elements
    check to see if element in hierarchy is in selectNodesWithEq result: this occurs 200 times total
```

Note that your current approach will be fastest when `Element.matches` can be used (the selector doesn't have `:eq`), whereas the proposed approach will be fastest otherwise. You might have to have a split code path here.

If you end up not implementing the proposed approach, I think you can optimize your current approach by changing this:

```
const nodes = selectNodesWithEq(selector, document);
```

to this:

```
const nodes = selectNodesWithEq(selector, element.parentNode);
```

So that you're querying within a smaller scope than `document`. However, that requires `selectNodesWithEq()` to work properly when you pass in a scope other than `document`, and I don't think it is right now. You can see this if you run this test inside `selectNodesWithEq.spec.js`:

```
it("should select when eq and just one element", () => {
    const content = `
      <div class="b">
        <div class="c">first</div>

        <div class="c">second</div>
        
        <div class="c">third</div>
      </div>
    `;

    appendNode(
      document.body,
      createNode("DIV", { id: "abc", class: "eq" }, { innerHTML: content })
    );

    const result = selectNodesWithEq(
      "#abc:eq(0) > div.b:eq(0) > div.c:eq(0)",
      document.getElementById("abc")
    );

    expect(result[0].tagName).toEqual("DIV");
    expect(result[0].textContent).toEqual("first");
  });
```

This fails to find an element. However, if you switch out:

```
const result = selectNodesWithEq(
      "#abc:eq(0) > div.b:eq(0) > div.c:eq(0)",
      document.getElementById("abc")
    );
```

with

```
const result = document
      .getElementById("abc")
      .querySelectorAll("#abc > div.b > div.c");
```

you'll see that it does find a result. This should either get fixed (might be super tricky) or remove the context parameter in `selectNodesWithEq`.

PR #253, opened 2019-10-29
Moved event sending logic up to core

Aaronius:
## Description
This moves the logic for sending an event up to core into an object called `eventManager`. Along with it, several other things changed as well:

### Opt-in object is now configured in core
Previously, the privacy component was in charge of configuring whether the opt-in object was enabled based off the configuration the user provided. Because the privacy component needed to configure whether opt-in was enabled before any other component could us it, we were passing in an `enableOptIn` function into the privacy component during construction and then we would provide the fully configured opt-in object to all the components on `onComponentsRegistered`.

The intention of this was to keep as much privacy-related logic inside the privacy extension, but it turned out to be more cumbersome that it was worth and made for a strange implicit relationship/timing between the `optIn` and the new `eventManager`, since `eventManager` depends on `optIn` and they're both created in core. I found it would be simpler and easier to understand if we just configured `optIn` in core when the object gets constructed and handed the configured opt-in into `eventManager` and all the components. Similarly, this allows us to pass `eventManager` to components through their constructor functions instead of through `onComponentsRegistered`.

Because I moved opt-in configuration into core, the users opt-in preferences could no longer be stored within the Alloy cookie under the privacy component's abbreviation namespace. I talked this over with @jfkhoury and we felt like it would be preferable to store opt-out information in a separate cookie anyway (I don't think we ever got our question about this officially answered by the privacy team), in case any customers were concerned about us reading the full Alloy cookie to check opt-in preferences if the user had previously opted into no purposes. For this reason, opt-in preferences are now stored in a separate cookie. Most of this logic should move to the server soon anyway as we migrate to Konductor writing cookies instead of Alloy.

### Tool factories are largely gone

We had this concept of [tool factories up in core](https://github.com/adobe/alloy/blob/master/src/core/index.js#L104-L117) that were used for instantiating/configuring the tools that could be immediately provided to components through their constructor functions as the components were constructed. As I was making yet another one for `eventManager` I felt it would be simpler to just have `core/index.js` pass a `getImmediatelyAvailableTools` function into `initializeComponents` and `getImmediatelyAvailableTools` would instantiate and configure the tools. I found this to be simpler and easier to understand than multi-level curry functions.

One of the factories, `cookieJarToolFactory`, still exists, but it's a bit simpler now and will possibly go away when cookie writing moves to the server.

### `network.sendRequest` API changed

I felt like I should be as consistent as possible between `eventManager` and `network`. `eventManager` provides a method to create an event and another to send the event, whereas `network` provides a method to create a payload and another to send a payload. `network`'s `sendRequest` signature was `sendRequest(payload, expectsResponse = true, documentUnloading = false)`. As I read through usage and tests it was difficult to remember what the booleans were for when seeing something like this: `network.sendRequest(payload, false, true)`. We've tried to resolve these situations previously by using an object for such options. I opted to do this with `network` such that the signature is now `.sendRequest(payload, { expectsResponse = true, documentUnloading = false })`. I then took a similar approach with `eventManager`'s sendEvent: `sendEvent(event, { isViewStart = false, applyUserProvidedData = noop })`.

### `componentRegistry` is no longer provided to components

I think it's a Bad Idea™ to have components accessing other components' commands (and potentially lifecycle methods). Now that Personalization is using `eventManager` from core, it no longer needs access to DataCollector's event command. I removed `componentRegistry` from being passed into components' `onComponentsRegistered` hook. If a component needs some functionality from another component, we should assess whether it's something that should be moved up to core. If that's not appropriate for some reason, then we can look into some sort of shared module pattern like in Launch extensions, but I'd rather not if we can help it.

### `setCustomerIds` still needs to be converted to use `eventManager`

`setCustomerIds` is currently basically duplicating the logic in `eventManager`. I kept that refactor out of the scope of this particular PR.

I think that's it.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-36617

## Motivation and Context
Allow extensions to send events without having to use DataCollector's public `event` command.

## Types of changes

## Checklist:

- [x] If someone wants to help double-check, that would be excellent. It seems to be functioning though. -> I have run the Sandbox successfully.


jfkhoury:
Thank you so much for the awesome description of this PR @Aaronius ! 🙇

src/components/ActivityCollector/index.js
```diff
@@ -14,19 +14,17 @@ import attachClickActivityCollector from "./attachClickActivityCollector";
...
 
-const EVENT_COMMAND = "event";
+const createActivityCollector = ({ config, eventManager }) => {
+  const linkClick = createLinkClick(window, config, eventManager);
 
-const createActivityCollector = ({ config }) => {
   return {
     lifecycle: {
       onComponentsRegistered(tools) {
-        const { lifecycle, componentRegistry } = tools;
```
jfkhoury: Love this change!

src/components/Identity/customerIds/createCustomerIds.js
```diff
@@ -6,7 +6,7 @@ import {
   clone
 } from "../../../utils";
 import { CUSTOMER_ID_HASH } from "../constants/cookieNames";
-import createEvent from "../../DataCollector/createEvent";
+import createEvent from "../../../core/createEvent";
```
jfkhoury: Let's please add a TODO to fix this.

src/core/createOptIn.js
```diff
@@ -12,7 +12,8 @@ governing permissions and limitations under the License.
 
 import { defer } from "../utils";
 
-const COOKIE_NAMESPACE = "optIn";
+// OptIn uses a different cookie than the rest of Alloy.
+const COOKIE_NAMESPACE = "alloy_optIn";
```
jfkhoury: I think we need to add the ORG ID to this cookie name right?

test/unit/specs/core/createEventManager.spec.js
```diff
@@ -0,0 +1,223 @@
...
+      });
+    });
+
+    it("calls sendRequest", () => {
+      return eventManager.sendEvent(event).then(() => {
+        expect(network.sendRequest).toHaveBeenCalledWith(payload, true, true);
+      });
+    });
+
+    // it("resolves with response information", () => {
```
jfkhoury: Should we uncomment those lines? or remove them?

Aaronius: I think you were commenting on the https://github.com/adobe/alloy/pull/253/files/61f641f7fec32013eca3253081596f7012cee294#diff-34fae68cfdead6324f76042dd9610713 commit directly instead of the latest. This was updated in a later commit.

src/components/Identity/customerIds/createCustomerIds.js
Aaronius: I'm working on an additional PR to fix this right now.

src/components/DataCollector/index.js
```diff
@@ -10,78 +10,36 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
-      .then(() => makeServerCall(event));
-  };
-
-  return {
-    lifecycle: {
-      onComponentsRegistered(tools) {
-        ({ lifecycle, optIn } = tools);
+        return eventManager.sendEvent(event, {
+          isViewStart: viewStart,
+          applyUserProvidedData() {
```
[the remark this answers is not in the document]
jonsnyder: Yes, I think a `event.setUserXdm`, and `event.setUserData` methods may be easier to follow.

src/core/index.js
```diff
@@ -95,28 +93,47 @@ if (instanceNamespaces) {
...
-          logger: loggerToolFactory(logController.createComponentLogger),
-          network: networkToolFactory(
-            createNetwork,
-            lifecycle,
-            logger,
-            networkStrategy
-          )
-        },
-        optIn
+        getImmediatelyAvailableTools(componentAbbreviation) {
```
jonsnyder: 👍

src/components/DataCollector/index.js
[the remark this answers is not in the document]
Aaronius: Yeah, I think that's a good idea. I'm going to log this as a separate issue and address it in a new PR.

src/components/ActivityCollector/index.js
```diff
@@ -14,19 +14,17 @@ import attachClickActivityCollector from "./attachClickActivityCollector";
...
-        const collect = componentRegistry.getCommand(EVENT_COMMAND);
-        attachClickActivityCollector(config, collect, lifecycle);
+        const { lifecycle } = tools;
+        attachClickActivityCollector(config, eventManager, lifecycle);
         // TODO: createScrollActivityCollector ...
       },
       onClick({ event, clickedObject }) {
```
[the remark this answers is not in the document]
Aaronius: I'll create a separate issue for this.

src/components/Identity/customerIds/createCustomerIds.js
Aaronius: For reference: https://jira.corp.adobe.com/browse/CORE-37059

src/components/DataCollector/index.js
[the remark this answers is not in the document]
Aaronius: For reference: https://jira.corp.adobe.com/browse/CORE-37060

src/components/ActivityCollector/index.js
```diff
@@ -14,19 +14,17 @@ import attachClickActivityCollector from "./attachClickActivityCollector";
 import createConfigValidators from "./createConfigValidators";
 import createLinkClick from "./createLinkClick";
```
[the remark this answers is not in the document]
Aaronius: For reference: https://jira.corp.adobe.com/browse/CORE-37062

src/components/ActivityCollector/index.js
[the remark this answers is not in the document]
Aaronius: For reference: https://jira.corp.adobe.com/browse/CORE-37062

PR #254, opened 2019-10-29
Feature/core 35990 Migration of ECID from legacy cookie

jileeshs:
## Description

This change reads ECID from existing AMCV cookie and set it in AlloyCookie if a `migrateIds` config is set to `true`

## Related Issue
https://jira.corp.adobe.com/browse/CORE-35990

## Motivation and Context

Customers might have pages with Alloy and some pages with legacy visitor changes. This change is an effort to copy existing IDs.

## Types of changes

## Checklist:


src/components/Identity/migration.js
```diff
@@ -0,0 +1,23 @@
+import cookie from "@adobe/reactor-cookie";
+
+export default imsOrgId => {
+  return {
+    readEcidFromAmcvCookie() {
+      const amcvCookieValue = cookie.get(`AMCV_${imsOrgId}`);
+      const reg = /(MCMID)\|(\d+)\|/;
```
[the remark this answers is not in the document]
Aaronius: I think it would be good to move this regex to inside the `if (amcvCookieValue) {` statement so we're only creating it if we're going to use it.

src/components/Identity/migration.js
```diff
@@ -0,0 +1,23 @@
...
+      let ecid = "";
+      if (amcvCookieValue) {
+        amcvCookieValue.replace(reg, (match, mid, value) => {
```
Aaronius: Why are we using `replace`? Should we be using `match` instead?

src/components/Identity/migration.js
```diff
@@ -0,0 +1,23 @@
+import cookie from "@adobe/reactor-cookie";
```
Aaronius: Change this to `import { cookieJar } from "../../utils"` please.

src/components/Identity/createComponent.js
```diff
@@ -0,0 +1,179 @@
...
+    }
+    return createEcidEvent();
+  };
+
+  return {
+    lifecycle: {
+      onComponentsRegistered(tools) {
+        ({ lifecycle, optIn } = tools);
+        if (migrateIds) {
+          getEcid(migrateIds);
```
Aaronius: I'm a bit confused here. `getEcid` doesn't take any arguments.

src/components/Identity/createComponent.js
```diff
@@ -0,0 +1,179 @@
...
+      .then(() => makeServerCall(payload));
+  };
+
+  const getEcid = () => {
+    const ecid = readEcid();
+    if (ecid) {
+      return ecid;
+    }
+    return createEcidEvent();
+  };
```
Aaronius: It looks like this function returns an ECID value sometimes and at other times returns a promise. This code: 

```
config.reactorRegisterGetEcid(() => {
  return optIn.isOptedIn() ? getEcid() : undefined;
});
```

provides a function to the Launch extension that allows the extension to access the ECID immediately. It isn't set up to handle the case where a promise is returned. I think in this case we need to change the code to use `readEcid` instead:

```
config.reactorRegisterGetEcid(() => {
  return optIn.isOptedIn() ? readEcid() : undefined;
});
```

Aaronius: I see this function is what's being exposed as the `getEcid` command. What if a user does not yet have an ECID and the customer does this:

```
alloy("getEcid");
alloy("getEcid");
```

Could this end up generating two ECIDs for the user? What about if a user did this:

```
alloy("event");
alloy("getEcid");
```

src/components/Identity/createComponent.js
```diff
@@ -0,0 +1,179 @@
...
+    return ecid;
+  };
+
+  const makeServerCall = payload => {
+    return lifecycle.onBeforeDataCollection({ payload }).then(() => {
+      return network.sendRequest(payload, payload.expectsResponse);
+    });
+  };
+
+  const createEcidEvent = () => {
```
jfkhoury: I'm a little worried about this and I'm not sure we thought the whole thing through. Creating those server calls to fetch one value are weird, because the response contains much more than the ECID. Let's talk this over with Rares.

src/components/Identity/createComponent.js
jfkhoury: I agree with @Aaronius. I think having a function that returns 2 different types is weird in general, but we might need to add guards against multiple calls in the same session.

sandbox/public/index.html
```diff
@@ -45,12 +45,13 @@
...
         // optInEnabled: true,
-        configId: "9999999",
+        propertyId: "9999999",
         imsOrgId: "53A16ACB5CC1D3760A495C99@AdobeOrg",
         // errorsEnabled: false,
         logEnabled: true,
         prehidingStyle: ".personalization-container { opacity: 0 !important }",
-        idSyncContainerId: 81
+        idSyncContainerId: 81,
+        migrateIds: true
```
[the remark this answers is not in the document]
jileeshs: If the customer is new (who doesn't have previous ids) and if they do-not want cross-site tracking.

src/components/Identity/createComponent.js
```diff
@@ -0,0 +1,179 @@
...
+  const createPayload = event => {
+    const payload = network.createPayload();
+    payload.addEvent(event);
+    return payload;
+  };
+
+  const readEcid = () => {
+    let ecid = cookieJar.get(EXPERIENCE_CLOUD_ID);
+    if (!ecid && migrateIds) {
+      ecid = migration(imsOrgId).readEcidFromAmcvCookie();
```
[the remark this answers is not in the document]
jileeshs: +:1

src/components/Identity/createComponent.js
jileeshs: It is for the `getEcid` command.
What is the better way to return an `Ecid` if there was no `viewstart`  event.

sandbox/public/legacy.html
```diff
@@ -0,0 +1,98 @@
...
+          <a href="/optIn">OptIn</a>
+        </li>
+        <li>
+          <a href="/largePayload">LargePayload</a>
+        </li>
+        <li>
+          <a href="/DualTag">DualTag</a>
+        </li>
+        <li>
+          <a href="http://localhost:3000/legacy.html">legacy</a>
```
liljenback: You don't need the full domain here right?

src/components/Identity/createComponent.js
liljenback: Should the legacy ID be removed at this point?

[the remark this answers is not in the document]
jileeshs: Done

jileeshs: I think we shouldn't. There could be pages from the same company using legacy libraries.

src/components/Identity/createComponent.js
jileeshs: Removed the network call.
We will add it if there is an endpoint just returning ID

sandbox/public/legacy.html
jileeshs: Your're right. Thanks!

src/components/Identity/createComponent.js
```diff
@@ -0,0 +1,142 @@
...
+      }
+    }
+    return ecid;
+  };
+
+  return {
+    lifecycle: {
+      onComponentsRegistered(tools) {
+        ({ lifecycle, optIn } = tools);
+        if (migrateIds) {
```
jfkhoury: No need for this `if` anymore ;)

src/components/Identity/createComponent.js
```diff
@@ -0,0 +1,142 @@
...
+  let optIn;
+  let deferredForEcid;
+  let lifecycle;
+  let customerIds;
+  let alreadyQueriedForIdSyncs = false;
+  const { migrateIds, imsOrgId } = config;
+  const migration = createMigration(imsOrgId, migrateIds);
+
+  const getEcid = () => {
+    let ecid = cookieJar.get(EXPERIENCE_CLOUD_ID);
```
jfkhoury: You can go: `let ecid = cokie... || migration.readF...`

src/components/Identity/createComponent.js
```diff
@@ -0,0 +1,142 @@
...
+export default (idSyncs, manualIdSyncs, cookieJar, config, logger, network) => {
+  let optIn;
+  let deferredForEcid;
+  let lifecycle;
+  let customerIds;
+  let alreadyQueriedForIdSyncs = false;
+  const { migrateIds, imsOrgId } = config;
+  const migration = createMigration(imsOrgId, migrateIds);
+
+  const getEcid = () => {
```
jfkhoury: Let's please add a TODO to remember to address the server side fetch of ECID.

PR #258, opened 2019-10-29
CORE-37060 Added setUserXdm and setUserData methods to event.

Aaronius:
## Description
Removed applyUserProvidedData and instead added `setUserXdm` and `setUserData` methods to the event. The event, upon serialization, will take care to override Alloy-produced data with user-provided data.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-37060
https://jira.corp.adobe.com/browse/CORE-36755

## Motivation and Context
Easier for engineers to understand.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have signed the [Adobe Open Source CLA](https://opensource.adobe.com/cla.html) or I'm an Adobe employee.
- [x] I have submitted a [documentation](https://github.com/AdobeDocs/alloy-docs) pull request or no changes are needed.
- [x] I have made any necessary test changes and all tests pass.
- [x] I have run the Sandbox successfully.

PR #259, opened 2019-10-30
onBeforeEventSend callback

jonsnyder:
This allows a user to modify the xdm just before the event is sent.  An old version of these changes are in #229 

## Description

This is the order of applied data to xdm:
1. The automatically collected data in context
2. The xdm option passed into the event command
3. The changes applied in the onBeforeEventSend callback

## Related Issue

https://jira.corp.adobe.com/browse/CORE-34882

## Motivation and Context

Allow customers to write code that gets run for every event sent to the server.  For link clicks and events sent through personalization, this is the only opportunity to change the data on the client.

## Types of changes

## Checklist:


Aaronius:
It looks like I just introduced an abundance of conflicts. 😢 Let me know if you want my help resolving them.

jonsnyder:
> It looks like I just introduced an abundance of conflicts. 😢 Let me know if you want my help resolving them.

No problem.  I just looked through your PR and figured it would.  The only thing I'm not sure about is how to make it so that the onBeforeEventSend callback is applied after the userXdm.

sandbox/public/index.html
```diff
@@ -50,7 +50,13 @@
         // errorsEnabled: false,
         logEnabled: true,
         prehidingStyle: ".personalization-container { opacity: 0 !important }",
-        idSyncContainerId: 81
+        idSyncContainerId: 81,
+        onBeforeEventSend(xdm) {
```
Aaronius: How do you see this changing when we start supporting `data` on the event command? Would there be any reason for a customer to access/modify that data here? If so, would `data` just be a second argument? Would it be better to make the signature `onBeforeEventSend({ xdm, data })`?

Would we need to support a use case where a user wants to always send their data layer on every event, so rather than passing it through each event command, they might just configure `onBeforeEventSend` to just send it every time? If so, how would they do that if their data layer were not XDM? For example:

```
onBeforeEventSend(xdm, data) {
  // I can't do "return dataLayer" because we don't support return values, therefore data would need to be an empty object to which I would need to copy properties from my data layer.
  Object.assign(data, dataLayer);
}
```

Anyway, I realize I was the one that said not to accept return values and still feel like that's probably the right way to go as long as this isn't a use case we need to consider. We can add support for return values later if we need to.

src/core/createEvent.js
```diff
@@ -24,6 +22,11 @@ export default () => {
       content.data = data;
     },
     mergeXdm: createMerger(content, "xdm"),
+    applyCallback(callback) {
+      const xdm = clone(Object(content.xdm));
+      callback(xdm);
+      content.xdm = xdm;
+    },
```
Aaronius: This will need some work now that user-provided XDM is merged on serialization: https://github.com/adobe/alloy/pull/258/files#diff-69cd2fefdc76adcc848d33b1263138dbR48-R57

Let me know if you want to talk through how to best approach this.

src/core/createEventManager.js
```diff
@@ -35,6 +35,11 @@ export default ({ createEvent, optIn, lifecycle, network, config }) => {
         })
         .then(() => {
           applyUserProvidedData(event);
+          try {
+            event.applyCallback(onBeforeEventSend);
+          } catch (e) {
+            logger.warn(e);
```
Aaronius: I think `logger.error` would be more appropriate since the callback failed and therefore may be resulting in erroneous data.

src/core/createConfigValidators.js
```diff
@@ -10,19 +10,29 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
-    },
-    edgeDomain: {
-      defaultValue: "beta.adobedc.net",
-      validate: string().domain()
-    },
-    imsOrgId: {
-      validate: string().unique()
-    }
-  };
-};
```
Aaronius: 👍 👍 👍

src/components/Context/index.js
```diff
@@ -28,25 +29,23 @@ const placeContext = placeContextFactory(() => new Date());
...
+  return createComponent(config, logger, optionalContexts, requiredContexts);
 };
 
 createContext.namespace = "Context";
 createContext.abbreviation = "CO";
 createContext.configValidators = {
   context: {
-    defaultValue: ["web", "device", "environment", "placeContext"]
+    defaultValue: Object.keys(optionalContexts),
+    validate: arrayOf(string())
```
Aaronius: 🎉

test/unit/specs/components/Context/deviceFactory.spec.js
```diff
@@ -11,13 +11,17 @@ describe("Context::deviceFactory", () => {
         height: 800
       }
     };
-    event = jasmine.createSpyObj("event", ["mergeXdm"]);
   });
 
+  const run = () => {
+    const xdm = {};
+    deepAssign(xdm, deviceFactory(window)(xdm));
```
Aaronius: I don't think the `deepAssign()` here is desired.

test/unit/specs/core/createConfigValidators.spec.js
```diff
@@ -3,16 +3,81 @@ Copyright 2019 Adobe. All rights reserved.
...
 governing permissions and limitations under the License.
 */
 
-import createConfigValidators from "../../../../../src/components/DataCollector/createConfigValidators";
-import createConfig from "../../../../../src/core/createConfig";
+import createConfigValidators from "../../../../src/core/createConfigValidators";
+import createConfig from "../../../../src/core/createConfig";
+
+describe("configValidators", () => {
+  const baseConfig = {configId: "1234", imsOrgId: "org1"}
```
Aaronius: Any idea how this line passed prettier formatting?

Aaronius:
@jonsnyder this is looking good. I'll review again after conflicts are resolved.

sandbox/public/index.html
jfkhoury: I agree that we should make the signature `{xdm, data}`.

Let's start without the return and see how it goes.

src/components/Context/implementationDetailsFactory.js
```diff
@@ -9,15 +9,15 @@ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTA
...
-    event.mergeXdm({
-      implementationDetails: {
-        name: "https://ns.adobe.com/experience/alloy",
-        version,
-        environment: "web"
-      }
-    });
+  return xdm => {
+    const implementationDetails = {
+      name: "https://ns.adobe.com/experience/alloy",
```
jfkhoury: Would sending `alloy` only as the name work?

src/core/createEvent.js
```diff
@@ -18,43 +18,76 @@ export default () => {
   let userData;
   let expectsResponse = false;
   let documentUnloading = false;
+  let frozen = false;
+
+  const guardFrozen = func => {
```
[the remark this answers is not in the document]
jonsnyder: You are right.  This isn't necessary as we don't expose the event externally, and once the event is in the payload object, it cannot be changed.  I'll remove the freezing.  When I was working with @Aaronius to come up with a solution for this, he seemed to think it would be necessary, but I agree it is over-kill.

src/components/DataCollector/index.js
```diff
@@ -41,6 +39,6 @@ const createDataCollector = ({ eventManager }) => {
 
 createDataCollector.namespace = "DataCollector";
 createDataCollector.abbreviation = "DC";
-createDataCollector.configValidators = createConfigValidators();
+createDataCollector.configValidators = {};
```
jileeshs: What is more readable?
This or
```
createDataCollector : {
  namespace: "DataCollector",
  abbreviation: "DC",
  configValidators: {}
}
}
```

jonsnyder: Doing it in an object like this is probably more readable, but I don't think there is a way to do it when you are adding properties to the function, and not just a generic object.  (createDataCollector is a function)

sandbox/public/index.html
```diff
@@ -51,7 +51,16 @@
         logEnabled: true,
         prehidingStyle: ".personalization-container { opacity: 0 !important }",
         idSyncContainerId: 81,
-        migrateIds: true
+        migrateIds: true,
+        onBeforeEventSend({ xdm, data }) {
```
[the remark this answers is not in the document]
jonsnyder: If it errors out in IE 11, then the  event will still send it just won't have these extra fields.  I think its okay to leave it in as an example.

PR #261, opened 2019-11-01
Added utility tests.

Aaronius:
## Description
Added several utility tests.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-36777
https://jira.corp.adobe.com/browse/CORE-36779
https://jira.corp.adobe.com/browse/CORE-36783
https://jira.corp.adobe.com/browse/CORE-36784
https://jira.corp.adobe.com/browse/CORE-36785
https://jira.corp.adobe.com/browse/CORE-36786

## Motivation and Context
Ensure code is working as expected.

## Types of changes

## Checklist:


src/utils/keys.js
```diff
@@ -1,21 +0,0 @@
...
-*/
-
-/**
- * Returns an array whose items are the provided object's own enumerable
- * property keys.
- * @param {Object} obj
- * @returns {Array}
- */
-export default obj => {
-  return Object.keys(obj);
```
Aaronius: Early in the Alloy project there were some conversations about what amount of complexity was required to be extracted into a utility. If I recall, the decision was that generally if it's a simple method call and no logic (conditions, etc) are involved, then it wouldn't qualify. If that's correct, I don't think this qualifies (although we do have a `noop` utility which probably would qualify when applying the same rules). Anyway, we were rarely using this utility and I've removed its usage.

test/unit/specs/utils/storageFactory.spec.js
```diff
@@ -10,5 +10,74 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+            }
+          };
+          const storage = storageFactory(window)("example.");
+          const result = storage[storageProperty].getItem("foo");
+          expect(result).toBeNull();
+        });
+      });
+    });
+  });
+});
```
jonsnyder: This test is great, but somewhere we should have a test that makes sure the storageFactory works with the real browser APIs.

jfkhoury:
I will merge this PR because one of the specs fixes the script that checks for missing tests.

PR #269, opened 2019-11-08
Removing element parameter from selectNodesWithEq() method, to avoid …confusion and incorrect usage

artur-ciocanu:
Select nodes with EQ bug fix.

## Description
Select nodes with EQ bug fix

## Related Issue

NONE

## Motivation and Context
Select nodes with EQ bug fix

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Personalization/helper/dom/matchesSelectorWithEq.js
+++ b/src/components/Personalization/helper/dom/matchesSelectorWithEq.js
@@ -21,7 +21,7 @@ export default (selector, element) => {
 
   // Using node selection vs matches selector, because of :eq()
   // Find all nodes using document as context
-  const nodes = selectNodesWithEq(selector, document);
+  const nodes = selectNodesWithEq(selector);
   let result = false;
 
   // Iterate through all the identified elements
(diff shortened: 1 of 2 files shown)


src/components/Personalization/helper/dom/matchesSelectorWithEq.js
```diff
@@ -21,7 +21,7 @@ export default (selector, element) => {
 
   // Using node selection vs matches selector, because of :eq()
   // Find all nodes using document as context
-  const nodes = selectNodesWithEq(selector, document);
+  const nodes = selectNodesWithEq(selector);
```
Aaronius: I sent this to @artur-ciocanu through Slack, but I'm going to paste it here to keep it in context:

My thought was that you would use the logic you have to find all elements matching the selector (using a document scope), and then you would filter down the result to only those elements that were descendants of the scope that was passed in. In other words, if I did `selectNodesWithEq("HTML > BODY > DIV.wrapper:eq(0) > HEADER.header:eq(0) > DIV.pagehead:eq(0) > P:nth-of-type(1)", scopeElement)`, then you would find all elements in the whole document that match that selector and then you would reduce that resultset using logic like this:

```
matchingElements.filter(matchingElement => scopeElement.contains(matchingElement))
```

It doesn’t make the query any more performant, but I think it would behave correctly.

artur-ciocanu: @Aaronius thansk for your suggestions. I'll leave as is for now, if we think we need this functionality I'll add it later.

PR #272, opened 2019-11-12
CORE-37168 Added support for identification via 3rd-party domain

Aaronius:
## Description
If the ECID isn't on the client, and it's the first request where we retrieve the ECID, and `thirdPartyCookiesEnabled` is `true`, and the browser actually supports third party cookies (more specifically, if we detect that the browser Chrome or IE), then we send the request to `adobedc.demdex.net` instead of the regular edge domain.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-37168

## Motivation and Context
Enhanced identification.

## Types of changes

## Checklist:

- [ ] I think this needs documentation? Anyway, I haven't done it yet. -> I have submitted a [documentation](https://github.com/AdobeDocs/alloy-docs) pull request or no changes are needed.

### DIFF
--- a/sandbox/public/index.html
+++ b/sandbox/public/index.html
@@ -15,7 +15,7 @@
             script-src 'self' 'unsafe-inline' assets.adobedtm.com cdn.tt.omtrdc.net;
             style-src 'self' 'unsafe-inline' cdn.tt.omtrdc.net;
             img-src * data:;
-            connect-src 'self' unifiedjslab.tt.omtrdc.net *.demdex.net http://localhost:8080 https://alpha.konductor.adobedc.net https://beta.adobedc.net https://edgegateway.azurewebsites.net konductor-qe.apps-exp-edge-npe-va6.experience-edge.adobeinternal.net"
+            connect-src 'self' ws://localhost:3000 http://localhost:8080 unifiedjslab.tt.omtrdc.net *.demdex.net https://alpha.konductor.adobedc.net https://beta.adobedc.net https://edgegateway.azurewebsites.net konductor-qe.apps-exp-edge-npe-va6.experience-edge.adobeinternal.net"
     />
 
     <title>Mock website hosting Alloy</title>
@@ -40,6 +40,15 @@
     <script src="alloy.js" async></script>
 
     <script>
+      function getUrlParameter(name) {
+        name = name.replace(/[\[]/, "\\[").replace(/[\]]/, "\\]");
+        var regex = new RegExp("[\\?&]" + name + "=([^&#]*)");
+        var results = regex.exec(location.search);
+        return results === null
+          ? ""
+          : decodeURIComponent(results[1].replace(/\+/g, " "));
+      }
+
       alloy("configure", {
         // edgeDomain: "ENTER YOUR COLLECTION CNAME IF AVAILABLE",
         // optInEnabled: true,
@@ -52,13 +61,16 @@
         prehidingStyle: ".personalization-container { opacity: 0 !important }",
         idSyncContainerId: 81,
         migrateIds: true,
-        onBeforeEventSend({ xdm, data }) {
-          const urlSearchParams = new URLSearchParams(window.location.search);
-          if (urlSearchParams.has("title")) {
-            xdm.web.webPageDetails.name = urlSearchParams.get("title");
+        onBeforeEventSend: function(options) {
+          const xdm = options.xdm;
+          const data = options.data;
+          const titleParam = getUrlParameter("title");
+          if (titleParam) {
+            xdm.web.webPageDetails.name = titleParam;
           }
-          if (urlSearchParams.has("campaign")) {
-            data.campaign = urlSearchParams.get("campaign");
+          const campaignParam = getUrlParameter("title");
+          if (campaignParam) {
+            data.campaign = campaignParam;
           }
         },
         debug: true,
--- a/src/components/Identity/createComponent.js
+++ b/src/components/Identity/createComponent.js
@@ -2,6 +2,7 @@ import createCustomerIds from "./customerIds/createCustomerIds";
 import { defer } from "../../utils";
 import { EXPERIENCE_CLOUD_ID } from "./constants/cookieNames";
 import createMigration from "./createMigration";
+import isThirdPartyCookieSupported from "../../utils/isThirdPartyCookieSupported";
 
 const addIdsContext = (payload, ecid) => {
   payload.addIdentity(EXPERIENCE_CLOUD_ID, {
@@ -101,6 +102,12 @@ export default (
             // promise so that future requests can know when the ECID has returned.
             deferredForEcid = defer();
             payload.expectResponse();
+            if (
+              config.thirdPartyCookiesEnabled &&
+              isThirdPartyCookieSupported(window)
+            ) {
+              payload.useIdThirdPartyDomain();
+            }
           }
           customerIds.addToPayload(payload);
           return promise;
--- a/src/components/Identity/createMigration.js
+++ b/src/components/Identity/createMigration.js
@@ -14,8 +14,11 @@ export default (imsOrgId, migrateIds) => {
       if (migrateIds) {
         const amcvCookieValue = cookieJar.get(`AMCV_${imsOrgId}`);
         if (amcvCookieValue) {
-          const reg = /MCMID\|(\d+)\|/;
-          [, ecid] = amcvCookieValue.match(reg);
+          const reg = /(^|\|)MCMID\|(\d+)($|\|)/;
+          const matches = amcvCookieValue.match(reg);
+          // Destructuring arrays breaks in IE
+          // eslint-disable-next-line prefer-destructuring
+          ecid = matches[2];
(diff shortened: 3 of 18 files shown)


sandbox/public/index.html
```diff
@@ -52,13 +61,16 @@
...
+          const data = options.data;
+          const titleParam = getUrlParameter("title");
+          if (titleParam) {
+            xdm.web.webPageDetails.name = titleParam;
           }
-          if (urlSearchParams.has("campaign")) {
-            data.campaign = urlSearchParams.get("campaign");
+          const campaignParam = getUrlParameter("title");
+          if (campaignParam) {
+            data.campaign = campaignParam;
```
Aaronius: I wanted to manually test the changes using the sandbox inside IE. In order to do that, I needed to convert this code to ES5.

src/components/Identity/createMigration.js
```diff
@@ -14,8 +14,11 @@ export default (imsOrgId, migrateIds) => {
       if (migrateIds) {
         const amcvCookieValue = cookieJar.get(`AMCV_${imsOrgId}`);
         if (amcvCookieValue) {
-          const reg = /MCMID\|(\d+)\|/;
-          [, ecid] = amcvCookieValue.match(reg);
+          const reg = /(^|\|)MCMID\|(\d+)($|\|)/;
```
Aaronius: I came across an amcv cookie value that was just `MCMID|123` where `123` was a long number. The regex that we had would only match if there was a pipe after the number, which there wasn't in this case. I think this was an unintentional bug or maybe I somehow had an unusual amcv cookie in my sandbox, but @jileeshs should double-check my work here.

src/components/Identity/createMigration.js
```diff
@@ -14,8 +14,11 @@ export default (imsOrgId, migrateIds) => {
...
+          const matches = amcvCookieValue.match(reg);
+          // Destructuring arrays breaks in IE
+          // eslint-disable-next-line prefer-destructuring
+          ecid = matches[2];
```
Aaronius: Array destructuring here was also breaking IE.

src/core/createEvent.js
```diff
@@ -33,7 +33,7 @@ export default () => {
     documentUnloading() {
       documentUnloading = true;
     },
-    isDocumentUnloading() {
+    get isDocumentUnloading() {
```
Aaronius: I changed this to a getter to be consistent with `expectsResponse`.

test/unit/specs/utils/assignIf.spec.js
```diff
@@ -0,0 +1,14 @@
...
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+// eslint-disable-next-line no-unused-vars
+import assignIf from "../../../../src/utils/assignIf";
```
Aaronius: I'd like to know how code was pushed without this spec file created.

test/unit/specs/utils/getBrowser.spec.js
```diff
@@ -0,0 +1,68 @@
...
+    "Mozilla/5.0 (Windows NT 6.3; Trident/7.0; rv:11.0) like Gecko"
+  ],
+  [SAFARI]: [
+    "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/12.0 Safari/605.1.15",
+    "Mozilla/5.0 (iPhone; CPU iPhone OS 12_2 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/13 Mobile/15E148 Safari/604.1"
+  ],
+  // Note that browsers like Opera will usually be reported as something like
+  // Chrome instead of Unknown because they contain other browser names in their
+  // user agent strings. For example, this one from Opera:
+  // Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/78.0.3904.97 Safari/537.36 OPR/64.0.3417.92
```
Aaronius: This note is important to understand. Due to how complicated user agent strings are now with practically every browser trying to mimic other browsers, some browsers that we aren't specifically trying to detect will end up being detected as one of our supported browsers. Unless we add regexes for more browsers, I think this is about the best we can do.

src/core/createConfigValidators.js
```diff
@@ -11,6 +11,8 @@ governing permissions and limitations under the License.
 */
 
 import { boolean, string, callback } from "../utils/configValidators";
+import { EDGE_DOMAIN } from "../constants/domains";
+import EDGE_BASE_PATH from "../constants/edgeBasePath";
```
Aaronius: It didn't look like `edgeBasePath.js` was being used at all previously.

test/unit/specs/utils/assignIf.spec.js
jfkhoury: I think this is on me @Aaronius, I'll double check why the script is not running on my machine.

src/components/Identity/createMigration.js
jfkhoury: I think it is possible to have an AMCV cookie with `MCMID|546736y47u239057230958` with no `|` at the end.

test/unit/specs/utils/getBrowser.spec.js
jfkhoury: Thank you @Aaronius. We can bring this up with Justin again and see if he'd like to be more granular.

sandbox/public/index.html
jonsnyder: At what point was this breaking in IE?  During the build?  I was hoping this would throw an error when called, which is handled by the eventManager.

test/unit/specs/utils/getBrowser.spec.js
jonsnyder: What do you think about using a library for this?

src/components/Identity/createComponent.js
```diff
@@ -101,6 +102,12 @@ export default (
             // promise so that future requests can know when the ECID has returned.
             deferredForEcid = defer();
             payload.expectResponse();
+            if (
+              config.thirdPartyCookiesEnabled &&
+              isThirdPartyCookieSupported(window)
+            ) {
+              payload.useIdThirdPartyDomain();
+            }
```
jileeshs: If I understand right, this code would call 3rd party server only when we know it supports 3rd party cookie (Chrome).

I think yesterday we were talking about call 3rd party end point on all browsers except the browsers known as blocking 3rd party (Example: Safari, Firefox and Edge).

sandbox/public/index.html
Aaronius: At runtime. It probably was eventManager throwing the error, but it seems odd to just ignore the error in IE rather than fix it.

test/unit/specs/utils/getBrowser.spec.js
Aaronius: The libraries I looked at all try to do a lot more than what we need and would end up bloating the library. If there's one you suggest though, I think we should consider it.

Aaronius: I do think some of the libraries have much more specific regexes though that could keep browsers like Opera in the "unknown" bucket. I'm going to try out some of them.

src/components/Identity/createComponent.js
Aaronius: If you look [here](https://github.com/adobe/alloy/pull/272/files#diff-672edb84f7055556d3a2b9055ab49876R19) you'll see that "unknown" browsers are assumed to have third-party cookie support.

test/unit/specs/utils/getBrowser.spec.js
jonsnyder01: I was looking at detect-browser from a quick google search. It would add some bloat to alloy, but having to keep this up to date on our own is a big investment.

Aaronius: I've been looking at detect-browser as well. Some of its regular expressions will keep browsers like Opera (but not Brave) in the Unknown bucket rather than the Chrome bucket, which is nice. I'm going to copy some of its regular expressions (I'll open a new PR for that work), but I don't feel like using the library directly is worth the cost.

sandbox/public/index.html
Aaronius: I took another look. Initially IE was throwing an error when parsing index.html because of this syntax:

`onBeforeEventSend({ xdm, data }) {`

so I converted that line to ES5, which fixed that particular error. After that, I ran into the error that was logged by the eventManager due to URLSearchParams being undefined and decided to fix that as well. Interestingly, URLSearchParams is sometimes defined and sometimes is undefined by the time the `onBeforeEventSend` function runs. It looks like there's code being loaded asynchronously, probably from the create-react-app stuff, that provides a polyfill for URLSearchParams. Considering onBeforeEventSend fails some of the time due to URLSearchParams being undefined, I think it's best to just avoid the problem altogether.

PR #273, opened 2019-11-12
Refactor config: simplify the API, introduce `isRequired`

jfkhoury:
## Description

- We were mixing up `config.get(prop)` and `config.prop`.
- We had to pass `defaultValue: undefined` for non-required configs which was leaking the abstraction. I'm not sure adding an `isRequired` prop to validators is the cleanest way to do this; maybe we should introduce a new `required()` validator util instead. For example:
`configId: required().string().unique()`. But this is a start.

## Related Issue

https://jira.corp.adobe.com/browse/CORE-37116

## Types of changes

## Checklist:

- [ x ] I have signed the [Adobe Open Source CLA](https://opensource.adobe.com/cla.html) or I'm an Adobe employee.
- [ x ] I have made any necessary test changes and all tests pass.


src/core/config/createValidator.js
```diff
@@ -0,0 +1,76 @@
...
+        const configValue = config[key];
+        const isConfigValueProvided = configValue !== undefined;
+        const configValidator = validators[key];
+        const hasDefault = configValidator.defaultValue !== undefined;
+
+        // 1: If no value provided, but there's a default, set default in config.
+        // TODO: It's weird that the validator sets defaults as well. This should happen
+        // when creating the config instead. Maybe pass `options` & `configValidators` to
+        // createConfig.
+        if (!isConfigValueProvided && hasDefault) {
```
jfkhoury: This stuff should be moved to `createConfig` and the validator should only validate. I will make this change in a separate PR.

src/core/config/createValidator.js
```diff
@@ -0,0 +1,76 @@
...
+          config[key] = configValidator.defaultValue;
+        }
+
+        // 2: Validate.
+        if (configValidator.isRequired) {
```
jfkhoury: I talk about this in the PR description, but let me know if having a `required` util is preferred here: `configIf: required().string().unique()`.

src/core/config/createValidator.js
```diff
@@ -0,0 +1,76 @@
...
+          if (!isConfigValueProvided) {
+            ac.push(`${key} is required.`);
+          } else if (configValidator.validate) {
+            executeValidator(configValidator.validate, key, ac);
+          }
+        } else if (isConfigValueProvided && configValidator.validate) {
+          executeValidator(configValidator.validate, key, ac);
+        }
```
jonsnyder: This could be simplified to the following: (note, I removed `ac.push...` because I don't want to duplicate the code in core/configValidator/required.js which gets called first for every validator)
```
if (!isConfigValueProvided && hasDefault) {
  config[key] = configValidator.defaultValue;
  // no need to validate the defaultValue
} else if (isConfigValueProvided || configValidator.isRequired) {
  // We need to validate if the value was provided by the user, but also if there was no value provided and this config
  // is required, the validator will give an error message.
  executeValidator(configValidator.validate, key, ac);
}
```
Truth Table: [Doc - Nov 12 2019 - 11-39 AM.pdf](https://github.com/adobe/alloy/files/3837676/Doc.-.Nov.12.2019.-.11-39.AM.pdf)

Also with this simplified we could just inline the executeValidator code.

test/unit/specs/core/buildAndValidateConfig.spec.js
```diff
@@ -1,14 +1,14 @@
...
-Unless required by applicable law or agreed to in writing, software distributed under
-the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
-OF ANY KIND, either express or implied. See the License for the specific language
-governing permissions and limitations under the License.
-*/
+// Unless required by applicable law or agreed to in writing, software distributed under
+// the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+// OF ANY KIND, either express or implied. See the License for the specific language
+// governing permissions and limitations under the License.
+// */
```
jonsnyder: weird Copyright.

src/core/config/createValidator.js
```diff
@@ -0,0 +1,76 @@
...
+
+const CONFIG_DOC_URI =
+  "https://launch.gitbook.io/adobe-experience-platform-web-sdk/fundamentals/configuring-the-sdk";
+
+const createConfigValidator = config => {
+  const validators = Object.create(null);
+
+  const validator = {
+    addValidators(newValidators) {
+      return assign(validators, newValidators);
```
jonsnyder: adding isRequired creates a weird edge case when isRequired is true, and there is a defaultValue.  We could throw an error or just ignore the isRequired field as this code currently does.

src/core/config/createValidator.js
```diff
@@ -0,0 +1,76 @@
...
+        const errorMessage = validate(key, config[key], defaultValue);
+
+        if (errorMessage) {
+          errors.push(errorMessage);
+        }
+      };
+
+      const errors = Object.keys(validators).reduce((ac, key) => {
+        const configValue = config[key];
+        const isConfigValueProvided = configValue !== undefined;
```
jonsnyder: Before, the logic was `const isConfigValueProvided  = configValue == null` which tested for null or undefined.  I would expect that if I passed in null for an optional config, the default value would be used, but that is not the case with this change.

test/unit/specs/core/buildAndValidateConfig.spec.js
```diff
@@ -1,14 +1,14 @@
...
```
Aaronius: Was this intentional?

test/unit/specs/core/config/createConfig.spec.js
```diff
@@ -0,0 +1,47 @@
...
+  it("returns undefined when a missing key is requested", () => {
+    const cfg = createConfig(testConfig);
+    expect(cfg.missing).toBe(undefined);
+  });
+  it("supports adding a key value mapping", () => {
+    const cfg = createConfig(testConfig);
+    cfg.d = "ABC";
+    expect(cfg.d).toEqual("ABC");
+  });
+});
```
Aaronius: You might want to test that the original `testConfig` object isn't modified when changing `cfg` and vice-versa.

src/core/config/createValidator.js
jfkhoury: Thank you @Aaronius !

src/core/config/createValidator.js
jfkhoury: Thank you @jonsnyder

src/core/config/createValidator.js
jfkhoury: Done.

test/unit/specs/core/buildAndValidateConfig.spec.js
jfkhoury: @Aaronius @jonsnyder I LOVE to double comment my comments. It's safer, you never know... 🤕

PR #275, opened 2019-11-12
CORE-37168 Improved user agent regular expressions. Improved tests.

Aaronius:
## Description
I'll add inline comments.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-37168

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Identity/createComponent.js
+++ b/src/components/Identity/createComponent.js
@@ -2,7 +2,7 @@ import createCustomerIds from "./customerIds/createCustomerIds";
 import { defer } from "../../utils";
 import { EXPERIENCE_CLOUD_ID } from "./constants/cookieNames";
 import createMigration from "./createMigration";
-import isThirdPartyCookieSupported from "../../utils/isThirdPartyCookieSupported";
+import areThirdPartyCookiesSupported from "../../utils/areThirdPartyCookiesSupported";
 import getBrowser from "../../utils/getBrowser";
 
 const addIdsContext = (payload, ecid) => {
@@ -105,7 +105,7 @@ export default (
             payload.expectResponse();
             if (
               config.thirdPartyCookiesEnabled &&
-              isThirdPartyCookieSupported(getBrowser(window))
+              areThirdPartyCookiesSupported(getBrowser(window))
             ) {
               payload.useIdThirdPartyDomain();
             }
--- a/src/components/Identity/createComponent.js
+++ b/src/components/Identity/createComponent.js
@@ -3,6 +3,7 @@ import { defer } from "../../utils";
 import { EXPERIENCE_CLOUD_ID } from "./constants/cookieNames";
 import createMigration from "./createMigration";
 import isThirdPartyCookieSupported from "../../utils/isThirdPartyCookieSupported";
+import getBrowser from "../../utils/getBrowser";
 
 const addIdsContext = (payload, ecid) => {
   payload.addIdentity(EXPERIENCE_CLOUD_ID, {
@@ -104,7 +105,7 @@ export default (
             payload.expectResponse();
             if (
               config.thirdPartyCookiesEnabled &&
-              isThirdPartyCookieSupported(window)
+              isThirdPartyCookieSupported(getBrowser(window))
             ) {
               payload.useIdThirdPartyDomain();
             }
--- a/src/utils/getBrowser.js
+++ b/src/utils/getBrowser.js
@@ -61,8 +61,10 @@ export default memoize(window => {
     FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
     DEALINGS IN THE SOFTWARE.
     */
-    /* eslint-disable */
+    // Regular expression patterns were copied from
+    // https://github.com/DamonOehlman/detect-browser
     // These are only the browsers that Alloy officially supports.
+    /* eslint-disable */
     [EDGE]: /Edge\/([0-9\._]+)/,
     // Edge Chromium can dynamically change its user agent string based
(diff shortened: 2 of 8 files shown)


src/components/Identity/createComponent.js
```diff
@@ -104,7 +105,7 @@ export default (
             payload.expectResponse();
             if (
               config.thirdPartyCookiesEnabled &&
-              isThirdPartyCookieSupported(window)
+              isThirdPartyCookieSupported(getBrowser(window))
```
Aaronius: Rather than `isThirdPartyCookieSupported` using `getBrowser` under the hood, I decided to pass in the browser. It makes it easier to unit test.

src/utils/getBrowser.js
```diff
@@ -35,13 +38,41 @@ const matchUserAgent = regexs => {
...
+    FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+    DEALINGS IN THE SOFTWARE.
+    */
+    /* eslint-disable */
+    // These are only the browsers that Alloy officially supports.
+    [EDGE]: /Edge\/([0-9\._]+)/,
+    // Edge Chromium can dynamically change its user agent string based
+    // on the host site:
+    // https://winaero.com/blog/microsoft-edge-chromium-dynamically-changes-its-user-agent/
+    [EDGE_CHROMIUM]: /Edg\/([0-9\.]+)/,
```
Aaronius: Notice I've added Edge Chromium, which uses a different user agent string than Edge. It may change, but since we know we'll be supporting Edge Chromium in the near future, I figured we should get a jump on it.

src/utils/isThirdPartyCookieSupported.js
```diff
@@ -10,13 +10,12 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
 import includes from "./includes";
 
 // Users could have also disabled third-party cookies in Chrome or IE, but we
 // don't know. We also assume "unknown" browsers support third-party cookies,
 // though we don't really know that either. We're making best guesses.
 const browsersSupportingThirdPartyCookie = [CHROME, IE, UNKNOWN];
 
-export default window =>
-  includes(browsersSupportingThirdPartyCookie, getBrowser(window));
+export default browser => includes(browsersSupportingThirdPartyCookie, browser);
```
Aaronius: Notice that `browser` is being passed in rather than this module calling `getBrowser`. Because the browser constants (e.g., `CHROME`, `IE`) are now used here whereas `getBrowser` is not, I felt it important to split the browser constants into their own constants file (`constants/browser.js`) instead of have them as named exports on `utils/getBrowser.js`

test/unit/specs/components/Identity/createMigration.spec.js
```diff
@@ -21,12 +19,20 @@ describe("createMigration(", () => {
...
-      const migration = createMigration("TEST_ORG", true);
-      expect(migration.getEcidFromAmcvCookie(identityCookieJar)).toEqual(
-        "1234"
-      );
+
+    [
+      "random|string|MCMID|1234|random|random",
+      "MCMID|1234|random|random",
+      "random|random|MCMID|1234",
+      "MCMID|1234"
```
Aaronius: This now tests the possible AMCV cookie variations.

src/utils/getBrowser.js
```diff
@@ -35,13 +38,41 @@ const matchUserAgent = regexs => {
...
+    [CHROME]: /(?!Chrom.*OPR)Chrom(?:e|ium)\/([0-9\.]+)(:?\s|$)/,
+    [FIREFOX]: /Firefox\/([0-9\.]+)(?:\s|$)/,
+    // This only covers version 11 of IE, but that's all we support.
+    [IE]: /Trident\/7\.0.*rv\:([0-9\.]+).*\).*Gecko$/,
+    [SAFARI]: /Version\/([0-9\._]+).*Safari/
```
Aaronius: Note that these regular expressions are more specific than the ones I had. These are better at not detecting Opera as Chrome, for example, but it's not perfect (Brave is still detected as Chrome, for example).

src/utils/isThirdPartyCookieSupported.js
Aaronius: BTW, does anyone feel like `isThirdPartyCookieSupported` should be renamed to `areThirdPartyCookiesSupported`?

Aaronius:
I added a commit that renames `isThirdPartyCookieSupported` to `areThirdPartyCookiesSupported`. I also added Edge and Edge Chromium to the does-support-3rd-party-coookies bucket. They both support 3rd-party cookies by default. They have a preview version that blocks 3rd-party cookies by default but I think we should consider them as supporting 3rd-party cookies until those previews are actually released as stable versions.

src/utils/getBrowser.js
liljenback: Default is always not to support 3rd party?

Aaronius: "Unknown" browsers are considered to support third-party cookies: https://github.com/adobe/alloy/pull/275/files#diff-32a1eddabd165b429781245af81fb630R24

Aaronius: Everything in this list is considered supporting 3rd party cookies:

```
[
  CHROME,
  EDGE,
  EDGE_CHROMIUM,
  IE,
  UNKNOWN
];
```

liljenback:
What would be the process when a browser in the approved list starts to block 3rd party? Update the implementations?

Aaronius:
Yeah, release a new version and the customers need will need to start using it.

PR #278, opened 2019-11-14
CORE-37378 add eventType and eventMergeId as event command options

jonsnyder:
## Description

If present, the eventType and eventMergeId options are added to the userXdm object of the event.

## Related Issue

https://jira.corp.adobe.com/browse/CORE-37378

## Motivation and Context

This will make using the Alloy extension easier as the user can specify their own eventType without modifying the xdm object they are passing in.

## Types of changes

## Checklist:

### DIFF
--- a/src/components/DataCollector/index.js
+++ b/src/components/DataCollector/index.js
@@ -21,8 +21,8 @@ const createDataCollector = ({ eventManager }) => {
           data,
           viewStart = false,
           documentUnloading = false,
-          eventType,
-          eventMergeId
+          type,
+          mergeId
         } = options;
         const event = eventManager.createEvent();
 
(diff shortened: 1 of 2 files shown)


src/components/DataCollector/index.js
```diff
@@ -14,18 +16,32 @@ const createDataCollector = ({ eventManager }) => {
...
       event(options) {
+        let { xdm } = options;
         const {
-          xdm,
           data,
           viewStart = false,
-          documentUnloading = false
+          documentUnloading = false,
+          eventType,
+          eventMergeId
```
Aaronius: Did you consider calling these `type` and `mergeId` instead of `eventType` and `eventMergeId` since we're already in an "event" command? I'm not sure which is better...be concise or be consistent with XDM names?

jonsnyder: I was just following the JIRA ticket.  @jfkhoury or @dancingcactus what do you think?

jfkhoury: I like concise a lot more, but the reason we are exposing them here is to be consistent with XDM. It’ll require a quick addition in the doc to mention that those Params are mapped to the XDM values. 

I’ll leave it up to @dancingcactus.

dancingcactus: Love the more concise language.

Aaronius: Is this what mobile will be doing as well? I want to be consistent.

jfkhoury: Talked to Pete, he doesn't think consistency is an issue.

Aaronius:
I approve again!

PR #283, opened 2019-11-15
CORE-37244 Handle konductor errors and warnings. Reduce request retry scope.

Aaronius:
## Description
Several things:

* Requests are considered successful if the status code is 200 or 204 (No Content). Request are considered retryable if the status code is in the 5xx range or if it's 429 (Too Many Requests). Requests of all other codes are considered failures and are not retried.
* For the promise we return to the customer from the `event` command, we now resolve it with nothing. Previously, we exposed some raw data about the request and the response. We decided to remove this to reduce the amount of API exposed to customers until we truly understand the use cases.
* We now log warnings coming from Konductor (Konductor is still working on returning warnings in the `warnings` key in the response but we don't need to wait for their changes).
* If an error occurs making a connection to the server, or JAG throws an error, or the response from Konductor is in a format that's not expected, then `lifecycle.onRequestFailed` is called (`lifecycle.onResponse` will not be called) and we reject the promise we return to the customer from the event command. 
* In the case that Konductor's response contains error objects (under the `errors` key), then `lifecycle.onResponse` will be called (`lifecycle.onRequestFailed` will not be called). This is because Konductor can return "handle" object that we still need to take action on even when there are error objects (in other words, some stuff may have succeeded while other stuff may have failed). If components need to watch for specific errors, we can open up a `getErrorsByCode` method on the response object we pass to `lifecycle.onResponse`. Note that if any error objects are found on the response, we reject the promise we return to the customer from the event command.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-37244 (although the JIRA issue only addresses a small portion of this work)

## Motivation and Context
Better communication to components and users about problems that occur.

## Types of changes

## Checklist:

- [] **I've logged a couple issues that I plan on doing next week but I wanted to get this PR submitted before the end of the week and before I need to work on cookie management. https://jira.corp.adobe.com/browse/CORE-37860 https://jira.corp.adobe.com/browse/CORE-37861**  I have submitted a [documentation](https://github.com/AdobeDocs/alloy-docs) pull request or no changes are needed.


jfkhoury:
> ## Description
> Several things:
> 

Thank you so much for the details. We should try to extract the things related to customers into the public doc, and the rest will go in the Technical Guide the we will start working on soon (https://jira.corp.adobe.com/browse/CORE-37874)

src/components/Personalization/index.js
```diff
@@ -84,7 +84,7 @@ const createPersonalization = ({ config, logger, eventManager }) => {
 
         showContainers();
       },
-      onResponseError() {
+      onRequestFailure() {
         showContainers();
```
jfkhoury: So is this the first use case of `response.getErrorsByCode`? Because if the status code is considered successful but Personalization errors out, those containers will stay hidden.

src/core/createEventManager.js
```diff
@@ -90,15 +90,31 @@ export default ({ createEvent, optIn, lifecycle, network, config, logger }) => {
...
-            returnData.responseBody = clone(response);
+          if (errors.length) {
+            const errorMessage = errors.reduce((memo, error) => {
+              return `${memo}\n• [Code ${error.code}] ${error.message}`;
+            }, "The server responded with the following errors:");
+            throw new Error(errorMessage);
           }
 
-          return returnData;
+          // We don't want to expose the response to the customer, so
```
jfkhoury: Nice!

src/core/network/createNetwork.js
```diff
@@ -12,26 +12,39 @@ governing permissions and limitations under the License.
...
 import apiVersion from "../../constants/apiVersion";
 import { ID_THIRD_PARTY_DOMAIN } from "../../constants/domains";
+import { RETRYABLE_ERROR, SUCCESS } from "../../constants/responseStatusType";
 
 export default ({ config, logger, lifecycle, networkStrategy }) => {
   const handleResponse = (requestId, responseBody) => {
     let parsedBody;
 
     try {
+      // One of the cases where the response body is not JSON is when
```
jfkhoury: Thank you for the comment, very useful.

src/core/network/createNetwork.js
```diff
@@ -12,26 +12,39 @@ governing permissions and limitations under the License.
...
-        `Error parsing server response.\n${e}\nResponse body: ${responseBody}`
+        `Unexpected server response.\n${e}\nResponse body: ${responseBody}`
       );
     }
 
     logger.log(`Request ${requestId}: Received response.`, parsedBody);
 
     const response = createResponse(parsedBody);
 
+    // TODO Document that onResponse will be called when Konductor
```
jfkhoury: 👍

src/core/network/createResponse.js
```diff
@@ -13,37 +13,45 @@ governing permissions and limitations under the License.
...
- *          example: "identity:persist"
+ * Creates a representation of a gateway response with the addition of
+ * helper methods.
+ * @returns Response
  */
-export default (content = { requestId: "", handle: [] }) => {
-  // TODO: Should we freeze the response to prevent change by Components?
-  // Object.freeze(response.handle.map(h => Object.freeze(h)));
+export default (content = {}) => {
+  const { handle = [], errors = [], warnings = [] } = content;
```
jfkhoury: Is it weird that it's `handle` and not `handles`? I think it's like that because we are matching the Konductor response... :)

Aaronius: "handle" and "handles" are both a little odd to me. Something like "actions" would seem more appropriate I think, but I probably prefer consistency with Konductor over being right here. 😄

src/components/Personalization/index.js
Aaronius: I don't see a case here where containers would stay hidden, but please point it out if I'm missing something. If the request itself fails, `onRequestFailure` is called, and containers will be shown. If the request succeeds (Konductor returns a valid response), then `onResponse` will get called, [`response.getPayloadsByType(DECISIONS_HANDLE);`](https://github.com/adobe/alloy/pull/283/files#diff-74183009f12144cdb4dccd827ce535d5L81) will return 0 fragments, and then containers will be shown.

If we wanted to detect that a personalization error occurred and do something specific about it, that would be an option and that's where we would use `response.getErrorsByCode`, but I don't know what specific action we would take at that point.

src/core/network/createResponse.js
jfkhoury: I agree 🤐  :)

package.json
```diff
@@ -34,7 +34,7 @@
     "sandbox:build": "cd sandbox && npm run build",
     "dev": "npm-run-all --parallel build:watch sandbox",
     "checkbundlesize": "npm run build:prod:standalone:minified && bundlesize",
-    "mockserver": "mockserver -serverPort 1080 -jvmOptions \"-Dmockserver.enableCORSForAllResponses=true\""
```
Aaronius: I'm not sure how you were getting this to work. The mockserver docs seem to show the library API but no CLI.

src/core/network/sendBeaconFactory.js
```diff
@@ -17,6 +17,13 @@ export default (navigator, fetch, logger) => {
...
     }
-    return Promise.resolve();
+
+    // Using sendBeacon, we technically don't get a response back from
+    // the server, but we'll resolve the promise with an object to maintain
+    // consistency with other network strategies.
+    return Promise.resolve({
+      status: 204,
+      body: ""
+    });
```
Aaronius: I didn't need to do this but it made code and tests simpler and more consistent.

package.json
jonsnyder: I must have installed mockserver with `brew`.  Thanks for fixing this!

PR #284, opened 2019-11-17
Migrate the ECID from the `s_ecid` cookie in addition to `AMCV`.

jfkhoury:
## Description

This is necessary in the case where the `AMCV` cookie might have expired on Safari, but the `s_ecid` still exists because it's created via a 1st party CNAME.

## Types of changes

## Checklist:

- [ X ] I have signed the [Adobe Open Source CLA](https://opensource.adobe.com/cla.html) or I'm an Adobe employee.
- [ X ] I have made any necessary test changes and all tests pass.


src/components/Identity/createMigration.js
```diff
@@ -2,27 +2,36 @@ import { cookieJar } from "../../utils";
...
+
+        if (legacyEcidCookieValue) {
           const reg = /(^|\|)MCMID\|(\d+)($|\|)/;
-          const matches = amcvCookieValue.match(reg);
+          const matches = legacyEcidCookieValue.match(reg);
           // Destructuring arrays breaks in IE
           // eslint-disable-next-line prefer-destructuring
-          ecid = matches[2];
-          identityCookieJar.set(EXPERIENCE_CLOUD_ID, ecid);
+          if (matches) {
```
jfkhoury: There are cases where the AMCV cookie might only have `version:1.2.3` by the time we read it, which will break `matches[2]`.

Aaronius: Can you add this as a test case?

jfkhoury: Done @Aaronius, thanks.

src/components/Identity/createComponent.js
```diff
@@ -21,7 +21,7 @@ export default (idSyncs, config, logger, cookieJar, optIn, eventManager) => {
   const getEcid = () => {
     const ecid =
       cookieJar.get(EXPERIENCE_CLOUD_ID) ||
-      migration.getEcidFromAmcvCookie(cookieJar);
+      migration.getEcidFromLegacyCookie(cookieJar);
```
liljenback: It's really cookie(s) here

test/unit/specs/components/Identity/createMigration.spec.js
```diff
@@ -10,14 +10,14 @@ describe("createMigration(", () => {
   beforeEach(() => {
     cookieJar.remove("AMCV_TEST_ORG");
   });
-  describe("getEcidFromAmcvCookie", () => {
+  describe("getEcidFromLegacyCookie", () => {
```
liljenback: Do you have a test when s_ecid is present?

src/components/Identity/createMigration.js
```diff
@@ -2,27 +2,36 @@ import { cookieJar } from "../../utils";
...
   return {
-    getEcidFromAmcvCookie(identityCookieJar) {
+    getEcidFromLegacyCookie(identityCookieJar) {
       let ecid = null;
+      const legacyItpCookieName = "s_ecid";
+
       if (idMigrationEnabled) {
-        const amcvCookieValue = cookieJar.get(`AMCV_${imsOrgId}`);
-        if (amcvCookieValue) {
+        const legacyEcidCookieValue =
```
jonsnyder: I just want to point out that if the legacy itp cookie doesn't match the format, we don't even look at the amcv cookie.  I don't know enough about these cookies to know if this is an issue or not.

jfkhoury: Thank you Jon. The legacy ITP can only have the correct format because the ECID is the only value that goes in there. That being said, I did consider going over the whole workflow one cookie at a time, so I might do that anyway, but in a separate PR.

jfkhoury:
@liljenback I will merge this PR, and work on a separate one with a quick refactor and additional test cases.

PR #290, opened 2019-11-22
C2581 - When ECID not available on client, allow the first request to be sent while queuing subsequent requests

shammowla:
This test only covers that the first command is sent to the server and that the subsequent commands are not sent to the server. I tried inflating the buffer from the demdex response to retrieve the ECID using several zlib options with no luck.


owenleej:
I added in a utility to get the decompressed response body from a request. I might have put it in the wrong directory so feel free to move it to wherever it should go.

test/functional/specs/desktop/C2581.js
```diff
@@ -23,6 +24,7 @@ test("Regression: Queue requests until we receive an ECID.", async () => {
   await t.click(Selector("#event-button"));
 
   await responseStatus(networkLogger.adobedcEndpointLogs.requests, 200);
+  console.log(getBody(networkLogger.adobedcEndpointLogs.requests[0]));
```
jfkhoury: Let's remove this console log please.

owenleej: I put this line in as an example for using the getBody utility. @shammowla please remove it when you finish out the test case.

PR #292, opened 2019-11-25
CORE-36752 & CORE-36754 add context test, replace topFrameSetFactory

jonsnyder:
## Description

Add context test for timestamp.  Replace topFrameSetFactory with window.top.  I copied topFrameSetFactory from AppMeasurement, but it looks like that `window.top` does the same thing in modern browsers.

## Related Issues
https://jira.corp.adobe.com/browse/CORE-37652
https://jira.corp.adobe.com/browse/CORE-37654

## Types of changes

## Checklist:


src/components/Context/webFactory.js
```diff
@@ -12,18 +12,14 @@ governing permissions and limitations under the License.
...
   return xdm => {
-    topFrameSet = topFrameSet || topFrameSetProvider();
-
     const web = {
       webPageDetails: {
         URL: window.location.href || window.location
       },
       webReferrer: {
-        URL: topFrameSet.document.referrer
+        URL: window.top.document.referrer
```
jfkhoury: Hey @jonsnyder, can we add a comment to the PR explaining why we removed this?

Aaronius: It's interesting to me that we use the top window's referrer instead of the the current window's referrer. I would have assumed that if a user clicked on something inside of an iframe and the iframe loaded up a document I was tracking, I'd want to track the referrer from the iframe rather than the referrer for the window containing the iframe. Do you know why we use the containing window's referrer?

Have you tested this in a cross-origin scenario where the parent window is in a different origin than the iframe's document?

jonsnyder: The history of the topFrameSet calculation in AppMeasurement.js is lost to time, but reading up on referrers I can guess on why it was done this way. On the initial page load an iframe's referrer will equal the containing page's location (i.e. window.top.location == window.document.referrer).  If we just did `document.referrer` we would lose the outside referrers on a page with frames.  

Integration tests are a good idea to make sure this does the same thing as the old topFrameSet calculation, and see exactly how it behaves with frames. I'll bring it up to @dancingcactus

Aaronius:
Merging on behalf of @jonsnyder who gave his approval to merge.

PR #310, opened 2019-12-11
Removing ids with hashEnabled when crypto.subtle is undefined

jileeshs:
https://jira.corp.adobe.com/browse/CORE-38489

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,5 +42,5 @@ For Adobe Employees that would like write access to this repository, please foll
 4. Clear your cookies.
 5. Refresh the browser. You'll get an error in the console, but you should now have some cookies set (at least one named something like `kndctr_53A16ACB5CC1D3760A495C99_AdobeOrg_identity`).
 6. Open `src/core/network/createNetwork.js` and remove `/${edgeBasePath}` from the url literal string template.
-7. Refresh the browser. Alloy should now function as expected. If you clear your cookies later, you'll need to add back `/${edgeBasePath}` and start fom step 5 again.
+7. Refresh the browser. Alloy should now function as expected.
 


src/utils/convertStringToSha256Buffer.js
```diff
@@ -10,6 +10,9 @@ const encodeText = str => {
 export default message => {
   const data = encodeText(message);
   const crypto = window.msCrypto || window.crypto;
+  if (!crypto.subtle) {
```
jfkhoury: I would log a warning here that we are not able to hash IDs on a non-secure page. Maybe we should double check that the protocol is HTTP before we log. So, inside `if (!crypto.subtle)..` add another if to check the protocol then warn the user.

jileeshs: Should we pass logger to the util?
What is the value in checking for HTTP, if the browser doesn't support hashing.
Shouldn't the warning be
`Your browser doesn't support hashing...`?

jfkhoury: As discussed on the call, let's not worry about the protocol.

Aaronius:
I believe the decision was to fail the command (throw an error) if hashing is enabled on any ID and `crypto.subtle` is undefined, right?

jfkhoury:
> I believe the decision was to fail the command (throw an error) if hashing is enabled on any ID and `crypto.subtle` is undefined, right?

Product wants to warn in the console that the IDs that need hashing have been dropped because hashing is not supported, but still make the `setCustomerIds` call with the remaining IDs, if any.

liljenback:
@Aaronius , @jfkhoury , @jonsnyder , I think Jon suggested earlier that it would be nice to get 3rd party cookie info. This is another example it would be nice to get feedback on how often this fails. It would be nice if we could have a module like this in alloy that collects misc valuable stats and piggybacks that info on the next outbound call.

Aaronius:
@jileeshs according to product requirements that @jfkhoury outlined in https://github.com/adobe/alloy/pull/310#issuecomment-566145063, we still need a warning message logged when an ID can't be hashed. Are you able to do this today? If not, let me know and I can help. I'd like to get it in for tomorrow's release.

PR #312, opened 2019-12-12
CORE-30826 Validating event command

liljenback:
## Related Issue

CORE-30826

## Motivation and Context

Provides some information to the user of the required attributes for the event command.

## Types of changes

## Checklist:

### DIFF
--- a/src/components/DataCollector/index.js
+++ b/src/components/DataCollector/index.js
@@ -13,7 +13,7 @@ governing permissions and limitations under the License.
 */
 
 const CONFIG_DOC_URI =
-  "https://launch.gitbook.io/adobe-experience-platform-web-sdk/fundamentals/executing-commands";
+  "https://www.adobe.io/apis/experienceplatform/home/services/web-sdk.html#!api-specification/markdown/narrative/edge/fundamentals/executing-commands.md";
 
 const createDataCollector = ({ eventManager, logger }) => {
   return {
--- a/src/components/DataCollector/index.js
+++ b/src/components/DataCollector/index.js
@@ -12,7 +12,10 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-const createDataCollector = ({ eventManager }) => {
+const CONFIG_DOC_URI =
+  "https://launch.gitbook.io/adobe-experience-platform-web-sdk/fundamentals/executing-commands";
+
+const createDataCollector = ({ eventManager, logger }) => {
   return {
     commands: {
       event(options) {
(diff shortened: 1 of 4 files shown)


src/components/DataCollector/index.js
```diff
@@ -12,7 +12,10 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-const createDataCollector = ({ eventManager }) => {
+const CONFIG_DOC_URI =
+  "https://launch.gitbook.io/adobe-experience-platform-web-sdk/fundamentals/executing-commands";
```
jfkhoury: I think it's time we use the bitly style links for our docs. This link for example is outdated because our docs moved here: https://www.adobe.io/apis/experienceplatform/home/services/web-sdk.html

@Aaronius was working on setting the links up, let's triple check with him; if not, I'll create a ticket and start that process. In any case, let's change this link to the `adobe.io` one pls.

src/core/createEvent.js
```diff
@@ -52,6 +52,17 @@ export default () => {
     set lastChanceCallback(value) {
       lastChanceCallback = value;
     },
+    validate() {
+      const warnings = [];
+      if (event.isEmpty()) {
+        warnings.push("No event xdm or event data specified.");
+      }
+      event.toJSON();
```
jfkhoury: I don't think we should be calling `toJSON` because it is currently being calling at a very specific point in the lifecyle so we can merge the event command's data with other data provided by the onBeforeSend callback and auto collected data.. 

@Aaronius @jonsnyder

liljenback: Yes, toJSON is doing a lot of things it probably shouldn't do given the function name by altering the event internals. However, I'll change the code to not rely on toJSON.

src/components/DataCollector/index.js
```diff
@@ -45,6 +48,15 @@ const createDataCollector = ({ eventManager }) => {
         event.userXdm = xdm;
         event.userData = data;
 
+        const warnings = event.validate();
+        if (warnings.length) {
+          logger.warn(
+            `Invalid event command options:\n\t - ${warnings.join(
+              "\n\t - "
+            )}\nFor documentation covering the event command see: ${CONFIG_DOC_URI}`
```
Aaronius: This seems brittle. We're not really validating the user input here but instead are asking the event whether it's valid. Whether the event reports itself as valid may not always be correlated to the options the user passed in through this command; we're making the assumption that it is related, but it might not be. Right now the two might match (when options are valid, the event is valid and vice versa), but if someone were to change the event.validate logic to incorporate something other than the options the user provided through this command, that assumption of correlation would break. Since we're reporting on the validity of the user input here, I'd rather we validate the user input rather than validate that the event object is valid. I think that makes even more sense since we want to use config validators for command options eventually (soon, hopefully).

Aaronius: Logged here: https://jira.corp.adobe.com/browse/CORE-39143

PR #317, opened 2019-12-17
Set the domain and expiry of the migration cookie.

jfkhoury:
## Description

- Without setting a `domain` when creating a cookie, the domain won't start with a `.` which makes it more restricted to sub domains.
- Without setting an `expires` attribute, the cookie will be created as a session cookie.

## Types of changes

## Checklist:

- [ x ] I have signed the [Adobe Open Source CLA](https://opensource.adobe.com/cla.html) or I'm an Adobe employee.
- [ x ] I have made any necessary test changes and all tests pass.

### DIFF
--- a/src/components/Identity/createMigration.js
+++ b/src/components/Identity/createMigration.js
@@ -10,7 +10,9 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import { cookieJar } from "../../utils";
+import { getApexDomain, cookieJar } from "../../utils";
+
+const apexDomain = getApexDomain(window, cookieJar);
 
 export default orgId => {
   const amcvCookieName = `AMCV_${orgId}`;
@@ -38,7 +40,9 @@ export default orgId => {
     createLegacyCookie(ecid) {
       const amcvCookieValue = cookieJar.get(amcvCookieName);
       if (!amcvCookieValue) {
-        cookieJar.set(amcvCookieName, `MCMID|${ecid}`);
+        cookieJar.set(amcvCookieName, `MCMID|${ecid}`, {
+          domain: apexDomain
+        });
       }
     }
   };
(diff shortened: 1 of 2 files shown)


src/components/Identity/createMigration.js
```diff
@@ -10,7 +10,10 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import { cookieJar } from "../../utils";
+import { getApexDomain, cookieJar } from "../../utils";
+// TODO: We are already retrieving the apex in core; find a way to reuse it.
+// Maybe default the domain in the cookieJar to apex while allowing overrides.
+const apexDomain = getApexDomain(window, cookieJar);
```
Aaronius: We have the current apex domain already calculated up in core (https://github.com/adobe/alloy/blob/master/src/core/index.js#L53) if you want to inject it down.

src/components/Identity/createMigration.js
```diff
@@ -38,7 +41,11 @@ export default orgId => {
     createLegacyCookie(ecid) {
       const amcvCookieValue = cookieJar.get(amcvCookieName);
       if (!amcvCookieValue) {
-        cookieJar.set(amcvCookieName, `MCMID|${ecid}`);
+        cookieJar.set(amcvCookieName, `MCMID|${ecid}`, {
+          domain: apexDomain,
+          // Without `expires` this will be a session cookie.
+          expires: 390 // days, or 13 months.
+        });
```
Aaronius: It might be good to test this. You'll need to inject cookieJar though.

src/components/Identity/createMigration.js
jfkhoury: > We have the current apex domain already calculated up in core (https://github.com/adobe/alloy/blob/master/src/core/index.js#L53) if you want to inject it down.

@Aaronius, what about the comment I added? Do you think defaulting it in the `cookieJar` is a valid approach?

Aaronius: Yeah, that's an interesting concept. I think it's a good idea.

jfkhoury: I will go ahead and merge this PR. Created a ticket to address the APEX default: https://jira.corp.adobe.com/browse/CORE-39130

PR #319, opened 2019-12-18
Adding a warning when hashing fails

jileeshs:
https://jira.corp.adobe.com/browse/CORE-38489

## Related Issue

https://jira.corp.adobe.com/browse/CORE-38489
## Types of changes

## Checklist:

### DIFF
--- a/src/components/Identity/createComponent.js
+++ b/src/components/Identity/createComponent.js
@@ -143,7 +143,9 @@ export default (processIdSyncs, config, logger, optIn, eventManager) => {
     },
     commands: {
       setCustomerIds(options) {
-        return optIn.whenOptedIn().then(() => customerIds.sync(options));
+        return optIn
+          .whenOptedIn()
+          .then(() => customerIds.sync(options, logger));
       }
     }
   };
--- a/src/components/Identity/customerIds/createCustomerIds.js
+++ b/src/components/Identity/customerIds/createCustomerIds.js
@@ -18,7 +18,7 @@ export default eventManager => {
         if (!hashedId) {
           delete finalIds[idsToHash[index]];
           logger.warn(
-            `This browser doesn't support hashing, alloy couldn't hash ${idsToHash[index]}`
+            `Unable to hash identity ${idsToHash[index]} due to lack of browser support.`
           );
         } else {
           finalIds[idsToHash[index]].id = convertBufferToHex(hashedId);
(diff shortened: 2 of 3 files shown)


src/components/Identity/customerIds/createCustomerIds.js
```diff
@@ -2,19 +2,27 @@ import { normalizeCustomerIds, validateCustomerIds } from "./util";
...
-      convertStringToSha256Buffer(normalizedIds[id].id)
+      convertStringToSha256Buffer(normalizedIds[id].id.trim().toLowerCase())
     );
     return Promise.all(idHashPromises).then(hashedIds => {
       return hashedIds.reduce((finalIds, hashedId, index) => {
-        finalIds[idsToHash[index]].id = convertBufferToHex(hashedId);
+        if (!hashedId) {
+          delete finalIds[idsToHash[index]];
+          logger.warn(
+            `This browser doesn't support hashing, alloy couldn't hash ${idsToHash[index]}`
```
Aaronius: No need to call out "alloy" here (we already prefix all messages). You could simplify it to something like `Unable to hash identity ${idsToHash[index]} due to lack of browser support.`

Also, isn't `idsToHash[index]` an object here? If so, I think it might be emitting `This browser doesn't support hashing, alloy couldn't hash [object Object]`. I don't know that printing the whole object is necessary anyway. Can we just print the actual ID or whatever is most identifiable to the user?

Aaronius: Tests too please!

jileeshs: I agree with the message.
`idsToHash` is an array and `idsToHash[index]` gives the namespace, i have verified it in Sandbox and it is like 
`This browser doesn't support hashing, alloy couldn't hash email`

Aaronius: In that case it might make more sense to say "Unable to hash email identity due to lack of browser support." Your call though.

src/components/Identity/createComponent.js
```diff
@@ -143,7 +143,9 @@ export default (processIdSyncs, config, logger, optIn, eventManager) => {
     },
     commands: {
       setCustomerIds(options) {
-        return optIn.whenOptedIn().then(() => customerIds.sync(options));
+        return optIn
+          .whenOptedIn()
+          .then(() => customerIds.sync(options, logger));
```
Aaronius: Hey @jileeshs, I just noticed you're passing through `logger` when you call `sync`. Instead, you should pass it through [when `customerIds` gets created](https://github.com/adobe/alloy/pull/319/files#diff-e3608b4e8d26a7ee90affbadd009744eR29) since the logger won't change across `sync` calls.

Aaronius: Logged here: https://jira.corp.adobe.com/browse/CORE-39144

PR #322, opened 2020-01-02
CORE-39107 E2E C2592 Event command sends a request

wgamboa:
E2E Testing
## Types of changes

## Checklist:

### DIFF
--- a/test/functional/sandbox/js/helper.js
+++ b/test/functional/sandbox/js/helper.js
@@ -111,18 +111,6 @@
   router.register("/event", "event", {
     onPostInit: function() {
       alloy("event", {
-        data: {
-          key: "value"
-        }
-      }).catch(ex => {
-        console.log(ex);
-      });
-    }
-  });
-
-  router.register("/C2592", "event-command", {
-    onPostInit: function() {
-      alloy("event", {
         // TODO: Change data to match final XDM schema.
         xdm: {
           key: "value"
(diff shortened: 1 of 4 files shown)


test/functional/sandbox/js/helper.js
```diff
@@ -120,6 +120,19 @@
     }
   });
 
+  router.register("/C2592", "event-command", {
```
Aaronius: I might be reading this wrong, but it seems like the other routes in this file are set up to be potentially useable by more than a single test. i think it would be useful to do the same here. In other words, instead of naming this route `/C2592`, maybe name it `/generic-event-with-xdm` or something like that.

Actually, scratch that. There's already an `/event` route--the only difference is that it sends generic data (using the `data` option) rather than xdm (using the `xdm` option), but none of the tests that are using the `event` route are testing anything specific to generic data yet. I would recommend just changing `data` to `xdm` in the `/event` route and using that route in your test. If we ever need to test sending generic data (using the `data` option), we an add that later.

test/functional/specs/desktop/C2592.js
```diff
@@ -0,0 +1,33 @@
+import { t, Selector } from "testcafe";
+import createNetworkLogger from "../../src/networkLogger";
+import { responseStatus } from "../../src/assertions/index";
+import fixtureFactory from "../../src/fixtureFactory";
+
+const urlCollector = `http://127.0.0.1:8080/test/functional/sandbox/html/alloySdk.html`;
```
liljenback: I added this constant in a different test:
https://github.com/adobe/alloy/blob/master/test/functional/src/constants/testServerUrl.js

PR #323, opened 2020-01-02
Opt-in and opt-out with server-managed cookies.

Aaronius:
## Description
**⚠️ Large PR inbound!**
This PR consists of the following:
* The opt-in and opt-out cookies are written by Konductor. Alloy reads these cookies in order to determine the user's consent preferences. Because of this, the format of the cookie's value has been agreed upon by Alloy and Konductor teams. Because of this, Konductor will treat any change to the key or value structure as a breaking change.
* The user can now opt out of all purposes using the `optOut` command. Once a user is opted out, they cannot reverse their opt-out (maybe this will change in the future?).
* Opt-out takes precedence over any opt-in. If the user opted out, it doesn't really matter what they've opted into; opt-out will prevent whatever the user is trying to do that's privacy-sensitive.
* When a user changes opt-in preferences or opt-out preferences, we immediately (optimistically) start using the new preferences, but also send the preferences to the server using the new `opt-in` and `opt-out` endpoints. If the server call fails for whatever reason, we revert back to the preferences stored in the cookies.
* The request payload structure for requests to the `opt-in` and `opt-out` endpoints are different from the request payloads to `interact` and `collect` endpoints. Because of this, there are now two types of request payloads: consent request payloads (`createConsentRequestPayload.js`) and data collection request payloads (`createDataCollectionRequestPayload.js`), both of which "inherit" from `createRequestPayload.js`. To make this inheritance simpler and because of some inconsistencies, it made sense to me to change our getters and setters on the payload and event objects to be actual methods (e.g., `getUseIdThirdPartyDomain` instead of `get useIdThirdPartyDomain`).
* If a data collection call is being made, the `onBeforeDataCollectionRequest` lifecycle method on components will be called whereas if a consent request is being made, the `onBeforeConsentRequest` lifecycle method will be called. Note that the response payload format from the server is the same for both data collection responses and consent responses. Therefore, the `onResponse` lifecycle method will be called for both data collection and consent responses.
* To support the above items, `createNetwork` has been reworked. There is now `sendNetworkRequestFactory.js`, which is focused on the "lower-level", edge-agnostic logic of sending and retrying requests. It is not knowledgable about edge URLs, component lifecycles, or payload structures (other than that the response is expected to be JSON). There is also `sendEdgeNetworkRequestFactory.js`, which leverages `sendNetworkRequestFactory.js`, but is more knowledgable about both the edge and component lifecycles. I don't feel 100% settled on the structure of these two modules, but I feel like they're decent.

I'll try to go through the PR and drop inline comments on smaller items.

If you run into unexpected behavior, be aware that there are some open issues in Konductor (or fixes that haven't been pushed to the Konductor environment you're using):

https://jira.corp.adobe.com/browse/EXEG-728
https://jira.corp.adobe.com/browse/EXEG-744
https://jira.corp.adobe.com/browse/EXEG-671
https://jira.corp.adobe.com/browse/EXEG-758
https://jira.corp.adobe.com/browse/EXEG-774
https://jira.corp.adobe.com/browse/EXEG-775

But please feel free to ask about anything unexpected you see anyway.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-37170
https://jira.corp.adobe.com/browse/CORE-38497

## Motivation and Context
Support consent management with cookies written by the server.

## Types of changes

## Checklist:

- [ ] **I haven't but I will start right after this!** I have submitted a [documentation](https://github.com/AdobeDocs/alloy-docs) pull request or no changes are needed.


.eslintrc.js
```diff
@@ -17,6 +17,7 @@ module.exports = {
     "prefer-destructuring": "off"
   },
   globals: {
-    expectAsync: "readonly" // newer jasmine feature
+    expectAsync: "readonly", // newer jasmine feature
+    spyOnAllFunctions: "readonly" // newer jasmine feature
```
Aaronius: I didn't end up using this but left it in the lint config anyway.

src/components/ActivityCollector/attachClickActivityCollector.js
```diff
@@ -10,18 +10,25 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+          return Promise.resolve();
+        }
 
-      eventManager.sendEvent(event);
-    });
+        return eventManager.sendEvent(event);
+      })
+      .catch(error => {
+        throw stackError("Failed to track click", error);
+      });
```
Aaronius: This is an interesting case where if the user opted into no purposes or opted out of all purposes, then click tracking will fail and the error will go unhandled, since the click tracking isn't actuated by the customer calling a command. Without adding context to the error, the error description was pretty generic (like "The user opted into no purposes").

src/components/ActivityCollector/createLinkClick.js
```diff
@@ -50,7 +50,7 @@ export default (window, config) => {
     }
 
     if (isValidLink) {
-      event.documentUnloading();
+      event.documentMayUnload();
```
Aaronius: Now that `event.documentUnloading()` is called for every link, regardless of whether the document is actually unloading, I renamed it to `event.documentMayUnload()`.

src/components/Identity/createComponent.js
```diff
@@ -17,135 +17,135 @@ const addEcidToPayload = (payload, ecid) => {
...
   );
   let deferredForEcid;
-  const migration = createMigration(orgId);
+  const migration = createMigration({ orgId, consent });
   const hasIdentityCookie = () => Boolean(cookieJar.get(identityCookieName));
-  const customerIds = createCustomerIds(eventManager);
+  const customerIds = createCustomerIds({ eventManager, consent });
+
+  // TO-DOCUMENT: We wait for ECID before trigger any events.
+  const accommodateIdentityOnRequest = payload => {
```
Aaronius: Identities are passed on both consent requests and data collection requests. The first request of either type should pause subsequent requests. For this reason, I broke this logic out and called it from both `onBeforeDataCollectionRequest` and `onBeforeConsentRequest`.

src/components/Identity/createComponent.js
```diff
@@ -17,135 +17,135 @@ const addEcidToPayload = (payload, ecid) => {
...
+    let promise;
+
+    if (!hasIdentityCookie()) {
+      const ecidToMigrate =
+        idMigrationEnabled && migration.getEcidFromLegacyCookies();
```
Aaronius: We deemed it okay to read the legacy cookie without consent, but we can't write it without consent.

src/components/Identity/createComponent.js
```diff
@@ -17,135 +17,135 @@ const addEcidToPayload = (payload, ecid) => {
...
+      onBeforeDataCollectionRequest({ payload }) {
+        return accommodateIdentityOnRequest(payload);
+      },
+      onBeforeConsentRequest({ payload }) {
+        return accommodateIdentityOnRequest(payload);
       },
-
-      // Waiting for opt-in because we'll be reading the ECID from a cookie
       onResponse({ response }) {
-        return optIn.whenOptedIn().then(() => {
```
Aaronius: This whole function body was wrapped in `optIn.whenOptedIn().then()`. I decided to push the logic that waits for consent closer to the actual thing that needs consent. In this case, we need consent when creating the legacy cookie and we need consent to process ID syncs, so I pushed the logic that waits for consent down into `migration.createLegacyCookie` and `processIdSyncs`. That prevents us from waiting for consent unnecessarily (like if we don't need to write a legacy cookie or process ID syncs) and helps ensure that we will wait for consent whenever we need to (for example, if some other module in the system started calling `migration.createLegacyCookie` or `processIdSyncs`).

src/components/Identity/createComponent.js
```diff
@@ -17,135 +17,135 @@ const addEcidToPayload = (payload, ecid) => {
...
+        );
+
+        return Promise.all(promises);
       }
     },
     commands: {
       setCustomerIds(options) {
-        return optIn
-          .whenOptedIn()
-          .then(() => customerIds.sync(options, logger));
```
Aaronius: Same thing here. I pushed the logic that waits for consent down into `customerIds.sync`.

src/core/createConsent.js
```diff
@@ -0,0 +1,264 @@
...
+    // We look for any purpose being true rather than just looking for the
+    // "general" purpose being true because it gives us some flexibility in the
+    // future. If the edge decides to split the "general" purpose into
+    // individual purposes, like "tracking" and "personalization", this code
+    // would conservatively consider the user opted out of everything if either
+    // "tracking" or "personalization" were set to true. This buys us time
+    // to release a new version of Alloy that appropriately behaves according
+    // to the more granular purposes and for customers to upgrade. In the
+    // case of opt-in, this is also why we look for all purposes being true
+    // rather than just the "general" purpose being true.
```
Aaronius: This is worth pointing out. Right now the opt-in and opt-out cookie values look like `GENERAL=true` or `GENERAL=false`, but that's just because we don't have more granular purposes. It's likely that `GENERAL` goes away and we start using more granular purposes.

src/core/edgeNetwork/requestPayloads/createConsentRequestPayload.js
```diff
@@ -0,0 +1,29 @@
...
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+import createRequestPayload from "./createRequestPayload";
+
+export default () => {
+  return createRequestPayload(content => {
```
Aaronius: Here's how the consent request payload "extends" the base request payload.

src/utils/stringToBoolean.js
```diff
@@ -1,5 +1,5 @@
 import isString from "./isString";
 
 export default str => {
-  return isString(str) && (str.toLowerCase() === "true" || str === "1");
+  return isString(str) && str.toLowerCase() === "true";
```
Aaronius: I felt like allowing a string of "1" was a too generous and if we're ever allowing `1` to signify `true` then we should probably reconsider why we're allowing it. We don't seem to be supporting it anywhere (I think it was considered for the `alloy_debug` query string parameter at one point but it's not documented as supported).

src/core/createConsent.js
```diff
@@ -0,0 +1,273 @@
...
+            payload,
+            action: "opt-in"
+          });
+        })
+        .then(() => {
+          // We optimistically use the preferences the user provided right away,
+          // but if the server doesn't agree and sends back a cookie value that
+          // conflicts, we adopt whatever the server said since that ultimately
+          // is what will be used if the user refreshes the page.
+          consentState.refreshFromCookies();
```
Aaronius: Because we're doing this, there is a race condition if the customer updates consent preferences in succession like this:

```
optIn({
  purposes: "all"
});
// At this point, consentState will show that the user has opted into all purposes, because we immediately use the user's preferences before we get a response back from the server.

optIn({
  purposes: "none"
});
// At this point, consentState will show that the user has opted into all purposes, because we immediately use the user's preferences before we get a response back from the server.
```

then let's say the response for the first command comes back from the server. If everything went well, the cookie will be set to `GENERAL=true`. Since we call `consentState.refreshFromCookies()` here, consentState will now show that the user has opted into all purposes until the response comes back from the second command. During this period, Alloy is technically not behaving according to user preferences, since the last thing the customer did was set `purposes: none`.

This is probably a very niche scenario, but one we should probably account for by using some sequencing logic. Thoughts?

src/components/ActivityCollector/attachClickActivityCollector.js
Aaronius: I'm still thinking about how to best handle this scenario. Because the error doesn't bubble up through generic error massaging logic, the error says nothing about Alloy and the error can't be silenced by `errorsEnabled: false`. I've logged this here: https://jira.corp.adobe.com/browse/CORE-39185

src/core/createConsent.js
```diff
@@ -0,0 +1,273 @@
...
+    onChange: processConsentStatus
+  });
+
+  const initialConsentStatus = consentState.getStatus();
+
+  if (initialConsentStatus.consented === undefined) {
+    logger.warn("Some commands may be delayed until the user opts in.");
+  } else if (!initialConsentStatus.consented) {
+    logger.warn(
+      `Some commands may be fail. ${initialConsentStatus.deniedConsentReason}`
```
wgamboa: may fail

src/core/createConsent.js
owenleej: I get an error message if I don't put {purposes:"all"} in the optOut command. Is it supposed to be all or none? Will opt-out always be all or nothing? Is specifying necessary?

Aaronius: Good questions. For now, you must and can only specify `purposes: "all"`. My understanding is that later this may be broken down into more granular purposes. At that point, you'll be able to do `purposes: "all"` still or control purposes more granularly like this:

```
alloy("optOut", {
  purposes: {
    purposeA: true,
    purposeB: false
  }
});
```

It's also possible that later on we'll allow users to, for lack of a better term, un-opt-out. But for now, un-opting-out isn't supported.

jonsnyder: This speaks to a greater set of race conditions where the server may set a cookie.  This is not just a consent problem.  Perhaps it could be solved in a part of the network object.  You could add a flag to the payload of "willUpdateConsentCookie" and/or "willUpdateIdentityCookie" which would sequence the requests.

src/core/createConsent.js
```diff
@@ -0,0 +1,273 @@
...
+        })
+        .then(() => {
+          // We optimistically use the preferences the user provided right away,
+          // but if the server doesn't agree and sends back a cookie value that
+          // conflicts, we adopt whatever the server said since that ultimately
+          // is what will be used if the user refreshes the page.
+          consentState.refreshFromCookies();
+        })
+        .catch(error => {
+          consentState.refreshFromCookies();
```
jonsnyder: If the consent endpoint is not working correctly, this stops us from collecting data.  This especially becomes a problem in a SPA.  Why do you call refreshFromCookies after the response?  Why would the server set the consent cookie to anything different?

src/core/createConsent.js
```diff
@@ -0,0 +1,273 @@
...
+    // will return a promise that will be resolved once the user has consented
+    // to the specified purposes.
+    whenConsented() {
+      const deferred = defer();
+      deferredsAwaitingConsent.push(deferred);
+      processConsentStatus(consentState.getStatus());
+      return deferred.promise;
+    }
+  };
+};
```
jonsnyder: There is a lot going on in this file.  Breaking these top-level functions into separate files under a `consent` folder may make unit testing easier.

src/components/Identity/createComponent.js
Aaronius: @jfkhoury are you suggesting adding a comment like "If you do anything privacy-sensitive within this lifecycle callback, be sure to call optIn.whenOptedIn first"? If so, it seems like that comment would apply to pretty much the whole Alloy codebase and not just this area of code. In other words, respecting privacy seems like something engineers need to keep in mind regardless of which part of Alloy they're working on and therefore a comment in this particular area of code doesn't really address the larger need to be privacy-conscious. Do you agree or do you still feel like we still need a comment here specifically?

src/core/createConsent.js
Aaronius: @jonsnyder let's try to discuss this during today's standup. I want to make sure I understand what you're describing.

src/core/createConsent.js
Aaronius: Yeah I think so too. I'll do that--probably in a different PR to make the changes more digestible.

src/components/Privacy/index.js
```diff
@@ -13,36 +13,57 @@ governing permissions and limitations under the License.
...
+        }
+
+        const lowerCasePurposes = purposes.toLowerCase();
+
+        if (lowerCasePurposes !== ALL && lowerCasePurposes !== NONE) {
+          throwInvalidOptInPurposesError(purposes);
+        }
+
+        return consent.setOptInPurposes({
+          GENERAL: lowerCasePurposes === ALL
```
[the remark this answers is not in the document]
Aaronius: I've logged this here: https://jira.corp.adobe.com/browse/EXEG-775

Aaronius:
As discussed in standup today, I've logged a few issues to address some of the feedback later:
https://jira.corp.adobe.com/browse/CORE-39219
https://jira.corp.adobe.com/browse/CORE-39218
https://jira.corp.adobe.com/browse/CORE-39217

PR #325, opened 2020-01-03
Simplify configuration validation especially regarding optional, required, and default values. 

jonsnyder:
## Description

The old style of validator declaration was an object with two properties: `defaultValue`, and `validate`.  This was confusing when an optional value was described.  In this case the defaultValue was set to `undefined`. This PR simplifies this by using a function instead of an object.  Here are some examples on how to use this:
```
import { objectOf, string } from "utils/validation";

const configValidator = objectOf({
  // optional string
  prop1: string(),
  // required string
  prop2: string().required(),
  // optional string with default value
  prop3: string().default("my default value")
}).required();

// this will throw an exception if the configCommandOptions are invalid
// if the options are valid, the returned object has the default values applied
const config = configValidator(configCommandOptions);
```

Validators can be chained together to create more complex validators.  For example:

```
import { number, string } from "utils/validation";

// a required integer greater than or equal to 0
const validator1 = number().integer().minimum(0).required();
// an optional non-empty string
const validator2 = string().nonEmpty();
```

## Related Issue

https://jira.corp.adobe.com/browse/CORE-36754

## Motivation and Context

This PR simplifies the validators.  Additionally this improves the re-usability of the validators so that they can be used to validate all of the command options.  Specifically pulling out the logic from `src/core/config/createValidator.js` into the `objectOf` validator allows the logic to be reused for the command options.

## Types of changes

## Checklist:


test/unit/specs/utils/validation/createDefaultValidator.spec.js
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 Adobe. All rights reserved.
+Copyright 2020 Adobe. All rights reserved.
```
Aaronius: FWIW, I've been told by our legal department that updating the year is unnecessary. That policy may have changed or if your IDE is doing it automatically, that's fine with me as well.

jonsnyder: I updated my IDE to use 2020 for new files.  I think the diff on GitHub mixed up the old files versus new files.

jonsnyder:
I added additional comments to chain.js and nullSafeChain.js.  I think those are key to understanding how the validation works.

PR #331, opened 2020-01-14
C2582 - When ECID is available on client, allow all requests to be pr…

shammowla:
## Description

Test C2582: Once the identity cookie is available, requests should go out promptly (in parallel) and should contain the ECID.

### DIFF
--- a/test/functional/specs/desktop/C2582.js
+++ b/test/functional/specs/desktop/C2582.js
@@ -1,0 +1,58 @@
+import { t, Selector, ClientFunction } from "testcafe";
+import createNetworkLogger from "../../helpers/networkLogger";
+import getResponseBody from "../../helpers/networkLogger/getResponseBody";
+import { responseStatus } from "../../helpers/assertions/index";
+import fixtureFactory from "../../helpers/fixtureFactory";
+import testServerUrl from "../../helpers/constants/testServerUrl";
+
+const urlCollector = `${testServerUrl}/test/functional/sandbox/html/alloySdk.html`;
+
+const networkLogger = createNetworkLogger();
+
+fixtureFactory({
+  title:
+    "C2582 - When ECID is available on client, allow all requests to be promptly sent using the ECID",
+  url: urlCollector,
+  requestHooks: [
+    networkLogger.edgeEndpointLogs,
+    networkLogger.gatewayEndpointLogs
+  ]
+});
+
+test.meta({
+  ID: "C2582",
+  SEVERITY: "P0",
+  TEST_RUN: "Regression"
+});
+
+test("Test C2582: Once the identity cookie is available, requests should go out promptly (in parallel) and should contain the ECID.", async () => {
+  await t.click(Selector("#debugEnabled-button"));
+  await t.click(Selector("#event-button"));
+
+  await responseStatus(networkLogger.edgeEndpointLogs.requests, 200);
+  await t.expect(networkLogger.edgeEndpointLogs.requests.length).eql(1);
+
+  const ecidResponse = getResponseBody(
+    networkLogger.edgeEndpointLogs.requests[0]
+  );
+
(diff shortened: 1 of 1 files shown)


test/functional/specs/desktop/C2582.js
```diff
@@ -0,0 +1,58 @@
...
+
+  await t
+    .expect(getCookie())
+    .contains("kndctr_53A16ACB5CC1D3760A495C99_AdobeOrg_identity");
+
+  await t.click(Selector("#debugEnabled-button"));
+  await t.click(Selector("#event-button"));
+
+  await responseStatus(networkLogger.gatewayEndpointLogs.requests, 204);
+  await t.expect(networkLogger.gatewayEndpointLogs.requests.length).eql(1);
```
jfkhoury: Why are we switching between `edgeEndpointLogs` & `gatewayEndpointLogs`?

test/functional/specs/desktop/C2582.js
```diff
@@ -0,0 +1,58 @@
...
+
+  const getCookie = ClientFunction(() => {
+    return document.cookie;
+  });
+
+  await t
+    .expect(getCookie())
+    .contains("kndctr_53A16ACB5CC1D3760A495C99_AdobeOrg_identity");
+
+  await t.click(Selector("#debugEnabled-button"));
```
jfkhoury: I don't think you need to re-configure Alloy since this is the same session. You only need to trigger an event.

test/functional/specs/desktop/C2582.js
```diff
@@ -0,0 +1,58 @@
...
+  ]
+});
+
+test.meta({
+  ID: "C2582",
+  SEVERITY: "P0",
+  TEST_RUN: "Regression"
+});
+
+test("Test C2582: Once the identity cookie is available, requests should go out promptly (in parallel) and should contain the ECID.", async () => {
```
Aaronius: The titles being passed into `fixtureFactory` and `test` (not in just this file but others) seem similar but not the same. What's our intended pattern/logic for choosing a title for each?

test/functional/specs/desktop/C2582.js
```diff
@@ -0,0 +1,58 @@
...
+
+  await responseStatus(networkLogger.edgeEndpointLogs.requests, 200);
+  await t.expect(networkLogger.edgeEndpointLogs.requests.length).eql(1);
+
+  const ecidResponse = getResponseBody(
+    networkLogger.edgeEndpointLogs.requests[0]
+  );
+
+  const stringifyResponse = JSON.stringify(ecidResponse);
+  await t.expect(stringifyResponse).contains("ECID");
```
Aaronius: It seems brittle to be searching the stringified response for `ECID`. Is there a reason we're not using the response body that has been parsed to a JavaScript object and then looking to see if the ECID is provided in the expected location within the object?

test/functional/specs/desktop/C2582.js
Aaronius: The title seems to suggest to me that the test would be as follows:

1. User clicks the event button three times.
2. Only one request goes out.
3. The response contains the ECID and the identity cookie.
4. Request #2 and #3 go out simultaneously with the ECID.

while this seems to be what's actually being tested:

1. User clicks the event button once.
2. One request goes out.
3. The response contains the ECID and the identity cookie.
4. The user clicks the event button again.
5. The second request goes out with the ECID.

I would prefer we test the former rather than the latter.

jonsnyder: > The titles being passed into `fixtureFactory` and `test` (not in just this file but others) seem similar but not the same. What's our intended pattern/logic for choosing a title for each?

If I understood the benefit that test rail gives us, perhaps I could have a better opinion on this.  I think we will make more work for ourselves if we require only one test per file.  More work to make more test rail items, and more work to setup each test in the test files.

PR #332, opened 2020-01-14
C2595 and updated paths for C12411.

shammowla:
C2595 Event command passes the org ID on the request

### DIFF
--- a/test/functional/specs/desktop/C2595.js
+++ b/test/functional/specs/desktop/C2595.js
@@ -24,9 +24,10 @@ test("Test C2595: Event command passes the org ID on the request.", async () => 
   await t.click(Selector("#event-button"));
 
   await responseStatus(networkLogger.edgeEndpointLogs.requests, 200);
-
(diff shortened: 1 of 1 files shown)


test/functional/specs/desktop/C2595.js
```diff
@@ -0,0 +1,32 @@
...
+test("Test C2595: Event command passes the org ID on the request.", async () => {
+  await t.click(Selector("#debugEnabled-button"));
+  await t.click(Selector("#event-button"));
+
+  await responseStatus(networkLogger.edgeEndpointLogs.requests, 200);
+
+  await t.expect(networkLogger.edgeEndpointLogs.requests.length).eql(1);
+
+  const request = networkLogger.edgeEndpointLogs.requests[0].request.body;
+  await t.expect(request).contains("orgId");
```
jfkhoury: I would make this more specific: Let's test that the `orgId` is being passed in the right expected location in the request: `{"meta":{"configOverrides":{"orgId":"...`

PR #335, opened 2020-01-17
TNT-34376 - implementing getDecisions() command and unit tests

ninaceban:
First stab at implementing `getDecisions` command

## Description

Adding `getDecisions` command implementation.
The current assumption is that `event` command will result in an `interact` call that will fetch audiences, decisions, etc. `getDecisions` command is built on top of that, it takes in an `options` parameter. 
`Options` can have:
- scopes - an array of strings
- viewStart - boolean - if we have this flag we'll know that we need to return page-wide scope decisions as well

On every `event:onResponse` , we will save in memory all the decisions by scope, if the decision doesn't have a scope property then it will be treated as a page-wide scope and will be saved under `page_wide_scope` key.
When a `getDecisions` command will be triggered we will return the decisions from the memory by the scope provided. 

- If `viewStart` is passed as well we will include the page-wide scope decisions in the response. 

- If `options` param is empty we will return an empty array.

Follow up task:
We will have to change `event` implementation to add the scopes parameter.
`event` command will have an additional parameter `scopes` that will be an array of strings.

## Related Issue

NONE

## Motivation and Context

Adding `getDecisions` command

## Types of changes

## Checklist:


ninaceban:
@artur-ciocanu, @jfkhoury and @Aaronius this is the first attempt to add `getDecisions`. This is still work in progress, but I wanted to have something out.

src/components/Personalization/index.js
```diff
@@ -10,7 +10,8 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import { isNonEmptyArray } from "../../utils";
+import CONFIG_DOC_URI from "../../constants/docUri";
```
Aaronius: It looks like this will link to the Tracking Events documentation page. Would you want to link to some sort of Retrieving Personalization Decisions page or something like that?

src/components/Personalization/index.js
```diff
@@ -44,35 +49,51 @@ const createCollect = eventManager => {
   };
 };
 
+const validateEvent = (event, logger) => {
+  const warnings = event.validate();
```
Aaronius: I had some thoughts on `event.validate()`. Let's talk about it in our next Alloy standup.

src/components/Personalization/index.js
```diff
@@ -44,35 +49,51 @@ const createCollect = eventManager => {
...
           event.expectResponse();
 
           // For viewStart we try to hide the personalization containers
           hideContainers(prehidingStyle);
+          return;
+        }
+
+        if (scopes) {
+          event.expectResponse();
+          mergeQuery(event, { scopes });
```
Aaronius: If `scopes` will only ever be added to the event when the `getDecisions` command is called, I would recommend running:

```
if (scopes) {
  event.expectResponse();
  mergeQuery(event, { scopes });
}
```

from within the `getDecisions` command down below rather than doing it here and sending the `scopes` to all the components through `lifecycle.onBeforeEvent`. If other components need access to `scopes` for some reason then maybe it still makes sense to send them through `lifecycle.onBeforeEvent`, but I don't know what that use case would be.

src/components/Personalization/index.js
```diff
@@ -93,6 +114,38 @@ const createPersonalization = ({ config, logger, eventManager }) => {
...
+        if (eventMergeId) {
+          assign(xdm, { eventMergeId });
+        }
+
+        event.setUserXdm(xdm);
+        event.setUserData(data);
+
+        validateEvent(event, logger);
+
+        return eventManager.sendEvent(event, { scopes });
```
Aaronius: I believe this would be exposing the full response to the user, which would be too much for the customer to have to sort through and too much public API we would need to maintain. We should sort through the response ourselves, gather the decisions in whatever structure the customer would find most useful, then resolve the promise we give the user with the nicely packaged decision information.

src/components/Personalization/index.js
ninaceban: yes, we will need a special URL for getDecisions command. Do you know how can we create a. adobe.ly URL?

src/components/Personalization/index.js
ninaceban: yes, you are right, thank you

src/components/Personalization/index.js
ninaceban: yes, we'll process and package the response that we'll get back to the customer in a nicely format.

src/components/Personalization/index.js
Aaronius: I think you can just go to https://bitly.com/ and paste in the documentation URL and it will spit out an adobe.ly URL. @liljenback, if the documentation structure shifts around later on, do you know if we can change the target URL that the adobe.ly URL redirects to?

ninaceban:
@Aaronius, @jonsnyder please take a look

src/components/Personalization/index.js
```diff
@@ -17,13 +17,57 @@ import { hideContainers, showContainers } from "./flicker";
...
+
+const storeDecisions = (storage, decisions) => {
+  if (!decisions) {
+    return;
+  }
+
+  const filteredDecisions = {};
+
+  decisions.forEach(decision => {
+    const key = decision.scope || ALL_SCOPES;
```
jonsnyder: This isn't really all of the scopes.  Would DEFAULT_SCOPE make more sense?

src/components/Personalization/index.js
```diff
@@ -93,6 +145,14 @@ const createPersonalization = ({ config, logger, eventManager }) => {
 
         collectClicks(merger, clickedElement, storage);
       }
+    },
+
+    commands: {
+      getDecisions(options = {}) {
+        const { scopes = [] } = options;
```
jonsnyder: Should the default be to get all decisions?  Currently, this will return all decisions with the default scope.  As we talked in standup, should there be a way to return default decisions and another type of decision?

src/components/Personalization/index.js
ninaceban: actually it is neither all scopes neither default, it is a page wide scope,  maybe PAGE_SCOPE?

test/unit/specs/components/Personalization/responsesMock/eventResponses.js
```diff
@@ -0,0 +1,201 @@
...
+        data: {
+          id: "A",
+          content: "Banner A ...."
+        }
+      }
+    ]
+  }
+];
+
+export const SCOPES_DECISIONS_2 = [
```
Aaronius: I found myself bouncing back and forth between the spec and this file to remember what was in `SCOPE_DECISIONS` vs `SCOPE_DECISIONS_2` vs `SCOPE_DECISIONS_3`. It might help to give more descriptive names. For example, in this case, maybe `SCOPES_FOO3_FOO1_DECISIONS` or something like that.

src/core/createEventManager.js
```diff
@@ -73,14 +73,15 @@ export default ({
      */
     sendEvent(event, options = {}) {
       event.setLastChanceCallback(onBeforeEventSendWithLoggedExceptions);
-      const { isViewStart = false } = options;
+      const { isViewStart = false, scopes } = options;
```
Aaronius: I don't think `scopes` will ever be defined unless you change https://github.com/adobe/alloy/pull/335/files#diff-bd7d44476add132a5476024d3975c3f8R59-R61 to pass `scopes` into `eventManager.sendEvent`.

src/components/Personalization/index.js
Aaronius: Just to not confuse anyone, my comment here is no longer applicable since scopes will now be passed through the `event` command.

src/components/Personalization/index.js
liljenback: @Aaronius , it doesn't look like you can update auto-generated links. In any case if documentation links mostly stay within alloy it seems like from a maintenance aspect it would be easier just generating new links and updating alloy. If we feel like these links need to work as well in old libraries we will have to switch to creating named links (e.g. we pick the name adobe.ly/alloy-events) which would enable us to update them if the structure changes.

liljenback: I vote that we'll just create new links if our doc structure change.

Aaronius: What would be the downsides with using named links? How do we create and update them? It seems like being able to update them would be the optimal way to go, all other things being equal.

ninaceban:
@Aaronius and @jonsnyder thank you for your review! I have addressed your comments and suggestions, could you please take a look again? I will add changes related `scopes` to the `event` command in another PR.

src/components/Personalization/index.js
```diff
@@ -11,13 +11,17 @@ governing permissions and limitations under the License.
...
-import { string } from "../../utils/validation";
+import { string, boolean, arrayOf, objectOf } from "../../utils/validation";
 import { initRuleComponentModules, executeRules } from "./turbine";
 import { hideContainers, showContainers } from "./flicker";
 import collectClicks from "./helper/clicks/collectClicks";
 
 const DECISIONS_HANDLE = "personalization:decisions";
 const PAGE_WIDE_SCOPE = "page_wide_scope";
+const GET_DECISIONS_OPTIONS_SCHEMA = {
+  viewStart: boolean().default(null),
```
Aaronius: Could you explain why `viewStart` is `null` by default instead of `false`?

src/components/Personalization/index.js
```diff
@@ -149,9 +133,18 @@ const createPersonalization = ({ config, logger, eventManager }) => {
...
+          if (viewStart) {
+            scopes.push(PAGE_WIDE_SCOPE);
+          }
+
+          return filterDecisions(decisionsStorage, scopes);
+        }
 
-        return filterDecisions(decisionsStorage, scopes);
+        throw new Error(
+          "Invalid getDecisions command options parameter. Please consult the documentation."
```
Aaronius: If we can, let's be explicit here about what the problem is so they might not have to consult the documentation. Something along the lines of "viewStart must set to true or scopes must be defined"

src/utils/groupBy.js
```diff
@@ -0,0 +1,31 @@
...
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+/**
+ * Group an array by a key getter provided
+ * @param {Array} arr Array to iterate over.
+ * @param {*} keyGetter The key getter by which to group.
```
Aaronius: It looks like this should be `@param {Function} keyGetter ...`

src/components/Personalization/index.js
ninaceban: My bad, I was thinking to make a difference when it is passed or not. but making it `false` by default is a better option.

src/components/Personalization/index.js
```diff
@@ -44,46 +79,59 @@ const createCollect = eventManager => {
   };
 };
 
+const validateOptions = options => {
+  const validate = objectOf(GET_DECISIONS_OPTIONS_SCHEMA);
+  const { viewStart, scopes } = validate(options);
+
+  if (viewStart == null && scopes.length === 0) {
```
Aaronius: This needs to change too right, since `viewStart` won't ever be `null`?

ninaceban: this is why I needed to make a difference when `viewStart` is missing. If `viewStart` is missing and `scopes` is empty or missing as well then we should trigger an error. OR we return always an empty array and we never throw an error. I would return an empty array rather than throw an error, what do you think?

src/components/Personalization/index.js
```diff
@@ -93,6 +141,20 @@ const createPersonalization = ({ config, logger, eventManager }) => {
 
         collectClicks(merger, clickedElement, storage);
       }
+    },
+
+    commands: {
+      getDecisions(options = {}) {
+        validateOptions(options);
+
+        const { viewStart, scopes = [] } = options;
```
jonsnyder: if you want to apply the defaults, you need to use the object returned from the validator.

src/components/Personalization/index.js
Aaronius: I think I don't really understand how viewStart is involved with scopes. Maybe you can provide some context here?

Aaronius: And could someone want to get all scopes?

ninaceban: We might have decisions for page-wide scope, those decisions don't have a scope parameter. Or we might have specific scopes decisions.
There will be use cases when the customer will want to getDecisions for "Foo1" scope, or for a page-wide scope, or for both "page-wide" and "Foo1". 
We talked about the use case when the customer will want to get everything he has in the decision storage and we are not going to support it now.

ninaceban: `viewStart` is a flag for page-wide scope decisions

src/utils/groupBy.js
ninaceban: yes, thank you

src/components/Personalization/index.js
ninaceban: yes, thank you

PR #338, opened 2020-01-23
Expect response during ECID migration

Aaronius:
## Description

When we have an ECID that we can migrate, we previously were not calling `payload.expectResponse()`. That means if no other components expected a response, the request would go to the `collect` endpoint instead of the `interact` endpoint and the identity cookie wouldn't be set.

I talked with @jfkhoury about this and we decided it would be best to expect a response in this case so that the identity cookie gets set on the first request+response. We also decided that it would be best to hold up other requests in this case as well, since ensuring an identity cookie exists before letting other requests go is probably the most consistent/robust way of managing identity. However, if we have an ECID to migrate, I opted to not make the request go to demdex, since I don't think it provides any particular value (please correct me if this might be wrong!).

## Related Issue
https://jira.corp.adobe.com/browse/CORE-39325

## Motivation and Context
Trying to make Alloy + Konductor communication as robust as possible.

## Types of changes

## Checklist:

- [ ] **I was going to add unit tests for identity, but noticed that JIRA shows that @jileeshs is starting on identity tests and I didn't want to cause a big conflict** I have made any necessary test changes and all tests pass.

### DIFF
--- a/src/components/Identity/createComponent.js
+++ b/src/components/Identity/createComponent.js
@@ -23,7 +23,7 @@ export default (processIdSyncs, config, logger, consent, eventManager) => {
     orgId,
     IDENTITY_COOKIE_KEY
   );
-  let deferredForEcid;
+  let deferredForIdentityCookie;
   const migration = createMigration({ orgId, consent });
   const hasIdentityCookie = () => Boolean(cookieJar.get(identityCookieName));
   const customerIds = createCustomerIds({ eventManager, consent, logger });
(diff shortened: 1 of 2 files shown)


src/components/Identity/createComponent.js
```diff
@@ -61,14 +55,23 @@ export default (processIdSyncs, config, logger, consent, eventManager) => {
...
+          // to the server. Since we have an ECID, it might be reasonable to
+          // allow other requests to go out in parallel, but to make sure
+          // there aren't unforeseen problems, we will hold back other
+          // requests until the identity cookie gets set.
+          addEcidToPayload(payload, ecidToMigrate);
+        } else if (
           config.thirdPartyCookiesEnabled &&
           areThirdPartyCookiesSupportedByDefault(getBrowser(window))
         ) {
+          // If third-party cookies are enabled by the customer and
```
jonsnyder: Can you add a comment here explaining why not to use the third party domain if we already have an ecid?

src/components/Identity/createComponent.js
```diff
@@ -61,14 +55,23 @@ export default (processIdSyncs, config, logger, consent, eventManager) => {
         // new visitor).
         deferredForEcid = defer();
         payload.expectResponse();
```
jonsnyder: So we are expecting a response because we need to get a cookie back?  Does /collect never set cookies, or is it that because we use sendBeacon when calling /collect we cannot get cookies back?  It seems like there are two concerns here 1. whether or not we expect a response (and therefore whether or not to call /interact vs /collect) and 2. whether or not we expect a cookie in return (and therefore whether to use fetch vs sendBeacon).

Aaronius: Good questions.

> So we are expecting a response because we need to get a cookie back?

Yes.

> Does /collect never set cookies, or is it that because we use sendBeacon when calling /collect we cannot get cookies back?

There are a couple cases where first-party cookies can't be set (either by Konductor or by Konductor telling Alloy to write a cookie):

1. When Alloy calls `/collect` when Konductor is on a different domain (e.g., edge.adobedc.net), Konductor can't set cookies on its own and won't send a response body so it can't tell Alloy to write a cookie.
2. When Alloy calls `/collect` using `sendBeacon`.

I brought this up with Rares and he said if we don't have an identity cookie, we need to hit `interact` instead of `collect`. It's an interesting scenario though, because what if Konductor needs to set cookies other than the identity cookie. Alloy isn't aware of all the cookies Konductor might need to set, and really shouldn't be aware. I brought this up with Rares too and he said the best decision we can make right now is to just focus on the identity cookie and hit `interact` endpoint if the identity cookie doesn't exist. This is sufficient for the cookies we have currently. He's not 100% sure this will be sufficient for the future, just 99%.

If you have ideas on how we could improve this, let us know. I can add some more comments around this.

src/components/Identity/createComponent.js
Aaronius: Yes. My reasoning was that we only hit demdex when we need an ECID value and can use the demdex cookie to help retrieve that ECID. Since we already have an ECID, I don't think hitting demdex would provide any value. I sent Rares an additional message just to verify that that's the case. Maybe @jfkhoury can help verify. I'll add a comment though.

src/components/Identity/createComponent.js
jonsnyder: Thank you for the clarification.  My thought here was maybe we should have an "expectCookie()" call on the payload instead of overloading the meaning of expectResponse().  I can see it either way though because you could think of a set-cookie header as part of the response.

Aaronius: > you could think of a set-cookie header as part of the response

Yes, partly this, and also we don't know if the cookie will come as a `set-cookie` header or in the response body (as a `state:store` handle) without looking to see if we're making a request to Konductor through a third-party domain. I thought it would be simpler to just stick with the concept we already have (`expectResponse()`) for both cases.

src/components/Identity/createComponent.js
Aaronius: I talked to Rares about this some more. He said that if we have an ECID to migrate, it would probably still be a good idea to hit demdex because if the demdex cookie isn't already set, demdex will use our ECID to set the demdex cookie. I've made the change and updated the comments to hopefully be clearer about why we're doing the things we're doing.

Aaronius:
@shammowla I keep getting `apt-get install failed` from the CI server. I remember this happening before. Do you know what's going on?

src/components/Identity/createComponent.js
```diff
@@ -23,7 +23,7 @@ export default (processIdSyncs, config, logger, consent, eventManager) => {
     orgId,
     IDENTITY_COOKIE_KEY
   );
-  let deferredForEcid;
+  let deferredForIdentityCookie;
```
jonsnyder: Love the name change!

PR #341, opened 2020-01-24
Limiting event validation to user submitted events

liljenback:
## Related Issue

CORE-39926

## Motivation and Context

Internal events hopefully won't need validation.

## Types of changes

## Checklist:

### DIFF
--- a/src/components/DataCollector/index.js
+++ b/src/components/DataCollector/index.js
@@ -1,4 +1,5 @@
 import { assign } from "../../utils";
+import validateUserEvent from "./validateUserEvent";
 
 /*
 Copyright 2019 Adobe. All rights reserved.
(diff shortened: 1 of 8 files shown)


src/components/DataCollector/index.js
```diff
@@ -47,7 +48,7 @@ const createDataCollector = ({ eventManager, logger }) => {
         event.setUserXdm(xdm);
         event.setUserData(data);
 
-        const warnings = event.validate();
+        const warnings = validateUserEvent(event);
```
Aaronius: Personally, I'd still prefer to validate the user options before they ever get into `event`. I'll try to explain why. Here we're making the assumption that `event.setUserXdm()` doesn't disperse some of the invalid data to some other module and continue to live on even after this call to `validateUserEvent`. We're also kind of implying that it's okay to set invalid data through `event.setUserXdm` or `event.setUserData` as long as something outside the event later calls `event.getUserXdm()` and `event.getUserData()` and validates the data. But who's responsibility is it to do that validation? If it's always the responsibility of the module that set the data on the event in the first place, then it begs the question why the module didn't just validate the data before it was ever set on event. I'd rather we just prevent invalid data from entering the library as early as possible--the idea being that we can stop bad input at the very edge so it doesn't have a chance to leak through the library in any way.

PR #351, opened 2020-02-02
Refactor the functional tests and remove unused sandbox pages.

jfkhoury:
## Description

This is a WIP, but as discussed, we would like to stop using the SPA sandbox along with separate sandbox pages per test, unless it's absolutely necessary.

## Related Issue

https://jira.corp.adobe.com/browse/CORE-40065

## Motivation and Context

It's becoming too confusing to add tests when using the SPA.

## Types of changes

- [ X ] Improvement (non-breaking change which does not add functionality)

## Checklist:

- [ X ] I have made any necessary test changes and all tests pass.


test/functional/specs/desktop/C2581.js
```diff
@@ -22,20 +23,43 @@ test.meta({
...
-  await t.click(Selector("#event-button"));
+  await configureAlloyInstance("alloy", debugEnabledConfig);
 
-  await responseStatus(networkLogger.edgeEndpointLogs.requests, 200);
+  await triggerAlloyEvents();
 
   await t.expect(networkLogger.edgeEndpointLogs.requests.length).eql(1);
 
-  const ecidResponse = getResponseBody(
+  await t.wait(1000);
```
Aaronius: This doesn't need to be addressed right now, but I'd like to see if we can get rid of arbitrary wait times. I don't think I know what this particular `wait` is for though so it's hard to provide ideas. Maybe you could add a comment or maybe I'm just dumb.

test/functional/specs/desktop/C2581.js
```diff
@@ -22,20 +23,43 @@ test.meta({
...
     networkLogger.edgeEndpointLogs.requests[0]
   );
+  const response = networkLogger.edgeEndpointLogs.requests[0].response;
+  const responseHeader = response.headers;
+
+  const stringifiedResponseBody = JSON.stringify(responseBody);
+
+  const hasIdentityCookieInBody = stringifiedResponseBody.includes(
+    identityCookieName
+  );
```
Aaronius: Do you know why we stringify the response body and then search the result rather than just looking in the `responseBody` object?

test/functional/specs/desktop/C2581.js
liljenback: Waiting for those events to resolve would probably do the trick in this case.

test/functional/specs/desktop/C2581.js
jfkhoury: Not really, it should be looked up in the object, I kept the logic as is and refactored only. I'm gonna make that change though, thank you!

PR #354, opened 2020-02-06
Added e2e test HTML page.

Aaronius:
## Description
According to https://wiki.corp.adobe.com/display/DMSArchitecture/E2E+Experience+Edge+Alloy.js and the thread at https://adobedx.slack.com/archives/CG7S5ENS2/p1580990004187300, there is a team that will run e2e tests which essentially just load HTML pages that we provide and then track that the data ended up in platform, etc. See the above links for more information.

## Related Issue
None that I'm aware.

## Motivation and Context
See the above links.

## Types of changes

## Checklist:

### DIFF
--- a/sandbox/public/e2e/index.html
+++ b/sandbox/public/e2e/index.html
@@ -1,0 +1,48 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <title>End-to-end Test Page</title>
+    <meta
+      http-equiv="Content-Security-Policy"
+      content="default-src 'self' *.demdex.net;
+            script-src 'self' 'unsafe-inline' assets.adobedtm.com cdn.tt.omtrdc.net;
+            style-src 'self' 'unsafe-inline' cdn.tt.omtrdc.net;
+            img-src * data:;
+            connect-src 'self' *.alloyio.com *.demdex.net *.adobedc.net"
+    />
+    <!-- prettier-ignore -->
+    <script>
+    !function(n,o){o.forEach(function(o){n[o]||((n.__alloyNS=n.__alloyNS||
+      []).push(o),n[o]=function(){var u=arguments;return new Promise(
+      function(i,l){n[o].q.push([i,l,u])})},n[o].q=[])})}
+    (window,["alloy"]);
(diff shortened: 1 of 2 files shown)


sandbox/public/e2e/index.html
```diff
@@ -0,0 +1,48 @@
...
+            connect-src 'self' *.alloyio.com *.demdex.net *.adobedc.net"
+    />
+    <!-- prettier-ignore -->
+    <script>
+    !function(n,o){o.forEach(function(o){n[o]||((n.__alloyNS=n.__alloyNS||
+      []).push(o),n[o]=function(){var u=arguments;return new Promise(
+      function(i,l){n[o].q.push([i,l,u])})},n[o].q=[])})}
+    (window,["alloy"]);
+  </script>
+    <script src="../alloy.js" async></script>
```
jfkhoury: I think using the Launch extension might be a better option here, especially that the AEP team won't have to make changes to the page often. We lose the ability to test while we're developing though. Thoughts?

Aaronius: Optimally the e2e tests would run against the latest code. With that in mind, they'll run against the latest Alloy standalone code if we keep it as-is, but it won't be won't be testing our Launch extension code in any way. If we were to use the Launch embed code here instead, the e2e test would likely be testing older Alloy extension code--whatever is in the Launch library. To ensure the e2e test is always testing the latest Alloy extension code, we would need to hook up some sort of automation script that watches for when a new Alloy extension version is available in Launch, have the script upgrade the Launch property to the latest Alloy extension version, publish a new library, then run tests. I think that would probably be ideal, but it will require some additional work to get that set up.

Aaronius: It'd be nice to have e2e tests run on both the most recent officially published version of Alloy and the version being prepared to be released (whatever's on master). The Alloy extension equivalent would be to test on both the officially published version of the extension and then also the latest that's in "development" availability.

jfkhoury: Exactly what I’ve been trying to achieve with Alloy and Konductor. But the edge base path and environment conversation is still in progress. 

Let’s discuss the extension approach some more, offline.

PR #356, opened 2020-02-06
CORE-40459 Test C14394: ID migration

wgamboa:
When ID migration is enabled and no identity cookie is found but legacy AMCV cookie is found, the ECID will be sent on the request

## Types of changes

## Checklist:


test/functional/specs/desktop/C14394.js
```diff
@@ -0,0 +1,62 @@
...
+    edgeBasePath: window.edgeBasePath,
+    debugEnabled: true,
+    idMigrationEnabled: true
+  });
+});
+
+test("Test C14394: When ID migration is enabled and no identity cookie is found but legacy AMCV cookie is found, the ECID will be sent on the request", async () => {
+  await apiCalls();
+
+  // NOTE: I assume a click is needed to get a response body. Putting the same event call in apiCalls() results in an undefined response body.
```
Aaronius: When you say "same event call", can you show me what the event call you were trying to put inside `apiCalls` looks like?

Aaronius: Are you talking about:

```
alloy("event", {
  viewStart: true,
  // TODO: Change data to match final XDM schema.
  xdm: {
    key: "value"
  }
}
```

test/functional/specs/desktop/C14394.js
```diff
@@ -0,0 +1,62 @@
...
+    .eql("16908443662402872073525706953453086963");
+
+  const response = JSON.parse(
+    getResponseBody(networkLogger.edgeEndpointLogs.requests[0])
+  );
+
+  await t.expect(response.handle[0].type).eql("identity:result");
+  await t
+    .expect(response.handle[0].payload[0].id)
+    .eql("16908443662402872073525706953453086963");
```
Aaronius: This is asserting that the first payload of the first handle will have the identity result. But if payloads or handles are in a different order, we really shouldn't fail the test, because it's still a proper Konductor response and everything will still behave properly. Instead, I would recommend we search through `response.handle` looking for any handles with the `identity:result` type. Then inside those handles we look for the first payload where `namespace.code` is `ECID`. Then check that the ID on that payload is correct. I suspect that searching the handles and their payloads will come up a lot, so it would be useful to create a method like `getPayloadsByType` just like we did in the Alloy source itself: https://github.com/adobe/alloy/blob/51d0f23286d11a6aff20597ad384a93a94dfeb6c/src/core/createResponse.js#L35-L39

test/functional/specs/desktop/C14394.js
```diff
@@ -0,0 +1,62 @@
+import { t, Selector, ClientFunction } from "testcafe";
+import createNetworkLogger from "../../helpers/networkLogger";
+import getResponseBody from "../../helpers/networkLogger/getResponseBody";
+import { responseStatus } from "../../helpers/assertions/index";
+import fixtureFactory from "../../helpers/fixtureFactory";
+import testServerUrl from "../../helpers/constants/testServerUrl";
+
+const urlCollector = `${testServerUrl}/test/functional/sandbox/html/alloySdk.html`;
```
jfkhoury: Hey Willi, let's please not use this SPA anymore. Let's switch to the format I have in my PR or in C8118.

test/functional/specs/desktop/C14394.js
wgamboa: I wasn't returning the event call promise in t.eval(). That was the problem.

wgamboa:
@jfkhoury I wanted to have a standalone utility that is not a sub function of createaResponse.

Aaronius:
@jfkhoury I think you're suggesting to use src/core/createResponse.js, is that right? I think that would be fine.

PR #360, opened 2020-02-10
Tnt 34831 - Render decisions functionality

ninaceban:
Render Decisions
## Description

At `event:onResponse` we will get an array of decisions. We will filter decision items that have DOM action schema. We will apply hiding styles if there is a prehiding selector, apply item action and remove prehiding styles. When decision was rendered successfully we will send a notification with the decisionId that was rendered.
## Types of changes

## Checklist:


src/components/Personalization/index.js
```diff
@@ -10,20 +10,20 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import { isNonEmptyArray, groupBy, values } from "../../utils";
+import { isNonEmptyArray, groupBy, values, assign } from "../../utils";
 import { string, boolean, arrayOf, objectOf } from "../../utils/validation";
-import { initRuleComponentModules, executeRules } from "./turbine";
+import { initDomActionsModules, executeActions } from "./turbine";
 import { hideContainers, showContainers } from "./flicker";
 import collectClicks from "./helper/clicks/collectClicks";
```
artur-ciocanu: May be it makes sense to add another constant like `ALL_SCHEMAS` that can be imported instead of wildcard import plus using `values()`.

Aaronius: I was the one that suggested the wildcard import with `values()`. The downside of `ALL_SCHEMAS` constant, at least if it's in the same file as the constants it contains, is you have to maintain the array. On the other hand, you don't have to call `values()` everywhere you want to use all the values. I don't feel feel strongly either way. I'll bring it up in the engineering checkin and see if we can quickly reach consensus.

Aaronius: We didn't get to it in the call, so I would recommend just picking one and running with it for now. We can change it later if we come to a consensus.

ninaceban: I'll leave it as it is, thank you

PR #379, opened 2020-03-10
CORE-39098 ECID/Alloy race condition causing mismatched demdex ids and ECIDs

wgamboa:
## Types of changes

## Checklist:


wgamboa:
Solution direction pending review by @jfkhoury and Justin.

src/components/Identity/createComponent.js
```diff
@@ -23,17 +25,78 @@ export default (processIdSyncs, config, logger, consent, eventManager) => {
     orgId,
     IDENTITY_COOKIE_KEY
   );
+  let deferredForOptInOld;
```
jfkhoury: I will review in more details asap but I think this whole logic should be in the `migration` module, and it should only run when migration is turned on.

src/components/Identity/createComponent.js
```diff
@@ -23,17 +25,78 @@ export default (processIdSyncs, config, logger, consent, eventManager) => {
...
+      }, true);
+    };
+
     if (!hasIdentityCookie()) {
-      if (deferredForIdentityCookie) {
+      if (doesVisitorExist && !deferredForOptInOld && !deferredForVisitorECID) {
+        if (isObject(window.adobe) && isObject(window.adobe.optIn)) {
+          const optInOld = window.adobe.optIn;
+
+          deferredForOptInOld = defer();
```
Aaronius: I would recommend not using `defer` for this case. Use a regularly constructed Promise where possible. In Aaron's Opinion™, a deferred should only be used when the thing resolving the promise is a task that the creator of the promise did not initiate. This is usually rare, resulting in deferreds often being abused and causing more problems than they try to solve (in fact, for a while you could create a deferred natively in some browsers until it was made obsolete).

Here, this code that's creating the promise is also the initiator of `optInOld.fetchPermissions`, so constructing a promise normally is most appropriate. I think you'll find your code to be simpler. I would start by breaking down each step in this process (1. fetching permissions and 2. Getting the visitor ID) into separate functions, each returning their own promise. Something like this:

```
const awaitVisitorOptIn = () => {
  return new Promise(resolve => {
    if (isObject(window.adobe) && isObject(window.adobe.optIn)) {
      const optInOld = window.adobe.optIn;
      optInOld.fetchPermissions(() => {
        if (optInOld.isApproved([optInOld.Categories.ECID])) {
          logger.log(
            "Received legacy opt in approval to let Visitor retrieve ECID from server."
          );

          resolve();
        }
      }, true);
      logger.log(
        "Delaying request while waiting for legacy opt in to let Visitor retrieve ECID from server."
      );
    } else {
      resolve();
    }
  });
};

const getVisitorECID = () => {
  return awaitVisitorOptIn().then(() => {
    return new Promise(resolve => {
      logger.log(
        "Delaying request while using Visitor to retrieve ECID from server."
      );

      const visitor = Visitor.getInstance(orgId, {});

      visitor.getMarketingCloudVisitorID(ecid => {
        if (!ecid) {
          return;
        }

        payload.expectResponse();
        addEcidToPayload(payload, ecid);

        if (
          areThirdPartyCookiesSupportedByDefault(getBrowser(window))
        ) {
          payload.useIdThirdPartyDomain();
        }

        logger.log(
          "Resuming previously delayed request that was waiting for ECID from Visitor."
        );

        resolve();
      }, true);
    });
  });
};
```

And put the functions inside the migration module. Then, like @jonsnyder said (and @jfkhoury eluded to), change `getEcidFromLegacy` to return promise. Then make `getEcidFromLegacy` leverage the two functions above. Let me know if that doesn't make sense.

src/components/Identity/createComponent.js
```diff
@@ -68,10 +68,10 @@ export default (processIdSyncs, config, logger, consent, eventManager) => {
         // the browser.
         payload.expectResponse();
 
-        if (ecidToMigrate) {
-          // We have an ECID, but we still want to establish an
-          // identity cookie before allowing other requests to be sent.
-          addEcidToPayload(payload, ecidToMigrate);
+        if (ecidToMigratePromise) {
+          promise = ecidToMigratePromise.then(ecidToMigrate => {
```
jonsnyder: Can you re-add the comment?

src/components/Identity/createMigration.js
```diff
@@ -10,15 +10,65 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+      });
+    });
+  };
 
   return {
+    getEcidFromLegacy() {
+      const ecid = this.getEcidFromLegacyCookies();
+
+      if (ecid) {
+        return new Promise(resolve => {
```
jonsnyder: You can just do
```
return Promise.resolve(ecid);
```

src/components/Identity/createMigration.js
```diff
@@ -10,15 +10,65 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+          resolve(ecid);
+        });
+      }
+
+      if (doesVisitorExist) {
+        return getVisitorECID();
+      }
+      return new Promise(resolve => {
```
jonsnyder: You can return something falsey here (false, null, or undefined) and then the addEcidToPayload won't run.

Aaronius:
I did a little pair programming with @wgamboa and pushed a commit with our changes.

Aaronius:
@jfkhoury Willi mentioned you might be demoing this to Justin? There's a bunch of testing work remaining to be done on this, so if there's a chance things could change, it would be good to find out sooner rather than later.

jfkhoury:
No more changes coming; test away! :)

src/components/Identity/visitorService/awaitVisitorOptIn.js
```diff
@@ -0,0 +1,35 @@
...
+        if (optInOld.isApproved([optInOld.Categories.ECID])) {
+          const message =
+            "Received legacy opt in approval to let Visitor retrieve ECID from server.";
+
+          logger.log(message);
+          resolve(message);
+        }
+      }, true);
+    } else {
+      resolve("Legacy opt in object does not exist.");
```
Aaronius: Hey @wgamboa, I noticed we're resolving these promises with a string, but we don't do anything with that resolved value. Can you explain why we don't just resolve the promise without the string?

wgamboa: Hey @Aaronius, I use it for the unit tests. I don't know how else to go about it.

wgamboa: I just remembered you advising me to avoid strings in the code. Maybe a numerical value?

Aaronius: I don't think a string or a number provides any value here. If you test:

1. The promise is rejected if `adobe.optIn` doesn't exist.
1. The promise is resolved if `adobe.optIn` does exist and the user approves.
1. The promise is rejected if `aodbe.optIn` does exist and the user disapproves. I think? Should it be rejected if the user disapproves? It doesn't seem to be resolved or rejected at this point.

...then I don't think there's any value in checking a resolution _value_.

wgamboa: 1. I don't know what Minimum version of Visitor we should support. Should it have optin? If that's not required, then I think this promise should be resolved with undefined.
3. Thanks for the catch. This should be rejected with undefined

test/unit/specs/components/Identity/visitorService/awaitVisitorOptIn.spec.js
```diff
@@ -0,0 +1,79 @@
...
+      );
+    });
+  });
+
+  describe("Legacy opt in object is present and gives approval", () => {
+    it("should return promise resolved with undefined", () => {
+      window.adobe = {
+        optIn: {
+          fetchPermissions(callback) {
+            setTimeout(callback, 10);
```
Aaronius: I recommend using a timeout of `0` here (and the test below) as anything more than `0` doesn't provide any advantage. Even a timeout of `0` will still cause the callback to be called asynchronously (on a different JavaScript frame/event loop).

src/components/Identity/visitorService/awaitVisitorOptIn.js
```diff
@@ -0,0 +1,35 @@
...
+        "Delaying request while waiting for legacy opt in to let Visitor retrieve ECID from server."
+      );
+      optInOld.fetchPermissions(() => {
+        if (optInOld.isApproved([optInOld.Categories.ECID])) {
+          logger.log(
+            "Received legacy opt in approval to let Visitor retrieve ECID from server."
+          );
+          resolve();
+        }
+        reject();
```
Aaronius: In the case where the user declined consent in Visitor, we're rejecting this promise which, if we follow the rejection up the promise chain, is going to cause the command that initiated the process to be rejected, I believe. Usually when we reject promises like this, we reject it with an error instance and a message explaining what happened. That would mean in this case we would do something like:

```
reject(new Error("Legacy opt-in was declined blah blah...");
```

Once that's in place, test it out manually in the sandbox to see if the command gets rejected:

```
alloy("event", {...}).catch(function(error) {
  // error.message should be the message you provided.
});
```

Then, if you haven't already, write up test cases in TestRail for all these scenarios so that we can write e2e tests for them.

src/components/Identity/visitorService/awaitVisitorOptIn.js
```diff
@@ -0,0 +1,35 @@
...
+  return new Promise((resolve, reject) => {
+    if (isObject(window.adobe) && isObject(window.adobe.optIn)) {
+      const optInOld = window.adobe.optIn;
+      logger.log(
+        "Delaying request while waiting for legacy opt in to let Visitor retrieve ECID from server."
+      );
+      optInOld.fetchPermissions(() => {
+        if (optInOld.isApproved([optInOld.Categories.ECID])) {
+          logger.log(
+            "Received legacy opt in approval to let Visitor retrieve ECID from server."
```
Aaronius: "opt in" is a phrasal verb (e.g., "The user is going to opt in."). We want the adjective or noun in this message here and the message a few lines above, so we should say "opt-in".

src/components/Identity/visitorService/getVisitorECID.js
```diff
@@ -0,0 +1,36 @@
...
+            "Resuming previously delayed request that was waiting for ECID from Visitor."
+          );
+          resolve(ecid);
+        }, true);
+      });
+    });
+  }
+
+  return Promise.resolve();
+};
```
Aaronius: This file should be called `getVisitorEcid` according to our camelCasing standards.

src/components/Identity/createMigration.js
```diff
@@ -10,33 +10,52 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import { getApexDomain, cookieJar } from "../../utils";
+import { getApexDomain, cookieJar, isFunction } from "../../utils";
+import getVisitorECID from "./visitorService/getVisitorECID";
+import awaitVisitorOptIn from "./visitorService/awaitVisitorOptIn";
```
Aaronius: Notice that when we get around to writing unit tests for createMigration.js, we'll need to inject these dependencies instead of importing them here, so we can mock them in the tests. In other words, all imports (except for utilities), will be in `Identity/index.js` and then injected down through the other Identity modules. We'll tackle that later though.

sandbox/public/index.html
```diff
@@ -58,6 +58,9 @@
         configId: "9999999",
         orgId: "53A16ACB5CC1D3760A495C99@AdobeOrg",
         debugEnabled: true,
+        idMigrationEnabled: !(
+          location.href.indexOf("idMigrationEnabled=false") >= 0
+        ),
```
Aaronius: Did you intend to leave this here? If so, you can use the `getUrlParameter` function to get the value of the `idMigrationEnabled` query param.

PR #382, opened 2020-03-19
Org id cleanup - Throw error if no ID set

jfkhoury:
## Description

1. Stop sending the ORG ID and sync validation flags
- There doesn't seem to be any use case for legitimately wanting to override a config's org ID. It would require the customer to also override a bunch of other stuff in the config, like datasetId, client code, etc., since those are all tied to their owning org.
- The org ID override isn't actually overriding the org ID right now anyway.

2. Use the final QE ORG and Config IDs in functional tests.
3. Throw an error if after receiving the first response we still don't have an identity cookie.

## Related Issue

https://jira.corp.adobe.com/browse/CORE-42551
https://jira.corp.adobe.com/browse/CORE-42550

## Types of changes

## Checklist:


src/components/Identity/createComponent.js
```diff
@@ -133,8 +133,16 @@ export default (processIdSyncs, config, logger, consent, eventManager) => {
...
-          deferredForIdentityCookie.resolve();
+        if (deferredForIdentityCookie) {
+          if (hasIdentityCookie()) {
+            deferredForIdentityCookie.resolve();
+          } else {
+            // This logic assumes that the code setting the cookie is working as expected and that
+            // the cookie was missing from the response.
+            const noIdentityCookieErrorMsg = `An identity was not set properly. Please verify that the org ID ${orgId} 
+            configured in Alloy matches the org ID specified in the edge configuration.`;
+            return deferredForIdentityCookie.reject(noIdentityCookieErrorMsg);
```
Aaronius: We've been rejecting promises with error instances for consistency.

src/core/createEventManager.js
```diff
@@ -32,23 +30,6 @@ export default ({
     }
   };
 
-  const addMetaTo = payload => {
-    const configOverrides = { orgId };
-
-    const dataCollection = Object.create(null);
-    assignIf(
-      dataCollection,
-      { synchronousValidation: true },
```
Aaronius: Why are we removing synchronous validation? Is thing something that we plan on enabling through the experience cloud debugger?

test/functional/specs/C14407.js
```diff
@@ -23,15 +24,13 @@ const setConsentIn = ClientFunction(() => {
...
   await cookies.clear();
   // await apiCalls(imsOrgId);
 
   const configure = await alloyEvent("configure", {
-    configId: "9999999",
-    orgId: imsOrgId,
-    debugEnabled: true,
-    idMigrationEnabled: false
+    idMigrationEnabled: false,
+    ...debugEnabledConfig
```
Aaronius: It looks like we should probably be using `debugEnabledConfig` in a bunch of the other tests as well. We can do that later though.

src/components/Identity/createComponent.js
Aaronius: It seems like the promise rejection is always unhandled, even when I try to handle it from sandbox/public/index.html. For example:

```
alloy("event", {
        viewStart: true,
        scopes: ["alloy-location-1", "alloy-location-2"],
        xdm: {
          // Demonstrates overriding automatically collected data
          device: {
            screenHeight: 1
          }
        }
      }).then(function(data) {
        console.log("Sandbox: View start event has completed.", data);
      }).catch(error => {
        console.log(error);
        debugger;
      });
```

We need to figure out why the promise that's handed back to customers isn't being rejected under this scenario and add an e2e test for it. Want my help with this?

jfkhoury: Yes please @Aaronius  :) If you can make the time.

src/core/createEventManager.js
jfkhoury: Yes, just sent out the invite for next week. But this wasn't working anyway, similar to how ORG ID override wasn't being applied.

src/components/Identity/createComponent.js
Aaronius: 👍 K, I'll hop on it when I'm done with my current task.

jfkhoury: Merci.

src/components/Identity/createComponent.js
```diff
@@ -133,8 +133,16 @@ export default (processIdSyncs, config, logger, consent, eventManager) => {
...
         // cookie at this point, but we check just to be sure.
-        if (deferredForIdentityCookie && hasIdentityCookie()) {
-          deferredForIdentityCookie.resolve();
+        if (deferredForIdentityCookie) {
+          if (hasIdentityCookie()) {
+            deferredForIdentityCookie.resolve();
+          } else {
+            // This logic assumes that the code setting the cookie is working as expected and that
+            // the cookie was missing from the response.
+            const noIdentityCookieErrorMsg = `An identity was not set properly. Please verify that the org ID ${orgId}
```
jonsnyder: Thanks for adding this! I need to add a similar check for the consent cookie.

src/components/Identity/createComponent.js
```diff
@@ -113,6 +113,24 @@ export default (processIdSyncs, config, logger, consent, eventManager) => {
...
+            deferredForIdentityCookie.resolve();
+          } else {
+            // This logic assumes that the code setting the cookie is working as expected and that
+            // the cookie was missing from the response.
+            const error = new Error(`An identity was not set properly. Please verify that the org ID ${orgId} 
+            configured in Alloy matches the org ID specified in the edge configuration.`);
+            deferredForIdentityCookie.reject(error);
+            return Promise.reject(error);
+          }
+        }
```
Aaronius: I moved this code up because I don't think we should create a legacy cookie if we don't have a legit identity cookie. I also changed this:

```
const noIdentityCookieErrorMsg = `An identity was not set properly. Please verify that the org ID ${orgId} 	
            configured in Alloy matches the org ID specified in the edge configuration.`;	
return deferredForIdentityCookie.reject(noIdentityCookieErrorMsg);
```

to this:

```
const noIdentityCookieError = new Error(
  `An identity was not set properly. Please verify that the org ID ${orgId} configured in Alloy matches the org ID specified in the edge configuration.`
);
deferredForIdentityCookie.reject(noIdentityCookieError);
return Promise.reject(noIdentityCookieError);
```

This is so that both the command that triggered the request fails with the error _and_ commands whose requests were queued also fail with the error. I believe this was the intention previously, but `deferredForIdentityCookie.reject` doesn't return anything, which is why the initial command wasn't failing properly.

BTW, this also exposed another bug that I logged: https://jira.corp.adobe.com/browse/CORE-42828

Aaronius:
I added an e2e test in TestRail for this: https://testrail.corp.adobe.com/index.php?/cases/view/32449

Aaronius:
@jfkhoury @jonsnyder could one of you review the commits I added before we merge?

jfkhoury:
All good thank you @Aaronius !!!

jonsnyder:
:+1

Aaronius:
Looks like we're getting an error from bundlesize-store.now.sh on the CI servers? @shammowla or @owenleej, have you seen this?

test/functional/helpers/assertions/responseStatus.js
```diff
@@ -1,19 +1,17 @@
...
+    // TODO: It looks like test cafe might not be handling gzip correctly.
+    if (!req.response) {
+      // eslint-disable-next-line no-await-in-loop
+      await t
         .expect(req.response.statusCode === status)
         .ok(`expected ${status} to be found in ${req.response.statusCode}`);
-    })
-  );
+    }
+  }
```
Aaronius: TestCafe was complaining about this. We're supposed to await each testcafe call individually so that order is maintained.

PR #385, opened 2020-03-19
Creating getEcid command

jileeshs:
Implementing `getEcid` command.
## Description
New `getEcid` which would return an `Ecid` if available, if not it would make a call to `identity/acquire` end point. 

Handling three scenarios.
 - Returning when alloy already has an `ecid` in memory
 - Waits for `deferredForIdentityCookie` state to complete
 - Calls `identity/acquire` when no `ecid` is available and not waiting for an identityCookie

## Related Issue
https://jira.corp.adobe.com/browse/EXEG-1068
## Types of changes

## Checklist:

### DIFF
--- a/src/components/Identity/configurePayload.js
+++ b/src/components/Identity/configurePayload.js
@@ -1,0 +1,39 @@
+import getBrowser from "../../utils/getBrowser";
+import areThirdPartyCookiesSupportedByDefault from "../../utils/areThirdPartyCookiesSupportedByDefault";
+import ecidNamespace from "../../constants/ecidNamespace";
+
+const addEcidToPayload = (payload, ecid) => {
+  payload.addIdentity(ecidNamespace, {
+    id: ecid
+  });
+};
+export default (
+  payload,
+  migration,
+  idMigrationEnabled,
+  thirdPartyCookiesEnabled
+) => {
+  const ecidToMigrate =
+    idMigrationEnabled && migration.getEcidFromLegacyCookies();
+  if (ecidToMigrate) {
+    // We have an ECID, but we still want to establish an
+    // identity cookie before allowing other requests to be sent.
+    addEcidToPayload(payload, ecidToMigrate);
+  }
+
+  if (
+    thirdPartyCookiesEnabled &&
+    areThirdPartyCookiesSupportedByDefault(getBrowser(window))
+  ) {
+    // If third-party cookies are enabled by the customer and
+    // supported by the browser, we will send the request to a
+    // a third-party identification domain that allows for more accurate
+    // identification of the user through use of a third-party cookie.
+    // If we have an ECID to migrate, we still want to hit the
+    // third-party domain because the third-party identification domain
+    // will use our ECID to set the third-party cookie if the third-party
+    // cookie isn't already set, which provides for better cross-domain
+    // identification for future requests.
+    payload.useIdThirdPartyDomain();
+  }
+};
--- a/src/components/Identity/createComponent.js
+++ b/src/components/Identity/createComponent.js
@@ -1,24 +1,20 @@
 import createCustomerIds from "./customerIds/createCustomerIds";
-import {
-  defer,
-  cookieJar,
-  getNamespacedCookieName,
-  areThirdPartyCookiesSupportedByDefault,
-  find
-} from "../../utils";
+import { defer, cookieJar, getNamespacedCookieName, find } from "../../utils";
 import { IDENTITY_COOKIE_KEY } from "../../constants/cookieDetails";
-import getBrowser from "../../utils/getBrowser";
 import createMigration from "./createMigration";
 import ecidNamespace from "../../constants/ecidNamespace";
+import getEcidCommand from "./getEcidCommand";
+import configurePayload from "./configurePayload";
 
-const addEcidToPayload = (payload, ecid) => {
-  payload.addIdentity(ecidNamespace, {
-    id: ecid
-  });
-};
-
-export default (processIdSyncs, config, logger, consent, eventManager) => {
-  const { idMigrationEnabled, orgId } = config;
+export default (
+  processIdSyncs,
+  config,
+  logger,
+  consent,
+  eventManager,
+  sendEdgeNetworkRequest
+) => {
+  const { idMigrationEnabled, orgId, thirdPartyCookiesEnabled } = config;
   const identityCookieName = getNamespacedCookieName(
     orgId,
     IDENTITY_COOKIE_KEY
@@ -27,6 +23,7 @@ export default (processIdSyncs, config, logger, consent, eventManager) => {
   const migration = createMigration({ orgId, consent });
   const hasIdentityCookie = () => Boolean(cookieJar.get(identityCookieName));
   const customerIds = createCustomerIds({ eventManager, consent, logger });
+  let ecid;
 
   // TO-DOCUMENT: We wait for ECID before trigger any events.
   const accommodateIdentityOnRequest = payload => {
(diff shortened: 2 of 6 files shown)


src/components/Identity/createComponent.js
```diff
@@ -150,6 +129,27 @@ export default (processIdSyncs, config, logger, consent, eventManager) => {
...
+      getEcid() {
+        if (ecid) {
+          return ecid;
+        }
+        if (deferredForIdentityCookie) {
+          // Should execute only after the first call resloves
+          return accommodateIdentityOnRequest().then(() => ecid);
+        }
+        if (!hasIdentityCookie()) {
+          // Should defer all calls if an
```
Aaronius: Need to finish the comment.

src/components/Identity/createComponent.js
```diff
@@ -150,6 +129,27 @@ export default (processIdSyncs, config, logger, consent, eventManager) => {
...
+          deferredForIdentityCookie = defer();
+        }
```
Aaronius: My biggest concern is that understanding how the deferred and `accommodateIdentityOnRequest` and `getEcidCommand` all coordinate together is getting somewhat complex. I would need to dig into it for a while to see how we can simplify things, but it's on my mind. I don't think it needs to hold up the PR.

src/components/Identity/getEcidCommand/createIdentityPayload.js
```diff
@@ -0,0 +1,18 @@
+import createRequestPayload from "../../../core/edgeNetwork/requestPayloads/createRequestPayload";
+
+export default () => {
+  return createRequestPayload(content => {
+    return {
+      addIdentity: (namespaceCode, identity) => {
+        content.identityMap = content.identityMap || {};
+        content.identityMap[namespaceCode] =
+          content.identityMap[namespaceCode] || [];
+        content.identityMap[namespaceCode].push(identity);
```
Aaronius: I see this was duplicated from `core/edgeNetwork/requestPayloads/createDataCollectionRequestPayload.js`. I think it would be good to create a module at `core/edgeNetwork/requestPayloads/contentModifiers/addIdentityFactory.js`. The module would export a function that could be used from both here and `core/edgeNetwork/requestPayloads/createDataCollectionRequestPayload.js` like this:

`addIdentity: createAddIdentity(content),`

src/components/Identity/getEcidCommand/createIdentityPayload.js
```diff
@@ -0,0 +1,18 @@
...
+      },
+      addQuery: () => {
+        content.query = content.query || {};
+        content.query.identity = { fetch: ["ECID"] };
+      }
```
Aaronius: If the query to fetch the identity is always required for this type of request, which I believe it is, then I think it would be best to avoid ever being able to create a request payload without the query. In other words, remove `addQuery` from here and enforce that `content.query.identity` is always set by just moving this code:

```
content.query = content.query || {};
content.query.identity = { fetch: ["ECID"] };
```
to before [this line](https://github.com/adobe/alloy/pull/385/files#diff-2b6fe5249e78819450208158631fd544R5).

src/components/Identity/index.js
```diff
@@ -16,12 +16,25 @@ import configValidators from "./configValidators";
...
   });
-  return createComponent(processIdSyncs, config, logger, consent, eventManager);
+  return createComponent(
+    processIdSyncs,
+    config,
+    logger,
+    consent,
+    eventManager,
+    sendEdgeNetworkRequest
+  );
```
Aaronius: This is a lot of arguments. Will you please change it to an object?

src/components/Identity/configurePayload.js
```diff
@@ -0,0 +1,39 @@
+import getBrowser from "../../utils/getBrowser";
+import areThirdPartyCookiesSupportedByDefault from "../../utils/areThirdPartyCookiesSupportedByDefault";
+import ecidNamespace from "../../constants/ecidNamespace";
+
+const addEcidToPayload = (payload, ecid) => {
+  payload.addIdentity(ecidNamespace, {
+    id: ecid
+  });
+};
+export default (
```
jfkhoury: It would be so cool to convert this module to a factory, since `migration, idMigrationEnabled & thirdPartyCookiesEnabled` are constants. The only thing that change is the `payload`. So a function taking the first 3 args that returns a function that takes the `payload`, and you can pass this second function around instead of passing those 4 args like on line 64 & 146 in `createComponent.js`

src/components/Identity/createComponent.js
```diff
@@ -150,6 +129,27 @@ export default (processIdSyncs, config, logger, consent, eventManager) => {
...
+        return getEcidCommand(
```
Aaronius: I think this will cause a race condition. If the getEcid request is sent, then a data collection request is sent before the getEcid response is returned from the server, then the user will be identified as two different users (two different ECIDs will have been created). I'm working on a refactor of Identity and I'm trying to take this into account.

src/components/Identity/configurePayload.js
```diff
@@ -0,0 +1,39 @@
...
+  payload,
+  migration,
+  idMigrationEnabled,
+  thirdPartyCookiesEnabled
+) => {
+  const ecidToMigrate =
+    idMigrationEnabled && migration.getEcidFromLegacyCookies();
```
Aaronius: The master branch no longer has `migration.getEcidFromLegacyCookies()`. Now it's `getEcidFromLegacy`, which returns a promise, which will cause race conditions if you're not careful.

src/components/Identity/configurePayload.js
```diff
@@ -0,0 +1,39 @@
...
+    // If third-party cookies are enabled by the customer and
+    // supported by the browser, we will send the request to a
+    // a third-party identification domain that allows for more accurate
+    // identification of the user through use of a third-party cookie.
+    // If we have an ECID to migrate, we still want to hit the
+    // third-party domain because the third-party identification domain
+    // will use our ECID to set the third-party cookie if the third-party
+    // cookie isn't already set, which provides for better cross-domain
+    // identification for future requests.
+    payload.useIdThirdPartyDomain();
```
Aaronius: It's very possible we already have an identity cookie. If we already have an identity cookie, there's no need to hit the third-party domain.

src/components/Identity/configurePayload.js
```diff
@@ -0,0 +1,39 @@
...
+  idMigrationEnabled,
+  thirdPartyCookiesEnabled
+) => {
+  const ecidToMigrate =
+    idMigrationEnabled && migration.getEcidFromLegacyCookies();
+  if (ecidToMigrate) {
+    // We have an ECID, but we still want to establish an
+    // identity cookie before allowing other requests to be sent.
+    addEcidToPayload(payload, ecidToMigrate);
+  }
```
Aaronius: This `configurePayload` module is used for the `getEcid` payload (as well as other payloads), but if we already have an ecid to migrate (`ecidToMigrate` exists), then there's no need to issue a getEcid request in the first place.

Aaronius:
Hey @jileeshs I've been going through this deeper as I restructure the identity component and there are race conditions that will be really difficult to solve without the restructure I'm working on. I would recommend you hold off on additional work. I'll take some of the concepts from this PR and work them into my PR with a simple way to implement the `getEcid` command. Feel free to close this PR or just leave it open until we figure out if we want to just build on top of the PR I'll be opening.

Aaronius:
@jileeshs I recommend we close this PR and build on top of https://github.com/adobe/alloy/pull/392. I've added a slot for you to make the ECID request: https://github.com/adobe/alloy/pull/392/files#diff-e3608b4e8d26a7ee90affbadd009744eR39-R48

PR #386, opened 2020-03-19
Consenting to no purposes should be persisted

alshaw:
## Description
1. Configure Alloy (you don't need to set default consent--by default it is "in").
2. Set consent to "out".
3. Reload the page.
4. Configure Alloy (you don't need to set default consent--by default it is "in").
5. Try to send an event.
6. Check that the promise from the command is rejected and check the error message to make sure it was due to the user being opted out.

## Related Issue
https://github.com/adobe/alloy/pull/353

## Motivation and Context

Testing that consenting to no purposes should be persisted

## Types of changes

- [X ] Improvement (non-breaking change which does not add functionality)

## Checklist:

- [X ] I have signed the [Adobe Open Source CLA](https://opensource.adobe.com/cla.html) or I'm an Adobe employee.


test/functional/specs/C14409.js
```diff
@@ -0,0 +1,96 @@
...
+  // expect that konductor cookie handle matches cookie name
+  await t.expect(cookieName in cookiesToSet).ok();
+
+  await t.expect(errorMessage).ok("Expected the event command to be rejected");
+  await t.expect(errorMessage).contains("The user declined consent.");
+
+  // expect that the cookie max age is greater than 0
+  await t.expect("maxAge" in cookiesToSet[cookieName]).ok();
+  await t.expect(cookiesToSet[cookieName].maxAge).gt(0);
+});
```
Aaronius: I think there are a couple things missing here:

1. We should be testing that `maxAge` is set to `15552000` (I didn't have this number when I wrote the TestRails tests). Ideally, we would be able to check the maxAge of the actual cookie (rather than what's in the edge response), because we'd like to test that Alloy actually writes the cookie using whatever maxAge was in the response. I don't know if this is actually possible with TestRail though. It's not normally possible with JavaScript in a regular browser. If it's not possible in TestRail, then checking the the maxAge in the response like you're doing here is probably the best we can do. Let's check that the value is actually `15552000` though.
1. We should be testing that the cookie `domain` is set to the apex (top-level) domain. Again, TestCafe may not let us do this by inspecting the cookie directly. You can't using JavaScript in a regular browser. However, we can still check that it's set using the apex domain if we navigate to a different subdomain before executing the event command. If the consent cookie is set on the apex domain, then the cookie will be accessible on the second subdomain and respected by Alloy. If the cookie were _not_ set on the apex domain, then the cookie would _not_ be accessible on the second domain and therefore not respected by Alloy. I'm not sure what the status is of being able to navigate to different subdomains within e2e tests, so I've posted a message in Slack.

test/functional/specs/C14409.js
```diff
@@ -0,0 +1,96 @@
...
+  await t.expect(bodyObject.handle.length).gt(0);
+
+  await bodyObject.handle.map(handle => {
+    if (handle.type !== "state:store") return handle;
+    handle.payload.map(payload => {
+      cookiesToSet[payload.key] = payload;
+      return payload;
+    });
+    return handle;
+  });
```
Aaronius: I think we should be using `createResponse(response).getPayloadsByType("store:handle");` then do a `reduce` to build the `cookiesToSet` object:

```
const storePayloads = createResponse(response).getPayloadsByType("store:handle");
const cookiesToSet = storePayloads.reduce((memo, storePayload) => {
  memo[storePayload.key] = storePayload;
  return memo;
, {});
```

It's possible to use a `forEach` instead of a `reduce` for building the `cookiesToSet` object like so:

```
const cookiesToSet = {};
storePayloads.forEach(storePayload => {
  cookiesToSet[storePayload.key] = storePayload;
}
```

but we had a team discussion about which pattern we wanted to follow for consistency and `reduce` received the majority support.

You'll probably want to update C14407.js with whatever you end up doing here.

Aaronius:
Yeah, thanks for bringing that up. I started to write a similar comment previously but then waffled on it because I wasn't sure how I felt about it myself. This is also probably my fault, since I wrote the test case description in TestRail and said to check the cookie value. https://testrail.corp.adobe.com/index.php?/cases/view/14409&group_by=cases:section_id&group_order=asc&group_id=237

I think what you're trying to portray is that the cookie and its value is somewhat of an implementation detail. Similarly, how the cookie gets set is also an implementation detail (the fact that it comes through the response payload and Alloy writes the cookie, in this case). I think that's largely true and probably something we don't need to test for.

The one thing that this approach would not test is the `maxAge` of the consent cookie that gets set. If you refresh the page and try to send an event and it gets rejected because the user is opted out, then that would verify that the cookie's max age is at least larger than 0, otherwise the cookie would no longer exist. On the other hand, it wouldn't test that the max age is `15552000`. Unless TestCafe offers a way to determine the max age of a cookie, the only way to do that would be to look at the `state:store` payloads on the consent response. So I think that part is a matter of whether we prefer (a) looking at the `state:store` payload to see what `maxAge` was set to or (b) being okay with only knowing that `maxAge` is greater than 0.

I for sure think we need to send the event _after_ a page reload, but preferably we load the page on a different subdomain, so that we can ensure that the cookie is accessible across subdomains.

Does this make sense?

test/functional/specs/C14409.js
alshaw: Saw your slack message. It looks like there aren't subdomains on alloyio.com for testing yet. Do you know if there's a way to simulate it locally?

Aaronius: You could probably mimic it on your machine by manually editing your `/etc/hosts` file, but I think @shammowla is working on something where you wouldn't have to do that. I wouldn't bother with it yet or let it hold up your PR. In your PR I would just reload the current page, then after we get subdomain support for our testing purposes, we can come back and update this test to navigate to a page on the different subdomain rather than just reloading the current page.

alshaw: Sounds good! I'll go with reload for now and update the `maxAge` assertion. Thanks!

test/functional/specs/C14409.js
```diff
@@ -0,0 +1,96 @@
...
+
+  const errorMessage = await t.eval(() =>
```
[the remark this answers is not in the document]
alshaw: I honestly was just following examples in other tests. In the TestCafe docs it looks like they work pretty similarly, except that with ClientFunctions you can reuse them.

test/functional/specs/C14409.js
```diff
@@ -0,0 +1,106 @@
...
+  await t.expect("handle" in request).ok();
+  await t.expect(request.handle.length).gt(0);
+
+  const storePayloads = createResponse(request).getPayloadsByType(
+    "state:store"
+  );
+  const cookiesToSet = storePayloads.reduce((memo, storePayload) => {
+    memo[storePayload.key] = storePayload;
+    return memo;
+  }, {});
```
Aaronius: Hey @alshaw, thanks for making these updates. We probably should have written down our decisions or concrete steps after our team discussion. I believe we decided that we weren't going to inspect the response payload or look at the cookies at all because they are implementation details only. This means you can remove the code I've highlighted here and any of the cookie checking below. I think it makes sense to remove the first `alloyEvent` in this test too. Here's how I see the test going:

1. Configure Alloy (you don't need to set default consent--by default it is "in").
1. Set consent to "out".
1. Reload the page.
1. Configure Alloy (you don't need to set default consent--by default it is "in").
1. Try to send an event.
1. Check that the promise from the command is rejected and check the error message like you did below to make sure it was due to the user being opted out.

Similarly, for C14407, I see it going like this:

1. Configure alloy with default consent set to "pending".
1. Set consent to "in".
1. Reload the page.
1. Configure alloy with default consent set to "pending".
1. Try to send an event.
1. Check that the promise from the command is resolved.

For both tests, I don't think there's any need to check the actual networks requests, network responses, or cookies. Does that make sense? I'm happy to work through this with you if you'd like. Just hit me up on Slack.

alshaw: Ah okay, I misunderstood. I'll update C14407 and C14409 to reflect these steps.

test/functional/specs/C14409.js
```diff
@@ -0,0 +1,65 @@
...
+  // [TODO] Navigate to a different subdomain when it is available
+  // https://github.com/DevExpress/testcafe/blob/a4f6a4ac3627ebeb29b344ed3a1793627dd87909/docs/articles/documentation/test-api/actions/navigate.md
+  await t.eval(() => document.location.reload());
+
+  const reconfigure = await alloyEvent("configure", {
+    configId: "9999999",
+    orgId: imsOrgId,
+    debugEnabled: true,
+    idMigrationEnabled: false
+  });
```
Aaronius: I recommend taking this moving this into a function above the test. Something like this:

```
const configureAlloy = async () => {
  const configure = await alloyEvent("configure", {
    configId: "9999999",
    orgId: imsOrgId,
    debugEnabled: true,
    idMigrationEnabled: false
  });
  await configure.promise;
});
```

Then inside the test in the two places where you configure Alloy, just do `await configureAlloy()`;

test/functional/specs/C14409.js
Aaronius: Thanks @alshaw! C14409 looks good. It looks like C14407 still needs to be updated though.

test/functional/specs/C14409.js
Aaronius: Actually, I don't think you need the `await configure.promise` either. Just this:

```
const configureAlloy = () => {
  return alloyEvent("configure", {
    configId: "9999999",
    orgId: imsOrgId,
    debugEnabled: true,
    idMigrationEnabled: false
  });
});
```

test/functional/helpers/getConsentCookieName.js
```diff
@@ -0,0 +1,3 @@
+export default imsOrgId => {
```
jfkhoury: Nice

Aaronius:
Looks like you have a conflict to resolve, but after that go ahead and merge it!

Aaronius:
Hey @alshaw, do you need help finishing off this PR?

alshaw:
> Hey @alshaw, do you need help finishing off this PR?

Just fixed the merge conflicts. Hopefully that's the last change that needs to be made on these tests

Aaronius:
Ship it! It looks like there's a separate test failing, but we can address it as a separate effort. Hit the merge button when you're ready.

PR #392, opened 2020-03-27
CORE-42757 Restructure identity for unit testing. Call interact except when document may unload.

Aaronius:
## Description

The initial goal of this effort was to restructure Identity so that it could be properly unit tested and to provide a more robust and simple structure for integrating a getEcid command (like @jileeshs has been working on in https://github.com/adobe/alloy/pull/385). To make this refactor as clean as possible, several other architectural changes needed to be made, most of which have been discussed as a team through Slack, tech check-ins, and standups. Here are the important bits:

### Calls `interact` for all data collection requests unless document may unload

Currently, Alloy is totally in control of whether a response should be sent by Konductor and we default to calling `collect` unless a component actively requires a response. This can potentially result in the first call being an `interact` call (because we need an identity) and all remaining calls being `collect` calls if no components require a response. This can be problematic, because Konductor needs to send back things like ID syncs and destinations on occasion (and potentially other things in the future). If we rarely request a response from the server, then Konductor doesn't have a chance to provide this type of information to Alloy. We discussed this as a team and the best proposal was to hit `interact` for every request unless the document may be unloading (a user clicked an exit link, for example).

(https://jira.corp.adobe.com/browse/CORE-43144)

### Add onResponse and onRequestFailure params to onBefore* lifecycle methods

There's an inherent relationship between a invocation of onBeforeEvent, onBeforeDataCollection, and onBeforeConsentRequest (things that are going to the server) and the response that comes through onResponse or onRequestFailure lifecycle methods (things that are coming back from the server). Because the request handling is found in different lifecycle methods from the response handling, it becomes difficult to robustly tie a response to its request. This leads to:

(1) Complex code. For example, identity must use a deferred because it returns a promise from onBeforeDataCollectionResponse or onBeforeConsentRequest and the promise must be resolved by code inside onResponse. This gets even more complex once we start introducing the getEcid command.

(2) Buggy code. For example, Personalization currently assumes that the first response is the response to an isViewStart request (I think). This is not guaranteed, however, and is therefore a race condition that could cause containers to be shown prematurely.

Now, two functions are passed into the onBefore* lifecycle methods: `onResponse` and `onRequestFailure`. You could call each function to register a callback.

The reason we would pass onResponse and onRequest as functions instead of passing promises to the components, is that doing a `.then` on the promises from within the components would cause a situation where errors would go unhandled, rather than bubbling up the error to the initiating command.

https://jira.corp.adobe.com/browse/CORE-42818

### Added `onBeforeRequest` and removed `onBeforeConsentRequest` lifecycle methods

While we had `onBeforeDataCollectionRequest` and `onBeforeConsentRequest`, it was problematic not having a general `onBeforeRequest` lifecycle method for any request that may be sent to the edge. For example, Identity hooked into `onBeforeDataCollectionRequest` and `onBeforeConsentRequest` to ensure that if we didn't have an identity yet, that only the first request could be sent to the server and subsequent requests were queued until the response to the first request came back. But what if in the future a different component started issuing a new type of request that had its own `onBeforeNewThingRequest` lifecycle? If the author of the component wasn't aware of how Identity behaved, they wouldn't know that they need to update the Identity component to hook into `onBeforeNewThingRequest` as well. This could result in the new type of request being sent to the server when it should have been queued by the Identity component. By exposing an `onBeforeRequest` lifecycle method, enforcing that it gets called from `sendEdgeNetworkRequest`, and hooking into it from Identity instead of hooking into `onBeforeDataCollectionRequest` and `onBeforeConsentRequest`, then any request that uses `sendEdgeNetworkRequest` (which should be all requests) will be properly queued by Identity.

I ended up eliminated `onBeforeConsentRequest`. I don't foresee any reason that a component would want to modify consent requests specifically. Technically, at this point we could also remove `onBeforeDataCollectionRequest` because no components are using it anymore either, but I suspect this is one that's more likely to be used by components in the future. If we find that it remains not used, we can remove it as well.

### More accurately handle Konductor responses with unsuccessful error codes

When Konductor is running as expected, it will always return a JSON body in the typical structure (handle, errors, warnings) regardless of which status code it returns, with the exclusion of an HTTP status code 204 No Response.

In other words, even if we get a status code 500 back, the response body may still contain "handles" that components should process.

As such, the definition between success and failure gets very blurry. Afterall, Konductor is an aggregator of success and failures from the downstream components. To accommodate, I've adjusted the response handling code. Now with the changes in this PR, if the response body is either (1) parseable JSON or (2) empty and the HTTP status code is a 204 No Response (like from a `collect` call), then the `onResponse` lifecycle method will be called. Otherwise, one of several things could have happened:

1. The request never made it to the server (e.g., no internet connection).
1. The request never made it to Konductor (e.g., JAG or load balancers failed).
1. Konductor is borked.

In this case, the `onResponse` lifecycle method will not be called, but instead the `onRequestFailure` lifecycle method will be called.

Even though the `onResponse` lifecycle method gets called, this does NOT mean that the promise returned to the customer from the command they were trying to execute gets resolved. This is because the response body may contain errors within it (the Konductor response can contain both `handles` and `errors`). After the `onResponse` lifecycle is called, if any error is found in the response, `sendEdgeNetworkRequest` will throw an error, which should reject the promise that was returned to the customer from the command they were trying to execute.

https://jira.corp.adobe.com/browse/CORE-43149

* In retrospect, I might have been able to get await with doing this change in another PR. 🤷‍♂ 

### Other notes

I discovered an issue related to ID migration because we can only request the ECID value from `interact` and `identity/acquire`, whereas we cannot through other types of requests (e.g., consent), even though those other types of requests may mint an ECID and set an identity cookie for the user. More details here: https://jira.corp.adobe.com/browse/EXEG-1234

## Related Issue
https://jira.corp.adobe.com/browse/CORE-42757
https://jira.corp.adobe.com/browse/CORE-42818
https://jira.corp.adobe.com/browse/CORE-43144
https://jira.corp.adobe.com/browse/CORE-43149
https://jira.corp.adobe.com/browse/CORE-43150

## Types of changes

## Checklist:


test/functional/specs/C2581.js
```diff
@@ -78,6 +78,8 @@ test.requestHooks(mockWithoutResponse)(
   }
 );
 
+// FIXME: Need to figure out how to check that the response from the first
+// request comes back before the other two requests go out.
```
jonsnyder: I did something similar here: https://github.com/adobe/alloy/blob/master/test/functional/specs/C14414.js

see `SequentialHook`.

src/components/Identity/createComponent.js
```diff
@@ -1,165 +1,51 @@
-import createCustomerIds from "./customerIds/createCustomerIds";
```
jfkhoury: Love this change!

src/core/createLifecycle.js
```diff
@@ -12,12 +12,22 @@ governing permissions and limitations under the License.
 
 // TO-DOCUMENT: Lifecycle hooks and their params.
 const hookNames = [
+  // Called after all components have been registered.
```
jfkhoury: Thank you for the docs!

src/core/edgeNetwork/requestPayloads/createDataCollectionRequestPayload.js
```diff
@@ -25,14 +24,7 @@ export default () => {
       },
       addEvent(event) {
         content.events = content.events || [];
-        expectResponse = expectResponse || event.getExpectResponse();
```
jfkhoury: Nice!!

Aaronius:
I spent some time manually testing things, particularly around waiting for identity and consent. Things seems to be working fine.

One thing I wanted to mention is that the unit tests for `createMigration.js` need to be improved (modules injected into it and properly mocked), which I've logged as a separate issue: https://jira.corp.adobe.com/browse/CORE-43419

jfkhoury:
> I spent some time manually testing things, particularly around waiting for identity and consent. Things seems to be working fine.
> 
> One thing I wanted to mention is that the unit tests for `createMigration.js` need to be improved (modules injected into it and properly mocked), which I've logged as a separate issue: https://jira.corp.adobe.com/browse/CORE-43419

Thank you!

jfkhoury:
@Aaronius Are you comfortable merging this PR at this point?

Aaronius:
I am, but technically I haven't had two people review the commits I made 5 days ago.

Aaronius:
C2660 is failing, but is supposed to be addressed in https://github.com/adobe/alloy/pull/397

PR #393, opened 2020-03-27
CORE-42341 Adding common validation support for all component commands

liljenback:
## Description

All commands can now supply a option validation function that will be run in core and provides common error or warning messages.

## Related Issue

CORE-42341

## Motivation and Context

Provides a standard way to implement command option validation.

## Types of changes

## Checklist:


src/components/DataCollector/index.js
```diff
@@ -13,62 +13,57 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
 
-import CONFIG_DOC_URI from "../../constants/docUri";
+import EVENT_COMMAND_DOC_URI from "../../constants/docUri";
 
-const createDataCollector = ({ eventManager, logger }) => {
+const createDataCollector = ({ eventManager }) => {
   return {
     commands: {
-      event(options) {
-        const { errors, warnings } = validateUserEventOptions(options);
```
jfkhoury: I have to think about this PR a little more, but from a first glance, and for the sake of keeping Components commands' API simple, I'd rather we convert the `validation message` into some sort of a template / util, and each component will use it to throw its errors if any, instead of changing the API of the commands and making it more verbose.

I'm open to discussing it though.

liljenback: If a component throws the error it is captured and presented as an error. The component command does not need to return errors and warnings. See the Personalization component as an example.

src/core/validateCommandOptions.js
```diff
@@ -0,0 +1,39 @@
...
+  const invalidOptionsMessage = `Invalid ${commandName} command options:\n\t - ${validationProblems.join(
+    "\n\t - "
+  )}\nFor command documentation see: ${documentationUri}`;
+  if (errors.length) {
+    throw new Error(invalidOptionsMessage);
+  }
+  if (warnings.length) {
+    logger.warn(invalidOptionsMessage);
+  }
+};
```
Aaronius: It looks like we're complicating this code and code within `validateUserEventOptions.js` to support warnings systemically and I'm not sure it's worth the cost. Let's discuss this and @jfkhoury's comment at standup.

src/core/validateCommandOptions.js
```diff
@@ -0,0 +1,39 @@
...
+  try {
+    const validationResult = command.optionsValidator(options);
+    errors = validationResult.errors || errors;
+    warnings = validationResult.warnings || warnings;
+  } catch (e) {
+    errors.push(e);
+  }
+  const validationProblems = errors.length > 0 ? errors : warnings;
+  const { commandName, documentationUri = COMMAND_DOC_URI } = command;
+  const invalidOptionsMessage = `Invalid ${commandName} command options:\n\t - ${validationProblems.join(
```
Aaronius: If only warnings exist, then the options should be considered "invalid".

Aaronius: Sorry, I meant to say _should not_.

src/components/DataCollector/index.js
```diff
@@ -13,62 +13,57 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import CONFIG_DOC_URI from "../../constants/docUri";
+import EVENT_COMMAND_DOC_URI from "../../constants/docUri";
```
Aaronius: Move out of core?

src/core/validateCommandOptions.js
```diff
@@ -13,27 +13,20 @@ governing permissions and limitations under the License.
...
+    documentationUri = COMMAND_DOC_URI,
+    optionsValidator
+  } = command;
+  let validatedOptions = options;
+  if (optionsValidator) {
+    try {
+      validatedOptions = optionsValidator({ options, logger });
+    } catch (validationError) {
+      const invalidOptionsMessage = `Invalid ${commandName} command options:\n\t - ${validationError}
+      \nFor command documentation see: ${documentationUri}`;
```
Aaronius: Because this is a template literal, I believe this will result in two line breaks after `validationError`. One would be due to the line break after `${validationError}` and another from `\n`. I would recommend putting the template literal on a single line:

```
const invalidOptionsMessage = `Invalid ${commandName} command options:\n\t - ${validationError}\nFor command documentation see: ${documentationUri}`;
```

src/core/executeCommandFactory.js
```diff
@@ -49,8 +49,12 @@ export default ({ logger, configureCommand, debugCommand, handleError }) => {
...
                 );
               }
-              validateCommandOptions({ command, options, logger });
-              return command.run(options);
+              const validatedOptions = validateCommandOptions({
+                command,
+                options,
+                logger
+              });
+              return command.run(validatedOptions);
```
Aaronius: Please update the related `executeCommandFactory` specs.

test/unit/specs/core/validateCommandOptions.spec.js
```diff
@@ -15,43 +15,39 @@ import validateCommandOptions from "../../../../src/core/validateCommandOptions"
...
-    };
+  it("supports commands not implementing command options validation.", () => {
     expect(() => {
-      validateCommandOptions({ command, options, logger });
-    }).toThrowError();
+      validateCommandOptions({ command, options });
+    }).not.toThrowError();
   });
-  it("should throw exception if error provided by command options validator.", () => {
+  it("should throw exception if command options validator throw exception.", () => {
```
Aaronius: validator throw exception -> validator throws exception

src/components/DataCollector/index.js
```diff
@@ -13,15 +13,13 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
 const createDataCollector = ({ eventManager }) => {
   return {
     commands: {
       event: {
-        documentationUri: EVENT_COMMAND_DOC_URI,
-        optionsValidator: options => {
-          return validateUserEventOptions(options);
+        documentationUri: "https://adobe.ly/2r0uUjh",
+        optionsValidator: ({ options, logger }) => {
+          return validateUserEventOptions({ options, logger });
```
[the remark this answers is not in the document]
liljenback: No, I should be able to use the other one.

test/unit/specs/core/executeCommandFactory.spec.js
```diff
@@ -101,6 +101,30 @@ describe("executeCommandFactory", () => {
...
+    };
+    const configureCommand = () => Promise.resolve(componentRegistry);
+    const executeCommand = executeCommandFactory({
+      logger,
+      configureCommand,
+      handleError
+    });
+    executeCommand("configure");
+    executeCommand("test").then(() => {
+      expect(runCommandSpy.toHaveBeenCalledWith("option"));
```
Aaronius: I believe this should be failing because `option` !== `options`. I think the test isn't waiting for your promise because you're not returning it. Hot tip: always check that your test can fail before making it pass.

liljenback: Hah, I actually changed that to force fail it to verify that it worked

test/unit/specs/core/executeCommandFactory.spec.js
```diff
@@ -101,6 +101,30 @@ describe("executeCommandFactory", () => {
...
+    const configureCommand = () => Promise.resolve(componentRegistry);
+    const executeCommand = executeCommandFactory({
+      logger,
+      configureCommand,
+      handleError
+    });
+    executeCommand("configure");
+    return executeCommand("test").then(() => {
+      expect(runCommandSpy).toHaveBeenCalledWith("options");
+    });
```
Aaronius: Hey, sorry to keep asking for changes, but I noticed we're skipping a fairly significant part here. `validateCommandOptions` should be injected into `executeCommandFactory` instead of imported. Then, `validateCommandOptions` should be mocked. Then the test should assert that:

1.  `validateCommandOptions` was called with `command` and `options`.
2.  `runCommandSpy` has been called with whatever the mock `validateCommandOptions` returns.

Without this, there's no way of verifying that:

1. `validateCommandOptions` is being called and that it's receiving what it's expecting.
1. We're passing the results from `validateCommandOptions` to the command (which is an important part of ensuring the command will receive options with their default values set).

PR #398, opened 2020-04-07
 returning an object in the event command 

ninaceban:
## Description

Adjusted `sendEdgeNetworkRequestFactory` to agregate into an object all the returned values from :

- `onResponse` as a function passed to `onBeforeRequest`
- `onResponse` as a lifecycle

Make sure to return `noop` in components that used to do `return ...` just for the sake of promise chaining.

## Types of changes

## Checklist:


src/components/ActivityCollector/attachClickActivityCollector.js
```diff
@@ -24,7 +24,7 @@ const createClickHandler = (eventManager, lifecycle) => {
           return Promise.resolve();
         }
 
-        return eventManager.sendEvent(event);
+        return eventManager.sendEvent(event).then(noop);
```
Aaronius, line 27: It might be worth adding a comment explaining why we're doing `.then(noop)`. I'm not sure if other developers will understand.

src/components/Personalization/index.js
```diff
@@ -140,6 +140,9 @@ const createPersonalization = ({ config, logger, eventManager }) => {
         showContainers();
 
         storeDecisions(decisionsStorage, decisions);
+
+        // eslint-disable-next-line consistent-return
```
Aaronius, line 144: Rather than disabling the linting, change your code at the top of `onResponse` to this:

```
if (authoringModeEnabled) {
  return undefined;
}
```

Whether the team wants to continue with this `consistent-return` lint rule or not is something I'll try to bring up at a tech checkin.

src/core/edgeNetwork/sendEdgeNetworkRequestFactory.js
```diff
@@ -94,7 +116,11 @@ export default ({
           // be thrown here which should ultimately reject the promise that
           // was returned to the customer for the command they executed.
           processWarningsAndErrors(response);
-          return response;
+
+          const first = processReturnValue(...segmentedReturnValues[0]);
+          const second = processReturnValue(...segmentedReturnValues[1]);
```
Aaronius, line 121: Maybe call these `lifecycleOnResponseReturnValues` and `consumerOnResponseReturnValues` instead of `first` and `second`?

src/core/edgeNetwork/sendEdgeNetworkRequestFactory.js
```diff
@@ -12,7 +12,23 @@ governing permissions and limitations under the License.
...
+} from "../../utils";
+
+const notNil = value => !isNil(value);
+
+const processReturnValue = values => {
+  if (isNil(values)) {
+    return {};
+  }
+
+  return values.filter(notNil).reduce((acc, item) => assign(acc, item), {});
```
Aaronius, line 30: I'm not understanding why this is necessary. I ran this little prototype in my browser console:

image

which seems to suggest you could do this down below:

```
const lifecycleOnResponseReturnValues = segmentedReturnValues[0] || [];
const consumerOnResponseReturnValues = segmentedReturnValues[1] || [];
return assign({}, ...lifecycleOnResponseReturnValues, ...consumerOnResponseReturnValues);
```

Did you try something like that?

src/components/ActivityCollector/attachClickActivityCollector.js
ninaceban: yes, you are right. I'll add a comment here and in other places where I have added `then(noop)`.

src/components/Personalization/index.js
ninaceban: ok, thank you.

src/core/edgeNetwork/sendEdgeNetworkRequestFactory.js
ninaceban: yes, sure

ninaceban:
@Aaronius thank you for your review, I have added your suggestions and some other changes to 'sendEdgeNetworkRequestFactory'. Please take a look. Thank you

src/components/ActivityCollector/attachClickActivityCollector.js
Aaronius: Nice. It may be good to be explicit about what you mean with "we are not returning a value". Specifically, we're trying to avoid returning a value _to the customer_. Consider something like:

```
// eventManager.sendEvent() will return a promise resolved to an
// object and we want to avoid returning any value to the customer
```

Whatever you think would most help other developers.

src/core/edgeNetwork/sendEdgeNetworkRequestFactory.js
```diff
@@ -81,21 +81,25 @@ export default ({
...
-          // server may ALSO containing errors. Because of this, we make
-          // sure we call lifecycle onResponse methods, then later
-          // process the warnings and errors.
-          // If there are errors in the response body, an error will
-          // be thrown here which should ultimately reject the promise that
-          // was returned to the customer for the command they executed.
-          processWarningsAndErrors(response);
-          return response;
-        });
+            return assign({}, ...returnValues[0], ...returnValues[1]);
```
Aaronius, line 101: This looks a lot better.

Do you mind adding a comment describing what this is all about? Something like:

```
// Merges all returned objects from all `onResponse` callbacks into
// a single object that can later be returned to the customer.
```

src/components/Audiences/index.js
```diff
@@ -22,7 +22,9 @@ const createAudiences = ({ logger }) => {
     lifecycle: {
       onResponse({ response }) {
         const destinations = response.getPayloadsByType("activation:push");
-        return processDestinations(destinations);
+        // we're now returning an object at onResponse
+        // here we need to add 'then(noop)' because we are not returning a value
+        return processDestinations(destinations).then(noop);
```
Aaronius, line 27: Hmm....why is the promise returned by processDestinations being resolved with a value of any kind? Rather than doing a `.then(noop)` here, this seems like something that should be fixed within `processDestinations`.

ninaceban: yes, actually we don't need it here

src/core/edgeNetwork/sendEdgeNetworkRequestFactory.js
ninaceban: yes sure. thank you

Aaronius:
FYI, as your PR stands right now, you'll get errors when you try to set consent:

image

That's because when sendEdgeNetworkRequest is used to send consent requests, `runOnResponseCallbacks` doesn't return an array.

ninaceban:
@Aaronius I have fixed those errors, and also adjusted one test. I'll have to add more tests to test different use cases.

src/components/Personalization/index.js
```diff
@@ -87,7 +87,7 @@ const createCollect = eventManager => {
 
     mergeMeta(event, meta);
 
-    eventManager.sendEvent(event);
+    // eventManager.sendEvent(event);
```
jonsnyder, line 90: Did you mean to leave this commented out?

src/core/edgeNetwork/sendEdgeNetworkRequestFactory.js
```diff
@@ -12,7 +12,20 @@ governing permissions and limitations under the License.
...
+  assign,
+  isNil,
+  isNonEmptyArray
+} from "../../utils";
+
+const notNil = value => !isNil(value) || isNonEmptyArray(value);
+const merge = (acc, value) => {
+  return Array.isArray(value) ? value.reduce(merge, acc) : assign(acc, value);
+};
+const mergeValues = values => values.filter(notNil).reduce(merge, {});
```
jonsnyder, line 28: mergeValues should be defined in a separate file so that we can do easy extensive testing.  It could be a util.

test/unit/specs/core/createLifecycle.spec.js
```diff
@@ -33,10 +33,16 @@ describe("createLifecycle", () => {
     });
   });
 
-  it("calls all callbacks for a given lifecycle method", () => {
+  fit("calls all callbacks for a given lifecycle method", () => {
```
jonsnyder, line 36: fit

src/components/Personalization/index.js
ninaceban: yes, otherwise we'll enter a loop. I'll fix this in next PR.

src/core/edgeNetwork/sendEdgeNetworkRequestFactory.js
ninaceban: ok, I'll extract that in another util

test/unit/specs/core/createLifecycle.spec.js
ninaceban: my bad, thank you

src/components/Audiences/processDestinationsFactory.js
```diff
@@ -34,7 +34,7 @@ const processUrls = (fireReferrerHideableImage, logger, destinations) => {
           logger.error(createResultLogMessage(urlDestination, false));
         });
     })
-  );
+  ).then(noop);
```
Aaronius, line 37: @ninaceban Would you please update the tests accordingly to assert that we're not resolving the promise with a value?

src/components/ActivityCollector/attachClickActivityCollector.js
Aaronius: @ninaceban Would you please update the tests accordingly to assert that we're not resolving the promise with a value?

src/components/Identity/customerIds/createCustomerIds.js
```diff
@@ -61,7 +62,9 @@ export default ({ eventManager, consent, logger }) => {
         const event = eventManager.createEvent();
         return consent.awaitConsent().then(() => {
           // FIXME: Konductor shouldn't require an event.
-          return eventManager.sendEvent(event);
+          // we're now returning an object at onResponse
+          // here we need to add 'then(noop)' because we are not returning a value
+          return eventManager.sendEvent(event).then(noop);
```
Aaronius, line 67: @ninaceban Would you please update the tests accordingly to assert that we're not resolving the promise with a value?

src/utils/merge.js
```diff
@@ -0,0 +1,25 @@
...
+ * Returns an object containing keys from all passed in values
+ * @param {*} values
+ * @returns {Object}
+ */
+export default values => {
+  if (isNonEmptyArray(values)) {
+    return values.reduce((acc, e) => assign(acc, e), {});
+  }
+  return {};
+};
```
Aaronius, line 25: @ninaceban My understanding is that this does basically the same thing:

```
assign({}, ...values);
```

except that it doesn't do the `isNonEmptyArray(values)` check, which shouldn't be necessary where you were using it anyway (https://github.com/adobe/alloy/pull/398/commits/604c60a5381d1104476f608ece3bc7161ca8d010#diff-f13c18241f022f5285cf29e5d815b506R109-R110). Did you try that and it didn't work or is this one of those doesn't-work-in-IE things?

test/unit/specs/core/edgeNetwork/sendEdgeNetworkRequestFactory.spec.js
```diff
@@ -414,14 +414,14 @@ describe("sendEdgeNetworkRequestFactory", () => {
 
   it("returns the merged object from  lifecycle::onBeforeRequest & lifecycle::onResponse", () => {
     lifecycle.onBeforeRequest.and.callFake(({ onResponse }) => {
-      onResponse(() => [{ a: 1 }, undefined, [], {}]);
+      onResponse(() => [{ a: 1 }, { b: 1 }, undefined]);
```
Aaronius, line 417: I wouldn't expect a component to return an array of objects from the `onResponse` callback. I'm either misunderstanding (very likely) or something is wrong.

Aaronius: I took a deeper look. The callback passed into `onResponse` should be returning a single object only. This test should really be failing as it is.

If you're trying to test that objects returned from the `onResponse` callbacks from multiple components get merged together properly, you would do something like this instead:

```
lifecycle.onBeforeRequest.and.callFake(({ onResponse }) => {
  onResponse(() => ({ a: 1 }));
  onResponse(() => ({ b: 1 })]);
  onResponse(() => undefined]);
});
```

test/unit/specs/core/edgeNetwork/sendEdgeNetworkRequestFactory.spec.js
```diff
@@ -392,4 +392,36 @@ describe("sendEdgeNetworkRequestFactory", () => {
...
+      return Promise.resolve();
+    });
+
+    return expectAsync(
+      sendEdgeNetworkRequest({ payload, action })
+    ).toBeResolvedTo({ a: 1 });
+  });
+
+  it("returns the object from lifecycle::onResponse", () => {
+    lifecycle.onResponse.and.returnValue([{ a: 1 }]);
```
Aaronius, line 408: The return value should probably be a promise resolved with the array to simulate actual behavior.

test/unit/specs/core/edgeNetwork/sendEdgeNetworkRequestFactory.spec.js
ninaceban: thank you, I'll fix my tests with your suggestions

src/utils/merge.js
ninaceban: `assign({}, ...values);` - this will work only if I pass values to a `flatMap`

src/components/Identity/customerIds/createCustomerIds.js
ninaceban: yes, sorry I forgot about these

src/utils/merge.js
ninaceban: I'll remove unused utils

src/components/Audiences/processDestinationsFactory.js
ninaceban: I'll add tests. thank you

test/unit/specs/core/edgeNetwork/sendEdgeNetworkRequestFactory.spec.js
ninaceban: yes, thank you

ninaceban:
@Aaronius I have addressed your comments/suggestions, could you please take a look again. Thank you

test/unit/specs/components/ActivityCollector/attachClickActivityCollector.spec.js
```diff
@@ -80,4 +79,15 @@ describe("ActivityCollector::attachClickActivityCollector", () => {
...
     });
   });
+  it("returns undefined", () => {
+    eventManager.createEvent = () => {
+      return {
+        isEmpty: () => false
+      };
+    };
+    attachClickActivityCollector(cfg, eventManager, lifecycle);
+    clickHandler({}).then(result => {
```
Aaronius, line 89: This promise needs to be returned, otherwise the test runner is not going to wait before moving on to the next test. It looks like other tests in the same spec file are suffering from the same problem. Would you please fix them as well while you're at it?

test/unit/specs/core/edgeNetwork/sendEdgeNetworkRequestFactory.spec.js
```diff
@@ -395,31 +395,36 @@ describe("sendEdgeNetworkRequestFactory", () => {
...
   it("returns the object from lifecycle::onResponse", () => {
-    lifecycle.onResponse.and.returnValue([{ a: 1 }]);
+    lifecycle.onResponse.and.returnValue(
+      Promise.resolve([{ c: 2 }, { h: 9 }, undefined])
+    );
 
     return expectAsync(
       sendEdgeNetworkRequest({ payload, action })
-    ).toBeResolvedTo({ a: 1 });
+    ).toBeResolvedTo({ c: 2, h: 9 });
```
Aaronius, line 417: The `returns the object from lifecycle::onBeforeRequest` and `returns the object from lifecycle::onResponse` tests seem like they're both already covered in the `returns the merged object from  lifecycle::onBeforeRequest & lifecycle::onResponse` test and are therefore redundant. I would prefer keeping `returns the merged object from  lifecycle::onBeforeRequest & lifecycle::onResponse` (because it tests that objects from both `lifecycle::onBeforeRequest` and `lifecycle::onResponse` are properly merged together) and getting rid of the other two.

We don't have any test that tests that the return values from `runOnResponseCallbacks` get merged into the final return value. I would recommend augmenting the `returns the merged object from  lifecycle::onBeforeRequest & lifecycle::onResponse` test to include that.

test/unit/specs/components/ActivityCollector/attachClickActivityCollector.spec.js
ninaceban: thank you, I'll fix the other tests as well

test/unit/specs/core/edgeNetwork/sendEdgeNetworkRequestFactory.spec.js
ninaceban: ok, I'll add that. thank you

Aaronius:
Ship it!

PR #399, opened 2020-04-07
Allow multiple instances with the same orgId

jonsnyder:
## Description

This PR allows multiple instances of Alloy to be created with the same orgId.  This is tricky because all the cookies are keyed off of the orgId, so if multiple instances of Alloy have the same org, they need to share cookies.  When cookies could be written by Konductor, each instance that share the org need to be notified.  This is important for Privacy and Identity because they are the two components that read cookies.

### Identity Configuration
`idMigrationEnabled` and `thirdPartyCookiesEnabled` are the two settings which control how an ECID is obtained.  In a scenario where two instances of Alloy are configured with the same orgId, but these config settings are different, the first instance which attempts to obtain an ECID will use its config.  

**Recommendation:** either each instance that shares an orgId have these settings set the same, or the instance with the desired settings always obtains the ECID first.

### Setting Customer Ids
Setting customer ids on one instance does not affect other instances which share an orgId.  Requests to Konductor will only have the customer ids set on that instance.

### Privacy Configuration
`defaultConsent` can be different across two instances of Alloy which share the same orgId; however, once consent is set, the consent state will be synchronized across the instances.

### Personalization
One thing missing from this that may be important is the ability to send Personalization decisions and rendering information to multiple configIds.  This is not tackled in this PR.

## Related Issue

https://jira.corp.adobe.com/browse/CORE-41753

## Motivation and Context

Customers want to be able to send data to two configIds within the same IMS org.

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Privacy/createComponent.js
+++ b/src/components/Privacy/createComponent.js
@@ -18,16 +18,20 @@ export default ({
   defaultConsent,
   consent,
   sendSetConsentRequest,
-  validateSetConsentOptions
+  validateSetConsentOptions,
+  consentObjects
 }) => {
   const consentByPurpose = assign({}, defaultConsent, readStoredConsent());
+  // only send to this instance's consent object because each instance can have a
+  // different default
   consent.setConsent(consentByPurpose);
 
   const readCookieIfQueueEmpty = () => {
     if (taskQueue.length === 0) {
       // Only read cookies when there are no outstanding setConsent
       // requests. This helps with race conditions.
-      consent.setConsent(readStoredConsent());
+      const storedConsent = readStoredConsent();
+      consentObjects.forEach(() => consent.setConsent(storedConsent));
     }
   };
 
@@ -36,7 +40,7 @@ export default ({
       setConsent: {
         optionsValidator: validateSetConsentOptions,
         run: options => {
-          consent.suspend();
+          consentObjects.forEach(() => consent.suspend());
           return taskQueue
             .addTask(() => sendSetConsentRequest(options))
             .catch(error => {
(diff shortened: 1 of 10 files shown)


jonsnyder:
I plan on writing unit and integration tests to exercise the new changes, but I wanted to get feedback first on what I have so far.

Aaronius:
@dancingcactus please review this PR description and weigh in on whether the behavior is as you would expect.

src/components/Privacy/createComponent.js
```diff
@@ -36,7 +40,7 @@ export default ({
       setConsent: {
         optionsValidator: validateSetConsentOptions,
         run: options => {
-          consent.suspend();
+          consentObjects.forEach(() => consent.suspend());
```
Aaronius, line 43: Should this be `consentObject => consentObject.suspend()`?

src/components/Privacy/createComponent.js
```diff
@@ -18,16 +18,20 @@ export default ({
...
+  // different default
   consent.setConsent(consentByPurpose);
 
   const readCookieIfQueueEmpty = () => {
     if (taskQueue.length === 0) {
       // Only read cookies when there are no outstanding setConsent
       // requests. This helps with race conditions.
-      consent.setConsent(readStoredConsent());
+      const storedConsent = readStoredConsent();
+      consentObjects.forEach(() => consent.setConsent(storedConsent));
```
Aaronius, line 34: Should this be `consentObject => consentObject.setConsent(storedConsent)`?

jfkhoury:
Awesome work @jonsnyder. 

I have the same concerns as @Aaronius in addition to few more, like the complexity in the code will make it really hard for future engineers to understand why the Identity & Privacy components are coded this way, and should they look for in their own components to account for multiple instances.

We will pick this up first thing next week to discuss with Justin and plan the next steps.

Thanks again.

jfkhoury:
@jonsnyder Thank you so much for this work! I will be closing this PR since we decided to support passing Data Sets on the event directly due to the maintenance cost this change can cause.

PR #402, opened 2020-04-10
C12412 / C12412 - Response should return Cookie & URL destinations if turned on in Blackbird

shammowla:
## Types of changes

## Checklist:

### DIFF
--- a/test/functional/specs/C12411.js
+++ b/test/functional/specs/C12411.js
@@ -1,0 +1,53 @@
+import { RequestLogger, t, ClientFunction } from "testcafe";
+import createNetworkLogger from "../helpers/networkLogger";
+import fixtureFactory from "../helpers/fixtureFactory";
+import { orgMainConfigMain } from "../helpers/constants/configParts";
+import configureAlloyInstance from "../helpers/configureAlloyInstance";
+import createResponse from "../../../src/core/createResponse";
+import getResponseBody from "../helpers/networkLogger/getResponseBody";
+
+const networkLogger = createNetworkLogger();
+
+const networkLoggerConfig = {
+  logRequestBody: true,
+  stringifyRequestBody: true
+};
+
+const destinationLogger = RequestLogger(
+  "https://cataas.com/cat",
+  networkLoggerConfig
+);
+
+fixtureFactory({
+  title:
+    "C12411 Response should return URL destinations if turned on in Blackbird",
+  requestHooks: [networkLogger.edgeEndpointLogs, destinationLogger]
+});
+
+test.meta({
+  ID: "C12411",
+  SEVERITY: "P0",
+  TEST_RUN: "Regression"
+});
+
+const triggerAlloyEvent = ClientFunction(() => {
+  return window.alloy("event", { xdm: { key: "value" } });
+});
+
+test("C12411 Response should return URL destinations if turned on in Blackbird", async () => {
+  await configureAlloyInstance("alloy", orgMainConfigMain);
+  await triggerAlloyEvent();
+  const response = JSON.parse(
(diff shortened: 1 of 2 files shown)


test/functional/specs/C12411.js
```diff
@@ -0,0 +1,37 @@
...
+  title:
+    "C12411 Response should return URL destinations if turned on in Blackbird",
+  requestHooks: [networkLogger.edgeEndpointLogs]
+});
+test.meta({
+  ID: "C12411",
+  SEVERITY: "P0",
+  TEST_RUN: "Regression"
+});
+const triggerAlloyEvent = ClientFunction(() =>
```
jfkhoury, line 20: Let's update this function to:

```
const triggerAlloyEvent = ClientFunction(() => {
  return new Promise(resolve => {
    window.alloy("event", { xdm: { key: "value" } }).then(resolve)
  });
});
```

jfkhoury: This will prevent any race conditions. We will be replacing all those `triggerEvents` with a util very soon.

test/functional/specs/C12412.js
```diff
@@ -0,0 +1,36 @@
...
+  title:
+    "C12412 Response should return Cookie destinations if turned on in Blackbird",
+  requestHooks: [networkLogger.edgeEndpointLogs]
+});
+test.meta({
+  ID: "C12412",
+  SEVERITY: "P0",
+  TEST_RUN: "Regression"
+});
+const triggerAlloyEvent = ClientFunction(() =>
```
jfkhoury, line 20: Same here.

test/functional/specs/C12411.js
Aaronius: I think you can just return the promise from `window.alloy` directly:

```
const triggerAlloyEvent = ClientFunction(() => {
  return window.alloy("event", { xdm: { key: "value" } });
});
```

jfkhoury:
Shamiul, let’s add the use cases mentioned by Aaron please.

shammowla:
@Aaronius Implemented the suggested changes.

test/functional/specs/C12411.js
```diff
@@ -0,0 +1,53 @@
...
+    getResponseBody(networkLogger.edgeEndpointLogs.requests[0])
+  );
+  const alloyResponse = createResponse(response);
+  const audiencesPayload = alloyResponse.getPayloadsByType("activation:push");
+  // This is a destination set for all visitors.
+  const destinationForAllVisitors = audiencesPayload.find(
+    payload =>
+      payload.type === "url" && payload.spec.url === "https://cataas.com/cat"
+  );
+  await t.expect(destinationForAllVisitors).ok();
```
Aaronius, line 50: Do you feel like this is providing value? As long as we see a request go to `https://cataas.com/cat`, which is your assertion below, that might be sufficient because we know that Alloy is receiving what it expects from Konductor.

test/functional/specs/C12412.js
```diff
@@ -0,0 +1,42 @@
...
+  const response = JSON.parse(
+    getResponseBody(networkLogger.edgeEndpointLogs.requests[0])
+  );
+  const alloyResponse = createResponse(response);
+  const audiencesPayload = alloyResponse.getPayloadsByType("activation:push");
+  // This is a destination set for all visitors.
+  const destinationForAllVisitors = audiencesPayload.find(
+    payload => payload.type === "cookie" && payload.spec.value === "test=C12412"
+  );
+  await t.expect(destinationForAllVisitors).ok();
```
Aaronius, line 39: Similar to my other comment: do you feel like this is providing value? I think the cookie assertion below may be sufficient.

PR #403, opened 2020-04-11
Core 42541 - change get decisions functionality 

ninaceban:
## Description

Changes related to get and render personalization decisions:
- renamed `viewStart` to `renderDecisions`
- renamed `scopes` to `decisionScopes`
- return decisions at the event command:
- if `renderDecisions` is true we add magical param `__view__` to `decisionScopes`. In the `event::onResponse` we render `dom-actions` decisions items. We return decisions with items that are not `dom-action` in the `event::command` promise.
- if `renderDecisions` is false or not defined and `decisionScopes` contains scopes, we return decisions in the `event:command` promise.

## Types of changes

## Checklist:


Aaronius:
I think you're trying to get rid of `viewStart` entirely in this PR, which is great, but still needs some work. Do a full search for `viewStart` in the repo and you'll see what still needs to be addressed. I would expect to no longer see `viewStart` anywhere in the repo. Some references that remain are just comments, but others are more important like in `validateUserEventOptions.js`.

src/components/DataCollector/index.js
```diff
@@ -25,11 +25,11 @@ const createDataCollector = ({ eventManager, logger }) => {
           let { xdm } = options;
           const {
             data,
-            viewStart = false,
             documentUnloading = false,
             type,
             mergeId,
-            scopes = []
+            renderDecisions = false,
+            decisionsScopes = []
```
Aaronius, line 32: Curious to know how others feel about this name. My initial reaction is that `decisionScopes` feels better than `decisionsScopes`.

src/components/DataCollector/index.js
```diff
@@ -53,13 +53,9 @@ const createDataCollector = ({ eventManager, logger }) => {
           event.setUserData(data);
 
           const details = {
-            isViewStart: viewStart
+            renderDecisionsEnabled: renderDecisions,
```
Aaronius, line 56: If we prefer the `Enabled` suffix for ourselves, should we prefer the `Enabled` suffix on the public API on behalf of our customers? Should it be `decisionRenderingEnabled` instead of `renderDecisionsEnabled`? Looking at some of the other configuration options we have:

* clickCollectionEnabled
* cookieDestinationsEnabled
* debugEnabled

I think `decisionRenderingEnabled` might be the most appropriate, though one could make the argument that `debugEnabled` should be `debuggingEnabled`. I hate naming things.

ninaceban: I think that options passed in the event command should be named `renderDecisions`,`viewStart` etc., anything related to configuration settings should be suffixed with `enabled` and we should follow the convention like `debuggingEnabled`, `cookieDestinationEnabled` etc.

Aaronius: 👍  I do like `renderDecisions` more than `decisionRenderingEnabled`. Maybe I like it more because "enabled" sounds like a long-running configuration option rather than whether something should be done a single time (in this case, decisions should be rendered). This would seem to agree with what you've stated.

I think I'm okay with the configuration options as they are though.

src/components/DataCollector/index.js
jfkhoury: @Aaronius  I like `decisionScopes` because I can pronounce it ;)

test/unit/specs/components/Personalization/index.spec.js
```diff
@@ -10,185 +10,94 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
 
-    personalization.lifecycle.onResponse({ response });
-
-    const result = personalization.commands.getDecisions.run({ scopes });
-
-    expect(Array.isArray(result)).toBeTrue();
-    expect(result.length).toEqual(0);
+    return expectAsync(deferred.promise).toBeResolvedTo({
+      decisions: SCOPES_FOO1_FOO2_DECISIONS
+    });
```
Aaronius, line 65: Nothing asynchronous is actually happening during this process, so there shouldn't be any need for a promise. I think you can re-work this to:

```
let result;
const onResponse = func => {
  result = func({ response });
};

personalization.lifecycle.onBeforeEvent({
  event,
  renderDecisionsEnabled,
  decisionsScopes,
  onResponse
});

expect(result.decisions).toEqual(expectedResponse);
```

I believe this applies to the other test in this file as well.

test/unit/specs/components/Personalization/index.spec.js
```diff
@@ -10,185 +10,94 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+      decisionsScopes,
+      onResponse
     });
 
-    expect(() =>
-      personalization.commands.getDecisions.validateOptions()
-    ).toThrow();
+    return expectAsync(deferred.promise).toBeResolvedTo({
+      decisions: expectedResponse
+    });
```
Aaronius, line 101: @ninaceban as you start to add more specs for personalization, I think you'll find that breaking down `Personalization/index.js` into smaller modules that get injected and then mocking those modules will provide a simpler way to test. See the Context or Identity components for examples. I think we'll talk about this in the engineering check-in today as well.

src/components/DataCollector/index.js
Aaronius: @ninaceban Mind changing it to `decisionScopes`? I think grammatically it probably makes more sense and @jfkhoury and I will be able to pronounce it. 😄

ninaceban: ok, I'll change that. thank you

src/components/DataCollector/index.js
ninaceban: ok, then I'll rename it to `renderDecisions`

src/components/Personalization/index.js
```diff
@@ -34,32 +29,27 @@ const mergeMeta = (event, meta) => {
...
+  return decisions.reduce((acc, decision) => {
+    const { items = [] } = decision;
+    const decisionItems = items.filter(isNotDomAction);
 
-  const filteredDecisions = groupBy(
-    decisions,
-    decision => decision.scope || PAGE_WIDE_SCOPE
-  );
+    if (isNonEmptyArray(decisionItems)) {
+      const newDecision = {};
```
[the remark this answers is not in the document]
ninaceban: ok, thank you

test/unit/specs/components/Personalization/index.spec.js
ninaceban: yes that makes sense, thank you Aaron

ninaceban:
@Aaronius could you please take a look, it is not done yet but I want to know if I'm on the right path. Thank you

src/components/Personalization/createComponent.js
```diff
@@ -0,0 +1,84 @@
...
+
+        // For renderDecisions we try to hide the personalization containers
+        if (renderDecisions) {
+          hideContainers(prehidingStyle);
+        }
+
+        mergeQuery(event, createQueryDetails(scopes));
+
+        onResponse(({ response }) =>
+          onResponseHandler({ renderDecisions, response, modules, logger })
```
Aaronius, line 69: Notice here that `modules` and `logger` are things that are not specific to `createComponent.js`. This is a sign that `modules` and `logger` may be things you could inject into a `createOnResponseHandler` module up in `index.js`, which would create an `onResponseHandler`, which you could then inject into `createComponent.js`. The easiest way to illustrate would probably be to open a PR showing you what that might look like.

ninaceban: thank you, I've got the idea :)

src/components/Personalization/createComponent.js
```diff
@@ -0,0 +1,84 @@
...
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+import { noop } from "../../utils";
+import { hasScopes, isAuthoringModeEnabled, getDecisionScopes } from "./utils";
+import { initDomActionsModules } from "./dom-actions";
+import collectClicks from "./dom-actions/clicks/collectClicks";
+import { hideContainers, showContainers } from "./flicker";
+import { mergeMeta, mergeQuery, createQueryDetails } from "./event";
```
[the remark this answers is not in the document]
ninaceban: thank you

src/components/Personalization/createComponent.js
```diff
@@ -0,0 +1,84 @@
...
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+import { noop } from "../../utils";
+import { hasScopes, isAuthoringModeEnabled, getDecisionScopes } from "./utils";
+import { initDomActionsModules } from "./dom-actions";
```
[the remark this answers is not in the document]
ninaceban: sure, thank you

ninaceban:
@Aaronius I have extracted all dependencies as separate modules and added tests for those. If you could take a look again. There are some other changes that I would do, in the next follow up PR. Thank you

src/components/Personalization/createExecuteDecisions.js
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 Adobe. All rights reserved.
+Copyright 2020 Adobe. All rights reserved.
```
Aaronius, line 2: If your editor is changing the year automatically, I don't think this is a problem, but if you're taking the time to manually change the year, you don't need to. Legal has said that the original year is fine and doesn't need to be updated as contents of the file change.

test/unit/specs/components/Personalization/createCollect.spec.js
```diff
@@ -0,0 +1,38 @@
...
+    id: 1,
+    decisionId: "foo"
+  };
+  const event = { type: "blah" };
+
+  beforeEach(() => {
+    eventManager = {
+      sendEvent: jasmine.createSpy("sendEvent"),
+      createEvent: jasmine.createSpy("createEvent").and.returnValue(event)
+    };
```
Aaronius, line 27: If you prefer, you can do this:

```
eventManager = jasmine.createSpyObj("eventManager", {
  sendEvent: undefined,
  createEvent: event
});
```

test/unit/specs/components/Personalization/createExecuteDecisions.spec.js
```diff
@@ -0,0 +1,84 @@
...
+    });
+    executeDecisions(decisions);
+    expect(executeActions).toHaveBeenCalledWith(
+      expectedAction,
+      modules,
+      logger
+    );
+  });
+
+  it("shouldn't trigger executeActions", () => {
```
Aaronius, line 71: I recommend adding more context here, like `shouldn't trigger executeActions when provided with empty array of actions` or something like that, just so it's easier to understand the scenario without sifting through the code.

test/unit/specs/components/Personalization/createOnClickHandler.spec.js
```diff
@@ -0,0 +1,39 @@
...
+  it("collects clicks", () => {
+    const handleOnClick = createOnClickHandler({
+      mergeMeta,
+      collectClicks,
+      clickStorage
+    });
+    const clickedElement = "foo";
+    handleOnClick({ event, clickedElement });
+    expect(collectClicks).toHaveBeenCalledWith(
+      jasmine.any(Function),
```
Aaronius, line 34: Because this function was created by the module, I recommend adding a test to make sure that the function does what you expect it to do. Specifically, it should call `mergeMeta` with `event` and `meta` and return the result.

test/unit/specs/components/Personalization/createOnResponseHandler.spec.js
```diff
@@ -0,0 +1,74 @@
...
+      extractDecisions,
+      executeDecisions,
+      showContainers
+    });
+
+    const result = onResponse({ renderDecisions, response });
+
+    expect(extractDecisions).toHaveBeenCalledWith(response);
+    expect(showContainers).toHaveBeenCalled();
+    expect(executeDecisions).toHaveBeenCalled();
```
Aaronius, line 52: Assert that `executeDecisions` was called with `renderableDecisions`.

test/unit/specs/components/Personalization/createOnResponseHandler.spec.js
```diff
@@ -0,0 +1,74 @@
...
+    expect(result).toEqual(expectedResult);
+  });
+
+  it("should not trigger executeDecisions and should return all decisions", () => {
```
Aaronius, line 56: Just add a little context here, like `when renderDecisions is false`.

test/unit/specs/components/Personalization/index.spec.js
```diff
@@ -10,96 +10,4 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
-      decisionScopes,
-      onResponse
-    });
-
-    return expectAsync(deferred.promise).toBeResolvedTo({
-      decisions: expectedResponse
-    });
-  });
-});
+describe("Personalization", () => {});
```
Aaronius, line 13: You can delete this file entirely. The script that runs that checks to see if a spec files exists for each source file ignores `index.js` files.

test/unit/specs/components/Personalization/createComponent.spec.js
```diff
@@ -10,94 +10,247 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+      onResponseHandler,
+      onClickHandler,
+      hideContainers,
+      showContainers,
+      hasScopes,
+      isAuthoringModeEnabled,
+      getDecisionScopes,
+      mergeQuery,
+      createQueryDetails
+    });
```
Aaronius, line 66: If you want, you can extract this out to a `build` function like was done here: https://github.com/adobe/alloy/blob/b7fc7d8e68eeb7cec68201c9205ef83584a638bb/test/unit/specs/components/Privacy/createComponent.spec.js#L37-L46

Completely your choice though.

Aaronius:
I noticed there are still a couple places where "view start" is mentioned that need to be cleaned up. One is in About.js:

image

Another is createEventManager.js:

image

src/components/Personalization/createExecuteDecisions.js
ninaceban: I see, I didn't know that! thank you!

test/unit/specs/components/Personalization/createCollect.spec.js
ninaceban: thank you

test/unit/specs/components/Personalization/createOnResponseHandler.spec.js
ninaceban: thank you

test/unit/specs/components/Personalization/index.spec.js
ninaceban: ok, thanks

test/unit/specs/components/Personalization/createComponent.spec.js
ninaceban: ok, thank you

test/unit/specs/components/Personalization/createOnResponseHandler.spec.js
ninaceban: ok, thank you

test/unit/specs/components/Personalization/createExecuteDecisions.spec.js
ninaceban: thank you

test/unit/specs/components/Personalization/createOnClickHandler.spec.js
ninaceban: there is a proposal to change the clickTracking functionality, and I would rather add these changes then, if it will still apply.

ninaceban:
@Aaronius I have created another branch with the changes since this one had too many conflicts with the master. Changes related to `onClickHandler` I'll address in another PR when I'll work on `click tracking`  and `notification` functionality since there might be some code design changes.
Here is the new PR: https://github.com/adobe/alloy/pull/409

Aaronius:
@ninaceban can we close this PR at this point?

Aaronius:
This was superseded by a different PR.

PR #406, opened 2020-04-13
CORE-43993 Fix bug where first request was not being sent to demdex.

Aaronius:
## Description
Initial requests from Alloy were never being sent to demdex. This is because `sendEdgeNetworkRequestFactory` is determining the domain that should be used before `lifecycle.onBeforeRequest` is called, which is when the Identity component calls `payload.useIdThirdPartyDomain`. This is a regression caused by yours truly in https://github.com/adobe/alloy/pull/392. It looks like we have functional tests planned that would have caught it (C10922).

## Related Issue
https://jira.corp.adobe.com/browse/CORE-43993

## Motivation and Context
Identity coordination using a third-party domain is important for improved identification.

## Types of changes

## Checklist:


Aaronius:
@jfkhoury and I talked and I'm going to add the functional tests to this PR.

package.json
```diff
@@ -126,7 +126,7 @@
     "rollup-plugin-replace": "^2.2.0",
     "rollup-plugin-terser": "^5.1.2",
     "start-server-and-test": "^1.10.6",
-    "testcafe": "^1.6.1",
+    "testcafe": "^1.8.4",
```
Aaronius, line 129: Needed to upgrade testcafe in order to get `t.browser`.

Aaronius:
I ended up amending and force pushing my last commit (and _only_ my last commit) because otherwise it would have been difficult for you to compare my changes with what you previously approved.

I've added e2e tests for demdex ("ID third-party domain") usage. I followed https://testrail.corp.adobe.com/index.php?/cases/view/10922 as a template, with a modification. The testrail test case has permutations where `idMigrationEnabled` is toggled on and off, but the expected results described in testrail don't reflect reality. For example, it says that when `idMigrationEnabled: false, thirdPartyCookiesEnabled: true, edgeDomain: "firstparty.alloyio.com"`, that the first request _shouldn't_ go to demdex, which isn't true with the code we have today in Alloy. I think back when the description was written we were planning on not hitting demdex if we had an ECID from an AMCV cookie that we could migrate. However, later we decided to hit demdex anyway because, according to my notes:

```
// If we have a identity to migrate, we still want to hit the
// third-party identification domain because the third-party identification
// domain will use our ECID to set the third-party cookie if the third-party
// cookie isn't already set, which provides for better cross-domain
// identification for future requests.
```

If we agree this is functioning as expected according to the latest requirements, I'll update the TestRail description when I merge this PR.

You may notice that in both the testrail description and the actual test case implementation we have permutations with `edgeDomain` switching between a first-party domain and third-party domain and `idMigrationEnabled` being toggled to `true` or `false`. Toggling them doesn't actually have any affect on whether we hit demdex. I included their variants because I want to make sure it stays that way; I could see us messing up the implementation when it comes to those two configuration options and whether we hit demdex.

@jfkhoury @jonsnyder do you mind reviewing my last commit?

Aaronius:
The test that's failing on the CI server is `C2579: Isolates multiple SDK instances`, which is unrelated. I'm going to start working on fixing that in a separate PR.

test/functional/helpers/networkLogger/DemdexProxy.js
```diff
@@ -38,6 +38,10 @@ class DemdexProxy extends RequestHook {
   async onRequest({ requestOptions }) {
     transformProdToInt(requestOptions);
   }
+
+  async onResponse() {
+    // Do nothing. TestCafe will throw an error if we don't provide this.
```
Aaronius, line 43: This is a new thing I had to do since I upgraded TestCafe.

test/functional/specs/C10922.js
```diff
@@ -0,0 +1,171 @@
...
+});
+
+const idThirdPartyDomainRegex = /\.demdex\.net/;
+
+const getHostForFirstRequest = () => {
+  const firstRequest = networkLogger.edgeInteractEndpointLogs.requests[0];
+  return firstRequest.request.headers.host;
+};
+
+const assertRequestWentToIdThirdPartyDomain = async () => {
```
jonsnyder, line 43: All these "third party" names hurt my head.  Do you think this would be easier to understand if we just said "Demdex". (i.e. `assertRequestWentToDemdex`)

Aaronius: I was hoping to keep the name brand-agnostic and more role-oriented but, yes, it would probably be easier to understand with "demdex". I'll change it.

test/functional/helpers/setLegacyIdentityCookie.js
```diff
@@ -0,0 +1,18 @@
...
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+import { ClientFunction } from "testcafe";
+
+export default ClientFunction(orgId => {
```
jfkhoury, line 15: Nice

PR #408, opened 2020-04-16
Fixes C2579 test.

Aaronius:
## Description
Fixes C2579 test. Here's what was happening. For a single instance, we were going through this process:

1. Alloy instance sends a request without an identity state entry. Konductor responds with an identity state entry containing the value `ABC` (as an example).
1. Alloy instance sends a request with identity state entry containing the value `ABC`. Konductor responds with identity state entry containing the value `DEF` **This is a different value than the one Konductor returned previously.**
1. The test refreshes the page.
1. Alloy instance sends a request with an identity state entry containing the value `DEF`.
1. We were asserting that the value of the identity state entry sent on the second request was the same as the value on the third request. `ABC` !== `DEF`, so this was failing.

On the second request, I'm not sure why Konductor is returning a different identity value than the value sent to Konductor, but I don't have any particular reason to believe this is invalid behavior. I'll ping Rares to look at this PR description and verify that nothing seems problematic here from his perspective.

I see one of two ways of changing our test to handle this case:

1. Change our assertion so that it compares the identity state entry value on the third request to the identity state entry value **returned** (rather than sent) from the second request.
1. Skip the page refresh and third request entirely. I think the rest of the test probably provides enough validation to ensure that separate ECIDs are used for multiple SDK instances, which is the described goal of the test.

I went with the second approach.

## Related Issue
None

## Motivation and Context
Ensure our code is working the way we think it is.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Improvement (non-breaking change which does not add functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have signed the [Adobe Open Source CLA](https://opensource.adobe.com/cla.html) or I'm an Adobe employee.
- [x] I have submitted a [documentation](https://github.com/AdobeDocs/alloy-docs) pull request or no changes are needed.
- [x] I have made any necessary test changes and all tests pass.
- [x] I have run the Sandbox successfully.

PR #412, opened 2020-04-24
CORE-44791 Prevent modification of passed in xdm and data objects.

Aaronius:
## Description
When xdm is passed as an option to the event command AND either type or mergeId are passed as options, DataCollector ends up modifying the xdm object passed by the customer. This is bad for a couple reasons:

1. The xdm object the customer passes is possibly used by other customer code. For example, it may be the customer's data layer and they may be sending the data layer to other endpoints as well. In this case, they probably don't want Alloy-event-specific information being added to the data layer.
2. It can cause incorrect event types and event merge IDs to be sent. For example, if I do this:

```
const dataLayer = { foo: "bar" };
alloy("event", { xdm: dataLayer, mergeId: "abc" });
alloy("event", { xdm: dataLayer, mergeId: "def" });
```

Then `def` will be sent for the event merge ID for both requests. This happens because the first command sets `dataLayer.eventMergeId` to `abc`, but before the object is sent to the server, the second command sets `dataLayer.eventMergeId` to "def".

## Related Issue
https://jira.corp.adobe.com/browse/CORE-44791

## Motivation and Context
Current behavior is unexpected.

## Types of changes

## Checklist:

### DIFF
--- a/src/components/DataCollector/index.js
+++ b/src/components/DataCollector/index.js
@@ -1,4 +1,3 @@
-import { assign } from "../../utils";
 import validateUserEventOptions from "./validateUserEventOptions";
 
 /*
@@ -22,8 +21,8 @@ const createDataCollector = ({ eventManager, logger }) => {
           return validateUserEventOptions({ options, logger });
         },
         run: options => {
-          let { xdm } = options;
           const {
+            xdm,
             data,
             documentUnloading = false,
             type,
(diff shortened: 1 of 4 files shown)


jfkhoury:
Hey @Aaronius, it seems like C2579 is failing for the first time, I wonder if your changes have something to do with that?

Aaronius:
@jfkhoury That's the same failure that I was getting inconsistent results with in https://github.com/adobe/alloy/pull/408. 😢

Aaronius:
I'm assuming, but haven't verified, that https://github.com/adobe/alloy/pull/416 will fix the e2e test that's intermittently failing. If not, we'll address it in a separate PR.

PR #414, opened 2020-04-24
Rename `event` to `sendEvent`; `configId` to `edgeConfigId`.

jfkhoury:
Docs PR: https://git.corp.adobe.com/AdobeDocs/experience-platform.en/pull/544

Launch Extension PR: https://github.com/adobe/reactor-extension-alloy/pull/71

## Related Issue

https://jira.corp.adobe.com/browse/CORE-44840
https://jira.corp.adobe.com/browse/CORE-44841

## Types of changes

## Checklist:


src/components/Privacy/createComponent.js
```diff
@@ -25,9 +25,13 @@ export default ({
 
   const readCookieIfQueueEmpty = () => {
     if (taskQueue.length === 0) {
+      const storedConsent = readStoredConsent();
```
liljenback, line 28: Is this supposed to be part of this commit?

jfkhoury: I sneaked this in there because I noticed a warning in the console when we were going through the demos :)  (@jonsnyder to confirm the fix)

test/functional/specs/C2579.js
```diff
@@ -62,14 +62,21 @@ const getIdentityCookieValue = request => {
 
 const instance1Config = () => configureAlloyInstance(altConfig);
 
-const instance1Event = ClientFunction(() =>
-  window.alloy("event", { data: { key: "value" } })
-);
+const instance1Event = ClientFunction(() => {
+  return new Promise(resolve => {
+    return window.alloy("sendEvent", { data: { key: "value" } }).then(resolve);
+  });
```
Aaronius, line 68: Can you help me understand why you had to wrap it in a new promise?

Aaronius:
I resolved the conflict because I felt bad making you do it.

src/components/Privacy/createComponent.js
jonsnyder: 👍

jfkhoury:
> I resolved the conflict because I felt bad making you do it.

You're too nice!

test/functional/specs/C2579.js
jfkhoury: @Aaronius  So I thought this is the only way we can prevent a race condition, because we are resolving the promise within the `ClientFunction` since we can return a promise into testcafe to await it there. No? Crazy?

Aaronius: I think there's a misunderstanding. If you return a promise, you can `await` it within your test, whether the promise is the one returned from Alloy or you wrap the Alloy one in another promise like you've done here. TestCafe has some special logic that says if the return value is a promise, it will wait for it and then serialize the resolved value.

The race condition problems we were having previously was because we were trying to return an object containing a promise, so TestCafe was serializing the object itself, and the promise never made it through the serialization/deserialization process because a promise can't be serialized.

jfkhoury: Ah yes, that makes more sense, thank you!

PR #415, opened 2020-04-26
Core 42887 - send bulk notifications

ninaceban:
## Description

Currently, we are sending a notification-call for each decision that we render. We need to collect metadata from all rendered decisions and send one notification call to Konductor. 
Metadata will look like this:
``` 
personalization: {
   decisions: [{
       id:1,
       scope: "___view__"
   }]
}
```
## Types of changes

## Checklist:

### DIFF
--- a/src/components/Personalization/createExecuteDecisions.js
+++ b/src/components/Personalization/createExecuteDecisions.js
@@ -22,18 +22,23 @@ const buildActions = decision => {
 
 export default ({ modules, logger, executeActions, collect }) => {
   return decisions => {
-    const decisionPromise = decisions.map(decision => {
+    const decisionMetasPromise = decisions.map(decision => {
       const actions = buildActions(decision);
 
       return executeActions(actions, modules, logger);
     });
-    return Promise.all(decisionPromise).then(result => {
-      if (isNonEmptyArray(result)) {
+    return Promise.all(decisionMetasPromise)
+      .then(result => {
         const metas = flatMap(result, identity);
-        const decisionMetas = metas.map(item => item.meta);
-
-        collect({ decisions: decisionMetas });
-      }
-    });
+        return metas.map(item => item.meta);
+      })
+      .then(metas => {
+        if (isNonEmptyArray(metas)) {
+          collect({ decisions: metas });
+        }
+      })
+      .catch(error => {
+        logger.error(error);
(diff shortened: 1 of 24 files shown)


src/components/Personalization/createExecuteDecisions.js
```diff
@@ -10,20 +10,28 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+  const meta = { id: decision.id, scope: decision.scope };
 
   return decision.items.map(item => assign({}, item.data, { meta }));
 };
 
-export default ({ modules, logger, executeActions }) => {
+export default ({ modules, logger, executeActions, collect }) => {
   return decisions => {
-    decisions.forEach(decision => {
+    const decisionPromise = decisions.map(decision => {
```
jfkhoury, line 25: Are those `decisionActionsPromises`?

src/components/Personalization/createExecuteDecisions.js
```diff
@@ -10,20 +10,30 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
 
-      executeActions(actions, modules, logger);
+      return executeActions(actions, modules, logger);
+    });
+    return Promise.all(decisionPromise).then(result => {
+      if (isNonEmptyArray(result)) {
+        const metas = flatMap(result, identity);
+        const decisionMetas = metas.map(item => item.meta);
+
+        collect({ decisions: decisionMetas });
```
Aaronius, line 35: Hey @ninaceban, I think we should be returning the promise from `collect`. If the `collect` call fails, I think we want to be failing the `sendEvent` command. I think. You would also need to handle the promise coming out of the `executeDecisions` call in `createOnResponseHandler.js` so that we wait for the promise to finish allowing the `onResponse` lifecycle to complete.

ninaceban: actually collect is a fire and forget action. I don't think onResponseHandler should wait for collect to complete. But we could add `then` and `catch` to collect call to at least log the success or failure that should be enough for debugging-troubleshooting.

src/components/Personalization/createExecuteDecisions.js
ninaceban: more like decisionMetasPromises

PR #416, opened 2020-04-26
Run e2e tests against local chrome.

jfkhoury:
## Description

Sauce labs keep failing on a regular basis, and the only browser we have been so far using is Chrome. I'm testing against the Chrome addon in this PR, and will be including FIrefox soon. After that, I'd like to switch the environment to mac osx instead of Linus, and test against Safari as well.

## Types of changes

## Checklist:


liljenback:
Is Saucelabs the problem, or something in our setup? Should we look at using another solution?

jfkhoury:
> Is Saucelabs the problem, or something in our setup? Should we look at using another solution?

SauceLabs fails intermittently which is causing false negatives.
The only reason we need SauceLabs or a similar solution is Edge & IE11. Edge is using Chromium now so I don't know how much we need to test on it, and IE11 is not a good enough reason to keep SauceLabs. Let me switch the env to mac osx, add Safari & Firefox and see how it goes, after we merge this PR.

Aaronius:
As long as we officially support IE, I don't think we can just not run automated tests in IE. We _will_ break customers. In my opinion, we either have to pull support for IE or figure out something else.

jfkhoury:
> As long as we officially support IE, I don't think we can just not run automated tests in IE. We _will_ break customers. In my opinion, we either have to pull support for IE or figure out something else.

I was thinking of using SauceLabs for IE11 only, using a separate npm script, and unblock all the other browsers. At some point next year we will revisit support for IE11. I just don't want to use SauceLabs for all the major browsers for no real reason.

Aaronius:
I support that. I still don't think we should go ahead with this PR until that's work is done though. I really don't want to be in a position where we're not running automated tests on IE. IMO, it's the browser where the automated CI tests may be the most important, since it's the most different of the browsers and we never (as far as I'm aware) run tests on IE locally.

Aaronius:
Do you want my help on this or do you just disagree with my premise?

jfkhoury:
> Do you want my help on this or do you just disagree with my premise?

I don't disagree at all. Just FYI, we haven't ran E2E tests on IE11 it's been few months, so this PR does not break anything we had before. We are running Unit Tests on IE11 currently though.

Aaronius:
Oh, that's news to me (I probably should have noticed). A bit sad, but somewhat reassuring that we're running unit tests still.

Aaronius:
Do you want to remove the saucelabs dependency, npm script, Karma-saucelabs config until we implement IE support or is that intentionally left in the repo?

Aaronius:
I resolved the conflict because I felt bad making you do it.

jfkhoury:
I am going to merge this PR so I can rebase the others on master. More work is needed here including:

- Use mac osx
- Install Safari
- Test on Safari & Firefox

PR #473, opened 2020-04-30
CORE-43157 Functional tests for ID migrations

liljenback:
## Related Issue

CORE-43157

## Types of changes

## Checklist:


test/functional/helpers/generateId.js
```diff
@@ -0,0 +1,19 @@
...
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+export default () => {
+  let id = "";
+  for (let i = 0; i < 38; i += 1) {
+    id += Math.floor((Math.random() * 10) % 10);
```
jonsnyder, line 16: Why include the modulus?  Seems like this would be sufficient.
`Math.floor(Math.random() * 10)`

test/functional/specs/C35448.js
```diff
@@ -0,0 +1,41 @@
...
+
+const config = compose(
+  orgMainConfigMain,
+  debugEnabled,
+  migrationEnabled
+);
+
+const getEcid = ClientFunction(() => {
+  return window.alloy("getIdentity", {}).then(result => {
+    return Promise.resolve(result.ECID);
```
jonsnyder, line 32: you can just `return result.ECID`. No need to wrap in a resolved promise when you are already in a promise then function.

test/functional/specs/C35450.js
```diff
@@ -0,0 +1,50 @@
...
+const getEcid = ClientFunction(() => {
+  return window.alloy("getIdentity", {}).then(result => {
+    return Promise.resolve(result.ECID);
+  });
+});
+
+test("C35450 - When ID migration is enabled and Visitor and Alloy are both awaiting consent, when consent is given to both, Alloy waits for Visitor to get ECID and then uses this value.", async () => {
+  const ecid = generateId();
+  await createMockVisitor(ecid);
+  await createMockOptIn(true);
```
jonsnyder, line 46: This doesn't match the description.  Visitor isn't awaiting consent.

liljenback: It is pending given that the callback has not been called.

test/functional/helpers/generateId.js
```diff
@@ -0,0 +1,19 @@
...
+governing permissions and limitations under the License.
+*/
+
+export default () => {
+  let id = "";
+  for (let i = 0; i < 38; i += 1) {
+    id += Math.floor(Math.random() * 10);
+  }
+  return id;
+};
```
Aaronius, line 19: Can you use the `uuid` npm package, since it's already a dependency?

test/functional/specs/C35450.js
Aaronius: The implementation in Alloy is kind of weird because of, in my opinion, some shortcomings of Visitor. When I talked to @wgamboa during implementation, he was saying that if Alloy calls `visitor.getMarketingCloudVisitorID` before opt-in is received, `visitor.getMarketingCloudVisitorID` will return--I think--an empty string. So we have to do this indirection where Alloy interfaces directly with the opt-in object to wait until the user opts in _then_ call `visitor.getMarketingCloudVisitorID`. :/

test/functional/helpers/optIn/createMockOptIn.js
```diff
@@ -0,0 +1,31 @@
...
+
+import { ClientFunction } from "testcafe";
+
+export default ClientFunction(approved => {
+  window.adobe = {
+    optIn: {
+      fetchPermissions(callback) {
+        setTimeout(() => {
+          callback();
+        }, 500);
```
Aaronius, line 21: Is the `500` important here versus, say `0`? Is there a way we can make it not important while still providing a reasonable test? I'm just always hesitant to add delays in automated tests because they start to add up fast.

test/functional/specs/C35450.js
Aaronius: In fact, @liljenback, do you mind adding my comment more or less to https://github.com/adobe/alloy/blob/296563ce509696478d6f33940dc23260b7045026/src/components/Identity/visitorService/getEcidFromVisitorFactory.js#L19 so that later engineers understand this relationship?

test/functional/helpers/optIn/createMockOptIn.js
liljenback: I just want the optIn to be pending for a little while. I could change it to explicitly calling it callback but I'd rather reduce it (which I think is also fine).

Aaronius: Unless it will actually influence the result of the test, I'd put it on a timeout of `0`.

liljenback: @Aaronius , regarding @wgamboa 's comment. That makes sense that we are looking at optin separately in Identity.

test/functional/specs/C35450.js
Aaronius: I think what @jonsnyder might be saying though is that Alloy _technically_ isn't waiting for Visitor to get the ECID. It's more like "When ID migration is enabled and Alloy receives consent, Alloy awaits legacy consent before requesting the ECID from Visitor." I think.

jonsnyder: Yes, for appMeasurement, when we call getVisitorValues and consent has not been given, it calls the callback with empty values.  We should duplicate that behavior in the mocks here.  This does bring up an interesting question though on whether we should be including the real Visitor.js on the page for these integration tests.

liljenback: I was thinking about including Visitor, but it would be nice if we could avoid it. We're only using the functionality to get the ECID.

Aaronius: I was also wondering why we weren't using a real Visitor instance. It would give us the most accurate results. What's your reasoning for wanting to avoid it even if it's just being used for ECID @liljenback?

liljenback: Well we call getMarketingCloudVisitorID with a callback and that resolves with an ID. The Visitor library has a lot more functionality and including that overhead didn't seem necessary.

jfkhoury: @liljenback @Aaronius @jonsnyder 

We already decided to use the real Visitor.js back when we planned this test, mainly because it's an E2E test. Meaning when Visitor changes, we'd love to to know if it's still compatible with Alloy. So let's use the `latest` tag from the Github repo where Visitor gets released. I believe it's this guy:

https://github.com/Adobe-Marketing-Cloud/id-service/releases/latest

Thanks.

PR #475, opened 2020-05-01
fixed - duplicates of metas in notification call

ninaceban:
## Description

If we have a decision with multiple items we should have only one insert in the `personalization.decisions` array
```
{
  "decisions": [
    {
      "id": "TNT:113866:1"
    },
    {
      "id": "TNT:115166:0"
    },
    {
      "id": "TNT:113712:1"
    }
  ]
}
```

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Personalization/createExecuteDecisions.js
+++ b/src/components/Personalization/createExecuteDecisions.js
@@ -20,6 +20,31 @@ const buildActions = decision => {
   return decision.items.map(item => assign({}, item.data, { meta }));
 };
 
+const dedupe = metas => {
+  const result = [];
+  const set = new Set();
+
+  metas.forEach(item => {
+    if (set.has(item.id)) {
+      return;
+    }
+
+    set.add(item.id);
+    result.push(item);
+  });
+
+  return result;
+};
+
+const collectMetas = (collect, decisionMetas) => {
+  const metas = flatMap(decisionMetas, identity);
+  const dedupedMetas = dedupe(metas.map(e => e.meta));
+
+  if (isNonEmptyArray(dedupedMetas)) {
+    collect({ decisions: dedupedMetas });
+  }
+};
+
 export default ({ modules, logger, executeActions, collect }) => {
   return decisions => {
     const decisionMetasPromise = decisions.map(decision => {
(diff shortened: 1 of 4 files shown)


src/components/Personalization/createExecuteDecisions.js
```diff
@@ -20,6 +20,31 @@ const buildActions = decision => {
...
+    if (set.has(item.id)) {
+      return;
+    }
+
+    set.add(item.id);
+    result.push(item);
+  });
+
+  return result;
+};
```
Aaronius, line 37: This should probably be extracted to a utility. Here's my thought:

```
// identity.js
const identity = value => value;

// dedupe.js
const dedupe = (items, keyResolver = identity) => {
  const set = new Set();
  return items.filter(item => {
    const key = keyResolver(item);
    if (set.has(key)) {
      return false;
    }

    set.add(key);
    return true;
  });
};
```

Example usage:

```
const dedupedMetas = dedupe(metas, meta => meta.id);
```

src/components/Personalization/createExecuteDecisions.js
```diff
@@ -20,6 +20,31 @@ const buildActions = decision => {
...
+
+const collectMetas = (collect, decisionMetas) => {
+  const metas = flatMap(decisionMetas, identity);
+  const dedupedMetas = dedupe(metas.map(e => e.meta));
```
Aaronius, line 41: I'm a little confused by the `metas` naming, because I would expect each item in `metas` to be a `meta`, but it seems like it's an object with a `meta` property. This might be something to improve is something comes to mind for you.

ninaceban:
@Aaronius could you please take a look again, I have made some other changes. We need to skip some results that contain errors from metas. We will discuss this later if we should send this data to konductor.

src/components/Personalization/createExecuteDecisions.js
```diff
@@ -20,23 +20,41 @@ const buildActions = decision => {
   return decision.items.map(item => assign({}, item.data, { meta }));
 };
 
+const processMetas = (collect, logger, actionResults) => {
+  const results = flatMap(actionResults, identity);
+  const finalMetas = [];
+  const set = new Set();
```
jfkhoury, line 26: Will this cause issues in IE11?

ninaceban: nope - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

PR #484, opened 2020-05-05
CORE-45102 Removed errorsEnabled. No longer failing commands due to consent.

Aaronius:
## Description
No longer reject command promises if the user has declined consent. Remove `errorsEnabled` configuration option. We may add it back later if we still find it is necessary, but now that promises aren't rejected due to consent, there will be much fewer failures and any failures that do occur probably deserve attention.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-45102

## Motivation and Context
I discussed this with the team, then with internal beta customers, then with Joe Khoury and Justin Grover. When a command is executed and the user is opted out, the promise returned from the command will now be resolved rather than rejected. The promise will be resolved with an object, but the object may not contain all the properties it would have had the user opted in. We also decided to remove the errorsEnabled configuration option. We may add it back later, but the primary reason for it, in my opinion and if I recall correctly, was because of the sheer quantity of errors we would be throwing when the customer attempts to execute commands and the user has opted out. Now that executing a command when the user is opted out is not considered an error, the number of errors thrown will be greatly reduced and really will deserve the customer's attention.

For all commands (even commands that have nothing to do with consent, like setDebug), we will always resolve the promise with an object as long as no error is thrown. When we have data to add to the object, we will do so and the data will be segregated into their own properties on the object (for example, decisions will go under result.decisions, identities will go under result.identities). This is somewhat akin to how we have an options parameter for every command for extensibility so that we can add options/properties later if needed.

We discussed always adding another property to the result object like optedOut or even including the consent status for each purpose, so that customers could programmatically know why the result object may not contain the full data they were expecting. This has its own shortcomings, however. We suspect that in the future, when granular consent purposes are introduced, that after the user opted into some purposes and opted out of other purposes, when a command like sendEvent is executed, Alloy may still send the request to Konductor, and Konductor would have logic based on the consent status of each purpose which results in a partial response. As an arbitrary example, maybe it returns identities but not decisions) for a particular request/response due to the users being opted into one purpose but not another. In such a case, Alloy would resolve the command's promise with a result object containing only identities and not decisions.

At that point, having a property on the result object like optedOut is too broad, since the user only opted out of some purposes and not others. Including the consent status for all purposes on the result object doesn't seem to be particularly useful. For example, would the customer use it to determine whether they can access result.decisions? Why wouldn't the customer check to see if result.decisions is defined directly? If customers are wanting to programmatically react to consent-based logic that is occurring within Konductor, then the best way to do that would be to have Konductor provide additional details in its response regarding why certain data was or was not included, which Alloy could then surface on the result object. Until we have granular purposes and this becomes more of a reality, any command that requires consent in order to complete will log a warning if the user is opted out and will resolve the command promise with an empty object.

## Types of changes

## Checklist:

### DIFF
--- a/src/components/EventMerge/index.js
+++ b/src/components/EventMerge/index.js
@@ -13,18 +13,24 @@ governing permissions and limitations under the License.
 import { uuid } from "../../utils";
 import { callback } from "../../utils/validation";
 
+const generateEventMergeIdResult = () => {
+  return {
+    eventMergeId: uuid()
+  };
+};
+
 const createEventMerge = ({ config }) => {
   // #if _REACTOR
   // This is a way for the Event Merge ID data element in the Reactor extension
   // to get an event merge ID synchronously since data elements are required
   // to be synchronous.
-  config.reactorRegisterCreateEventMergeId(uuid);
+  config.reactorRegisterCreateEventMergeId(generateEventMergeIdResult);
   // #endif
 
   return {
     commands: {
       createEventMergeId: {
-        run: uuid
+        run: generateEventMergeIdResult
       }
     }
   };
--- a/src/components/Identity/createComponent.js
+++ b/src/components/Identity/createComponent.js
@@ -42,20 +42,26 @@ export default ({
       syncIdentity: {
         optionsValidator: validateSyncIdentityOptions,
         run: options => {
-          return identityManager.sync(options.identities);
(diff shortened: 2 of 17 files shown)


Aaronius:
Hold on. I still need to update functional tests.

Aaronius:
I'll open a separate PR.

PR #490, opened 2020-05-06
CORE-45102 Removed errorsEnabled. No longer failing commands due to consent.

Aaronius:
## Description
No longer reject command promises if the user has declined consent. Remove `errorsEnabled` configuration option. We may add it back later if we still find it is necessary, but now that promises aren't rejected due to consent, there will be much fewer failures and any failures that do occur probably deserve attention.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-45102

## Motivation and Context
I discussed this with the team, then with internal beta customers, then with Joe Khoury and Justin Grover. When a command is executed and the user is opted out, the promise returned from the command will now be resolved rather than rejected. The promise will be resolved with an object, but the object may not contain all the properties it would have had the user opted in. We also decided to remove the errorsEnabled configuration option. We may add it back later, but the primary reason for it, in my opinion and if I recall correctly, was because of the sheer quantity of errors we would be throwing when the customer attempts to execute commands and the user has opted out. Now that executing a command when the user is opted out is not considered an error, the number of errors thrown will be greatly reduced and really will deserve the customer's attention.

For all commands (even commands that have nothing to do with consent, like setDebug), we will always resolve the promise with an object as long as no error is thrown. When we have data to add to the object, we will do so and the data will be segregated into their own properties on the object (for example, decisions will go under result.decisions, identities will go under result.identities). This is somewhat akin to how we have an options parameter for every command for extensibility so that we can add options/properties later if needed.

We discussed always adding another property to the result object like optedOut or even including the consent status for each purpose, so that customers could programmatically know why the result object may not contain the full data they were expecting. This has its own shortcomings, however. We suspect that in the future, when granular consent purposes are introduced, that after the user opted into some purposes and opted out of other purposes, when a command like sendEvent is executed, Alloy may still send the request to Konductor, and Konductor would have logic based on the consent status of each purpose which results in a partial response. As an arbitrary example, maybe it returns identities but not decisions) for a particular request/response due to the users being opted into one purpose but not another. In such a case, Alloy would resolve the command's promise with a result object containing only identities and not decisions.

At that point, having a property on the result object like optedOut is too broad, since the user only opted out of some purposes and not others. Including the consent status for all purposes on the result object doesn't seem to be particularly useful. For example, would the customer use it to determine whether they can access result.decisions? Why wouldn't the customer check to see if result.decisions is defined directly? If customers are wanting to programmatically react to consent-based logic that is occurring within Konductor, then the best way to do that would be to have Konductor provide additional details in its response regarding why certain data was or was not included, which Alloy could then surface on the result object. Until we have granular purposes and this becomes more of a reality, any command that requires consent in order to complete will log a warning if the user is opted out and will resolve the command promise with an empty object.

## Types of changes

## Checklist:


src/core/consent/createConsentStateMachine.js
```diff
@@ -13,6 +13,13 @@ governing permissions and limitations under the License.
 import { defer } from "../../utils";
 
 export const DECLINED_CONSENT = "The user declined consent.";
+export const DECLINED_CONSENT_ERROR_CODE = "declinedConsent";
+
+const createDeclinedConsentError = () => {
+  const error = new Error(DECLINED_CONSENT);
+  error.code = DECLINED_CONSENT_ERROR_CODE;
```
Aaronius, line 20: Note that we add a code to the error. This is how, up in Core, we can segregate errors due to consent being declined versus other kinds of errors. There are a few ways we could have done this (e.g., using a different error "class" or detecting using a regex ran against the error's message instead of a checking a code), but I felt like this was a decent way of handling it.

src/core/executeCommandFactory.js
```diff
@@ -94,6 +94,13 @@ export default ({
       const executor = getExecutor(commandName, options);
       logger.log(`Executing ${commandName} command.`, "Options:", options);
       resolve(executor());
-    }).catch(handleError);
+    })
+      .then(result => {
+        // We should always be returning an object from every command.
+        return isObject(result) ? result : {};
```
Aaronius, line 100: This is important to call out. Any time a component's command does not return an object or a promise which resolves to an object, we make sure we return an object to the customer. An alternative to this is to make every component command return an object from within the component's code itself, but we run a greater risk of an engineering creating a new command and forgetting to return an object.

We always return a `result` object just to be consistent across commands.

src/core/handleErrorFactory.js
```diff
@@ -11,17 +11,20 @@ governing permissions and limitations under the License.
...
 import { toError } from "../utils";
+import { DECLINED_CONSENT_ERROR_CODE } from "./consent/createConsentStateMachine";
 
-const suppressionNote =
-  "Note: Errors can be suppressed by setting the errorsEnabled configuration option to false.";
+export default ({ instanceNamespace, logger }) => (error, commandName) => {
+  // In the case of declined consent, we've opted to not reject the promise
+  // returned to the customer, but instead resolve the promise with an
+  // empty result object.
+  if (error.code === DECLINED_CONSENT_ERROR_CODE) {
```
Aaronius, line 20: Here's where we use the code to segregate consent problems vs non-consent problems.

test/functional/specs/C2587.js
```diff
@@ -26,16 +24,13 @@ test.meta({
...
 });
 
-test.before(async () => {
+test("Test C2587: Throw error when executing command that doesn't exist", async t => {
   await configureAlloyInstance("alloy", config);
-  await bogusCommand();
-})(
-  "Test C2587: Throw error when executing command that doesn't exist",
-  async t => {
-    const { error } = await t.getBrowserConsoleMessages();
```
Aaronius, line 38: This was testing that an error was logged to the console rather than testing that the command's promise was actually rejected.

test/functional/specs/C2588.js
```diff
@@ -23,14 +12,18 @@ test.meta({
...
+    .alloy("configure", config)
+    .then(() => {}, error => error.message);
+});
 
-  await configureAlloyInstance("alloy", config);
+test("Test C2588: Throw error when configure is executed multiple times.", async t => {
+  await configureAlloyInstance(orgMainConfigMain);
+  const errorMessage = await configureAlloyInstance(orgMainConfigMain);
 
-  const { error } = await t.getBrowserConsoleMessages();
```
Aaronius, line 31: This was testing that an error was logged to the console rather than testing that the command's promise was actually rejected.

test/unit/specs/components/Identity/createComponent.spec.js
```diff
@@ -30,10 +30,10 @@ describe("Identity::createComponent", () => {
 
   beforeEach(() => {
     addEcidQueryToEvent = jasmine.createSpy("addEcidQueryToEvent");
-    identityManager = jasmine.createSpyObj("identityManager", [
-      "addToPayload",
-      "sync"
-    ]);
+    identityManager = jasmine.createSpyObj("identityManager", {
+      addToPayload: undefined,
+      sync: Promise.resolve()
```
Aaronius, line 35: This is to match how `identity` is currently functioning.

test/unit/specs/components/Identity/createComponent.spec.js
```diff
@@ -122,22 +122,22 @@ describe("Identity::createComponent", () => {
...
 
   it("getIdentity command should make a request when ecid is not available", () => {
-    let ecid;
+    const onResolved = jasmine.createSpy("onResolved");
     component.commands.getIdentity
       .run({ namespaces: ["ECID"] })
-      .then(identities => {
-        ecid = identities.ECID;
-      });
+      .then(onResolved);
```
Aaronius, line 135: I switched this to use a spy because I think it's easier to read and, if something fails (like `identity` is supposed to be `identities`, the jasmine output is easier to understand.

src/components/EventMerge/index.js
```diff
@@ -13,18 +13,24 @@ governing permissions and limitations under the License.
 import { uuid } from "../../utils";
 import { callback } from "../../utils/validation";
 
+const generateEventMergeIdResult = () => {
+  return {
```
jfkhoury, line 17: WHAT!!???? Now I have to do `result.eventMergeId`?? 🤣

test/unit/specs/core/executeCommandFactory.spec.js
```diff
@@ -116,29 +118,90 @@ describe("executeCommandFactory", () => {
...
   });
 
   it("executes the core commands", () => {
     const configureCommand = jasmine
       .createSpy()
       .and.returnValue(Promise.resolve("configureResult"));
-    const setDebugCommand = jasmine
-      .createSpy()
-      .and.returnValue(Promise.resolve("logResult"));
+    const setDebugCommand = jasmine.createSpy();
```
Aaronius, line 204: Not returning a promise or any result reflects the current reality of the `setDebugCommand`.

src/core/consent/createConsentStateMachine.js
jfkhoury: It is decent, and easy to extend and add other important codes in the future, and simple than sub classing. 👍

src/components/EventMerge/index.js
Aaronius: :rage4:

test/functional/helpers/consoleLogger/index.js
```diff
@@ -1,28 +1,113 @@
-const createConsoleLogger = (context, logLevel) => {
-  const t = context;
-  const level = logLevel;
-  let cursor = 0;
-
-  const getNewMessages = async () => {
```
jfkhoury, line 6: I think Martin's E2E test is using this `getNewMessages` helper and that's why it's failing in CI.

test/functional/specs/C2588.js
jfkhoury: Nice!

test/functional/specs/C2594.js
```diff
@@ -26,18 +29,14 @@ test.meta({
...
+  const logger = await createConsoleLogger();
+  const sendEventPromise = t.eval(() => window.alloy("sendEvent"));
   await t.eval(() => window.alloy("setConsent", { general: "out" }));
-  const errorMessage = await errorMessagePromise;
-
-  await t.expect(errorMessage).ok("Expected the event command to be rejected");
-  await t.expect(errorMessage).contains("The user declined consent.");
+  const result = await sendEventPromise;
+  await t.expect(result).eql({});
+  await logger.warn.expectMessageMatching(/user declined consent/);
```
jfkhoury, line 39: Sick!!

test/functional/specs/C35448.js
```diff
@@ -28,14 +28,20 @@ const config = compose(
...
+const getIdentity = ClientFunction(() => {
+  return window.alloy("getIdentity");
 });
 
 test("C35448 - When ID migration is enabled and Visitor is on the page, Alloy waits for Visitor to get ECID and then uses this value.", async () => {
   await configureAlloyInstance("alloy", config);
-  const visitorEcid = await getVisitorEcid(orgMainConfigMain.orgId);
-  await t.expect(getAlloyEcid()).eql(visitorEcid);
+  // Don't await the visitor ECID before executing the getIdentity command.
+  // This helps ensure that Alloy is actually waiting for Visitor.
```
jfkhoury, line 38: Nice!

test/functional/specs/C14412.js
```diff
@@ -1,41 +0,0 @@
...
-  });
-  const errorMessage = await t.eval(() =>
-    window
-      .alloy("sendEvent", { data: { a: 1 } })
-      .then(() => undefined, e => e.message)
-  );
-
-  await t.expect(errorMessage).ok("Expected the event command to be rejected");
-  await t.expect(errorMessage).contains("user declined consent");
-});
```
Aaronius, line 41: I deleted this test because it was a duplicate of what we're doing in C14414.js. I'll delete the corresponding test case in test rail assuming this PR gets merged like this.

test/functional/helpers/consoleLogger/index.js
```diff
@@ -1,28 +1,113 @@
...
-    let consoleMessages = await t.getBrowserConsoleMessages();
-    consoleMessages = consoleMessages[level];
-    const newMessages = consoleMessages.slice(cursor);
-    cursor = consoleMessages.length;
-    return newMessages;
+import { t } from "testcafe";
```
jfkhoury, line 1: Love this revamped version!!!

test/functional/specs/C2580.js
```diff
@@ -48,8 +48,7 @@ test("C2580: Command queueing test.", async () => {
   await getLibraryInfoCommand();
   await t.expect(getAlloyCommandQueueLength()).eql(2);
   const alloyLibrary = fs.readFileSync("dist/standalone/alloy.js", "utf-8");
-  const logger = createConsoleLogger(t, "log");
+  const logger = await createConsoleLogger();
   await injectScript(alloyLibrary);
-  const newMessages = await logger.getNewMessages();
-  await t.expect(newMessages).match(/Executing getLibraryInfo command/);
+  await logger.log.expectMessageMatching(/Executing getLibraryInfo command/);
```
jfkhoury, line 53: 🥇

PR #491, opened 2020-05-07
Support for target VEC click tracking metrics

ninaceban:
## Description

I have added click tracking support for target VEC click tracking metrics.
I have added an event `type` : 
- `display` for visual experiences
- `click` for click tracking metrics

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Personalization/createCollect.js
+++ b/src/components/Personalization/createCollect.js
@@ -13,7 +13,7 @@ governing permissions and limitations under the License.
 export default ({ eventManager, mergeMeta }) => {
   return meta => {
     const event = eventManager.createEvent();
-
+    event.mergeXdm({ eventType: "display" });
     mergeMeta(event, meta);
 
     return eventManager.sendEvent(event);
--- a/src/components/Personalization/createOnClickHandler.js
+++ b/src/components/Personalization/createOnClickHandler.js
@@ -12,7 +12,10 @@ governing permissions and limitations under the License.
 
 export default ({ mergeMeta, collectClicks, clickStorage }) => {
   return ({ event, clickedElement }) => {
-    const merger = meta => mergeMeta(event, meta);
+    const merger = meta => {
+      event.mergeXdm({ eventType: "click" });
+      mergeMeta(event, meta);
+    };
 
     collectClicks(merger, clickedElement, clickStorage);
   };
--- a/src/components/Personalization/createOnClickHandler.js
+++ b/src/components/Personalization/createOnClickHandler.js
@@ -12,11 +12,8 @@ governing permissions and limitations under the License.
 
 export default ({ mergeMeta, collectClicks, clickStorage }) => {
   return ({ event, clickedElement }) => {
-    const merger = meta => {
-      event.mergeXdm({ eventType: "click" });
-      mergeMeta(event, meta);
-    };
-
-    collectClicks(merger, clickedElement, clickStorage);
+    const clickMetas = collectClicks(clickedElement, clickStorage);
+    event.mergeXdm({ eventType: "click" });
+    mergeMeta(event, { decisions: clickMetas });
   };
 };
--- a/test/unit/specs/components/Personalization/createCollect.spec.js
+++ b/test/unit/specs/components/Personalization/createCollect.spec.js
@@ -18,7 +18,10 @@ describe("Personalization::createCollect", () => {
     id: 1,
     decisionId: "foo"
   };
-  const event = { type: "blah" };
+  const event = {
+    type: "blah",
+    mergeXdm: jasmine.createSpy()
+  };
 
   beforeEach(() => {
     eventManager = jasmine.createSpyObj("eventManager", {
@@ -32,6 +35,7 @@ describe("Personalization::createCollect", () => {
     const collect = createCollect({ eventManager, mergeMeta });
     collect(meta);
     expect(eventManager.createEvent).toHaveBeenCalled();
+    expect(event.mergeXdm).toHaveBeenCalledWith({ eventType: "display" });
     expect(mergeMeta).toHaveBeenCalledWith(event, meta);
     expect(eventManager.sendEvent).toHaveBeenCalled();
   });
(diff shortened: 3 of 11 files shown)


test/unit/specs/components/Personalization/createExecuteDecisions.spec.js
```diff
@@ -143,4 +143,28 @@ describe("Personalization::createExecuteDecisions", () => {
       expect(logger.error).toHaveBeenCalledWith(error);
     });
   });
+
+  it("should not trigger collect when dom-action click", () => {
+    executeActions = jasmine
+      .createSpy()
+      .and.returnValues([undefined], [undefined]);
+    const spy = jasmine.createSpy();
```
Aaronius, line 151: I noticed these module spies (in this and other tests) aren't actually used for spying and maybe not used for anything at all. Maybe do this instead if you need `foo` to be a function:

```
const modules = {
  foo() {}
}
```

Or if you do need foo to be a spy for one of the tests:

```
const modules = jasmine.createSpyObj("modules", ["foo"]);
```

test/unit/specs/components/Personalization/createOnClickHandler.spec.js
```diff
@@ -14,8 +14,11 @@ import createOnClickHandler from "../../../../../src/components/Personalization/
 describe("Personalization::createOnClickHandler", () => {
   let mergeMeta;
   let collectClicks;
-  const event = { type: "blah" };
-  const clickStorage = {};
+  const event = {
+    type: "blah",
+    mergeXdm: jasmine.createSpy()
+  };
+  const clickStorage = [];
```
Aaronius, line 21: It doesn't seem like anything is testing the `merger` function that the `createOnClickHandler` creates and passes to `collectClicks`. Rather than just checking that the first argument to `collectClicks` is a function of any kind, you should instead call the function and ensure that `event.mergeXdm` and `mergeMeta` get called as a result.

test/unit/specs/components/Personalization/createOnClickHandler.spec.js
```diff
@@ -14,8 +14,11 @@ import createOnClickHandler from "../../../../../src/components/Personalization/
 describe("Personalization::createOnClickHandler", () => {
   let mergeMeta;
   let collectClicks;
-  const event = { type: "blah" };
-  const clickStorage = {};
+  const event = {
+    type: "blah",
```
Aaronius, line 18: Is `type: "blah"` here serving any purpose?

test/unit/specs/components/Personalization/createCollect.spec.js
```diff
@@ -18,7 +18,10 @@ describe("Personalization::createCollect", () => {
     id: 1,
     decisionId: "foo"
   };
-  const event = { type: "blah" };
+  const event = {
+    type: "blah",
```
Aaronius, line 22: Is `type: "blah"` here serving any purpose?

src/components/Personalization/createCollect.js
```diff
@@ -13,7 +13,7 @@ governing permissions and limitations under the License.
 export default ({ eventManager, mergeMeta }) => {
   return meta => {
     const event = eventManager.createEvent();
-
+    event.mergeXdm({ eventType: "display" });
```
jonsnyder, line 16: We should discuss adding an eventType to the known list in xdm, here: https://github.com/adobe/xdm/blob/b092f1c71966d1c42ca59f8f863379fdaf033bc5/schemas/data/time-series.schema.json#L35

src/components/Personalization/createOnClickHandler.js
```diff
@@ -12,7 +12,10 @@ governing permissions and limitations under the License.
 
 export default ({ mergeMeta, collectClicks, clickStorage }) => {
   return ({ event, clickedElement }) => {
-    const merger = meta => mergeMeta(event, meta);
+    const merger = meta => {
+      event.mergeXdm({ eventType: "click" });
```
jonsnyder, line 16: {eventType: "web.webinteraction.linkClicks"} is what we have used for the ActivityMap stuff.

src/components/Personalization/createCollect.js
ninaceban: We have discussed the `eventType` and we know that we don't have yet XDM mixins for it. We'll go with some defaults and work on XDM mixins later on.

ninaceban: @jfkhoury @dancingcactus

test/unit/specs/components/Personalization/createExecuteDecisions.spec.js
ninaceban: yes, I'll refactor these tests. Thank you

test/unit/specs/components/Personalization/createOnClickHandler.spec.js
ninaceban: I'll add more coverage. Thank you

src/components/Personalization/createOnClickHandler.js
ninaceban: True, but this is not really a `linkClick` it can be anything, so ideally XDM should allow us to send either freeform event type or we should add other types along `linkClick`. This should apply to mobile as well.

test/unit/specs/components/Personalization/createCollect.spec.js
ninaceban: Removed.

test/unit/specs/components/Personalization/createOnClickHandler.spec.js
ninaceban: Removed

ninaceban: Removed.

ninaceban:
@Aaronius and @jfkhoury we discussed `display` and `click` for `eventType` with David when we presented the proposed changes for notifications. The decision was to use this for now, once we figure out the XDM piece we will adjust it.

PR #493, opened 2020-05-09
Adding support for redirect items (DON'T MERGE YET)

ninaceban:
Adding support for redirect items

## Description

The personalization component is now able to handle redirect items. The way this is implemented is the following:
- if we have `renderDecisions` and we get a redirect item for page-wide scope, then we will skip all other items processing, we will mark the notification event as `documentMayUnload` and we will execute a `window.location.replace()`.
- in all other cases, we will just return a list of decisions with all the items

## Related Issue

None

## Motivation and Context

Supporting redirect items is one of the basic uses cases

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Personalization/createRedirectDecisionHandler.js
+++ b/src/components/Personalization/createRedirectDecisionHandler.js
@@ -9,16 +9,21 @@ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTA
 OF ANY KIND, either express or implied. See the License for the specific language
 governing permissions and limitations under the License.
 */
-import { isNonEmptyArray } from "../../utils";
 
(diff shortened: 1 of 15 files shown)


src/components/Personalization/createRedirectDecisionHandler.js
```diff
@@ -0,0 +1,29 @@
...
+  const { content } = items[0].data;
+
+  return { content, meta: { id, scope } };
+};
+
+export default ({ collect, win = window }) => {
+  return redirectDecisions => {
+    const { content, meta } = getRedirectDetails(redirectDecisions);
+
+    return collect({ decisions: [meta] }, true).then(() => {
```
Aaronius, line 25: I recommend making the parameter to `collect` be an options object so you don't end up with this type of scenario where the reader doesn't know what `true` means without looking at the `collect` code.

PR #497, opened 2020-05-11
CORE-45716 - sendEvent command promise is resolved with undefined

ninaceban:
Fixing the result of `sendEvent` command promise

## Description

Fixing the result of `sendEvent` command promise

## Related Issue

NONE

## Motivation and Context

Fixing the result of `sendEvent` command promise

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Identity/createComponent.js
+++ b/src/components/Identity/createComponent.js
@@ -1,4 +1,5 @@
 import getIdentityOptionsValidator from "./getIdentity/getIdentityOptionsValidator";
+import { noop } from "../../utils";
 
 export default ({
   addEcidQueryToEvent,
@@ -35,7 +36,7 @@ export default ({
           }
         }
 
-        return handleResponseForIdSyncs(response);
+        return handleResponseForIdSyncs(response).then(noop);
       }
     },
     commands: {
--- a/src/components/Identity/createComponent.js
+++ b/src/components/Identity/createComponent.js
@@ -1,5 +1,4 @@
 import getIdentityOptionsValidator from "./getIdentity/getIdentityOptionsValidator";
-import { noop } from "../../utils";
 
 export default ({
   addEcidQueryToEvent,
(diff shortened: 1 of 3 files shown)


Aaronius:
It should already be handled systemically. @ninaceban, can you help figure out why this code isn't causing an object to return instead of `undefined`: https://github.com/adobe/alloy/blob/610f1203b5f2c8bc0926c2127898f9869c105608/src/core/executeCommandFactory.js#L99-L100

src/components/Identity/createComponent.js
```diff
@@ -35,7 +36,7 @@ export default ({
           }
         }
 
-        return handleResponseForIdSyncs(response);
+        return handleResponseForIdSyncs(response).then(noop);
```
Aaronius, line 39: @ninaceban I think I understand the problem now. It seems that `processIdSyncs` is resolving the returned promise with an array of `undefined` values. I think the proper thing to do here would be to move the `.then(noop)` to inside `processIdSyncsFactory`. The reasoning is that an array of `undefined` values isn't useful outside that module, so we should be keep it from leaking out of that module. If I'm understanding this right.

ninaceban: yes, thats right, that promise gets resolved with an array of `undefined`. I'll change `then(noop)` to `processIdSyncsFactory`. Thank you

test/unit/specs/components/Identity/createComponent.spec.js
```diff
@@ -111,7 +115,7 @@ describe("Identity::createComponent", () => {
     const response = { type: "response" };
     const result = component.lifecycle.onResponse({ response });
     expect(handleResponseForIdSyncs).toHaveBeenCalledWith(response);
-    expect(result).toBe(idSyncsPromise);
+    expectAsync(result).toBeResolvedTo(response);
```
Aaronius, line 118: I think you have to do `return expectAsync`

ninaceban: thank you!

PR #526, opened 2020-06-05
Monitor support

jonsnyder:
## Description

Allow a better integration with external debugging tools like the Adobe Debugger.

## Related Issue

https://jira.corp.adobe.com/browse/CORE-43399

## Motivation and Context

Adobe debugger is currently using some fragile integrations to capture this data.

## Types of changes

## Checklist:

### DIFF
--- a/src/core/createLogger.js
+++ b/src/core/createLogger.js
@@ -13,9 +13,9 @@ governing permissions and limitations under the License.
 import { assign } from "../utils";
 
 export default ({ getDebugEnabled, console, getMonitors, context }) => {
-  let prefix = `[${context.instanceNamespace}]`;
-  if (context.componentNamespace) {
-    prefix += ` [${context.componentNamespace}]`;
+  let prefix = `[${context.instanceName}]`;
+  if (context.componentName) {
+    prefix += ` [${context.componentName}]`;
   }
 
   const notifyMonitors = (method, data) => {
--- a/src/core/createLogger.js
+++ b/src/core/createLogger.js
@@ -31,6 +31,7 @@ export default ({ getDebugEnabled, console, getMonitors, context }) => {
   };
 
   const log = (level, ...rest) => {
+    notifyMonitors("onBeforeLog", { level, arguments: rest });
     if (getDebugEnabled()) {
       console[level](prefix, ...rest);
     }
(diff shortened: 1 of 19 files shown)


src/core/createLogger.js
```diff
@@ -10,38 +10,107 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+      notifyMonitors("onInstanceConfigured", data);
+      log("info", "Instance configured. Computed configuration:", data.config);
+    },
+    logOnBeforeCommand(data) {
+      notifyMonitors("onBeforeCommand", data);
+      log(
+        "info",
+        `Executing ${data.commandName} command. Options:`,
+        data.options
+      );
```
Aaronius, line 57: Considering this will log a message to the console, shouldn't monitors' `onBeforeLog` also be called? Maybe I'm overlooking where that's happening.

src/core/index.js
```diff
@@ -39,25 +39,33 @@ import validateNetworkResponseIsWellFormed from "./edgeNetwork/validateNetworkRe
...
-const instanceNamespaces = window.__alloyNS;
+const instanceNames = window.__alloyNS;
 
 const createNamespacedStorage = injectStorage(window);
 
 const { console } = window;
 
+// set this up as a function so that monitors can be added at anytime
+// eslint-disable-next-line no-underscore-dangle
+const getMonitors = () => window.__alloyMonitors || [];
```
jfkhoury, line 50: Let's add a comment OR an entry in the Tech WIKI regarding the interface of the monitors.

src/core/createLogger.js
```diff
@@ -10,38 +10,107 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-export default (console, getDebugEnabled, prefix) => {
-  const process = (level, ...rest) => {
+import { assign } from "../utils";
```
jfkhoury, line 13: I love this change!

src/core/createLogger.js
```diff
@@ -10,38 +10,107 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+        data.error
+      );
     },
     /**
      * Outputs a message to the web console.
      * @param {...*} arg Any argument to be logged.
      */
-    log: process.bind(null, "log"),
+    log(...args) {
+      notifyMonitors("onBeforeLog", { level: "log", arguments: args });
```
jfkhoury, line 86: @Aaronius It's happening here.

Aaronius: `logOnBeforeCommand` isn't calling this `log` function though.

test/unit/specs/core/injectExecuteCommand.spec.js
```diff
@@ -219,4 +225,6 @@ describe("injectExecuteCommand", () => {
       expect(setDebugResult).toEqual({});
     });
   });
+
+  it("logs onBeforeCommand", () => {});
```
Aaronius, line 229: Should this be filled out?

src/core/createLogger.js
jonsnyder: I was thinking that the monitor was already notified of the event so then didn't need the log notification as well.  I think you are right though.

test/unit/specs/core/injectExecuteCommand.spec.js
jonsnyder: Yes, good catch.

PR #535, opened 2020-06-16
Remove syncIdentity command

ninaceban:
## Description

Removed syncIdentity command and cleaned up the functional and unit tests
## Types of changes

## Checklist:


src/components/Identity/createComponent.js
```diff
@@ -39,12 +38,6 @@ export default ({
       }
     },
     commands: {
-      syncIdentity: {
-        optionsValidator: validateSyncIdentityOptions,
-        run: options => {
-          return identityManager.sync(options.identity);
```
liljenback, line 45: So there is nothing in the identityManager that is specific of how the options were formatted in the user exposed command that we should remove as well?

jfkhoury: Great question. I think we should remove `identityManager.sync` as well since we won't be using it in the future.

jfkhoury: I mean we can even remove all the hashing and validation logic, and bring them back once we implement Phase 1 which adds a new config that specifies which Identity Namespaces the customer would like to mutate (Hash...)

@ninaceban What do you think?

ninaceban: will remove them

jfkhoury:
@ninaceban It looks like the E2E tests are failing, the ID Migration tests specifically. Can you pls build:prod locally and run the tests to see what's going on?

PR #537, opened 2020-07-07
CORE-47269 - Unhide the body if no personalization payload

ninaceban:
## Description

The issue was happening on the pages with the prehiding snippet and where `sendEvent` command fires at page load with `renderDecisions: false; decisionsScope:[]`.
Also, I have changed the behavior of `onResponsehandler` so that if personalization payload is empty we unhide the body immediately and return from the lifecycle.
## Types of changes

## Checklist:

### DIFF
--- a/src/components/Personalization/createComponent.js
+++ b/src/components/Personalization/createComponent.js
@@ -47,6 +47,7 @@ export default ({
         const scopes = getDecisionScopes(renderDecisions, decisionScopes);
 
         if (!hasScopes(scopes)) {
+          showContainers();
           return;
         }
 
--- a/src/components/Personalization/createOnResponseHandler.js
+++ b/src/components/Personalization/createOnResponseHandler.js
@@ -15,16 +15,14 @@ const DECISIONS_HANDLE = "personalization:decisions";
 
 export default ({ extractDecisions, executeDecisions, showContainers }) => {
   return ({ renderDecisions, response }) => {
-    const decisionsResponse = response.getPayloadsByType(DECISIONS_HANDLE);
-    if (!isNonEmptyArray(decisionsResponse)) {
+    const unprocessedDecisions = response.getPayloadsByType(DECISIONS_HANDLE);
+    if (!isNonEmptyArray(unprocessedDecisions)) {
       showContainers();
       return { decisions: [] };
     }
-    const [
-      renderableDecisions,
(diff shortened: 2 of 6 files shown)


src/components/Personalization/extractDecisions.js
```diff
@@ -56,8 +54,6 @@ const splitDecisions = (decisions, predicate) => {
   return [matchedDecisions, nonMatchedDecisions, decisions];
```
jonsnyder, line 54: No need to return decisions here as it is the same as the parameter that you passed in.

src/components/Personalization/createComponent.js
```diff
@@ -47,6 +47,7 @@ export default ({
         const scopes = getDecisionScopes(renderDecisions, decisionScopes);
 
         if (!hasScopes(scopes)) {
+          showContainers();
```
jonsnyder, line 50: Isn't there an implicit "__view__" scope?  How would a user just get the "__view__" scope and render those decisions?

src/components/Personalization/createOnResponseHandler.js
```diff
@@ -9,14 +9,22 @@ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTA
...
   return ({ renderDecisions, response }) => {
+    const decisionsResponse = response.getPayloadsByType(DECISIONS_HANDLE);
+    if (!isNonEmptyArray(decisionsResponse)) {
+      showContainers();
+      return { decisions: [] };
+    }
     const [
       renderableDecisions,
       decisions,
       unprocessedDecisions
```
jonsnyder, line 26: unprocessedDecisions === decisionsResponse, so no need to set it here.

src/components/Personalization/extractDecisions.js
ninaceban: yeah, good catch, thank you

src/components/Personalization/createComponent.js
ninaceban: yes, `getDecisionScopes(renderDecisions, decisionScopes);` if renderDecisions is `true` we add `__view__` scope in `scopes` array.

jonsnyder: ah, I see.

PR #543, opened 2020-07-09
Query for the ECID in set Consent command.

jfkhoury:
## Description

The `setConsent` command was not querying for the ECID, which was affecting the ID Migration workflow.

## Related Issue

https://jira.corp.adobe.com/browse/CORE-48247

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Identity/createComponent.js
+++ b/src/components/Identity/createComponent.js
@@ -2,21 +2,21 @@ import getIdentityOptionsValidator from "./getIdentity/getIdentityOptionsValidat
 
 export default ({
   addEcidQueryToPayload,
-  ensureRequestHasIdentity,
+  ensureSingleIdentity,
   setLegacyEcid,
   handleResponseForIdSyncs,
   getEcidFromResponse,
   getIdentity,
   consent
 }) => {
-  let ecid; 
+  let ecid;
   return {
     lifecycle: {
       onBeforeRequest({ payload, onResponse }) {
-        // Querying the ECID at the payload level so would return it on all requests,
-        // like `setConsent`.
+        // Querying the ECID on every request to be able to set the legacy cookie, and make it
+        // available for the `getIdentity` command.
         addEcidQueryToPayload(payload);
-        return ensureRequestHasIdentity({ payload, onResponse });
+        return ensureSingleIdentity({ payload, onResponse });
       },
       onResponse({ response }) {
         if (!ecid) {
--- a/src/components/Identity/createComponent.js
+++ b/src/components/Identity/createComponent.js
@@ -1,25 +1,22 @@
 import getIdentityOptionsValidator from "./getIdentity/getIdentityOptionsValidator";
 
 export default ({
-  addEcidQueryToEvent,
-  ensureSingleIdentity,
+  addEcidQueryToPayload,
+  ensureRequestHasIdentity,
   setLegacyEcid,
   handleResponseForIdSyncs,
   getEcidFromResponse,
   getIdentity,
   consent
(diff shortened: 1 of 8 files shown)


src/components/Privacy/injectSendSetConsentRequest.js
```diff
@@ -10,12 +10,19 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
 
+import ecidNamespace from "../../constants/ecidNamespace";
+
 export default ({
   createConsentRequestPayload,
   sendEdgeNetworkRequest
 }) => consent => {
   const payload = createConsentRequestPayload();
   payload.setConsent(consent);
+  payload.mergeQuery({
```
jfkhoury, line 21: I thought about doing this in the `onBeforeRequest` lifecycle hook instead, but thought we might not need to query for the ECID in ALL requests.. or do we? @jonsnyder @Aaronius @jileeshs

jonsnyder: I agree that this would be better put within the identity component.  Currently the query is done at the event level inside "onBeforeEvent", but according to this ticket: https://jira.corp.adobe.com/browse/EXEG-1234, the query can be at the payload level.  I propose adding a "mergeQuery" method to "createRequestPayload", and then within the identity component, it can add the ecid query inside the "onBeforeRequest" method.

Aaronius: @jfkhoury Sorry for not answering sooner. My understanding is that any request could result in an ECID being established for the user. If we always want the legacy ECID to be created whenever an ECID is established for a user, then I think we _have_ to request the ECID in all requests as long as the legacy ECID hasn't been established previously.

I agree with what I think is being said though, which is that it's weird that Identity isn't the one responsible for requesting the ECID in this PR. It should be Identity's responsibility and not Privacy's.

Aaronius: I see you switched it. Nice. 👍

src/components/Identity/createComponent.js
```diff
@@ -12,13 +12,10 @@ export default ({
...
     lifecycle: {
-      // TODO: It would probably be best to query on the data collection payload level
-      // rather than the event. It seems like a payload-level thing and would save
-      // space whenever we start supporting multiple events per payload.
-      onBeforeEvent({ event }) {
-        addEcidQueryToEvent(event);
-      },
       onBeforeRequest({ payload, onResponse }) {
+        // Querying the ECID at the payload level so would return it on all requests,
+        // like `setConsent`.
```
jonsnyder, line 17: I think you missed a word in this comment.  We should add in the comment why we need to get it:
1. to set the legacy ecid cookie.
2. to make it available for the getIdentity command.

And also add in the comment that the ecid can change in response to changes in the identityMap. (i.e. when a user logs in)

test/unit/specs/components/Privacy/injectSendSetConsentRequest.spec.js
```diff
@@ -11,7 +11,7 @@ describe("Privacy:injectSendSetConsentRequest", () => {
       "createConsentRequestPayload"
     );
     sendEdgeNetworkRequest = jasmine.createSpy("sendEdgeNetworkRequest");
-    payload = jasmine.createSpyObj("payload", ["setConsent"]);
+    payload = jasmine.createSpyObj("payload", ["setConsent", "mergeQuery"]);
```
jonsnyder, line 14: This change is no longer necessary

PR #547, opened 2020-07-13
CORE-47161 call onRequestFailure for 400 & 500 responses

jonsnyder:
## Description

This changes the behavior of the onResponse and onRequestFailure callbacks:
* onResponse - called when Konductor returns a 2** response
* onRequestFailure - called for all other scenarios. (request failure and 400/500 errors)

Update all the components to use onRequestFailure appropriately
## Related Issue

https://jira.corp.adobe.com/browse/CORE-47161

## Motivation and Context

The response handler for the identity component was throwing an error on a 400 status code because there was no identity set on the response.  This was showing an identity error to the user, when the real problem was that the request was malformed.

## Types of changes

## Checklist:


jonsnyder:
Looking over this now, what do you think about changing "onResponse" to "onRequestSuccess"?

src/components/Audiences/index.js
```diff
@@ -18,12 +18,17 @@ const createAudiences = ({ logger }) => {
...
+    return processDestinations(destinations);
+  };
   return {
     lifecycle: {
-      onResponse({ response }) {
-        const destinations = response.getPayloadsByType("activation:push");
-        return processDestinations(destinations);
-      }
+      onResponse: handleResponse,
+      onRequestFailure: handleResponse
```
ninaceban, line 31: maybe we should rename `handleResponse` to something more generic that applies to both lyfecycles?

src/components/Identity/injectAwaitIdentityCookie.js
```diff
@@ -38,6 +38,15 @@ export default ({ orgId, doesIdentityCookieExist }) => {
           throw noIdentityCookieError;
         }
       });
+      onRequestFailure(() => {
+        if (doesIdentityCookieExist()) {
+          resolve();
+        } else {
+          // The error from the request failure will be logged separately. Rejecting this here
+          // will tell ensureSingleIdentity to send the next request without identity
+          reject(new Error("No identity was set on response."));
```
jfkhoury, line 47: nice

src/core/edgeNetwork/injectSendEdgeNetworkRequest.js
```diff
@@ -74,11 +74,18 @@ export default ({
...
         // Note that networkResponse.parsedBody may be undefined if it was a
         // 204 No Content response. That's fine.
-        const response = createResponse(networkResponse.parsedBody);
+        const response = createResponse(parsedBody);
         cookieTransfer.responseToCookies(response);
+
+        if (statusCode >= 400) {
+          const throwError = () => processWarningsAndErrors(response);
+          return onRequestFailureCallbackAggregator
+            .call({ response })
```
Aaronius, line 86: So it looks like the onRequestFailure lifecycle hook will get called with an object that contains either `error` or `response`. That feels a little weird, but I get it and don't have a better alternative. It would probably be good to document that in `createLifecycle` where we have a little bit of documentation about each hook.

PR #551, opened 2020-07-14
Set up CI with Azure Pipelines

shammowla:


jfkhoury:
@shammowla Can you please add a comment to this PR to understand why we need it?

PR #564, opened 2020-08-04
POC - native VEC support for Target - postMessage impl

dcottingham:
This is POC to show how Alloy can support loading external scripts using postMessage when run in an iframe. This is necessary in order to support the Target visual experience composer (VEC). More details here:
https://wiki.corp.adobe.com/pages/viewpage.action?spaceKey=Target&title=Alloy+VEC+Integration

* Once the design is agreed upon I will create test cases.

## Checklist:

### DIFF
--- a/src/core/preloadHook.js
+++ b/src/core/preloadHook.js
@@ -1,5 +1,7 @@
 const ALLOW_LIST = [
-  "http://localhost.corp.adobe.com:4502" // TODO - Target UI
+  "http://localhost.corp.adobe.com:4502",
+  "http://experience.adobe.com",
+  "https://atag.experiencecloud.adobe.com" // TODO - Target UI all envs incl. shell
 ];
 
 let preloadListener;
--- a/src/core/preloadHook.js
+++ b/src/core/preloadHook.js
@@ -1,7 +1,10 @@
+import loadScript from "@adobe/reactor-load-script";
+
+// TODO
 const ALLOW_LIST = [
   "http://localhost.corp.adobe.com:4502",
   "http://experience.adobe.com",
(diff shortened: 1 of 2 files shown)


src/core/preloadHook.js
```diff
@@ -0,0 +1,44 @@
...
+    if (
+      event.data.type !== "preloadHook" ||
+      !ALLOW_LIST.includes(event.origin)
+    ) {
+      return;
+    }
+
+    try {
+      // eslint-disable-next-line no-eval
+      window.eval(event.data.value);
```
jonsnyder, line 24: This is a big red flag for any security scanner.

dcottingham: So there are two overall goals I'm trying to achieve - outlined in this wiki I put together:  https://wiki.corp.adobe.com/pages/viewpage.action?spaceKey=Target&title=Alloy+VEC+Integration

The SPA support - which is what uses the monitoring hooks - does use postMessage to send those events to the VEC in the parent window.  That will work fine without any explicit changes to Alloy.  

This PR addresses the second goal - native VEC support.  So basically we need some means of executing our target.js script inside of the iframe before Alloy instances are created.  Given the constraints, I can't think of any way to implement this where Alloy doesn't use eval to run the script.

dcottingham: I've updated this to use the reactor-load-script instead of eval

jonsnyder: Wasn't one of the goals not requiring the customer to allow scripts loaded from target?

jfkhoury:
@dcottingham Can we close this PR at this point?

dcottingham:
Going with browser extension instead of script injection

PR #574, opened 2020-08-12
Switch to using Pre Prod instead of int for testing. 

shammowla:
1) Switch to using Pre Prod instead of int for testing. 
2) Fixed Target VEC test cases by adding "latestAlloyTestPage.html" end points.


package.json
```diff
@@ -16,9 +16,9 @@
...
     "test:watch:chrome": "karma start --browsers Chrome",
-    "functional:ci": "EDGE_ENV=\"int\" testcafe chrome",
-    "functional:ci:bs": "EDGE_ENV=\"int\" testcafe 'browserstack:ie@11:Windows 10'",
-    "functional:local": "EDGE_ENV=\"int\" testcafe chrome",
+    "functional:ci": "EDGE_ENV=\"prod\" EDGE_BASE_PATH=\"ee-pre-prd\" ALLOY_ENV=\"prod\" testcafe chrome", 
+    "functional:ci:bs": "EDGE_ENV=\"prod\" EDGE_BASE_PATH=\"ee-pre-prd\" ALLOY_ENV=\"prod\" testcafe 'browserstack:ie@11:Windows 10'",
+    "functional:local": "EDGE_ENV=\"prod\" EDGE_BASE_PATH=\"pre-prd\" testcafe chrome",
     "functional:cron:prod": "EDGE_ENV=\"prod\" ALLOY_ENV=\"prod\" testcafe chrome",
     "functional:local:allure": "EDGE_ENV=\"int\" testcafe chrome -r allure",
     "functional:local:headless": "EDGE_ENV=\"int\" testcafe chrome:headless",
```
jfkhoury, line 24: Let's do the same here pls, and `allure`.

PR #581, opened 2020-08-21
 Personalization component - SPA Support 

ninaceban:
## Description

So that we had feature parity with what at.js 2.x offers now, we have to add SPA support for Target. 
How it works right now using at.js:
Suppose you have 4 views: home, products, cart, and about.
When the page is loading we fetch page-load decisions and decisions for all views customer has.
All the decisions that are for "page-wide" scope, aka targetGlobalMbox we auto-render and we send a "display" notification call with decision ID and scope. Decisions for other views are cached in memory. At a view change customer has to implement something like this:
``` 
history.listen(location => {
  const viewName = location.pathname.split("/")[1];
  adobe.target.triggerView(viewName);
});
```
This means that at every view change we will look up in cache for decisions based on a viewName and render them, then we will send a display notification call. 
Also, VEC is listening to this API (`adobe.target.triggerView`) and recognizes a view and is coupling decisions to a viewName. 

We should have pretty much the same approach in Alloy.
The customer will have to trigger a `sendEvent` command at page load with `renderDecisions: true`, this will fetch from target the `page-load` decisions and will prefetch the views decisions.  
We'll auto-render `__view__` scope decisions and store in cache other views decisions.
Similar to `triggerView` in at.js, customer will have to trigger a `sendEvent` command with `xdm.web.webPageDetails.viewName` and `renderDecisions: true`.
``` 
alloy("sendEvent", {
  renderDecisions: true,
  xdm: {
    web: {
      webPageDetails: {
        viewName: viewName
       }
    },
    eventType: "page-view"
  }
});
```
At every view change we will look up in cache and render decisions for the `xdm.web.webPageDetails.viewName` view

## Types of changes

## Checklist:


src/components/DataCollector/index.js
```diff
@@ -11,6 +11,25 @@ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTA
 OF ANY KIND, either express or implied. See the License for the specific language
 governing permissions and limitations under the License.
 */
+const getViewName = xdm => {
```
jfkhoury, line 14: This seems like a fun use case for a util that returns object properties safely instead of all the ifs :) 

Like a util with a try/catch that takes a callback?

src/components/Personalization/createComponent.js
```diff
@@ -44,6 +46,11 @@ export default ({
           return;
         }
 
+        if (renderDecisions && viewName) {
```
jfkhoury, line 49: Let's add an inline comment pls to explain why we're doing this.

src/components/Personalization/createViewCollect.js
```diff
@@ -0,0 +1,35 @@
...
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+import { isNonEmptyArray, assign } from "../../utils";
+
+export default ({ eventManager, mergeMeta }) => {
+  return ({ meta, xdm = {} }) => {
+    const { decisions = [] } = meta;
+    const data = { eventType: "display" };
```
jfkhoury, line 18: Do we need a more descriptive event type here? `display` can mean anything.

jfkhoury: Or is it because we are sending this information in `data`, only Target cares about this? Which I don't understand why we are not logging this as a regular experience event.

src/components/Personalization/createViewStorage.js
```diff
@@ -0,0 +1,22 @@
+const viewStorage = {};
```
jfkhoury, line 1: Can we pls add a comment for this module, and do we need a `clearStorage` api?

jfkhoury (commented):
Awesome work; I added few comments here and there. I think in general we need few more inline comments & functional tests.

src/components/Personalization/createViewChangeHandler.js
```diff
@@ -0,0 +1,28 @@
...
+    const viewDecisions = getView(viewName);
+
+    if (isNonEmptyArray(viewDecisions)) {
+      executeViewDecisions(viewDecisions);
+      return;
+    }
+
+    const xdm = { web: { webPageDetails: { viewName } } };
+
+    collect({ meta: {}, xdm });
```
jonsnyder, line 26: Why send a display event when no decisions were rendered?

src/components/Personalization/createViewStorage.js
jonsnyder, line 1: When there are multiple instances, the viewStorage should be unique to the instance. You can achieve this by putting the viewStorage, addView, and storeViews declarations inside the default function.

src/components/Personalization/constants/scope.js
```diff
@@ -0,0 +1,13 @@
+/*
```
jonsnyder, line 1: Consider renaming this file to "pageWideScope.js"

src/components/Personalization/createViewCollect.js
jonsnyder, line 18: We should add this to the xdm as a known event type. Perhaps it should be more specific like "personalization.display"

Also, calling this variable "data" is confusing, because this is actually merged into the xdm. Consider "viewXdm"

src/components/Personalization/createViewCollect.js
```diff
@@ -0,0 +1,35 @@
...
+      const viewName = decisions[0].scope;
+
+      data.web = {
+        webPageDetails: { viewName }
+      };
+
+      mergeMeta(event, meta);
+    }
+
+    event.mergeXdm(assign(data, xdm));
```
jonsnyder, line 31: I'm thinking you want a deepAssign here. You could do that with this:

event.mergeXdm(data);
event.mergeXdm(xdm);

jonsnyder (commented):
Looking good. I noted a few suggestions regarding naming and simplifying code. My main concerns are:

1. Sending a display event when there are no decisions to render for that view.
2. Keeping the view cache for different instances separated.

Also consider adding some functional tests.

src/components/Personalization/extractPageWideScopeDecisions.js
```diff
@@ -0,0 +1,25 @@
...
+      pageWideDecisions.push(decision);
+    } else {
+      if (!viewDecisions[decision.scope]) {
+        viewDecisions[decision.scope] = [];
+      }
+      viewDecisions[decision.scope].push(decision);
+    }
+  });
+
+  return [pageWideDecisions, viewDecisions];
```
Aaronius, line 20: Are we returning an array instead of an object in attempt to optimize? It seems like an object would have been the default choice.

Aaronius (commented):
I agree with what @jfkhoury and @jonsnyder said. From today's call, there's a design problem you were going to take to the Personalization team, so I'll wait to approve until that's settled.

src/components/Personalization/createOnResponseHandler.js
```diff
@@ -10,22 +10,41 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
-      unprocessedDecisions
-    );
 
     if (renderDecisions) {
-      executeDecisions(renderableDecisions);
+      const [renderableDecisions, decisions] = extractRenderableDecisions(
+        unprocessedDecisions
+      );
+
+      if (isNonEmptyArray(renderableDecisions)) {
```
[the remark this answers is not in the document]
Aaronius: Yeah, I tend to think we're abusing this `isNonEmptyArray` utility a bit. Every time I see `isNonEmptyArray` I think, "When would it not be an array?" and then I dig into it and the answer seems to be that it _is_ always an array. In other words, we're really just using the `isNonEmpty` part of `isNonEmptyArray`, which can simply be done by checking `.length` and would make it clear that `renderableDecisions` will always be an array.

src/components/Personalization/createViewChangeHandler.js
ninaceban: We send a notification of a view "displayed" even if we didn't have anything in the cache to render for that specific view. This is needed for Target reporting.

src/components/Personalization/createOnResponseHandler.js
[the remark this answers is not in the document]
ninaceban: I see! will check the length only if it is an array!

src/components/Personalization/createViewCollect.js
ninaceban: this is not something that has been introduced recently. The notifications for `__view__` are already sent as `xdm.eventType: "display"` and it was discussed in the wiki a long time ago.

src/components/DataCollector/index.js
ninaceban: ok, I will extract it as a util

src/components/Personalization/createViewCollect.js
ninaceban: `xdm.eventType: "display"` has been added a while ago, this is not something that has been introduced in this PR. Konductor has logic around this `eventType` so we can't simply rename it.

src/components/Personalization/createViewCollect.js
ninaceban: ok thank you

src/components/Personalization/constants/scope.js
ninaceban: I have renamed it to `scope` also as part of a review.

src/components/Personalization/createViewStorage.js
```diff
@@ -0,0 +1,22 @@
...
+
+const addView = (viewName, decisions) => {
+  viewStorage[viewName] = decisions;
+};
+
+const storeViews = decisions => {
+  Object.keys(decisions).forEach(scope => {
+    addView(scope, decisions[scope]);
+  });
```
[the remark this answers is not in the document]
ninaceban: ok, thank you

src/components/Personalization/createViewCacheManager.js
```diff
@@ -0,0 +1,37 @@
...
+      viewStorage = {};
+    }
+    assign(viewStorage, decisions);
+  };
+
+  const getView = viewName => {
+    return viewStorage[viewName];
+  };
+
+  const isInitialized = () => {
```
jonsnyder, line 29: This implementation doesn't take into account when the cache is initializing, i.e. a request has gone out to fetch decisions that will be put in the cache. If I were doing this, I'd setup the cache with promises like this:

```
export default () => {
  let viewStorage;
  let viewStorageDeferred = defer();

  const storeViews = decisionsPromise => {
    if (viewStorage === undefined) {
      viewStorage = {};
    }
    decisionsPromise.then((decisions) => {
      assign(viewStorage, decisions);
      viewStorageDeferred.resolve();
    })
  }

  const getView = viewName => {
    viewStorageDeferred.then(() => {
      return viewStorage[viewName];
    })
  };

  const isInitialized = () => {
    return !(viewStorage === undefined);
  }

  return {
    storeViews,
    getView,
    isInitialized
  };
};
```

So with this implementation, storeViews accepts a promise that should return the decisions to store. This should be called before the fetchDecisions request returns. Also the getView returns a promise so that if the cache is initializing you can wait for the first request to return.

src/components/Personalization/createViewChangeHandler.js
```diff
@@ -0,0 +1,32 @@
...
+    if (personalization.isRenderDecisions()) {
+      executeCachedViewDecisions({
+        viewName: personalization.getViewName()
+      });
+      return;
+    }
+
+    onResponse(() => {
+      return {
+        decisions: viewCache.getView(personalization.getViewName())
```
jonsnyder, line 24: This is inefficient for customers that want to render their own decisions. They have to wait until the request returns before being able to get the cached decisions. I don't know how many people would want the decisions sooner, but it is something to think about for future improvements.

test/unit/specs/components/Personalization/createComponent.spec.js
```diff
@@ -163,49 +125,35 @@ describe("Personalization", () => {
...
-    expect(mergeQuery).toHaveBeenCalledWith(event, eventQueryDetails);
-    expect(hideContainers).toHaveBeenCalled();
-    expect(onResponseHandler).toHaveBeenCalled();
-
+    expect(pageLoadHandler).toHaveBeenCalled();
+    expect(viewChangeHandler).not.toHaveBeenCalled();
+    expect(mergeQuery).not.toHaveBeenCalled();
     expect(onClickHandler).not.toHaveBeenCalled();
-    expect(showContainers).not.toHaveBeenCalled();
+  }); */
```
Aaronius, line 152: Can we get rid of this or make it a useful test?

src/components/Personalization/createViewCacheManager.js
Aaronius: Could `storeViews` be called multiple times on the same page? I'm having a hard time figuring this out. If so, then we may need to think about this even more.

src/components/Personalization/createViewChangeHandler.js
ninaceban: I agree! but since we log an impression in Target when we return view decisions we can't just ignore them, we should return them. As a future improvement, I think that having a `render` command that will render decisions (similar to `applyOffers`) will solve this issue.

src/components/Personalization/createExecuteCachedViewDecisions.js
```diff
@@ -0,0 +1,28 @@
...
+export default ({ viewCache, executeViewDecisions, collect }) => {
+  return ({ viewName }) => {
+    const viewDecisions = viewCache.getView(viewName);
+
+    if (isNonEmptyArray(viewDecisions)) {
+      executeViewDecisions(viewDecisions);
+      return;
+    }
+
+    const xdm = { web: { webPageDetails: { viewName } } };
```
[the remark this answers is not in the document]
ninaceban: ok I will add that

Aaronius (commented):
I think I generally grasp what's going on and I think it's on the right track. I'm not sure how much you've through the various concurrency problems, but there are probably dark corners to think about, like what happens when `sendEvent` is called in quick succession with different options (`renderDecisions`, `decisionScopes`, `viewName`) and different states (cache is initialized, cache is not initialized). Were you planning on adding functional tests? I think making functional tests that cover potential concurrency problems would be pay off.

test/unit/specs/components/Personalization/createComponent.spec.js
ninaceban: yes, thank you, somehow I forgot it

src/components/Personalization/utils.js
```diff
@@ -10,20 +10,7 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
-  const scopes = [...decisionScopes];
-
-  if (renderDecisions && !includes(scopes, PAGE_WIDE_SCOPE)) {
-    scopes.push(PAGE_WIDE_SCOPE);
-  }
-
-  return scopes;
+// eslint-disable-next-line import/prefer-default-export
+export const isAuthoringModeEnabled = (doc = document) => {
+  return doc.location.href.indexOf("mboxEdit") !== -1;
```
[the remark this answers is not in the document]
ninaceban: ok, I can move it. Thank you

src/components/Personalization/createOnResponseHandler.js
```diff
@@ -9,27 +9,63 @@ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTA
...
+    if (unprocessedDecisions.length === 0) {
+      showContainers();
+      return { decisions: [] };
+    }
 
-    if (!renderDecisions) {
-      return { decisions: unprocessedDecisions };
+    const {
+      schemaDecisions,
+      otherDecisions
```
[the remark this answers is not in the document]
ninaceban: I tried to be generic, since we group decisions by a schema. I can rename again as they were domAction and nonDomAction.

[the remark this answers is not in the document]
Aaronius: As long as you've considered my input, feel free to name them what you'd like.

src/components/Personalization/createComponent.js
```diff
@@ -44,28 +39,27 @@ export default ({
...
+        });
 
-        if (!hasScopes(scopes)) {
+        if (personalization.shouldFetchData()) {
+          pageLoadHandler({
+            personalization,
+            event,
+            onResponse,
+            onRequestFailure
+          });
```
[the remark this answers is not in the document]
ninaceban: yes, should we pause all the other calls till we get back the response for the first call (and cache is initialized)?

src/components/Personalization/createExecuteCachedViewDecisions.js
```diff
@@ -0,0 +1,28 @@
...
+
+    collect({ meta: {}, xdm });
```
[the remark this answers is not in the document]
ninaceban: no, this behavior is only related to SPA and views.

src/components/Personalization/createPersonalizationDetails.js
```diff
@@ -0,0 +1,94 @@
...
+
+  if (!webPageDetails) {
+    return undefined;
+  }
+
+  return webPageDetails.viewName;
+};
+
+export default ({ renderDecisions, decisionScopes, event, viewCache }) => {
+  const xdm = event.toJSON().xdm;
```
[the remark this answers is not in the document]
jonsnyder: I like getUserXdm. You could also do getViewName on the event object.

This is a great scenario to bring up. I'm not sure how to solve it either.

src/components/Personalization/createComponent.js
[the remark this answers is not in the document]
Aaronius: From the customer perspective, they would probably prefer the request go out ASAP to ensure the data gets tracked before the visitor navigates away from the page, but that means we would need to properly deal with personalization coming back on both responses. Does that mean we'll try to render the decisions coming back on both responses? Does that mean we'll send a `display` notification after trying to render decisions both times? What happens if the personalization content is the same on both responses? What if it's different? What if some of it is the same and some of it is different? Would `decisions` be returned to the customer on the `result` object from both `sendEvent` commands and is that what they would expect? We need to nail those answer down and probably answers to more questions I'm not thinking about. What we actually do probably depends on the answers.

src/components/Personalization/createExecuteCachedViewDecisions.js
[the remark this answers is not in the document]
Aaronius: I also see collect logic here though: https://github.com/adobe/alloy/blob/e32c6f8acfaf809fb880ed4bb4ad57646b24ddd3/src/components/Personalization/createExecuteDecisions.js#L49

I'm trying to figure out how what we're doing here in createExecuteCachedViewDecisions.js relates to what we're doing in `createExecuteDecisions`. Anyway, my question isn't very clear because I don't really have a very specific question. I think what I'm really trying to figure out is if we could/should call `executeViewDecisions` from this module _regardless_ of whether the `viewDecisions` array is empty and let `executeViewDecisions` be the only module that deals with `display` notifications. I don't understand `executeViewDecisions` enough to know what that would look like though, so my question was a shot in the dark and quite vague. Don't worry about spending a bunch of time trying to answer. It's probably fine as it is.

test/unit/specs/components/Personalization/event.spec.js
```diff
@@ -41,3 +42,4 @@ describe("Personalization::event", () => {
     expect(result).toEqual({ schemas: expectedSchemas, decisionScopes });
   });
 });
+*/
```
Aaronius, line 45: These tests should be re-enabled or removed.

src/components/Personalization/createComponent.js
```diff
@@ -39,25 +39,34 @@ export default ({
...
-        if (personalization.shouldFetchData()) {
-          pageLoadHandler({
-            personalization,
+        if (personalizationDetails.shouldFetchData()) {
+          const decisionsDeferred = defer();
+
+          viewCache.storeViews(decisionsDeferred.promise);
+
+          fetchDataHandler({
+            decisionsDeferred,
```
Aaronius, line 55: It would probably be simpler to just inject `viewCache` directly into `createResponseHandler` from index.js and had `createResponseHandler` call `viewCache.storeViews` with the decisions rather than a promise.

src/components/Personalization/constants/scope.js
Aaronius: That was me that [suggested naming it scope.js](https://github.com/adobe/alloy/pull/594/files/66c5115b7517740afa1bb5b78cd23ad3e690536d#r486373572). Just to clarify my stance on this: If the file is intended to potentially contain multiple scopes, then having it called `scope.js` and exporting a constant (`export const PAGE_WIDE_SCOPE = "__view__"`) makes sense, because then other scopes can fall in line as exported constants and it's consistent with https://github.com/adobe/alloy/wiki/Conventions#enumerations.

If the "page wide scope" is the only scope we'll have as a constant (now or in the future), then naming the file `pageWideScope.js` and exporting a default value (`export default "__view__"`) might make more sense sense. 

Exporting the page wide scope as the `default` export value from a file named `scope.js` like we are currently probably doesn't make as much sense in my opinion just because it's less clear that `scope.js` is, specifically, the page wide scope.

Anyway, take it for what it's worth, which is probably not much.

Aaronius:
Note from a conversation @ninaceban and I just had: On the first request for the page view, we request `__view__` scope decisions. On some subsequent calls for the same page view (I think when a customer requests decisions for a specific scope?), we request `__view__` scope decisions again. I'm not sure why, but it might be something we don't need to do, since we already have `__view__` scope decisions in cache. @ninaceban was going to discuss it with David Son.

Aaronius (approved):
It looks like you're still fixing tests after the rebase, but everything else seems to look good.

test/functional/specs/Personalization/C782718.js
```diff
@@ -0,0 +1,189 @@
...
+const networkLogger = createNetworkLogger();
+const config = compose(
+  orgMainConfigMain,
+  debugEnabled
+);
+const PAGE_WIDE_SCOPE = "__view__";
+createFixture({
+  title:
+    "C782718 SPA support functional test with auto-rendering and view notifications",
+  url: `${testPageUrl}?spaImplementationTest=true`,
```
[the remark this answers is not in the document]
ninaceban: makes sense, I'll add a comment

test/functional/specs/Personalization/C782718.js
```diff
@@ -0,0 +1,189 @@
...
+
+  const personalizationSchemas =
+    requestBody.events[0].query.personalization.schemas;
+
+  const result = [
+    "https://ns.adobe.com/personalization/dom-action",
+    "https://ns.adobe.com/personalization/html-content-item",
+    "https://ns.adobe.com/personalization/json-content-item",
+    "https://ns.adobe.com/personalization/redirect-item"
+  ].every(schema => !!personalizationSchemas.find(s => s === schema));
```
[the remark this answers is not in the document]
ninaceban: thank you, will change it

test/functional/specs/Personalization/C782718.js
```diff
@@ -0,0 +1,189 @@
...
+
+const networkLogger = createNetworkLogger();
+const config = compose(
+  orgMainConfigMain,
+  debugEnabled
+);
+const PAGE_WIDE_SCOPE = "__view__";
+createFixture({
+  title:
+    "C782718 SPA support functional test with auto-rendering and view notifications",
```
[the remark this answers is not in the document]
ninaceban: thank you, I'll change it

jonsnyder (approved):

PR #594, opened 2020-09-10
Add "dom-action" schema only if PAGE_WIDE_SCOPE decisions are requested

ninaceban:
## Description

Personalization component should be sending "dom-action" schema ONLY when `sendEvent(...)` `renderDecisions` is set to `true`.

## Related Issue

None

## Motivation and Context

We should not send "dom-action" schema if whoever implements Alloy doesn't want decisions to be auto-rendered. This should also help the backend be a little bit more selective about the data it processes and returns.

None

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Personalization/constants/personalizationConstants.js
+++ b/src/components/Personalization/constants/personalizationConstants.js
@@ -17,3 +17,4 @@ export const JSON_CONTENT_ITEM =
   "https://ns.adobe.com/personalization/json-content-item";
 export const REDIRECT_ITEM =
   "https://ns.adobe.com/personalization/redirect-item";
+export const PAGE_WIDE_SCOPE = "__view__";
--- a/src/components/Personalization/constants/schema.js
+++ b/src/components/Personalization/constants/schema.js
@@ -10,20 +10,10 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import { includes, isNonEmptyArray } from "../../utils";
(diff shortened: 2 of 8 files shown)


src/components/Personalization/constants/personalizationConstants.js
```diff
@@ -17,3 +17,4 @@ export const JSON_CONTENT_ITEM =
   "https://ns.adobe.com/personalization/json-content-item";
 export const REDIRECT_ITEM =
   "https://ns.adobe.com/personalization/redirect-item";
```
Aaronius, line 19: A filename like `personalizationConstants.js` suggests that the focus of this file might not be narrow enough, especially since it's located under `Personalization/constants`. Rather than a bucket of all potential personalization constants, would it make more sense to have `schema.js` and `scope.js` files instead? Another thing to note is that `personalizationConstants.js` is plural whereas our convention is that the file should be singular (https://github.com/adobe/alloy/wiki/Conventions#enumerations).

Aaronius (approved):

jonsnyder (approved):

jonsnyder:
The title of this PR is a bit misleading. Its really doing: Add "Dom-action" schema only if page wide scope is requested explicitly
or implicitly by setting renderDecisions = true.

src/components/Personalization/constants/personalizationConstants.js
ninaceban: thank you, will extract those in separate files

PR #597, opened 2020-09-23
Core 51328 - fix flickering issue

ninaceban:
## Description

Personalization flickering issue when having multiple sequencing `sendEvent` calls:
If we have 2 sendEvent calls:
  1. sendEvent call with renderDecisions: true
  2. sendEvent call with renderDecisions: false
Second call response returns first and unhides the body.

The proposal is to unhide the body only when receiving response from a call with `renderDecisions:true`.
## Types of changes

## Checklist:


src/components/Personalization/createOnResponseHandler.js
```diff
@@ -16,7 +16,7 @@ const DECISIONS_HANDLE = "personalization:decisions";
 export default ({ extractDecisions, executeDecisions, showContainers }) => {
   return ({ renderDecisions, response }) => {
     const unprocessedDecisions = response.getPayloadsByType(DECISIONS_HANDLE);
-    if (!isNonEmptyArray(unprocessedDecisions)) {
+    if (renderDecisions && !isNonEmptyArray(unprocessedDecisions)) {
```
Aaronius, line 19: This isn't a big deal, but consider switching out `!isNonEmptyArray(unprocessedDecisions)` for `unprocessedDecisions.length`. We know `unprocessedDecisions` will always be an array and `!isNonEmptyArray` involves more mental load than necessary (it's a double negative and can also lead the reader to wonder whether we're checking if `unprocessedDecisions` has items or whether `unprocessedDecisions` is not an array or both).

Aaronius: I noticed you have another condition with `renderDecisions` down below. We may want to consolidate all this. If I understand the logic correctly, maybe something like this?

```
let unprocessedDecisions = response.getPayloadsByType(DECISIONS_HANDLE);

if (renderDecisions) {
  let renderableDecisions;
  ([renderableDecisions, unprocessedDecisions] = extractDecisions(unprocessedDecisions));
  if (renderableDecisions.length) { // Not sure if you need this. If you need this check it all, it would probably be best to move it to inside executeDecisions
    executeDecisions(renderableDecisions);
  }
  showContainers();
}

return { decisions: unprocessedDecisions };
```

Aaronius: Have these been thought through:

Do we always want to add `decisions` to the object that customers receive in response to executing a `sendEvent` command? Are we doing so currently? Is the value always an array? Do we have tests that ensure that this is the case?

src/components/Personalization/createComponent.js
```diff
@@ -47,7 +47,6 @@ export default ({
         const scopes = getDecisionScopes(renderDecisions, decisionScopes);
 
         if (!hasScopes(scopes)) {
-          showContainers();
```
Aaronius, line 50: Help me understand this: If a customer only executes one `sendEvent` command and that command does not have scopes and `renderDecisions` is set to `false`, then the containers won't be shown until the timeout is hit? Is that what we want? Do we have tests that cover this?

src/components/Personalization/createOnResponseHandler.js
ninaceban: we return in the `sendEvent` promise the decisions array that contains decisions that can't be rendered by us. `decisions` is always an array.

src/components/Personalization/createComponent.js
ninaceban: We need the prehiding snippet to be added only if bundle is loaded async, they care about flickering, and they have `sendEvent` with `renderDecisions: true`.
If they don't have a `sendEvent` command with `renderDecisions: true` on a page, they don't need this prehiding snippet.
If they load the bundle sync they also don't need the prehiding snippet, because all the prehiding/unhiding manipulations are done by us and they don't have to worry about it.

Aaronius: Got it. Thanks.

src/components/Personalization/createOnResponseHandler.js
Aaronius: It might be good to have a functional test for this. Specifically, if no decisions are returned that can't be auto-rendered, ensure that the `result` object that comes through the promise still has a `decisions` array. It would also be helpful to update [the personalization documentation](https://docs.adobe.com/content/help/en/experience-platform/edge/fundamentals/rendering-personalization-content.html) to show that `decisions` will always be provided and will always be an array. Our documentation currently suggests that `result.decisions` is only sometimes provided, because we have this example that checks if `result.decisions` exists:

```
alloy("sendEvent",{
    xdm:{...},
    decisionScopes:['demo-1', 'demo-2']
  }).then(function(result){
    if (result.decisions){
      // Do something with the decisions.
    }
  })
```

Aaronius (approved):

src/components/Personalization/createComponent.js
Aaronius: Can we make this clearer in [the documentation](https://docs.adobe.com/content/help/en/experience-platform/edge/solutions/target/flicker-management.html#managing-flicker-when-sdk-is-loaded-asynchronously)? Currently, it states:

> When the SDK is loaded asynchronously, it is required to use the prehiding snippet.

src/components/Personalization/createOnResponseHandler.js
ninaceban: `decisions` will always be provided if personalization on the call.

src/components/Personalization/createComponent.js
ninaceban: yes I will, I have a special ticket for that.

src/components/Personalization/createOnResponseHandler.js
```diff
@@ -9,27 +9,28 @@ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTA
...
 export default ({ extractDecisions, executeDecisions, showContainers }) => {
   return ({ renderDecisions, response }) => {
     const unprocessedDecisions = response.getPayloadsByType(DECISIONS_HANDLE);
-    if (!isNonEmptyArray(unprocessedDecisions)) {
+
+    if (!renderDecisions) {
+      return { decisions: unprocessedDecisions };
+    }
+
+    if (isEmptyArray(unprocessedDecisions)) {
```
jonsnyder, line 23: unprocessedDecisions will always be an array, so you can just do

if (unprocessedDecisions.length === 0) {

src/utils/isEmptyArray.js
```diff
@@ -0,0 +1,18 @@
...
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+/**
+ * Returns whether the value is a empty array.
+ * @param {*} value
+ * @returns {boolean}
+ */
+export default value => Array.isArray(value) && value.length === 0;
```
jonsnyder, line 18: No need for this util because unprocessedDecisions will always be an array.

test/unit/specs/components/Personalization/createComponent.spec.js
```diff
@@ -106,7 +106,6 @@ describe("Personalization", () => {
     expect(isAuthoringModeEnabled).toHaveBeenCalled();
     expect(getDecisionScopes).toHaveBeenCalled();
     expect(hasScopes).toHaveBeenCalled();
-    expect(showContainers).toHaveBeenCalled();
```
jonsnyder, line 109: can you add this line back in, but change it to "not.toHaveBeenCalled()". Following Test Driven Design, you want the test to fail before making the change.

jonsnyder (approved):

test/unit/specs/components/Personalization/createComponent.spec.js
ninaceban: ok, I have added it

src/utils/isEmptyArray.js
ninaceban: thank you, make sense

src/components/Personalization/createOnResponseHandler.js
ninaceban: thank you

PR #598, opened 2020-09-25
CORE-51982 Use sendBeacon for link tracking.

Aaronius:
## Description

We weren't using `sendBeacon` for click tracking and now we are.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-51982

## Motivation and Context
We want to be using sendBeacon on link clicks because there's a high likelihood the user is navigating away from the page and we want to ensure the beacon reaches the server.

## Types of changes

## Checklist:

- [ ] I have made any necessary test changes and all tests pass. ** Some integration tests are failing but should be fixed by https://github.com/adobe/alloy/commit/d4e7cc237333a291c4ff44c45a36c34a375a5689


sandbox/src/LargePayload.js
```diff
@@ -19,20 +19,39 @@ export default function LargePayload() {
       <h2>Large Payloads</h2>
       <p>This page tests send really large payloads to the edge.</p>
       <p>
-        All those requests should not use beacon calls, and should not fail.
```
Aaronius, line 22: This wasn't accurate. The command options include `documentUnloading: true`, which means Alloy _does_ use `sendBeacon`. Depending on the browser, some payload will fail to be sent using `sendBeacon` and Alloy will fall back to `fetch`.

src/core/edgeNetwork/requestPayloads/createDataCollectionRequestPayload.js
```diff
@@ -19,7 +19,12 @@ export default () => {
       addIdentity: createAddIdentity(content),
       addEvent(event) {
         content.events = content.events || [];
-        content.events.push(event.toJSON());
+        content.events.push(event);
```
Aaronius, line 22: This change here probably deserves some explanation. I believe we were calling `event.toJSON()` here because when we log the payload to the console, we don't want the event's methods and stuff showing up on the logged object. Since the time this was written, we've adjusted how we log the payload to the console to be smarter: https://github.com/adobe/alloy/blob/master/src/core/network/injectSendNetworkRequest.js#L24-L39

When we log the payload, we serialize and then deserialize the object. When the object is serialized, `JSON.parse` recursively goes through the payload object, including its events, calling `toJSON`. Because of this, there's no reason to do `event.toJSON()` here anymore. We also don't want to do `event.toJSON()` here anymore because we need access to its methods in the new `getDocumentMayUnload` method below.

src/core/network/injectSendNetworkRequest.js
```diff
@@ -39,7 +39,11 @@ export default ({ logger, networkStrategy, isRetryableHttpStatusCode }) => {
     }
 
     const executeRequest = (retriesAttempted = 0) => {
-      return networkStrategy(url, stringifiedPayload).then(response => {
+      return networkStrategy({
+        url,
+        body: stringifiedPayload,
+        documentMayUnload: payload.getDocumentMayUnload()
```
Aaronius, line 45: This is the crux of why `sendBeacon` wasn't being used before. We weren't passing `documentMayUnload`.

test/functional/helpers/constants/testServerUrl.js
```diff
@@ -16,14 +16,17 @@ const alloyWithVisitorTestPageUrl = `https://alloyio.com/functional-test/${alloy
...
   console.log("EDGE ENV:", env);
+  // eslint-disable-next-line no-console
   console.log("ALLOY ENV:", alloyEnv);
   let pageUrl;
   if (alloyEnv) {
     pageUrl = alloyPages[alloyEnv];
   } else {
     pageUrl = alloyPages[env];
   }
+  // eslint-disable-next-line no-console
```
Aaronius, line 29: I was seeing lint warnings.

test/unit/specs/components/Identity/getIdentity/createIdentityPayload.spec.js
```diff
@@ -24,12 +25,17 @@ describe("createIdentityPayload", () => {
...
+    const payload = createDataCollectionRequestPayload();
+    expect(payload.getDocumentMayUnload()).toBeFalse();
+  });
+
   it("serializes properly", () => {
     const payload = createIdentityPayload(["ECID"]);
     payload.addIdentity("IDNS", {
       id: "ABC123"
     });
-    expect(payload.toJSON()).toEqual({
```
Aaronius, line 32: This technically wasn't really serializing. `JSON.stringify` recursively calls `toJSON` on the content passed to it, which is a better representation of what we're trying to test.

src/core/edgeNetwork/requestPayloads/createDataCollectionRequestPayload.js
jonsnyder: This also affects when the lastChanceCallback gets called. The cloning of the object is also important so that if the event is queued, when the event is sent, it contains the data as it was when sendEvent was called. We should probably break out some of the toJSON code into a "freeze" method.

jonsnyder (changes requested):

src/core/edgeNetwork/requestPayloads/createDataCollectionRequestPayload.js
Aaronius: Thanks for pointing this out. I thought I might be missing something. So, with the changes in this PR, the last chance callback would get called [when the payload gets serialized](https://github.com/adobe/alloy/blob/6658c2625983ae191b63a476277fcfad33185eb1/src/core/network/injectSendNetworkRequest.js#L24), which seems to me to be more appropriate than the way it was previously, since the callback will now be called be as late as possible.

Regarding cloning of customer data, I feel like it should be the responsibility of the data collector to clone the data ASAP, even before it makes its way into the event.

Thoughts on that?

jonsnyder: Pros of the approach in this PR:
- using the lastChanceCallback to set the consent and preferences on pre-consent events is possible
- Customers have the flexibility of giving an object to sendEvent that can change up until the event is serialized or providing a copy

Cons:
- Customers may be confused about when the xdm and data objects are frozen. (We can solve this with documentation)

I'm sold on the change in this PR.

src/core/createEvent.js
```diff
@@ -52,24 +52,29 @@ export default () => {
...
-        // so setting content.xdm and content.data is inside this try
 
-        // We only set content.xdm if content.xdm was already set or
-        // if content.xdm was empty and the lastChanceCallback added items to it.
-        if (content.xdm || !isEmptyObject(xdm)) {
-          content.xdm = xdm;
-        }
-        if (content.data || !isEmptyObject(data)) {
-          content.data = data;
+      if (lastChanceCallback) {
```
Aaronius, line 56: This modification was to limit the amount we're cloning `xdm` and `data`.

src/components/DataCollector/index.js
```diff
@@ -37,8 +38,17 @@ const createDataCollector = ({ eventManager, logger }) => {
             event.documentMayUnload();
           }
 
-          event.setUserXdm(xdm);
-          event.setUserData(data);
+          if (xdm) {
+            // Clone xdm in case the customer modifies the object
+            // between now and the time it gets sent to the server.
+            event.setUserXdm(clone(xdm));
```
Aaronius, line 44: Like the comment says, DataCollector is now cloning `xdm` and `data` as soon as possible so that if the customer modifies the object between now and the time it gets sent to the server, we use the original state and not the state after modification. What's interesting is that the command option validator we're using (which leverages Yup) _also_ clones this data in some fashion, making this particular `clone` call unnecessary to accomplish our goal, but yet I prefer that we explicitly clone here anyway so that our intentions are explicit and we're not relying on the validator to do the cloning.

jonsnyder (approved):

sandbox/src/EventMerge.js
```diff
@@ -4,12 +4,12 @@ export default function EventMerge() {
   const eventMergeIdPromise = useRef(window.alloy("createEventMergeId"));
 
   useEffect(() => {
-    eventMergeIdPromise.current.then(eventMergeId => {
+    eventMergeIdPromise.current.then(result => {
```
liljenback, line 7: This is a separate fix for something else?

Aaronius: Yeah, I thought I left a comment somewhere but I don't know where it is now. The event merge ID view in the sandbox was failing because the promise is resolved with an object containing the event merge ID rather than the event merge ID itself.

sandbox/src/LargePayload.js
liljenback: Wouldn't this be configured without documentUnloading to avoid sendBeacon limit then?

Aaronius: I'm not sure I understand your question, but this page is to test what happens when `sendBeacon` _is_ used, so I believe the configuration (`documentUnloading: true`) is correct, but this description we had previously didn't make sense and didn't correlate with our configuration.

jfkhoury (approved):
👍  💯

PR #599, opened 2020-09-28
Fix broken personalization functional specs 

shammowla:
The expected Target personalization schema urls specs failed due to the order of URLs. 

## Types of changes

## Checklist:


shammowla:
@ninaceban When I run C28756 locally, the personalization schema returns "https://ns.adobe.com/personalization/dom-action", but not on production. Is this expected?

ninaceban:
this test fails because we add "dom-action" schema only when we send `__view__` scope or `renderDecisions:true` in the latest alloy. Is the alloy version you are using the latest?

PR #612, opened 2020-10-15
CORE-53556 simplify referrer calculation to avoid cross domain iframe errors

jonsnyder:
## Description

Getting "window.top.document.referrer" throws an error when called within an iframe with a different domain than its parent. We could wrap it inside a try/catch, but the change in this PR is more straight-forward. For example, within an iframe it's not clear what the referrer should be. Should it be the referrer of the parent? Or if the user navigated within the iframe, should it be the referrer of the iframe?

## Related Issue

https://jira.corp.adobe.com/browse/CORE-53556
https://jira.corp.adobe.com/browse/CORE-53562

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Context/injectWeb.js
+++ b/src/components/Context/injectWeb.js
@@ -19,7 +19,7 @@ export default window => {
         URL: window.location.href || window.location
       },
       webReferrer: {
-        URL: window.top.document.referrer
+        URL: window.document.referrer
       }
     };
     deepAssign(xdm, { web });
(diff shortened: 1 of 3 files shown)


src/components/Context/injectWeb.js
```diff
@@ -19,7 +19,7 @@ export default window => {
         URL: window.location.href || window.location
       },
       webReferrer: {
-        URL: window.top.document.referrer
+        URL: window.document.referrer
```
jfkhoury, line 22: Should I create a doc ticket to mention this change?

jfkhoury (approved):
Thank yoU!!

Aaronius (approved):
🎉

src/components/Context/injectWeb.js
jonsnyder: All the current docs say is this: "The URL of the previous page visited." so we probably don't need to rush out a doc change. I think the release notes will be fine. It would be good to have a ticket to to mention how an iframe will affect this value in the docs.

PR #613, opened 2020-10-15
CORE-53556 Don't attempt to change the message on a error with a read-only message property.

Aaronius:
## Description

As errors are bubbling up through Alloy, we modify the message of the error to provide additional context. Some errors, like DOMExceptions, have a `message` property that is read-only. When trying to change the `message` property value on these types of errors, a new error is thrown because doing so is not allowed.

This PR modified the behavior so that if we _can_ modify the `message` property, we'll add context. If we can't, we leave the error as it is.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-53556

## Motivation and Context
In certain cases, a DOMException is thrown (one case is due to a bug when Alloy is running within an iframe that has a different origin than the parent). We cause more issues by trying to update its `message` property which masks the underlying error.

## Types of changes

## Checklist:

- [ ] (some functional tests fail due to pending error code changes in konductor) I have made any necessary test changes and all tests pass.

### DIFF
--- a/src/core/injectHandleError.js
+++ b/src/core/injectHandleError.js
@@ -10,21 +10,25 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import { toError } from "../utils";
+import { toError, updateErrorMessage } from "../utils";
 import { DECLINED_CONSENT_ERROR_CODE } from "./consent/createConsentStateMachine";
 
 export default ({ errorPrefix, logger }) => (error, operation) => {
+  const err = toError(error);
+
   // In the case of declined consent, we've opted to not reject the promise
   // returned to the customer, but instead resolve the promise with an
   // empty result object.
-  if (error.code === DECLINED_CONSENT_ERROR_CODE) {
+  if (err.code === DECLINED_CONSENT_ERROR_CODE) {
     logger.warn(
       `The ${operation} could not fully complete because the user declined consent.`
     );
     return {};
   }
 
-  const err = toError(error);
-  err.message = `${errorPrefix} ${err.message}`;
+  updateErrorMessage({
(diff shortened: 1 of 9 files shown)


src/core/injectHandleError.js
```diff
@@ -10,21 +10,25 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import { toError } from "../utils";
+import { toError, updateErrorMessage } from "../utils";
 import { DECLINED_CONSENT_ERROR_CODE } from "./consent/createConsentStateMachine";
 
 export default ({ errorPrefix, logger }) => (error, operation) => {
+  const err = toError(error);
```
Aaronius, line 17: Moving this line wasn't necessary, but I thought it was safer since it ensures that's we're dealing with an error object before we attempt to access the `code` property below.

src/core/injectHandleError.js
```diff
@@ -10,21 +10,25 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
     );
     return {};
   }
 
-  const err = toError(error);
-  err.message = `${errorPrefix} ${err.message}`;
+  updateErrorMessage({
+    error: err,
+    message: `${errorPrefix} ${err.message}`
+  });
```
Aaronius, line 32: Notice we're not using `stackError` here since we're just trying to add a prefix to the first line in the error message rather than doing what `stackError` does: `${message}\nCaused by: ${errorToStack.message}`.

I think we probably have an opportunity to modify `stackError` so that we could use it here as well, but that's something that may be worth investigating later.

jonsnyder (approved):

jfkhoury (approved):
Thank you!

Issue #611, opened 2020-10-15
AEP Web SDK 2.2.0 - checkout- Uncaught (in promise) TypeError: [alloy] Cannot convert object to primitive value  at String (<anonymous>)


stewartschillingsdi:
### Expected Behaviour
It is expected that this error does not happen (which prevent the xdm from being sent).

### Actual Behaviour
On my test implementation (https://www.seven-goats.com - Password : "goats") , when placing an order and being shown the order confirmation page, there is an "AEP Web SDK - Send Event" action.  In the console, there are a few errors:

![Screen Shot 2020-10-15 at 11 32 32 AM](https://user-images.githubusercontent.com/22940746/96159438-42203b00-0eda-11eb-882e-649bf9a18f4c.png)

Because of this error, the action does not actually send the xdm event. 
If I reload the page, everything works as expected. 

Other notes: I have "Disable cache" enabled on my Network Settings in Chrome. 

### Reproduce Scenario (including but not limited to)
0) Open Chrome Dev Tools, enable Launch Debug and Alloy debug, set log level to include verbose)
1) Navigate to [https://www.seven-goats.com ](url). Use the password "goats" for entry.
2) Add a goat or two to the shopping cart.
3) Proceed to Checkout. 
4) Use any valid address for shipping / billing.  On the payment page enter "1" as the credit card number. Enter "Bogus Gateway" as the Name on the Credit card.  Expiration 02/22,  CVV: 222
5) Place order and observe the errors in the console. 

#### Steps to Reproduce
See above. 

#### Platform and Version
Launch / AEP Web SDK 2.2.0

jfkhoury:
Thank you @stewartschillingsdi . We will look into this and get back to you shortly.

Jira: https://jira.corp.adobe.com/browse/CORE-53571

jonsnyder:
I was able to reproduce the error. You also need to be debugging for the error to occur. (i.e. `_satellite.setDebug(true)`). This is what is happening:

1. The console logger is overwritten by some code on the page (Maybe Shopify code?)
2. During the configure command, Alloy is logging the computed configuration object to the console. This object was created with "Object.create(null)".
3. The overwritten console logger calls `String(loggedObject)` which throws an error because the loggedObject does not have a "toString" object.

I am working on a fix to make sure the object will have a toString method. This will probably go out with the next Alloy/Alloy Extension release. In the mean time here are some workarounds:

Workarounds:
1. Turn off debugging (_satellite.setDebug(false))
[...]

PR #614, opened 2020-10-16
CORE-53571 use regular object prototype for config objects

jonsnyder:
## Description

We were logging an object created with Object.create(null) which caused problems on a Shopify site where the logging was overwritten. This change fixes that case and adds a functional test to make sure it doesn't happen again.

## Related Issue

https://github.com/adobe/alloy/issues/611
https://jira.corp.adobe.com/browse/CORE-53571

## Types of changes

## Checklist:

### DIFF
--- a/test/functional/specs/Logging/C532204.js
+++ b/test/functional/specs/Logging/C532204.js
@@ -12,6 +12,12 @@ const debugEnabledConfig = compose(
   debugEnabled
 );
 
+/*
+ * Some pages will redefine the console logging methods with implementations
+ * that aren't as forgiving as the built in logger. We ran into this issue
+ * on a Shopify site with a redefined logger. This test runs through some basic
+ * scenarios and makes sure the logged objects can be stringified
+ */
 createFixture({
   title: "C532204: Logged objects can be stringified"
 });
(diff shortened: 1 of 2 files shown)


test/functional/specs/Logging/C532204.js
```diff
@@ -0,0 +1,45 @@
...
+  title: "C532204: Logged objects can be stringified"
+});
+
+test.meta({
+  ID: "C532204",
+  SEVERITY: "P0",
+  TEST_RUN: "Regression"
+});
+
+const setupLogger = ClientFunction(() => {
```
liljenback, line 25: This will cover all scenarios we need from the `triggerAlloyEvent` output?

liljenback (approved):

Aaronius (approved):
Thanks @jonsnyder. This looks good. I would recommend adding a comment to the functional test explaining why it's important that logged information can be stringified, lest the reasoning be long to the sands of time.

test/functional/specs/Logging/C532204.js
jonsnyder: Actually, simply configuring alloy reproduced the bug, but triggering an alloy event creates a lot of other logging statements. This test doesn't exercise all the logging scenarios, but I felt like it was good enough.

jfkhoury:
> Thanks @jonsnyder. This looks good. I would recommend adding a comment to the functional test explaining why it's important that logged information can be stringified, lest the reasoning be lost to the sands of time.

+1 :)

jfkhoury (approved):
👍  💯

PR #616, opened 2020-10-20
CORE-45072 Improved functional testing workflow.

Aaronius:
## Description
The primary goal was to make an npm script that would (1) build Alloy (2) run functional tests (3) re-build Alloy when source code changes and make it easy to re-run tests afterward by hitting Ctrl+R (4) re-run tests automatically if test files changed.

During the process, I also wanted to simplify our functional test setup by reducing the number of test pages to one and dynamically including scripts depending on the test. Part of this was to also simplify the API that functional tests needed to use to configure fixtures.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-45072

## Motivation and Context
Simpler test workflows.

## Types of changes

## Checklist:


.gitignore
```diff
@@ -1,72 +1,10 @@
...
-
-# Rollup plugin script
-.rts2_cache/
-.rpt2_cache/
-
-TMP_DIR/
-
-bin/
-
-deploy_key
```
Aaronius, line 72: Anything not project-specific but is instead system-specific should be added to the global gitignore on your machine. Some of these things look system-specific while other look unrelated to our project entirely (e.g., bower, sass, typescript). https://gist.github.com/subfuzion/db7f57fff2fb6998a16c

package.json
```diff
@@ -119,15 +119,15 @@
...
     "rollup-plugin-commonjs": "^10.1.0",
     "rollup-plugin-glob-import": "^0.4.5",
     "rollup-plugin-istanbul": "^2.0.1",
-    "rollup-plugin-jscc": "^1.0.0",
-    "rollup-plugin-license": "^0.12.1",
+    "rollup-plugin-jscc": "^2.0.0",
+    "rollup-plugin-license": "^2.2.0",
     "rollup-plugin-node-resolve": "^5.2.0",
-    "rollup-plugin-terser": "^5.1.2",
+    "rollup-plugin-terser": "^7.0.2",
```
Aaronius, line 130: I needed to upgrade these dependencies in order to get the API I needed to write `scripts/watchFunctionalTests.js`.

rollup.config.js
```diff
@@ -61,7 +61,9 @@ if (buildTarget !== buildTargets.DEV) {
   plugins.push(
     license({
       banner: {
-        file: path.join(__dirname, "LICENSE_BANNER")
+        content: {
+          file: path.join(__dirname, "LICENSE_BANNER")
+        }
```
Aaronius, line 66: This API was changed in one of the major versions of rollup-plugin-license.

scripts/deploy_sandbox.sh
```diff
@@ -1,5 +0,0 @@
-ssh -t mowla@20.186.4.13 "cd /var/www/alloyio.com && git reset HEAD --hard && 
-git pull origin master && npm i && npm run build:prod && npm run sandbox:install && 
-npm run sandbox:build && 
-cp /var/www/alloyio.com/dist/standalone/alloy.js /var/www/alloyio.com/sandbox/build &&
-cp /var/www/alloyio.com/dist/standalone/alloy.js /var/www/alloyio.com/sandbox/build/demo"
```
Aaronius, line 5: @shammowla said we didn't need this anymore.

scripts/qe_status.sh
```diff
@@ -1,4 +0,0 @@
-ssh -t mowla@20.186.4.13 "cd /var/www/alloyio.com && 
-npm run test:slreport && npm run functional:local && 
-npm run allure:generate && 
-cp /var/www/alloyio.com/allure/allure-report /var/www/qe.alloyio.com"
```
Aaronius, line 4: @shammowla said we didn't need this anymore.

test/functional/helpers/constants/testServerUrl.js
```diff
@@ -1,29 +0,0 @@
...
-export { alloyWithVisitorTestPageUrl };
-
-// eslint-disable-next-line no-console
-console.log("EDGE ENV:", env);
-// eslint-disable-next-line no-console
-console.log("ALLOY ENV:", alloyEnv);
-// eslint-disable-next-line no-console
-console.log("ALLOY PAGE:", alloyTestPageUrl);
-// eslint-disable-next-line no-console
-console.log("ALLOY VISITOR PAGE:", alloyWithVisitorTestPageUrl);
```
Aaronius, line 29: We now just deal with one test page.

test/functional/helpers/createFixture/clientScripts.js
```diff
@@ -0,0 +1,166 @@
...
+      "document.write('<script src=\"https://github.com/Adobe-Marketing-Cloud/id-service/releases/latest/download/visitorapi.min.js\"></script>')"
+  });
+};
+
+const getFixtureClientScriptsForInt = options => {
+  const clientScripts = [];
+  const promisePolyfillPath = path.join(
+    __dirname,
+    "../promisePolyfill/promise-polyfill.min.js"
+  );
```
Aaronius, line 57: We could just use the CDN-hosted polyfill like we do in `getFixtureClientScriptsForProd`, but it might be a bit slower because we would have to go over the internet to load it.

test/functional/helpers/createFixture/clientScripts.js
```diff
@@ -0,0 +1,166 @@
...
+    document.getElementsByTagName("head")[0].appendChild(scriptElement);
+  },
+  {
+    dependencies: {
+      remoteAlloyLibraryUrl
+    }
+  }
+);
+
+const injectAlloyDuringTestByEnvironment = {
```
liljenback, line 146: 👍
Adding nonce here as well would be desired I guess.

test/functional/helpers/createFixture/clientScripts.js
```diff
@@ -0,0 +1,166 @@
+/*
```
liljenback, line 1: Maybe add a bit more comments for the different functions making the library available and why it is being loaded that way.

liljenback (approved):

test/functional/helpers/createFixture/clientScripts.js
Aaronius: I added some more comments and shifted a couple things around. Not sure if that's any more helpful though. I tried!

test/functional/helpers/createFixture/clientScripts.js
```diff
@@ -0,0 +1,206 @@
...
+  if (options.includeVisitorLibrary) {
+    addRemoteUrlClientScript({
+      clientScripts,
+      url: remoteVisitorLibraryUrl
+    });
+  }
+
+  clientScripts.push({
+    path: alloyPageSnippetPath
+  });
```
Aaronius, line 132: Note that we're always using the alloy page snippet that's in the file committed to the repo. There's not really a "production" page snippet.

test/functional/helpers/createFixture/clientScripts.js
```diff
@@ -0,0 +1,206 @@
...
+  "../../../../dist/standalone/alloy.js"
+);
+const remoteAlloyLibraryUrl =
+  "https://cdn1.adoberesources.net/alloy/latest/alloy.js";
+
+let localAlloyCode;
+
+// We're only testing against the alloy code found in /dist when alloyEnv
+// is "int". Otherwise, we're testing against the prod alloy that's on the CDN
+// and in those cases alloy code may not event exist in /dist. This is why
```
jonsnyder, line 48: event => even

test/functional/helpers/createFixture/clientScripts.js
jonsnyder: Why not use the page snippet that is built into dist/ ?

sandbox/public/functional-test/alloyVisitorTestPageInt.html
```diff
@@ -1,15 +0,0 @@
-<!DOCTYPE html>
```
jonsnyder, line 1: Are these files moving to the repo that @shammowla is creating?

jonsnyder (approved):
I'm excited to start running the tests with the new watch.

sandbox/public/functional-test/alloyVisitorTestPageInt.html
shammowla: I only moved the "perf" directory from "alloy/sandbox/public/".

Aaronius:
Regarding:

> Why not use the page snippet that is built into dist/ ?

Yeah, that's a great question. One thing to note is that the snippet isn't ever build to `dist`. We do have a `build:basecode` npm script that takes `/src/baseCode/index.js`, runs it through a minifier, and outputs the result to the console. The idea is we'd copy it out of the console to update the documentation, for example.

But....that doesn't mean we couldn't change our script to dump a minified version into `dist` for at least functional test purposes. Or, more simply, we could just read the unminified code from `/src/baseCode/index.js` and use it. Part of the reason that wasn't done, I presume, was that the snippet we use for functional tests isn't the same as the snippet that's in `/src/baseCode/index.js` or that's produced by `build:basecode`. The difference is that the snippet used for functional tests configures two instances: `alloy` _and_ `instance2`.

That said, I'm thinking it might be best if, for functional testing, we read the base code from `/src/baseCode/index.js` and then just did a simple string replace to modify `["alloy"]` to be `["alloy", "instance2"]`. Thoughts?

test/functional/helpers/createFixture/clientScripts.js
Aaronius: Yeah, that's a great question. One thing to note is that the snippet isn't ever build to `dist`. We do have a `build:basecode` npm script that takes `/src/baseCode/index.js`, runs it through a minifier, and outputs the result to the console. The idea is we'd copy it out of the console to update the documentation, for example.

But....that doesn't mean we couldn't change our script to dump a minified version into `dist` for at least functional test purposes. Or, more simply, we could just read the unminified code from `/src/baseCode/index.js` and use it. Part of the reason that wasn't done, I presume, was that the snippet we use for functional tests isn't the same as the snippet that's in `/src/baseCode/index.js` or that's produced by `build:basecode`. The difference is that the snippet used for functional tests configures two instances: `alloy` _and_ `instance2`.

That said, I'm thinking it might be best if, for functional testing, we read the base code from `/src/baseCode/index.js` and then just did a simple string replace to modify `["alloy"]` to be `["alloy", "instance2"]`. Thoughts?

sandbox/public/functional-test/alloyVisitorTestPageInt.html
Aaronius: What @shammowla said. The need for `alloyVisitorTestPageInt.html` and its variants goes away entirely due to the other changes in this PR. Now we just have a single test page and we just inject scripts dynamically as needed based on the purmutation needed for the test (all handled in `clientScripts.js`.

shammowla (approved):

scripts/watchFunctionalTests.js
```diff
@@ -0,0 +1,71 @@
...
+
+  const watcher = rollup.watch(options);
+  watcher.on("event", event => {
+    const effect = effectByEventCode[event.code];
+
+    if (effect) {
+      effect(event);
+    }
+  });
+})();
```
jfkhoury, line 71: SICK!!!!

test/functional/helpers/createFixture/clientScripts.js
```diff
@@ -0,0 +1,206 @@
...
+
+/**
+ * Injects the Alloy library while a test is running. This is useful if you
+ * want to test what happens when, for example, commands are queued before
+ * the Alloy library finishes loading. If you use this, you'll want to set
+ * includeAlloyLibrary to false when configuring client scripts for the
+ * fixture so that Alloy isn't injected into <head>.
+ */
+export const injectAlloyDuringTest =
+  injectAlloyDuringTestByEnvironment[alloyEnv];
```
jfkhoury, line 206: Love this module.

jfkhoury (approved):
Awesome work!

test/functional/helpers/createFixture/clientScripts.js
```diff
@@ -0,0 +1,206 @@
...
+// and in those cases alloy code may not event exist in /dist. This is why
+// we only try to load the file from the file system if alloyEnv is int.
+if (alloyEnv === "int") {
+  localAlloyCode = fs.readFileSync(localAlloyLibraryPath, "utf8");
+}
+
+const addRemoteUrlClientScript = ({ clientScripts, url, async = false }) => {
+  // TestCafe client scripts don't "natively" support loading a script
+  // from a remote URL, so we have to make our local script add another script
+  // element that loads Visitor from the remote server.
```
Aaronius, line 57: Note to self: I mention "Visitor" here because I wrote the comment before I made this function generic for any remote URL. I should remove "Visitor" from the comment.

test/functional/helpers/createFixture/clientScripts.js
Aaronius: Unless someone is against the approach I proposed, I'm going to make it happen, but I'm going to do it in a separate PR just so we can get this PR merged to master so @liljenback can start building on top of it.

test/functional/helpers/createFixture/clientScripts.js
Aaronius: Fixed! Thanks.

PR #618, opened 2020-10-20
Remove EDGE_ENV environment variable

Aaronius:
## Description
When testing against Edge in its Int environment, we would have to use a different host than when testing against Edge in its Prod environment. This difference in host was driven by the `EDGE_ENV` environment variable. Recently, this need for using different hosts went away and we now only need to test against the production host. We now differentiate whether we want to test against Edge in its "pre-prod" environment or Edge in its prod environment through the "edge base path", which is driven by the `EDGE_BASE_PATH` environment variable.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-52757

## Motivation and Context
Simplified 

## Types of changes

## Checklist:


test/functional/helpers/constants/domain.js
```diff
@@ -0,0 +1,14 @@
...
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+export const THIRD_PARTY_DOMAIN = "edge.adobedc.net";
+export const FIRST_PARTY_DOMAIN = "firstparty.alloyio.com";
```
jonsnyder, line 14: We had a question yesterday from the Konductor team if they could run our tests against a locally running version of Konductor. @shammowla I think we could allow these domains to be set via a environment variable

jonsnyder (approved):

test/functional/helpers/constants/domain.js
jfkhoury: Yes pls :)

jfkhoury (approved):
Love it! If we can make the domains be set via env variable, that would be awesome.

test/functional/helpers/constants/domain.js
Aaronius: It looks like they want a custom port and no edge base path as well. I'm going to submit a separate PR.

Aaronius: Oh and they want to run on http instead of https. Some of this will require changes to source code.

PR #620, opened 2020-10-22
CORE-53745 Cleaned up npm scripts. Swapped out mockServer-node with miragejs

Aaronius:
## Description
The primary goal was to clean up our npm scripts because there were some that weren't being used and some that weren't consistent in naming with the Launch extension repo. Some npm scripts were being used, but only through other npm scripts. I thought it would be nice to trim down the npm scripts to _just_ what we as developers would run directly or some automation server somewhere would run directly. I imagine there will be some opinions on this and we may need to adjust it over time, but that was the idea.

In the process, I started looking at our mockServer.js script. I remembered we weren't running it automatically when unit tests were being run, which is unfortunate because then we have some unit tests that aren't being run very often. In fact, when I ran tests with mockServer.js, tests were failing! 😱  The source is fine, but the tests had fallen out of date because apparently nothing and nobody has been running tests recently with mockServer.js running. I started to make the scripts so that mockServer.js would run every time we ran unit tests, but it took about 23 seconds just to start the mock server on my machine, which is unacceptable every time I'm running tests locally IMO. 

I then started to look into what we would lose if we used [Mirage JS](https://miragejs.com/) instead. Mirage works differently than how our mockServer.js script works in that it doesn't actually run a server but instead overrides the browser's `fetch` and `XMLHttpRequest` methods. It, however, does NOT override `sendBeacon` (Mirage uses Pretender and it's a [requested feature of Pretender](https://github.com/pretenderjs/pretender/issues/249)). I decided to implement Mirage and feel okay about the trade-offs. We lost the sendBeacon tests, which is unfortunate, but those are pretty well covered in functional tests. The test for a dropped internet connection also is slightly hacky. The main thing we probably lose is it's not testing actual browser network behavior, but I think I prefer less-than-perfect tests that always run over more-perfect tests that rarely run. We also have the option of reverting this move to Mirage and just running the mockServer.js script whenever tests are run on the CI server, so at least we can guarantee the tests that require the mock server are running sometimes. @jonsnyder I'd like to know what you think about this since you're more familiar with the these tests and the mock server we were using.

@shammowla I'm sure some of our automation scripts will need to be updated to work with these new script names. We'll need to be careful about when those changes are made though since the CI servers will probably start _not_ working for other branches that have the old npm script names.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-53745

## Motivation and Context
A better development experience because it's easier to figure out which npm scripts are useful. Using Miragejs allows us to always run some unit tests that were previously rarely run.

## Types of changes

## Checklist:


package.json
```diff
@@ -4,47 +4,27 @@
   "description": "Client-Side SDK for Unified Data Collection",
   "main": "src/core/index.js",
   "scripts": {
+    "install": "cd sandbox && npm install",
```
Aaronius, line 7: This didn't exist before, which meant a newcomer developer would need to run `npm run sandbox:install` to get sandbox functionality. Now the sandbox dependencies get installed automatically when installing the top-level dependencies.

package.json
```diff
@@ -4,47 +4,27 @@
...
+    "test": "npm run test:unit && npm run test:functional",
+    "test:unit": "karma start --single-run",
+    "test:unit:watch": "karma start",
+    "test:unit:saucelabs": "karma start karma-saucelabs.conf.js",
+    "test:unit:coverage": "karma start --single-run --reporters spec,coverage",
+    "test:functional": "rollup -c --environment BUILD:prodStandalone && EDGE_BASE_PATH=\"ee-pre-prd\" ALLOY_ENV=\"int\" testcafe chrome",
+    "test:functional:watch": "EDGE_BASE_PATH=\"ee-pre-prd\" ALLOY_ENV=\"int\" BUILD=\"prodStandalone\" ./scripts/watchFunctionalTests.js",
+    "test:functional:ci": "EDGE_BASE_PATH=\"ee-pre-prd\" ALLOY_ENV=\"int\" testcafe chrome && EDGE_BASE_PATH=\"ee-pre-prd\" ALLOY_ENV=\"int\" testcafe 'browserstack:ie@11:Windows 10'",
+    "test:functional:cron:prod": "ALLOY_ENV=\"prod\" testcafe chrome",
+    "test:functional:api": "testcafe chrome:headless",
```
Aaronius, line 20: This was still needed by the Konductor team. I'm not sure if `test:functional:api` is the right name though. Maybe @shammowla could chime in.

package.json
```diff
@@ -4,47 +4,27 @@
...
+    "dev": "concurrently --names build,sandbox \"rollup -c -w\" \"cd sandbox && npm start\"",
+    "build": "npm run format && npm run lint && npm run clean && concurrently \"rollup -c --environment BUILD:prodStandalone\" \"rollup -c --environment MINIFY,BUILD:prodStandalone\" \"rollup -c --environment BUILD:prodReactor\" && echo \"Base Code:\" && terser src/baseCode/index.js --mangle --compress unused=false",
+    "checkthattestfilesexist": "./scripts/checkThatTestFilesExist.js"
```
Aaronius, line 23: I decided to keep this as an npm script instead of only running it on the pre-commit hook. When I create a bunch of new modules, it's helpful to understand which spec files I still need to create (without having to try to commit first).

package.json
```diff
@@ -4,47 +4,27 @@
...
+    "test:unit": "karma start --single-run",
+    "test:unit:watch": "karma start",
+    "test:unit:saucelabs": "karma start karma-saucelabs.conf.js",
+    "test:unit:coverage": "karma start --single-run --reporters spec,coverage",
+    "test:functional": "rollup -c --environment BUILD:prodStandalone && EDGE_BASE_PATH=\"ee-pre-prd\" ALLOY_ENV=\"int\" testcafe chrome",
+    "test:functional:watch": "EDGE_BASE_PATH=\"ee-pre-prd\" ALLOY_ENV=\"int\" BUILD=\"prodStandalone\" ./scripts/watchFunctionalTests.js",
+    "test:functional:ci": "EDGE_BASE_PATH=\"ee-pre-prd\" ALLOY_ENV=\"int\" testcafe chrome && EDGE_BASE_PATH=\"ee-pre-prd\" ALLOY_ENV=\"int\" testcafe 'browserstack:ie@11:Windows 10'",
+    "test:functional:cron:prod": "ALLOY_ENV=\"prod\" testcafe chrome",
+    "test:functional:api": "testcafe chrome:headless",
+    "dev": "concurrently --names build,sandbox \"rollup -c -w\" \"cd sandbox && npm start\"",
```
Aaronius, line 21: I swapped out `npm-run-all` with `concurrently` because `npm-run-all` would only run npm scripts, whereas `concurrently` can run npm scripts or any other script, which helps us reduce the number of npm scripts needed.

package.json
```diff
@@ -4,47 +4,27 @@
...
+    "build": "npm run format && npm run lint && npm run clean && concurrently \"rollup -c --environment BUILD:prodStandalone\" \"rollup -c --environment MINIFY,BUILD:prodStandalone\" \"rollup -c --environment BUILD:prodReactor\" && echo \"Base Code:\" && terser src/baseCode/index.js --mangle --compress unused=false",
+    "checkthattestfilesexist": "./scripts/checkThatTestFilesExist.js"
   },
   "husky": {
     "hooks": {
-      "pre-commit": "echo 'Running pre-commit scripts...' && lint-staged",
-      "pre-push": "echo 'Running pre-push scripts...' && npm run checkthattestfilesexist && npm run test"
```
Aaronius, line 47: Someone requested that we check that test files exists before committing rather than before pushing, which makes sense. I also decided to run unit tests before commit as well. Note that we're (still) NOT running functional tests before commit, but maybe we should?

Aaronius:
I removed deploy_key.enc because @shammowla said it's not being used anymore.

test/unit/specs/core/network/injectNetworkStrategy.spec.js
```diff
@@ -57,221 +53,112 @@ const eventually = (
...
+  const scenarios = [
+    {
+      description: "window with fetch support",
+      simulateWindowWithoutFetch: true
+    },
+    {
+      description: "window without fetch support",
+      simulateWindowWithoutFetch: false
+    }
+  ];
```
Aaronius, line 66: I think we may want to rethink this. It's interesting that we're running a scenario _with_ fetch support and another scenario _without_ fetch support. If we run our unit tests on a browser that support fetch and another browser  which doesn't (Internet Explorer), then should we just _not_ simulate browsers in the test? If you think about it, when we run these tests in IE, we're running a "window with fetch support" test on a browser that doesn't and _can't_ provide fetch support.

According to our SauceLabs setup, our unit tests should be running on both kinds of browsers. Because of this, I think I would prefer _not_ simulating a window without fetch support here. Thoughts?

test/unit/specs/karmaEntry.spec.js
```diff
@@ -15,4 +15,4 @@ governing permissions and limitations under the License.
 // and file system errors.
 
 // eslint-disable-next-line import/no-unresolved,import/extensions
-import "./**/*.spec.js";
+import "./*/**/*.spec.js";
```
Aaronius, line 18: The Rollup Karma preprocessor was printing this warning:

```
Circular dependency: test/unit/specs/karmaEntry.spec.js -> test/unit/specs/_46__47__42__42__47__42__46_spec_46_js -> test/unit/specs/karmaEntry.spec.js
```

Aaronius:
After this PR is reviewed but before merging, I'll update the [Development Scripts wiki page](https://github.com/adobe/alloy/wiki/Development-Scripts).

package.json
jonsnyder: Doesn't this prevent you from installing the dependencies from this package.json? I think this should be put in "postinstall"

test/unit/specs/core/network/injectNetworkStrategy.spec.js
jonsnyder: I agree. I think we should just simulate the things that are available in the running environment.

jonsnyder (approved):
Nice. Thank you for simplifying the scripts and removing the need for the mock server.

jfkhoury (approved):
@Aaronius can you please stop making things better? Thank you! 

Wanna move or rename the install script per Jon’s comment?

package.json
Aaronius: No, it installs the dependencies from package.json and then runs the `install` script. I could have put this on `preinstall`, `install`, or `postinstall`. Post-install seems to usually be used for doing things (often using dependencies) after dependencies are installed. In this case though, we're actually still in the process of installing dependencies (the sandbox's dependencies), so that's why I figured it should be in the install stage. I don't feel strongly though. If you still feel like it would be more appropriate in `postinstall` I can make the change.

test/unit/specs/core/network/injectNetworkStrategy.spec.js
Aaronius: Thanks for the feedback. I'll make this update before merging.

package.json
jonsnyder: Thanks for the explanation. I didn't know that it would install the dependencies and then run the install script. Disregard my comment.

Aaronius:
So it turns out Mirage doesn't support IE. 😭  They don't seem to state their browser support anywhere explicitly, but they're using things like `Object.assign` which break the tests in IE. Is anyone opposed to me just making `injectNetworkStrategy.spec.js` straight unit tests rather than trying to mimic functional tests?

jonsnyder:
That's too bad you went through the work of using mirage. I think setting up jasmine mocks is fine. The functional tests make sure that the requests are working correctly.

Aaronius:
I changed the `injectNetworkStrategy` tests so they're no longer using Mirage. @jfkhoury and @jonsnyder do you mind reviewing cb6f470? Thank you! 🙏

src/core/network/injectNetworkStrategy.js
```diff
@@ -10,22 +10,29 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+export default ({
+  window,
+  logger,
+  injectSendFetchRequest,
+  injectSendXhrRequest,
+  injectSendBeaconRequest
+}) => {
+  const { fetch, navigator, XMLHttpRequest } = window;
+  const sendFetchRequest = isFunction(fetch)
+    ? injectSendFetchRequest({ fetch })
```
Aaronius, line 24: I renamed `injectFetch` and `injectSendBeacon` to `injectSendFetchRequest` and `injectSendBeaconRequest` to distinguish their resulting functions (`sendFetchRequest` and `sendBeaconRequest` from their underlying native APIs (`fetch` and `sendBeacon`) and I also thought it would be nice if they were consistent with `injectSendXhrRequest`.

test/unit/specs/core/network/injectSendBeaconRequest.spec.js
```diff
@@ -24,46 +24,60 @@ const guardForSendBeaconAvailability = spec => {
...
+      const sendBeaconRequest = injectSendBeaconRequest({
+        navigator,
+        sendFetchRequest
       });
+      // eslint-disable-next-line consistent-return
+      return sendBeaconRequest("https://example.com/endpoint", body).then(
+        () => {
+          expect(sendFetchRequest).not.toHaveBeenCalled();
+        }
+      );
```
Aaronius, line 80: Nothing really changed as far as logic goes in these tests. It's just renaming and sending arguments into `injectSendBeaconRequest ` as an object.

test/unit/specs/core/network/injectSendFetchRequest.spec.js
```diff
@@ -10,33 +10,35 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
 
   it("rejects returned promise upon network failure", () => {
-    const nativeFetch = jasmine
+    const fetch = jasmine
       .createSpy()
       .and.returnValue(Promise.reject(new Error("No connection")));
-    const fetch = injectFetch(nativeFetch);
-    return fetch("http://example.com/endpoint", { a: "b" })
+    const sendFetchRequest = injectSendFetchRequest({ fetch });
+    return sendFetchRequest("http://example.com/endpoint", { a: "b" })
```
Aaronius, line 41: Nothing really changed as far as logic goes in these tests. It's just renaming and sending arguments into `injectSendFetchRequest ` as an object.

src/core/network/injectNetworkStrategy.js
```diff
@@ -10,22 +10,29 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+    ? injectSendBeaconRequest({
+        navigator,
+        sendFetchRequest,
+        logger
+      })
+    : sendFetchRequest;
 
   return ({ url, body, documentMayUnload }) => {
-    const method = documentMayUnload ? sendBeacon : fetch;
+    const method = documentMayUnload ? sendBeaconRequest : sendFetchRequest;
```
Aaronius, line 35: The logic in this module didn't really change except that we're no longer checking to see if `window.navigator` is truthy, since it will always be truthy. The rest is just how dependencies are injected into this module and how dependencies are injected into those modules (I hope that makes sense).

Aaronius:
I'm down with it. I'll make it happen.

Aaronius:
@jonsnyder I made that change. I'm not too concerned about the size of index.js. While it's getting larger, the logical complexity of it seems to be staying pretty low. Want to review my change before I merged?

src/core/network/injectSendBeaconRequest.js
```diff
@@ -10,11 +10,11 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-export default ({ navigator, sendFetchRequest, logger }) => {
+export default ({ sendBeacon, sendFetchRequest, logger }) => {
   return (url, body) => {
     const blob = new Blob([body], { type: "text/plain; charset=UTF-8" });
-    if (!navigator.sendBeacon(url, blob)) {
-      logger.log("The `beacon` call has failed; falling back to `fetch`");
```
Aaronius, line 17: The previous message felt a bit alarmist.

jonsnyder (approved):
Thanks! I like the changes.

PR #622, opened 2020-10-26
CORE-45072 Fix functional test failure when running against Alloy prod.

Aaronius:
## Description
Fixes a functional tests that would incorrectly fail when run against Alloy prod. The test that was failing was C2580. Its source is as follows:

```
  await configureAlloy(debugEnabledConfig);
  await getLibraryInfoCommand();
  await t.expect(getAlloyCommandQueueLength()).eql(2);
  const logger = await createConsoleLogger();
  await injectAlloyDuringTest();
  await logger.info.expectMessageMatching(/Executing getLibraryInfo command/);
```

The problem here is that when we call `injectAlloyDuringTest()`, it wouldn't wait for the prod Alloy library to finish loading before moving on to `await logger.info.expectMessageMatching(/Executing getLibraryInfo command/);` and therefore that assertion would fail because the log message hadn't been seen yet. We don't have this problem when running tests against Alloy int, because the Alloy code in that case is added as an inline script, which means it would be executed before we run the logger assertion.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-45072

## Motivation and Context
Failing tests are no good.

## Types of changes

## Checklist:

### DIFF
--- a/test/functional/helpers/createFixture/clientScripts.js
+++ b/test/functional/helpers/createFixture/clientScripts.js
@@ -169,9 +169,14 @@ const injectAlloyDuringTestForInt = ClientFunction(
  */
 const injectAlloyDuringTestForProd = ClientFunction(
   () => {
-    const scriptElement = document.createElement("script");
-    scriptElement.src = remoteAlloyLibraryUrl;
-    document.getElementsByTagName("head")[0].appendChild(scriptElement);
+    return new Promise(resolve => {
+      const scriptElement = document.createElement("script");
+      scriptElement.src = remoteAlloyLibraryUrl;
+      scriptElement.addEventListener("load", () => {
+        resolve();
+      });
+      document.getElementsByTagName("head")[0].appendChild(scriptElement);
+    });
   },
   {
     dependencies: {
--- a/test/functional/specs/Data Collector/C8118.js
+++ b/test/functional/specs/Data Collector/C8118.js
@@ -33,6 +33,7 @@ test("Test C8118: Load page with link. Click link. Verify event.", async () => {
 
   await t.click(Selector("#alloy-link-test"));
   await t.expect(getLocation()).contains("blank.html");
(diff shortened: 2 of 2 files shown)


test/functional/specs/Data Collector/C8118.js
```diff
@@ -33,6 +33,7 @@ test("Test C8118: Load page with link. Click link. Verify event.", async () => {
 
   await t.click(Selector("#alloy-link-test"));
   await t.expect(getLocation()).contains("blank.html");
+  await t.expect(networkLogger.edgeCollectEndpointLogs.requests.length).eql(1);
```
Aaronius, line 36: This is unrelated to the test being fixed in the PR. In this case, the test is failing when run against prod as expected because Alloy prod doesn't handle sendBeacon properly. However, the failure message was `TypeError: Cannot read property 'request' of undefined` because we try to access `networkLogger.edgeCollectEndpointLogs.requests` on the line below. I decided to add this check for length just to make the failure message easier to understand.

test/functional/helpers/createFixture/clientScripts.js
```diff
@@ -169,9 +169,14 @@ const injectAlloyDuringTestForInt = ClientFunction(
...
-    scriptElement.src = remoteAlloyLibraryUrl;
-    document.getElementsByTagName("head")[0].appendChild(scriptElement);
+    return new Promise(resolve => {
+      const scriptElement = document.createElement("script");
+      scriptElement.src = remoteAlloyLibraryUrl;
+      scriptElement.addEventListener("load", () => {
+        resolve();
+      });
+      document.getElementsByTagName("head")[0].appendChild(scriptElement);
+    });
```
Aaronius, line 179: This is the fix for C2580.

jfkhoury (approved):
Thank you!!

ninaceban (approved):

PR #634, opened 2020-11-10
CORE-54518 Support not calling setConsent

jonsnyder:
## Description

This is a proof of concept to find any problems with the idea of removing the need for the setConsent call. This PR changes Alloy to support only two default consents ("in", and "out"). When consent is out, only events that have different consents are sent to experience edge. While an event with different consent is outstanding, other requests are held up. When Alloy detects a consent change, it passes the new consent hash under the meta key. Alloy expects this hash to be written to the consent cookie.

## Related Issue

https://jira.corp.adobe.com/browse/CORE-54518

## Motivation and Context

With Safari's newest security change cookies set from a CNAME domain are only allowed to last for 7 days. Currently we are telling customers to call setConsent only when the consent changes, but if the consent cookie expires, Alloy will not know what the current consent level is. If consent is always passed in through the event XDM, Alloy can determine when it needs to get a refreshed consent cookie from conductor.

## Types of changes

## Checklist:

### DIFF
--- a/src/constants/consentPurpose.js
+++ b/src/constants/consentPurpose.js
@@ -12,3 +12,4 @@ governing permissions and limitations under the License.
 
 // eslint-disable-next-line import/prefer-default-export
 export const GENERAL = "general";
+export const CONSENT_HASH = "consentHash";
--- a/src/core/config/createCoreConfigs.js
+++ b/src/core/config/createCoreConfigs.js
@@ -14,11 +14,11 @@ import { boolean, string, callback, enumOf } from "../../utils/validation";
 import { noop } from "../../utils";
 import { EDGE as EDGE_DOMAIN } from "../../constants/domain";
 import EDGE_BASE_PATH from "../../constants/edgeBasePath";
-import { IN, PENDING } from "../../constants/consentStatus";
+import { IN, OUT, PENDING } from "../../constants/consentStatus";
 
 export default () => ({
   debugEnabled: boolean().default(false),
-  defaultConsent: enumOf(IN, PENDING).default(IN),
+  defaultConsent: enumOf(IN, PENDING, OUT).default(IN),
   edgeConfigId: string()
     .unique()
     .required(),
--- a/src/core/consent/hash.js
+++ b/src/core/consent/hash.js
@@ -1,0 +1,1 @@
+export default object => JSON.stringify(object);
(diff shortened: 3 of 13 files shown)


src/constants/consentPurpose.js
```diff
@@ -12,3 +12,4 @@ governing permissions and limitations under the License.
 
 // eslint-disable-next-line import/prefer-default-export
 export const GENERAL = "general";
+export const CONSENT_HASH = "consentHash";
```
jonsnyder, line 15: I'm expecting the consent cookie to be like: "general=in;consentHash=1234abcd"

src/core/config/createCoreConfigs.js
```diff
@@ -14,11 +14,11 @@ import { boolean, string, callback, enumOf } from "../../utils/validation";
 import { noop } from "../../utils";
 import { EDGE as EDGE_DOMAIN } from "../../constants/domain";
 import EDGE_BASE_PATH from "../../constants/edgeBasePath";
-import { IN, PENDING } from "../../constants/consentStatus";
+import { IN, OUT, PENDING } from "../../constants/consentStatus";
 
 export default () => ({
   debugEnabled: boolean().default(false),
-  defaultConsent: enumOf(IN, PENDING).default(IN),
+  defaultConsent: enumOf(IN, PENDING, OUT).default(IN),
```
jonsnyder, line 21: I left PENDING here to be backward compatible, but PENDING default consent behaves just like OUT.

src/core/consent/hash.js
```diff
@@ -0,0 +1 @@
+export default object => JSON.stringify(object);
```
jonsnyder, line 1: This is just a placeholder; it will need to change to a real hashing function. Preferably where the order of the object keys doesn't change the hash output, and where the resulting string is small (maybe base64 encoded).

src/core/createEvent.js
```diff
@@ -52,6 +52,13 @@ export default () => {
 
       return userXdm.web.webPageDetails.viewName;
     },
+    getConsent() {
+      if (!userXdm || !userXdm.consentStrings) {
```
jonsnyder, line 56: Russ is still working on the XDM mixin for consent and preferences. I'm not sure where it will be, but this is the location for IAB consent strings.

Aaronius (commented):
Nice job! I'm sad it doesn't really decrease our code, but I still prefer the general strategy. I'll approve once we get hashing and such in place.

jfkhoury:
Hey @jonsnyder, we can close this PR right?

jonsnyder:
We will revisit this when it becomes top priority.

PR #637, opened 2020-11-13
Get ready to publish Alloy as an NPM repo

jonsnyder:
## Description

This PR gets the repo ready to be published as an NPM library. Other PRs will automate the build process. The NPM library publishes an ES6 module which exposes the baseCode, core initialization, and the createEventMergeId functions. The built baseCode and standalone javascript files are not published in the NPM library. If you want to use the built versions you should use the version on the CDN.

See #632 

There are three entry points now:
  1. src/baseCode.js - used to generate the base code snippet
  2. src/standAlone.js - used to generate the standalone code
  3. src/index.js - used to generate the ES6 modules

I was also able to remove the REACTOR pre-processor statements, as the launch extension can use the ES6 module and the createEventMergeId directly.
## Related Issue
https://jira.corp.adobe.com/browse/CORE-52500

## Motivation and Context

1. We don't want to have to copy built files as part of the Launch Extension release process. The launch extension can use the NPM library to integrate.
2. Publishing an NPM library allows customers to integrate with their javascript build system.
3. If we were to export more modules we could allow customers to do custom builds. (i.e. exclude the personalization module).

## Types of changes

## Checklist:


package.json
```diff
@@ -1,8 +1,13 @@
 {
   "name": "@adobe/alloy",
   "version": "2.3.0",
-  "description": "Client-Side SDK for Unified Data Collection",
-  "main": "src/core/index.js",
+  "description": "Adobe Experience Platform Web SDK",
```
jfkhoury, line 4: Finally :)

package.json
```diff
@@ -53,18 +58,25 @@
   "license": "Apache-2.0",
   "bundlesize": [
     {
-      "path": "./dist/standalone/alloy.min.js"
+      "path": "./dist/alloy.min.js"
     }
   ],
   "dependencies": {
+    "css.escape": "^1.5.1",
+    "typescript": "^4.0.5",
```
jfkhoury, line 66: Does that pull typescript into the library?

jonsnyder: Oops, I didn't mean to add this as a dependency. Thanks for catching.

README.md
```diff
@@ -2,8 +2,56 @@
...
+```javascript
+import { baseCode, core } from "@adobe/alloy";
+baseCode(["alloy"]);   // creates the window.alloy function
+core();                // runs
+
+window.alloy("config", { ... });
+window.alloy("sendEvent", { ... });
+```
+
+The ES6 exports are exposed via the NPM "module" property, so you may need to adjust your rollup system to look for the module property.
```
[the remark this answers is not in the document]
Aaronius: Couple questions here: 

(1) Are you aware of any bundlers that don't support `module` by default? I'm aware it isn't standard, but I think it's been supported by default for 2+ years on the major bundlers. If so, this line may not be providing value and just introducing cognitive load, but who knows. If you keep it, it might be better to say something like "are exposed by the NPM package via the `module` field of its package.json, so you may...". If I'm a consumer having problems, that might give me more context for googling. 🤷‍♂️ 


(2) I'm trying to think through why we're creating `main.js` and `modules.js` as bundles. Normally what I've seen is that you would just run all the source code files through babel and dump all the resulting files into, say, an `es5` directory, then point the `main` field in package.json to `es5/index.js`. For the es6 variant, you may just be able to point the `module` field in package.json to `src/index.js`.

One of the benefits of not generating bundles for the npm package consumers is that if the consuming application already has a dependency that matches one of our dependencies, the final application bundle only needs to include the dependency code once. Currently, we only have a couple dependencies, so that chance is somewhat low, but that may grow over time

One effect of taking this approach is that the extension will need to do some bundling before getting packaged and uploaded. But, I believe this would mean that you would be able to move the Launch core module dependencies from `peerDependencies` back to `dependencies` in this repo. When you do the bundling for the extension, I believe you'd just mark Launch core module dependencies as "externals" so that Launch can provide them instead of bundling them in from `node_modules`.

Sorry, I should've thought through this during my previous review as it definitely affects what I was saying in https://github.com/adobe/alloy/pull/632/files#r518529523.

src/baseCode/index.js
```diff
@@ -59,4 +59,4 @@ governing permissions and limitations under the License.
       window[instanceName].q = [];
     }
   });
-})(window, ["alloy"]);
+};
```
Aaronius, line 62: One of the consequences of doing this is that the code emitted to the console during our build process, which I believe is intended to be copied into the installation instructions within our user docs, is different/larger than the base code we already have in the instructions. For example, here's what we have currently:

```
!function(n,o){o.forEach(function(o){n[o]||((n.__alloyNS=n.__alloyNS||
[]).push(o),n[o]=function(){var u=arguments;return new Promise(
function(i,l){n[o].q.push([i,l,u])})},n[o].q=[])})}
(window,["alloy"]);
```

And here's what's now emitted to the console during the build process:

```
!function(){"use strict";var n;(function(n){n.forEach((function(n){window[n]||(window.__alloyNS=window.__alloyNS||
[],window.__alloyNS.push(n),window[n]=function(){var o=arguments;return new Promise(
(function(w,i){window[n].q.push([w,i,o])}))},window[n].q=[])}))})
(["alloy"])}();
```

It seems we would rather use the old one than the new one. I'm not sure how to best solve this problem. We might be able to maintain this file as it previously was, and then do some magic during build time to manipulate the file content into a module. Another option would be to maintain two separate files with similar content (both the old file and this new module-based one). Anyway, it's something to think about.

test/unit/specs/components/EventMerge/createComponent.spec.js
```diff
@@ -0,0 +1,26 @@
...
+    const createEventMergeId = () => undefined;
+    expect(createComponent({ createEventMergeId })).toEqual({
+      commands: {
+        createEventMergeId: {
+          run: createEventMergeId
+        }
+      }
+    });
+  });
+});
```
Aaronius, line 26: I'm still seeing the error I mentioned in https://github.com/adobe/alloy/pull/632/files#r518533046.

Aaronius (commented):
Thanks for moving this forward Jon. I left some comments that we should probably resolve.

src/baseCode/index.js
jonsnyder: I modified the rollup & terser options and code so that the same output is generated.

jonsnyder:
@Aaronius can you take another look at this PR?

jonsnyder:
@Aaronius can you take another look at this PR? I implemented the createInstance method like you suggested in the docs PR.

test/unit/specs/core/createInstanceFunction.spec.js
```diff
@@ -10,7 +10,7 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import createInstance from "../../../../src/core/createInstance";
+import createInstance from "../../../../src/core/createInstanceFunction";
```
Aaronius, line 13: You probably want to rename `createInstance` to `createInstanceFunction` here.

Aaronius (approved):
I like.

package.json
```diff
@@ -68,6 +74,8 @@
     "uuid": "^3.3.2"
   },
   "devDependencies": {
+    "@adobe/alloy": "file:./",
```
jonsnyder, line 77: For now this just references the current directory, but when we publish the NPM library I will change this to reference the actual version.

jfkhoury (approved):
Great work @jonsnyder !! 

I recommend testing the built library manually in the sandbox since there's so many changes :)

PR #646, opened 2020-12-14
Initial work related to adding onBeforeLinkCollect() callback handler

garyr:
Initial work related to adding a callback handler mechanism for onBeforeLinkCollect

## Types of changes

## Checklist:

### DIFF
--- a/src/components/ActivityCollector/attachClickActivityCollector.js
+++ b/src/components/ActivityCollector/attachClickActivityCollector.js
@@ -12,16 +12,34 @@ governing permissions and limitations under the License.
 
 import { noop } from "../../utils";
 
-const createClickHandler = ({ eventManager, lifecycle, handleError }) => {
+const createClickHandler = ({
+  eventManager,
+  lifecycle,
+  handleError,
+  onBeforeLinkCollect
+}) => {
   return clickEvent => {
     // TODO: Consider safeguarding from the same object being clicked multiple times in rapid succession?
     const clickedElement = clickEvent.target;
     const event = eventManager.createEvent();
+
     return (
       lifecycle
         .onClick({ event, clickedElement })
         .then(() => {
-          if (event.isEmpty()) {
+          let shouldCollectLinkClick = null;
+          if (onBeforeLinkCollect) {
+            shouldCollectLinkClick = onBeforeLinkCollect({
+              xdm: event.xdm ? event.xdm : null,
+              element: clickedElement
+            });
+          }
+
+          return shouldCollectLinkClick;
+        })
+        .then(shouldCollectLinkClick => {
+          // don't send event if empty or onBeforeLinkCollect returned false
+          if (event.isEmpty() || shouldCollectLinkClick === false) {
             return Promise.resolve();
           }
           return eventManager.sendEvent(event);
@@ -43,10 +61,13 @@ export default ({ config, eventManager, lifecycle, handleError }) => {
     return;
   }
 
+  const onBeforeLinkCollect = config.onBeforeLinkCollect;
+
   const clickHandler = createClickHandler({
     eventManager,
     lifecycle,
-    handleError
+    handleError,
+    onBeforeLinkCollect
   });
 
   document.addEventListener("click", clickHandler, true);
--- a/src/core/config/createCoreConfigs.js
+++ b/src/core/config/createCoreConfigs.js
@@ -31,5 +31,6 @@ export default () => ({
   orgId: string()
     .unique()
     .required(),
+  onBeforeLinkCollect: callback().default(noop),
   onBeforeEventSend: callback().default(noop)
 });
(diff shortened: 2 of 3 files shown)


src/core/config/createCoreConfigs.js
```diff
@@ -31,5 +31,6 @@ export default () => ({
   orgId: string()
     .unique()
     .required(),
+  onBeforeLinkCollect: callback().default(noop),
```
Aaronius, line 34: It kind of bugs me we're using `Collect` here and `Send` in `onBeforeEventSend`. I suppose that's because we have a config option called `clickCollectionEnabled` and are aligning `onBeforeLinkCollect` with that.  Maybe `onBeforeLinkClickSend` would be better? Or maybe even `onLinkClick`? I'm curious to hear what you or others think.

test/unit/specs/components/ActivityCollector/attachClickActivityCollector.spec.js
```diff
@@ -88,6 +89,51 @@ describe("ActivityCollector::attachClickActivityCollector", () => {
...
 
+  it("Sends an event if onBeforeLinkCollect does not return false", () => {
+    eventManager.createEvent = () => {
+      return {
+        isEmpty: () => false
+      };
+    };
+    spyOn(eventManager, "sendEvent").and.callThrough();
+    config.onBeforeLinkCollect = function() {};
+    spyOn(config, "onBeforeLinkCollect").and.callThrough();
```
Aaronius, line 100: The easiest way to do these two lines is like this: 

`config.onBeforeLinkCollect = jasmine.createSpy();`

test/unit/specs/components/ActivityCollector/attachClickActivityCollector.spec.js
```diff
@@ -88,6 +89,51 @@ describe("ActivityCollector::attachClickActivityCollector", () => {
...
+    eventManager.createEvent = () => {
+      return {
+        isEmpty: () => false
+      };
+    };
+    spyOn(eventManager, "sendEvent").and.callThrough();
+    config.onBeforeLinkCollect = function() {
+      return false;
+    };
+    spyOn(config, "onBeforeLinkCollect").and.callThrough();
```
Aaronius, line 129: and then here you would do:

`config.onBeforeLinkCollect = jasmine.createSpy().and.returnValue(false);`

test/unit/specs/components/ActivityCollector/attachClickActivityCollector.spec.js
```diff
@@ -20,6 +20,7 @@ describe("ActivityCollector::attachClickActivityCollector", () => {
   let handleError;
   beforeEach(() => {
     config.clickCollectionEnabled = true;
+    config.onBeforeLinkCollect = undefined;
```
Aaronius, line 23: I would recommend removing this line. The reasoning is that without this line, we can test the most common case of a customer providing a config with no `onBeforeLinkCollect` property at all. As you have it, we're testing what happens when a customer sets a property of `onBeforeLinkCollect` with a value of `undefined`, which wouldn't be common.

Aaronius (commented):
You're barking up the right tree!

src/components/ActivityCollector/attachClickActivityCollector.js
```diff
@@ -12,16 +12,34 @@ governing permissions and limitations under the License.
...
+
     return (
       lifecycle
         .onClick({ event, clickedElement })
         .then(() => {
-          if (event.isEmpty()) {
+          let shouldCollectLinkClick = null;
+          if (onBeforeLinkCollect) {
+            shouldCollectLinkClick = onBeforeLinkCollect({
+              xdm: event.xdm ? event.xdm : null,
```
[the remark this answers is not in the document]
Aaronius: From today's standup discussion, but maybe modifying our conclusion a bit, I think you would create a new method on event called `executeOnBeforeLinkClick`. From here in attachClickActivityCollector.js, you would then call it:

```
event.executeOnBeforeLinkCollect(onBeforeLinkCollect);
```

And event would execute `onBeforeLinkCollect`.

You might run into the question of which XDM do we pass into `onBeforeLinkCollect`. We have this concept of user-provided XDM vs automatically-collected XDM. What @jonsnyder was saying during standup was that the XDM that we generate in https://github.com/adobe/alloy/blob/e30caf3a6f520410221b6008f797561b641a23b1/src/components/ActivityCollector/createLinkClick.js#L60-L72 is really acting like "user-provided data", since we're the one generating the event on behalf of the customer. Usually user-provided XDM is added to the event using `event.setUserXdm`. 

Right now, createLinkClick.js is calling `event.mergeXdm` instead of `event.setUserXdm`. We could switch it to call `event.setUserXdm` and then when event calls `onBeforeLinkCollect`, it could provide `onBeforeLinkCollect` with the "user XDM", which is currently stored in a `userXdm` variable inside event. We'll need to make sure this doesn't cause other problems though. I would poke around the scenario where a user clicks a link that's also part of a Personalization offer, since Personalization also attempts to augment the event XDM (https://github.com/adobe/alloy/blob/e30caf3a6f520410221b6008f797561b641a23b1/src/components/Personalization/createOnClickHandler.js#L20-L21).

Anyway, go with that and see where you get. It's probably worth bringing up your findings in standup or a tech checkin and just validate that we're all on the same page.

PR #648, opened 2020-12-17
CORE-52954 Always use interact endpoint with fetch when an identity hasn't been established.

Aaronius:
## Description
If a new user comes to a website and the first command executed is a `sendEvent` command with `documentUnloading` set to `true`, Alloy uses `sendBeacon` to send the event. Since no response is returned from Konductor, no identity cookie is set, which results in no identity being established and Alloy logging the message:

`An identity was not set properly. Please verify that the org ID 334F60F35E1597910A495EC2@AdobeOrg configured in Alloy matches the org ID specified in the edge configuration.`

We should never use `sendBeacon` or the `collect` endpoint if an identity hasn't been established for the user.

In order to determine which endpoint action to use (e.g., `interact` vs `collect`) and which browser network API to use (`fetch` or `sendBeacon`), the Identity component has to somehow indicate whether an identity has been established as the payload makes its way through the Alloy lifecycle. Rather than adding more methods, like `setIsIdentityEstablished` to the payload object that gets delivered to components through the lifecycle, I decided there was a better way to structure payloads. My theory is that the payload's responsibility should be just that: the payload. In other words, it should be responsible for the content that actually gets sent within the request. It should **not** be responsible for things related to the request itself, like which domain, endpoint action, or browser network API should be used. As such, I split out the objects so that we have:

* Data Collection Request Payload
* Identity Request Payload
* Consent Request Payload
* Data Collection Request
* Identity Request
* Consent Request

Another aspect is that `sendNetworkRequest` and `networkStrategy` modules used `documentMayUnload` off the payload to determine whether `fetch` or `sendBeacon` should be used, while `eventManager` used `documentMayUnload` to determine whether `interact` or `collect` should be used. It was a bit odd that two things that are so inter-related (which browser network API to use and which endpoint to hit) were so distal. I would have had to add logic to both of these places to take into consideration whether an identity had already been established, since both the browser API and the endpoint action would be determined based on this piece of data.

As a result, I moved the logic determining which endpoint action should be used and which browser API should be used into the request objects themselves. So, for example, the Data Collection request is responsible for indicating whether to use the `interact` or `collect` endpoint and whether to use `fetch` or `sendBeacon`.

Most of this PR is a reflection of the changes I just described. I'll provide comments on some of the other changes that were made.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-52954

## Motivation and Context
Customers would appreciate the SDK making its best effort to establish an identity. Also, showing fewer errors in the console is valuable.

## Types of changes

## Checklist:

- [x] I have made any necessary test changes and all tests pass. (except some functional tests still failing in `main`)


src/core/createLifecycle.js
```diff
@@ -16,9 +16,6 @@ const hookNames = [
   "onComponentsRegistered",
   // Called before an event is sent on a data collection request
   "onBeforeEvent",
-  // Called before each data collection request
-  // (`interact` or `collect` endpoints)
-  "onBeforeDataCollectionRequest",
```
Aaronius, line 21: I removed the `onBeforeDataCollectionRequest` hook entirely because we didn't end up using it.

src/core/edgeNetwork/requests/createDataCollectionRequest.js
```diff
@@ -0,0 +1,67 @@
...
+    // 2. The document is unloaded and Alloy receives a response with an
+    // identity back from the interact endpoint before navigation is completed.
+    // No error will be thrown and an identity is established. This is good.
+    // 3. The document is unloaded and Alloy does not receive the empty response
+    // back from the collect endpoint before navigation is completed. In this
+    // case, no error is thrown, but no identity was established and it's
+    // more likely that the request never makes it to the server because we're
+    // using fetch instead of sendBeacon.
+    //
+    // The second approach seemed preferable.
```
Aaronius, line 50: This is useful in understanding some of the context that went into the decisions made.

src/core/network/injectSendNetworkRequest.js
```diff
@@ -12,38 +12,37 @@ governing permissions and limitations under the License.
...
     // insufficient to get all the nested raw data, because it's
     // not recursive (it doesn't call toJSON() on the event objects).
     // Parsing the result of JSON.stringify(), however, gives the
     // fully recursive raw data.
-    // JSON.parse is expensive so we short circuit if logging is disabled.
-    if (logger.enabled) {
-      logger.logOnBeforeNetworkRequest({
-        url,
-        requestId,
-        payload: JSON.parse(stringifiedPayload)
```
Aaronius, line 37: The idea behind wrapping this particular statement in `if (logger.enabled) {` was so that we would only run `JSON.parse` if we were actually going to log the result to the console (because it was considered expensive). It's not really helpful though, because we end up running `JSON.parse(stringifiedPayload)` eventually within this module anyway. So now we just do it once up front.

test/functional/helpers/createFixture/clientScripts.js
```diff
@@ -35,15 +36,22 @@ const localAlloyLibraryPath = path.join(
...
+  // point it will retrieve fresh file content, cache it, and return it.
+  return readCache.sync(localAlloyLibraryPath, "utf8");
+};
 
-// We're only testing against the alloy code found in /dist when alloyEnv
-// is "int". Otherwise, we're testing against the prod alloy that's on the CDN
-// and in those cases alloy code may not even exist in /dist. This is why
-// we only try to load the file from the file system if alloyEnv is int.
-if (alloyEnv === "int") {
-  localAlloyCode = fs.readFileSync(localAlloyLibraryPath, "utf8");
```
Aaronius, line 45: The changes in this file were to improve watch mode when running functional tests. I noticed that depending on what changes I was making in our source code and how I was focusing a particular test, that my changes weren't being used when the test was re-run. This was happening because we were reading the produced Alloy build file once using this line of code and then TestCafe wasn't running this line again when re-running the test in watch mode.

I got around this by making this a function (see `getLocalAlloyCode` in my changes) that reads the Alloy build file every time TestCafe injects scripts into the test page, which happens every time a test is run or re-run. I optimized it though by using the `read-cache` npm package, which caches a file's contents and returns that content as long as the file hasn't changed since the last call (it uses the file's last modified date to determine if it's changed rather than having to read all its contents again). That's just so we're not reading the full file contents from disk for every test.

test/functional/specs/Data Collector/C455258.js
```diff
@@ -20,33 +23,49 @@ test.meta({
...
+
+  if (isSendBeaconSupported()) {
+    await t.expect(sendBeaconMock.getCallCount()).eql(1);
+  }
+
+  // documentUnloading is not set to true. The request should go to the
+  // interact endpoint.
+  await sendEvent({});
+
+  await t.expect(networkLogger.edgeInteractEndpointLogs.requests.length).eql(2);
```
Aaronius, line 68: This functional test and the next one are what really test the meat of the problem we're trying to solve.

test/functional/specs/Logging/C532204.js
```diff
@@ -30,7 +30,9 @@ test.meta({
 
 const setupLogger = ClientFunction(() => {
   ["log", "info", "warn", "error"].forEach(methodName => {
+    // eslint-disable-next-line no-console
     const origConsoleMethod = console[methodName];
+    // eslint-disable-next-line no-console
```
Aaronius, line 35: We were getting eslint warnings.

test/unit/constants/uuidV4Regex.js
```diff
@@ -10,9 +10,4 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-export default ({ sendFetchRequest, sendBeaconRequest }) => {
-  return ({ url, body, documentMayUnload }) => {
-    const method = documentMayUnload ? sendBeaconRequest : sendFetchRequest;
-    return method(url, body);
-  };
-};
```
Aaronius, line 18: Git correlated these files, but they're not related at all. injectNetworkStragy.js has been removed. At one point it used to do more, but it wasn't providing much value anymore.

test/unit/specs/components/Privacy/createConsentRequestPayload.spec.js
```diff
@@ -11,68 +11,51 @@ governing permissions and limitations under the License.
...
+  });
+
+  it("sets consent", () => {
+    const payload = createConsentRequestPayload();
     payload.setConsent([
       {
         standard: "Adobe",
         version: "1.0",
         value: {
-          general: "out"
```
Aaronius, line 50: I didn't really intend to switch this from `out` to `in`, but it doesn't hurt anything.

test/unit/specs/core/createEventManager.spec.js
```diff
@@ -131,33 +136,24 @@ describe("createEventManager", () => {
...
+      const consentDeferred = defer();
+      lifecycle.onBeforeEvent.and.returnValue(onBeforeEventDeferred.promise);
+      consent.awaitConsent.and.returnValue(consentDeferred.promise);
       eventManager.sendEvent(event);
+      expect(lifecycle.onBeforeEvent).toHaveBeenCalled();
       return flushPromiseChains()
         .then(() => {
-          expect(lifecycle.onBeforeEvent).toHaveBeenCalled();
-          expect(
-            lifecycle.onBeforeDataCollectionRequest
```
Aaronius, line 142: `onBeforeDataCollectionRequest` has been removed so I had to rework this a bit. I also improved it a bit to consider `awaitConsent`'s effect on the lifecycle.

test/unit/specs/components/Identity/injectEnsureSingleIdentity.spec.js
```diff
@@ -23,9 +23,9 @@ describe("Identity::injectEnsureSingleIdentity", () => {
   let ensureSingleIdentity;
 
   let sentIndex;
-  let recievedIndex;
-  let payloads;
-  let requestsSentYet;
+  let receivedIndex;
```
jonsnyder, line 26: 👍 "i" before "e" except after "c"

jonsnyder (approved):
The new refactor looks great! The only weird bit now is the "documentMayUnload" being part of the payload, but that can be solved another day. I'm also excited to be able to use the auto-reload when testing.

Aaronius:
@jonsnyder FYI, I've moved getDocumentMayUnload the the data collection request payload.

jonsnyder (approved):
👍

jfkhoury (approved):

PR #649, opened 2021-01-07
HOLD: CORE-56247 Encoding web.webpageDetails.URL and web.webReferrer.URL to conform with URI spec.

Aaronius:
## Description
URLs reported by browser APIs may not be URI-compliant because browsers are more lenient than the [URI RFC](https://tools.ietf.org/html/rfc3986). The XDM fields we're populating, however, should conform to the URI spec. For this reason, we need to encode the URL.  We decode the URL first so it doesn't end up double encoded if it was previously encoded.

## Related Issue
https://jira.corp.adobe.com/browse/CORE-56247

## Motivation and Context
Ingestion would fail when non-URI complaint URLs were encountered.

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Context/injectWeb.js
+++ b/src/components/Context/injectWeb.js
@@ -12,14 +12,20 @@ governing permissions and limitations under the License.
 
 import { deepAssign } from "../../utils";
 
+// URLs reported by browser APIs may not be URI-compliant because browsers are more lenient
+// than the URI RFC (https://tools.ietf.org/html/rfc3986). The XDM fields we're populating,
+// however, should conform to the URI spec. For this reason, we need to encode the URL.
+// We decode the URL first so it doesn't end up double encoded if it was previously encoded.
+const encode = url => encodeURI(decodeURI(url));
+
 export default window => {
   return xdm => {
     const web = {
       webPageDetails: {
-        URL: window.location.href || window.location
+        URL: encode(window.location.href || window.location)
       },
       webReferrer: {
-        URL: window.document.referrer
+        URL: encode(window.document.referrer)
       }
     };
     deepAssign(xdm, { web });
--- a/test/unit/specs/components/Context/injectWeb.spec.js
+++ b/test/unit/specs/components/Context/injectWeb.spec.js
@@ -1,23 +1,43 @@
 import injectWeb from "../../../../../src/components/Context/injectWeb";
 
-describe("Context::injectWeb", () => {
-  const window = {
-    location: { href: "http://mylocation.com" },
-    document: {
-      referrer: "http://myreferrer.com"
-    }
-  };
-
+fdescribe("Context::injectWeb", () => {
   it("works", () => {
+    const window = {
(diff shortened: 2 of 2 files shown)


test/unit/specs/components/Context/injectWeb.spec.js
```diff
@@ -1,23 +1,43 @@
...
 
-describe("Context::injectWeb", () => {
-  const window = {
-    location: { href: "http://mylocation.com" },
-    document: {
-      referrer: "http://myreferrer.com"
-    }
-  };
-
+fdescribe("Context::injectWeb", () => {
```
jfkhoury, line 3: Should we change to “describe”?

jfkhoury (approved):

src/components/Context/injectWeb.js
```diff
@@ -12,14 +12,20 @@ governing permissions and limitations under the License.
 
 import { deepAssign } from "../../utils";
 
+// URLs reported by browser APIs may not be URI-compliant because browsers are more lenient
+// than the URI RFC (https://tools.ietf.org/html/rfc3986). The XDM fields we're populating,
+// however, should conform to the URI spec. For this reason, we need to encode the URL.
+// We decode the URL first so it doesn't end up double encoded if it was previously encoded.
+const encode = url => encodeURI(decodeURI(url));
```
liljenback, line 19: So when/if another decodeURI were to be applied on the result (e.g. decodeURI(encodeURI(decodeURI(url)))) we're sure that the end result would end up being the original url?

liljenback, line 19: By doing this we'll avoid double encoding, but we lose a way to provide a representation of the original url we obtained. This is probably less important than avoiding multiple encoding but something we need to be aware of.

Aaronius: Thanks @liljenback. I acknowledge this is the case. Do you have any ideas on how to more appropriately handle this situation?

test/unit/specs/components/Context/injectWeb.spec.js
Aaronius: 😢 Yes. Thanks for the catch. I've created https://jira.corp.adobe.com/browse/CORE-56252 so that we can start catching this through linting.

Aaronius:
Hold tight. We may end up going a different route by removing the URI restriction on the field.

src/components/Context/injectWeb.js
liljenback: I think it'll be fine to do it this way. Avoiding multiple encodings will prevent security related problems. Keeping it simple is most important.

jonsnyder (approved):
I think the better reason for us to be doing this is so that the reported URLs are in a canonical form. For example, two URLs that decode to the same thing should be reported as the same page.

Just one note about decodeURL possibly throwing an exception.

jfkhoury:
We can close this PR now @Aaronius

PR #656, opened 2021-02-08
PDCL-1245 - Support redirect Personalization offer types

ninaceban:
## Description

Redirect offers are usually the experiences that customer set to redirect from the experience's entry criteria URL to another URL and if `renderDecisions` is true we'd like to execute it as quick as possible. If there will be more than one `redirect decision` then we will execute the first one. After the redirect decision was executed we will send a notification call. 

- If `renderDecisions` is true we will execute the first redirect decision. 
- If `renderDecisions` is false we will return these decisions to the customer at the `sendEvent`.

If the Redirect Offer is A4T enabled, the redirect URL will contain analytics `tnt_a` encoded payload in the `alloy_redirect` query string param. Analytics data exchange will happen on the Konductor side.

## Types of changes

## Checklist:


jfkhoury:
@ninaceban Can we please add this feature to the sandbox? thanks!

jonsnyder (approved):

src/components/Personalization/createRedirectHandler.js
```diff
@@ -0,0 +1,30 @@
...
+  return { content, decisions: [{ id, scope }] };
+};
+
+export default ({ collect, win = window }) => {
+  return redirectDecisions => {
+    const { content, decisions } = getRedirectDetails(redirectDecisions);
+    const documentMayUnload = true;
+
+    return collect({ meta: { decisions } }, documentMayUnload).then(() => {
+      win.location.replace(content);
```
Aaronius, line 27: I'm noticing this in the [docs for `location.replace`](https://developer.mozilla.org/en-US/docs/Web/API/Location/replace):

> If the assignment can't happen because of a security violation, a DOMException of the SECURITY_ERROR type is thrown. This happens if the origin of the script calling the method is different from the origin of the page originally described by the Location object, mostly when the script is hosted on a different domain.

I haven't verified, but that seems to suggest that if Alloy is loaded from a different origin than the page loading Alloy, that an error will be thrown when Alloy calls `location.replace`. At least, that's my interpretation. Have you tested this scenario?

Aaronius (approved):
I have a concern about an error being thrown when `location.replace` is called, which I added a comment about. Otherwise, the rest is just small tweaks. Nice work @ninaceban.

src/components/Personalization/createOnResponseHandler.js
```diff
@@ -30,18 +32,34 @@ export default ({
...
+        decisions: unprocessedDecisions,
+        schema: REDIRECT_ITEM
+      }
+    );
+
+    if (
+      isNonEmptyArray(decisionsGroupedByRedirectItemSchema.matchedDecisions) &&
+      personalizationDetails.isRenderDecisions()
+    ) {
+      decisionsDeferred.resolve({});
```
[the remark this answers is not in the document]
ninaceban: yes, if we have at least one redirect offer and we want to render it (renderDecisions=true) then it doesn't make any sense to store the view related offers. If we don't want to render them then we will store SPA offers and will resolve the sendEvent promise with `__view__` and `REDIRECT_ITEM` decisions.

test/functional/specs/Personalization/C205528.js
```diff
@@ -0,0 +1,54 @@
...
+      });
+  });
+});
+
+test("Test C205528: A redirect offer should redirect the page to the URL in the redirect decision", async () => {
+  await configureAlloyInstance("alloy", config);
+  try {
+    await triggerAlloyEvent();
+  } catch (e) {
+    await t.expect(redirectLogger.requests.length).eql(1);
```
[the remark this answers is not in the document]
ninaceban: yes sure! thank you

ninaceban:
> @ninaceban Can we please add this feature to the sandbox? thanks!
sure!

src/components/Personalization/createOnResponseHandler.js
```diff
@@ -30,18 +32,34 @@ export default ({
...
+      return handleRedirectDecisions(
+        decisionsGroupedByRedirectItemSchema.matchedDecisions
+      );
+    }
+
+    const decisionsGroupedByDomActionSchema = decisionsExtractor.groupDecisionsBySchema(
+      {
+        decisions: decisionsGroupedByRedirectItemSchema.notMatchedDecisions,
```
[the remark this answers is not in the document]
ninaceban: ok, I'll rename it

src/components/Personalization/createCollect.js
```diff
@@ -11,11 +11,15 @@ governing permissions and limitations under the License.
 */
 
 export default ({ eventManager, mergeMeta }) => {
-  return ({ meta }) => {
+  return ({ meta }, documentMayUnload = false) => {
```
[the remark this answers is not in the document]
ninaceban: ok, I'll refactor

src/components/Personalization/createRedirectHandler.js
```diff
@@ -0,0 +1,30 @@
...
+
+const getRedirectDetails = redirectDecisions => {
+  const decision = redirectDecisions[0];
+  const { items, id, scope } = decision;
+  const { content } = items[0].data;
+
+  return { content, decisions: [{ id, scope }] };
+};
+
+export default ({ collect, win = window }) => {
```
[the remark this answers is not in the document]
ninaceban: thank you, I will refactor it

src/components/Personalization/createOnResponseHandler.js
```diff
@@ -10,15 +10,17 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 import { isEmptyObject } from "../../utils";
-import { DOM_ACTION } from "./constants/schema";
+import { DOM_ACTION, REDIRECT_ITEM } from "./constants/schema";
```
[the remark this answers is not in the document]
ninaceban: These are exactly the schema names. The only difference is that the ones defined by DOM_ACTION, are actions that we can execute against DOM, while the 3 others are items that the customer knows how to handle, and based on a schema they have a slightly different structure or goals.

src/components/Personalization/createRedirectHandler.js
```diff
@@ -0,0 +1,30 @@
...
+  return redirectDecisions => {
+    const { content, decisions } = getRedirectDetails(redirectDecisions);
```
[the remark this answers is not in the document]
ninaceban: Thank you, I really appreciate it! I will do a refactor in this area, maybe as part of another PR.

src/components/Personalization/createRedirectHandler.js
ninaceban: a try/catch wouldn't cause any harm, but currently, we don't try-catch in at.js. I will add a try/catch, it would catch exceptions like malformed URL as well.

Aaronius: I'm more concerned that the navigation won't actually occur. You've verified that the navigation actually occurs in this scenario though? I'm probably just misunderstanding MDN's documentation, but now I want to know what they're documentation is trying to say. If no error is being thrown in this scenario and it's actually navigating the user, then there's no need to add a catch block unless we know what/why we're catching.

sandbox/src/RedirectOffers.js
```diff
@@ -0,0 +1,23 @@
...
+import ContentSecurityPolicy from "./ContentSecurityPolicy";
+
+export default function RedirectOffers() {
+  window
+    .alloy("sendEvent", {
+      renderDecisions: true
+    })
+    .then(({ decisions = [] }) => {
+      console.log("personalized decisions", decisions);
+    });
```
Aaronius, line 11: This is going to send an event every time this component renders, which probably isn't really what you're after (the component can re-render many times while the user is on the view). If you want to send an event every time the user navigates to this view, you'll probably need to use the `useEffect` React hook. Something like:

```
useEffect(() => {
    window
      .alloy("sendEvent", {
        renderDecisions: true
      })
      .then(({ decisions = [] }) => {
        console.log("personalized decisions", decisions);
      });
  }, []);
  ```

File: test/unit/specs/components/Personalization/createRedirectHandler.spec.js [hunk 10]
```diff
@@ -29,21 +30,41 @@ describe("Personalization::createRedirectDecisionHandler", () => {
...
+      logger,
+      showContainers
     });
     return handleRedirectDecisions(decisions).then(() => {
-      expect(collect).toHaveBeenCalledWith({ meta }, documentMayUnload);
+      expect(collect).toHaveBeenCalledWith({ meta, documentMayUnload });
       expect(replace).toHaveBeenCalledWith(decisions[0].items[0].data.content);
     });
   });
+  it("should trigger showContainers and logger at redirect", () => {
```
Aaronius, line 55: This description should probably mention something about an error being thrown when attempting to redirect.

src/components/Personalization/createRedirectHandler.js
Aaronius: Ignore my last sentence. It might still valuable to catch. We can leave it. Is it a problem that if we try to redirect and it failes, that we show container but won't have stored any of the views in the cache?

Aaronius (approved):
I suggested some tweaks. You probably want to fix the redirect offer component not to send an event on every render.

jfkhoury (approved):
Awesome work @ninaceban !! ✌️  I will follow up with the testing group.

Aaronius:
Let's wait until Eric and others finish testing this feature before merging this PR.

jfkhoury:
@ninaceban Latest Update: We tested this feature and it seems like there's a disconnect between Target & Analytics. We will pull this feature from the release and start design meetings to finish it.

PLEASE DON'T MERGE.

sandbox/src/App.js
```diff
@@ -59,6 +61,9 @@ function BasicExample() {
           <li>
             <a href="/legacy.html">Legacy Visitor ID</a>
           </li>
+          <li>
+            <a href="/redirectOffers">Redirect Offers</a>
+          </li>
```
Aaronius, line 66: You'll probably need to update the menu found within legacy.html as well. Keeping the menu in sync in these two files is a chore, but hopefully will be improved in https://jira.corp.adobe.com/browse/PDCL-4822.

test/unit/specs/components/Personalization/createRedirectHandler.spec.js
```diff
@@ -0,0 +1,70 @@
...
+    replace.and.throwError("Malformed url");
+
+    const handleRedirectDecisions = createRedirectHandler({
+      collect,
+      window,
+      logger,
+      showContainers
+    });
+    return handleRedirectDecisions(decisions).then(() => {
+      expect(collect).toHaveBeenCalled();
```
Aaronius, line 65: It might be good to assert that `collect` was called with `{ meta, documentMayUnload }`.

Aaronius (approved):
Nice work @ninaceban!

As a side note, there might be a valid reason for having force-pushed your commits, but I would recommend trying to avoid force pushing unless you're reasonably confident that nobody has looked at your PR yet. Force pushing makes it harder for reviewers to know what has changed since the last time they reviewed, among other things. If you're not sure how to accomplish whatever you were trying to do without force-pushing, let me know and we can talk through it.

Thanks Nina!

src/components/Personalization/decisionsExtractor.js
```diff
@@ -32,22 +32,22 @@ const createDecision = (decision, items) => {
 };
```
jonsnyder, line 33: I'd love to do a refactor here. Not important for this PR, but I think it would simplify this code and make it easier to understand. If you think this will work, I'll create a JIRA ticket.

```
const createDecision = (decision, items) => {
  return { id: decision.id, scope: decision.scope, items };
 };

const findMatchingIndex = (decision, item, predicates) => {
  return predicates.reduce((matchingIndex, predicate, currentIndex) => {
    // Check if we already matched one of the predicates
    if (matchingIndex !== predicates.length) {
      return matchingIndex;
    }
    // Check if this predicate matches
    if (predicate({ decision, item })) {
      return currentIndex;
    }
    // This predicate doesn't match so return index for unmatched items
    return predicates.length;
  }, predicates.length);
}

// returns an array of decisions one greater than the number of predicates.
// the predicates are functions that return true or false, and are given {decision, item}
// index 0 holds the decisions that match the first predicate
// index 1 holds the decisions that match the second predicate
// and so forth
// the last index holds the decisions that didn't match any predicate
const extractDecisions = (decisions, ...predicates) => {
  const matchedDecisions = new Array(predicates.length + 1).map(() => []);
  decisions.forEach(decision => {
    const matchedItems = new Array(predicates.length + 1).map(() => []);
    // note: if a decision doesn't have items, it will be dropped.
    const { items = [] } = decision;
    items.forEach(item => {
      matchedItems[findMatchingIndex(decision, item, predicates)].push(item);
    });
    matchedItems.forEach((matchedItemArray, i) => {
      if (isNonEmptyArray(matchedItemArray)) {
        matchedDecisions[i].push(createDecision(decision, matchedItemArray));
      }
    });
  });
  return matchedDecisions;
};

// inside createOnResponseHandler
const [redirectItems, pageWideDomActions, nonPageWideDomActions, otherDecisions] =
  extractDecisions(
    unprocessedDecisions,
    ({ item }) => item.schema === REDIRECT_ITEM,
    ({ item, decision }) => item.schema === DOM_ACTION && decision.scope === PAGE_WIDE_SCOPE,
    ({ item, decision }) => item.schema === DOM_ACTION && decision.scope !== PAGE_WIDE_SCOPE
  );

```

jonsnyder (changes requested):
We need to use the proposition XDM mixin to record redirects.

test/unit/specs/components/Personalization/createRedirectHandler.spec.js
ninaceban: I did that in the test above, I just wanted to assert here that when the redirect itself will throw an error we would show containers and log a warning.

sandbox/src/App.js
ninaceban: thank you

src/components/Personalization/decisionsExtractor.js
ninaceban: yes, I would love to improve that part

src/components/Personalization/createRedirectHandler.js
```diff
@@ -0,0 +1,35 @@
...
+
+  return { content, decisions: [{ id, scope }] };
+};
+
+export default ({ collect, window, logger, showContainers }) => {
+  return redirectDecisions => {
+    const { content, decisions } = getRedirectDetails(redirectDecisions);
+    const documentMayUnload = true;
+
+    return collect({ meta: { decisions }, documentMayUnload })
```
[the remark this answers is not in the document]
ninaceban: thank you

src/components/Personalization/decisionsExtractor.js
jonsnyder: https://jira.corp.adobe.com/browse/PDCL-5983

jonsnyder (approved):

PR #657, opened 2021-02-11
PDCL-1094 Implement support for Konductor's new error API.

Aaronius:
## Description
Konductor has modified the way they report warnings and errors. They have asked us to support this new strategy, which required changes to the Alloy code. The changes are outlined in [this JIRA ticket comment](https://jira.corp.adobe.com/browse/PDCL-1094?focusedCommentId=23462616&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-23462616). The other comments in that issue are also useful for additional context and answering other questions you may have. The documentation for Konductor's new warning and error strategy can be found [here](https://git.corp.adobe.com/pages/experience-edge/konductor/#/apis/errors?id=handling-4xx-and-5xx-responses).

Probably the most notable change that should be called out is that non-fatal errors (as described in Konductor's error documentation) will no longer cause promises returned by Alloy to be rejected. Instead, errors have been split into fatal and non-fatal categories (again, see the docs). Fatal errors will cause the promise to be rejected, while non-fatal errors will only be logged to the console.

Note that I have no way of forcing Konductor to produce many of these error states, so other than unit tests, I just manually tested them by mocking Konductor responses using Charles. I brought this up in Slack to see if we want to try to mock Konductor responses in our functional tests. @raresvla, if there's a way we can force non-fatal errors, warnings, status codes, retry-after headers, etc., please let us know. And if there isn't away, maybe you could consider adding a way via a querystring parameter or something like that? I know [mobile is looking for something similar](https://adobedx.slack.com/archives/CLC5WRT7Z/p1609883540312000).

I'd be happy to answer any questions you may have as you go through this PR.

@raresvla, please review.

## Related Issue
https://jira.corp.adobe.com/browse/PDCL-1094

## Motivation and Context
Support Konductor's new warning and error strategy, which should also be better for our users.

## Types of changes

**In some ways, all of the above.**

## Checklist:


.eslintrc.js
```diff
@@ -14,7 +14,8 @@ module.exports = {
     // We enable the rule in src/.eslintrc.js since that's the only place we
     // want to disallow importing extraneous dependencies.
     "import/no-extraneous-dependencies": "off",
-    "prefer-destructuring": "off"
+    "prefer-destructuring": "off",
+    "import/prefer-default-export": "off"
```
Aaronius, line 18: Whenever we start a constants file where we know there may eventually be multiple constant values (like `httpHeaderNames.js`), we'd run into this eslint error and end up ignore it. I decided to disable the rule and expect us to just be judicious about when we use a default export vs named exports. If you disagree with this change, please let me know.

src/core/createLogger.js
```diff
@@ -67,7 +67,7 @@ export default ({ getDebugEnabled, console, getMonitors, context }) => {
         data.parsedBody || data.body ? `response body:` : `no response body.`;
       log(
         "info",
-        `Request ${data.requestId}: Received response with status code ${data.status} and ${messagesSuffix}`,
+        `Request ${data.requestId}: Received response with status code ${data.statusCode} and ${messagesSuffix}`,
```
Aaronius, line 70: You'll see `status` renamed to `statusCode` in a few places. That's because the object we were returning from `sendBeaconRequest`, `sendFetchRequest`, and `sendXhrRequest` previously had a `status` field on it, while the object returned from `sendNetworkRequest` had a `statusCode` field on it. I just wanted to make the names consistent.

src/core/edgeNetwork/injectProcessWarningsAndErrors.js
```diff
@@ -10,33 +10,21 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
-    // Regardless of whether the response had a successful status code,
-    // the response payload may have warnings and errors we still
-    // want to process.
-
-    const warnings = response.getWarnings();
-    const errors = response.getErrors();
-
-    warnings.forEach(warning => {
-      logger.warn(
-        `Warning received from server: [Code ${warning.code}] ${warning.message}`
```
Aaronius, line 31: This is where we are depending on the old format (warnings and errors had `code` and `message` properties). With the new format, there's a lot of good information in the warning and error objects that Konductor returns so I decided to log the objects themselves, rather than try to morph or condense them into a string of some sort.

src/core/edgeNetwork/injectSendEdgeNetworkRequest.js
```diff
@@ -67,28 +67,28 @@ export default ({
         return networkResponse;
```
Aaronius, line 67: The line right above this one is a call to `validateNetworkResponseIsWellFormed`. That's where make sure the response status code is a 2xxx and that the content looks like something we would expect. I question whether `validateNetworkResponseIsWellFormed` is still the right name, but I struggled coming up with a more appropriate name. I'm all 👂👂  if you feel like there's a better name. `validateNetworkResponseIsSuccessful` perhaps?

src/core/network/getRequestRetryDelay.js
```diff
@@ -0,0 +1,80 @@
...
+
+  if (headerValue) {
+    if (isIntegerString(headerValue)) {
+      const headerValueInt = parseInt(headerValue, 10);
+      delayInMillis = headerValueInt * 1000;
+    } else {
+      delayInMillis = Math.max(
+        0,
+        new Date(headerValue).getTime() - new Date().getTime()
+      );
```
Aaronius, line 54: @raresvla, if you can guarantee that the `Retry-After` header will always be a number and not a data, I could cut some of this out, but it might be best to just keep the support in case things change.

src/core/network/getRequestRetryDelay.js
```diff
@@ -0,0 +1,80 @@
...
+
+import { RETRY_AFTER } from "../../constants/httpHeaderNames";
+import { isIntegerString } from "../../utils";
+
+// The retry gets incrementally (but not exponentially) longer for each retry.
+const FIRST_DELAY_MILLIS = 1000;
+const INCREMENTAL_DELAY_MILLIS = 1000;
+
+// When the target delay is randomized, make it within the range of this percentage above or below the target delay.
+const MAX_RANDOM_VARIANCE_PERCENTAGE = 0.3;
```
Aaronius, line 21: @raresvla, you might want to provide input on these three constants.

src/core/network/injectSendNetworkRequest.js
```diff
@@ -43,11 +44,21 @@ export default ({
...
+
+        if (requestIsRetryable) {
+          const requestRetryDelay = getRequestRetryDelay({
+            response,
+            retriesAttempted
+          });
+          return new Promise(resolve => {
+            setTimeout(() => {
+              resolve(executeRequest(retriesAttempted + 1));
+            }, requestRetryDelay);
```
Aaronius, line 60: Previously, we were retrying the request right away. Now we have an incremental, moderately randomized backoff.

test/functional/specs/Command Logic/C13819.js
```diff
@@ -0,0 +1,52 @@
...
+test("Test C13819: Sending invalid config ID rejects command promise with useful error", async t => {
+  await configureAlloyInstance("alloy", config);
+  const errorMessage = await sendEvent();
+  await t
+    .expect(errorMessage)
+    .contains(
+      "Unexpected server response with status code 400 and response body"
+    );
+  await t.expect(errorMessage).contains("EXEG-0003-400");
+});
```
Aaronius, line 52: I figured I'd try to cover at least this fatal error case. We have some other functional test that cover other fatal errors as well like problems setting consent.

src/core/edgeNetwork/validateNetworkResponseIsWellFormed.js
```diff
@@ -19,6 +19,8 @@ import { NO_CONTENT } from "../../constants/httpStatusCode";
 export default networkResponse => {
   const { statusCode, body, parsedBody } = networkResponse;
   if (
+    statusCode < 200 ||
+    statusCode >= 300 ||
```
jonsnyder, line 23: When there is a 400 or 500 status code, could there be a json body? If so we may want to throw an error that has those fields as properties. Now that Konductor's error responses are structured, Alloy's responses could be structured as well.

jonsnyder (commented):
We need to think through these changes in regards to the onRequestFailure callbacks in the individual components. Particularly, I'm thinking about the personalization component. When the call to fetch decisions fails, it should show the containers. This would be a non-fatal error though that isn't triggering the onRequestFailure callbacks.

src/core/edgeNetwork/validateNetworkResponseIsWellFormed.js
Aaronius: For others reading this, I talked to @jonsnyder over BlueJeans about this. If Konductor was able to respond, then it will have a JSON body, but if it's JAG that's responding, it won't be a JSON body (it'll just be a string of text). While we could add a property on the error object that contains more details, we decided that including the response body in the error message like we're doing now is sufficient until we understand more about use cases where customers might want to programmatically react to certain errors coming back from the server.

src/core/edgeNetwork/injectSendEdgeNetworkRequest.js
```diff
@@ -67,28 +67,28 @@ export default ({
...
         const response = createResponse(parsedBody);
         cookieTransfer.responseToCookies(response);
 
-        if (statusCode >= 400) {
-          const throwError = () => processWarningsAndErrors(response);
-          return onRequestFailureCallbackAggregator
-            .call({ response })
-            .then(throwError, throwError);
-        }
+        processWarningsAndErrors(response);
```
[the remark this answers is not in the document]
Aaronius: I talked to @jonsnyder over BlueJeans about this. We agreed that `processWarningsAndErrors` and `validateNetworkResponseIsWellFormed` would probably be best combined. I'll add a commit for that.

Aaronius:
> We need to think through these changes in regards to the onRequestFailure callbacks in the individual components. Particularly, I'm thinking about the personalization component. When the call to fetch decisions fails, it should show the containers. This would be a non-fatal error though that isn't triggering the onRequestFailure callbacks.

I talked to @jonsnyder over BlueJeans about this and talked to the rest of the team about it in standup yesterday. I believe we're covered here. Containers will be shown if there is a non-fatal error, since there will be no decisions payload, and personalization [shows containers in this case](https://github.com/adobe/alloy/blob/2bee677f4b7296c936d64aeee2a419c786cd23f9/src/components/Personalization/createOnResponseHandler.js#L25-L31).

I looked over the other components as well and don't see any problems. I did [log an issue](https://jira.corp.adobe.com/browse/PDCL-4473) for documenting when `result.decisions` will or will not exist (as a result of a `sendEvent` command) and what its value may be and why. This should help our customers be able to write code that can handle various scenarios correctly.

src/core/edgeNetwork/injectSendEdgeNetworkRequest.js
```diff
@@ -94,17 +90,6 @@ export default ({
...
-            // As long as we received a properly structured response,
-            // we consider the response sucessful enough to call lifecycle
-            // onResponse methods. However, a structured response from the
-            // server may ALSO containing errors. Because of this, we make
-            // sure we call lifecycle onResponse methods, then later
-            // process the warnings and errors.
-            // If there are errors in the response body, an error will
-            // be thrown here which should ultimately reject the promise that
-            // was returned to the customer for the command they executed.
-            processWarningsAndErrors(response);
```
Aaronius, line 107: In case you're looking at this diff and noticing that we were calling `processWarningsAndErrors` twice. Yeah, I should have removed this line with my prior commit. Leaving it there was causing warnings and errors to be logged twice. I didn't catch it previously because I thought they were being logged for separate events during my manual testing. Anyway, what was `validateNetworkResponseIsWellFormed` has now been merged with `processWarningsAndErrors` and it's only being called once up higher.

src/core/edgeNetwork/injectProcessWarningsAndErrors.js
```diff
@@ -10,21 +10,37 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-const MESSAGE_PREFIX = "The server responded with the following";
+import { NO_CONTENT } from "../../constants/httpStatusCode";
+
+const MESSAGE_PREFIX = "The server responded with a";
```
Aaronius, line 15: I decided to use this message prefix for fatal errors, non-fatal errors, and warnings. This resulted in some of the text changes you see in tests.

Aaronius:
@jonsnyder I've consolidated validateNetworkResponseIsWellFormed and processWarningsAndErrors as we discussed.

Something I was thinking about: We no longer use the `getWarnings` and `getErrors` methods on the response object created by the `createResponse` module EXCEPT from inside some of our functional tests. You can see how we're using `createResponse` from `src` within our functional tests [here](https://github.com/adobe/alloy/blob/2bee677f4b7296c936d64aeee2a419c786cd23f9/test/functional/specs/Privacy/IAB/C224678.js#L66). It looks like `createResponse` is the _only_ module from `src` that we import from functional tests. We have a few options:

1. Leave things as they are, which means `getWarnings` and `getErrors` won't _currently_ be used by anything in `src`, but they will be used by functional tests.
2. Remove the `getWarnings` and `getErrors` from the response object created by `createResponse`, and functional tests will just need to look directly at the parsed payload for warnings and errors. It's a little weird though, because we'll still be using `createResponse` in functional tests to get access to the `response.getPayloadsByType` method.
3. Remove the `getWarnings` and `getErrors` from the response object created by `createResponse`. Stop using `src`'s `createResponse` from inside functional tests. Instead, we could make a `createResponse` module that's just for use within functional tests, which would expose `getPayloadsByType`, `getWarnings`, and `getErrors` methods.

I prefer the third option. Thoughts?

src/core/network/getRequestRetryDelay.js
```diff
@@ -0,0 +1,80 @@
...
+
+  // Note that the value of delay may be 0 at this point, which would be a valid delay we want to use
+  // and not override, which is why we don't do:
+  // if (!delay) { ... }
+  if (delayInMillis === undefined) {
+    delayInMillis = calculateRetryDelay(retriesAttempted);
+  }
+
+  return delayInMillis;
+};
```
jfkhoury, line 80: Thank you for the great comments!

src/core/network/injectSendNetworkRequest.js
jfkhoury: Nice!

jfkhoury (approved):
Awesome work, comments & attention to details.

Do we need to update the public docs?

Thanks.

src/core/network/getRequestRetryDelay.js
raresvla: 👍 The values look good to me as they are set right now.

src/core/network/injectSendNetworkRequest.js
raresvla: Just want to double check this: when retrying, is the same `requestId` being used or a new one would be generated?

test/unit/specs/core/edgeNetwork/injectProcessWarningsAndErrors.spec.js
```diff
@@ -13,63 +13,113 @@ governing permissions and limitations under the License.
...
-        code: "general:100",
-        message: "General error occurred."
+        title: "General warning",
+        detail: "General warning detail"
       },
       {
-        code: "personalization:204",
-        message: "Personalization error occurred."
+        title: "Personalization warning",
+        message: "Personalization warning detail"
```
raresvla, line 104: We now use `title` and `detail`, instead of `message` -- is this a leftover?

test/unit/specs/core/edgeNetwork/injectProcessWarningsAndErrors.spec.js
```diff
@@ -13,63 +13,113 @@ governing permissions and limitations under the License.
...
-        code: "general:100",
-        message: "General warning."
+        title: "General warning",
+        detail: "General warning detail"
       },
       {
-        code: "personalization:204",
-        message: "Personalization warning."
+        title: "Personalization warning",
+        message: "Personalization warning detail"
```
raresvla, line 74: We now use `title` and `detail`, instead of `message` -- is this a leftover?

raresvla (approved):
💯  With a few small comments, looks good to me!

src/core/edgeNetwork/injectProcessWarningsAndErrors.js
Aaronius: Fatal errors are handled higher up in this same file where we detect if the status is outside the 2xx range: https://github.com/adobe/alloy/blob/e975b9d4415b32469e78e6f48e38af0c4f521461/src/core/edgeNetwork/injectProcessWarningsAndErrors.js#L22-L23

src/core/network/injectSendNetworkRequest.js
Aaronius: The request ID will be re-used. Would you expect a new ID to be generated?

test/unit/specs/core/edgeNetwork/injectProcessWarningsAndErrors.spec.js
Aaronius: Technically, our source code no longer cares what the contents of the errors or warnings are. We just log the full object. In other words, it doesn't really matter what the warning and error objects look like in this spec. I'll update the tests though to either match the new warning and error object structure or perhaps deviate from it further to accentuate the point that the object structure doesn't matter.

Aaronius: Oh, I see what you're pointing out now. I renamed one of the object's `title` property to `detail`, but not the `title` property on the other object. While the object structure doesn't actually matter, I do want these to be consistent and having them not consistent was an oversight. I'll update that. Thanks.

Aaronius:
Alright, I'm merging this.  @raresvla, is there a way we can force non-fatal errors, warnings, status codes, retry-after headers, etc.? The more we can test without mocking Konductor responses, the better.

Aaronius:
Thank you all for your feedback.

PR #660, opened 2021-02-22
PDCL-4522 - The onBeforeEventSend callback should be user cancellable

garyr:
## Description

The user configurable callback known as `onBeforeEventSend()`, aka the "last chance" callback needs to be amended so that returning false or an error condition will cancel the event. This change provides the following features:

returning false in `onBeforeEventSend()`  will cause the event to not be sent
if an error is thrown inside of `onBeforeEventSend()` the event will not be sent

Note that if a user is currently returning false in the callback, when they upgrade those events may no longer be sent

## Types of changes

## Checklist:


src/core/createEvent.js
```diff
@@ -10,28 +10,88 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
-  let lastChanceCallback;
+  let isFinalized = false;
+  let shouldSendEvent = true;
 
   const event = {
+    getUserXdm() {
+      return userXdm;
+    },
     setUserXdm(value) {
+      if (isFinalized) {
```
jfkhoury, line 28: I wonder if we should extract those 3 lines into a quick helper, like `throwIfEventFinalized`..

src/core/createEvent.js
```diff
@@ -10,28 +10,88 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+      if (onBeforeEventSend) {
+        const xdm = content.xdm || {};
+        const data = content.data || {};
+
+        // this allows the user to replace the object passed into the callback
+        const tempContent = { xdm, data };
+        const result = onBeforeEventSend(tempContent);
+        if (result === false) shouldSendEvent = false;
+
+        if (Object.keys(tempContent).length) {
```
jfkhoury, line 84: Are we missing `tempContent.xdm` ?

src/core/createEvent.js
```diff
@@ -53,39 +113,7 @@ export default () => {
       return userXdm.web.webPageDetails.viewName;
     },
     toJSON() {
-      if (userXdm) {
```
jfkhoury, line 56: Love this change!!!

src/core/createEventManager.js
```diff
@@ -69,6 +58,17 @@ export default ({
...
         .then(() => {
+          // NOTE: this calls onBeforeEventSend callback (if configured)
+          event.finalize(onBeforeEventSend);
+
+          // if the callback returns false, the event should not be sent
+          return event.shouldSend();
+        })
+        .then(shouldSendEvent => {
+          if (shouldSendEvent === false) {
+            return Promise.resolve();
```
jfkhoury, line 69: Do we need any sort of cleanup at this point? I can't think of anything, just thought I'd ask :)

jfkhoury (commented):
One comment about `tempContent.xdm`, otherwise awesome work @garyr !! 

Do you think there's anything in this PR that should be documented in our tech WIKI?

test/unit/specs/core/createEventManager.spec.js
```diff
@@ -108,31 +108,51 @@ describe("createEventManager", () => {
...
+      });
+      return eventManager.sendEvent(eventToIgnore).then(() => {
+        expect(sendEdgeNetworkRequest).not.toHaveBeenCalled();
+      });
+    });
 
-      let wrappedLastChanceCallback;
-      event.setLastChanceCallback.and.callFake(callback => {
-        wrappedLastChanceCallback = callback;
+    it("returns null onBeforeEventSend and event should be sent", () => {
```
Aaronius, line 130: I *think* what you're saying is that if `onBeforeEventSend` returns null, the event should still be sent. I don't think that's what's really being tested here though. This test is currently testing that if `shouldSend` returns null, that the event will still be sent, but I think this is an invalid test case because `shouldSend` always returns a boolean, as far as I can tell.

I believe what we need to do is remove this test case  and instead make sure that `createEvent.spec.js` has a test that asserts that when `onBeforeEventSend` returns `null`, that the event then returns `true` from `shouldSend`.

test/unit/specs/core/createEventManager.spec.js
```diff
@@ -108,31 +108,51 @@ describe("createEventManager", () => {
...
-        wrappedLastChanceCallback();
+      return eventManager.sendEvent(event).then(() => {
         expect(config.onBeforeEventSend).toHaveBeenCalled();
       });
     });
 
-    it("logs errors in the config.onBeforeEventSend callback", () => {
-      const error = new Error("onBeforeEventSend error");
-      config.onBeforeEventSend.and.throwError(error);
+    it("returns false onBeforeEventSend and event should not be sent", () => {
```
Aaronius, line 120: From a unit perspective, this isn't really testing what happens when `onBeforeEventSend` returns `false`. I think a more appropriate description would be: `does not send event when event.shouldSend returns false`. The test for whether `event.shouldSend` returns `false` when `onBeforeEventSend` returns `false` belongs in createEvent.spec.js.

test/unit/specs/core/createEventManager.spec.js
```diff
@@ -108,31 +108,51 @@ describe("createEventManager", () => {
         });
     });
 
-    it("sets the lastChanceCallback, which wraps config.onBeforeEventSend, on the event", () => {
-      let wrappedLastChanceCallback;
-      event.setLastChanceCallback.and.callFake(callback => {
-        wrappedLastChanceCallback = callback;
+    it("events call onBeforeEventSend callback", () => {
```
Aaronius, line 111: I think this test is mainly just testing your test, since your event mock is the one actually calling `onBeforeEventSend` and then you're asserting that `onBeforeEventSend` was called. Instead, from a unit perspective, I think here you just want test that `event.sendEvent` calls `event.finalize` with `onBeforeEventSend`. The test for whether `onBeforeEventSend` actually gets called by `event.finalize` belongs in `createEvent.spec.js` instead. Let me know if this doesn't make sense.

test/unit/specs/core/createEventManager.spec.js
```diff
@@ -108,31 +108,51 @@ describe("createEventManager", () => {
...
+      const eventToIgnore = jasmine.createSpyObj("event", {
+        finalize: () => {},
+        shouldSend: null
+      });
+      return eventManager.sendEvent(eventToIgnore).then(() => {
+        expect(sendEdgeNetworkRequest).toHaveBeenCalled();
       });
+    });
+
+    it("returns true onBeforeEventSend and event should be sent", () => {
```
Aaronius, line 140: From a unit perspective, this isn't really testing what happens when `onBeforeEventSend` returns `true`. I think a more appropriate description would be: `sends when event.shouldSend returns true`. The test for whether `event.shouldSend` returns `true` when `onBeforeEventSend` returns `true` belongs in createEvent.spec.js.

test/unit/specs/core/createEventManager.spec.js
```diff
@@ -108,31 +108,51 @@ describe("createEventManager", () => {
...
+      const errorMsg = "Expected Error";
+      event.finalize.and.throwError(errorMsg);
 
-      return eventManager.sendEvent(event, {}).then(() => {
-        expect(() => {
-          wrappedLastChanceCallback();
-        }).toThrowError("onBeforeEventSend error");
-        expect(logger.error).toHaveBeenCalledWith(error);
+      return eventManager.sendEvent(event).catch(() => {
+        expect(sendEdgeNetworkRequest).not.toHaveBeenCalled();
```
Aaronius, line 155: You'll need to make sure that the promise returned to Jasmine is rejected if the promise from `eventManager.sendEvent(event)` is resolved. Here's why: Jasmine sees you returned a promise and will wait until the promise is resolved or rejected before moving to the next test. If the promise is resolved and all assertions pass, then it considers the test as passed. If the promise is rejected or any assertion fails, then it considers the test as failed. The problem here is that if the promise coming from `eventManager.sendEvent(event)` is resolved rather than rejected, then your `.catch` will never be called and the promise jasmine is observing will be resolved. Therefore, the test would be considered as passed, when it should be considered as failed. I would recommend something like this:

```js
return eventManager.sendEvent(event).then(
  () =>
    throw new Error("Should not have resolved.")
  }, 
  () => {
    expect(sendEdgeNetworkRequest).not.toHaveBeenCalled();
  }
);

File: src/core/createEvent.js (hunk 2 above)
Comment (garyr) [reply src/core/createEvent.js:84]: Good catch, yes

**Review by Aaronius — CHANGES_REQUESTED** [review]
Nice work Gary! I requested some changes. Let me know if you want to talk through them.

File: src/core/createEventManager.js (hunk 4 above)
Comment (garyr) [reply src/core/createEventManager.js:69]: @Aaronius - can you spot anything else we need to do here?

File: test/unit/specs/core/createEventManager.spec.js (hunk 9 above)
Comment (Aaronius) [reply test/unit/specs/core/createEventManager.spec.js:155]: I think another options is using `expectAsync`. It works something like this:

await expectAsync(boom()).toBeRejectedWith(new TypeError("BOOM"))

and then you could do `expect(sendEdgeNetworkRequest).not.toHaveBeenCalled();` afterward.

File: src/core/createEvent.js (hunk 1 above)
Comment (garyr) [reply src/core/createEvent.js]: Done

File: test/unit/specs/components/ActivityCollector/attachClickActivityCollector.spec.js [hunk 10]
```diff
@@ -110,12 +105,12 @@ describe("ActivityCollector::attachClickActivityCollector", () => {
...
-      return {
+    eventManager = jasmine.createSpyObj("eventManager", {
+      createEvent: {
         isEmpty: () => false
-      };
-    };
-    spyOn(eventManager, "sendEvent").and.returnValue(Promise.reject(error));
+      },
+      sendEvent: Promise.reject(error)
+    });
```
Aaronius, line 113: Rather than recreating the full mock, you should be able to do:

```
eventManager.createEvent.and.returnValue({
  isEmpty: () => false
});
```

The same can be applied to the previous test.

test/unit/specs/core/createEventManager.spec.js
```diff
@@ -108,32 +111,43 @@ describe("createEventManager", () => {
...
       });
+    });
 
-      return eventManager.sendEvent(event, {}).then(() => {
-        expect(() => {
-          wrappedLastChanceCallback();
-        }).toThrowError("onBeforeEventSend error");
-        expect(logger.error).toHaveBeenCalledWith(error);
+    it("does send event when event.shouldSend returns true", () => {
+      const eventToIgnore = jasmine.createSpyObj("event", {
```
Aaronius, line 131: Does the name `eventToIgnore` here apply since we're not ignoring it?

test/unit/specs/core/createEventManager.spec.js
```diff
@@ -39,7 +39,10 @@ describe("createEventManager", () => {
     consent = jasmine.createSpyObj("consent", {
       awaitConsent: Promise.resolve()
     });
-    event = jasmine.createSpyObj("event", ["setLastChanceCallback"]);
+    event = jasmine.createSpyObj("event", {
+      finalize: undefined,
+      shouldSend: true
+    });
```
Aaronius, line 45: If you want, you can do `finalize() {}` here and then in your tests below you could use this event rather than creating a new `eventToIgnore` event. To tweak the return value of this event's `shouldSend` method for a specific test, you would do `event.shouldSend.and.returnValue(false)`. Your choice though.

Aaronius (approved):
Nice job @garyr. I can tell you're getting more comfortable with the codebase.

src/core/createEvent.js
```diff
@@ -10,28 +10,92 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+  const throwIfEventFinalized = methodName => {
+    if (isFinalized) {
+      throw new Error(
+        `${methodName} cannot be called after event is finalized`
+      );
+    }
+  };
 
   const event = {
+    getUserXdm() {
```
jonsnyder, line 32: It doesn't look like this is used anywhere.

src/core/createEvent.js
```diff
@@ -53,39 +117,7 @@ export default () => {
...
-          }
-          if (content.data || !isEmptyObject(args.data)) {
-            content.data = args.data;
-          }
-        } catch (e) {
-          // the callback should have already logged the exception
-        }
-      }
-
+      if (!isFinalized) this.finalize();
```
jonsnyder, line 120: If you want to guard against the event being used improperly, you could throw an error here instead. Silently running finalize without the onBeforeEventSend callback could be bad.

if (!isFinalized) {
  throw new Error("Attempt to call toJSON when the event has not yet been finalized.");
}

src/core/createEventManager.js
```diff
@@ -69,6 +58,14 @@ export default ({
...
+          event.finalize(onBeforeEventSend);
+
+          // if the callback returns false, the event should not be sent
+          if (!event.shouldSend()) {
+            return undefined;
+          }
+
           return sendEdgeNetworkRequest({
             request,
             runOnResponseCallbacks: onResponseCallbackAggregator.call,
```
jonsnyder, line 71: If there is an exception thrown in the onBeforeEventSend callback, it should call the onRequestFailure component callbacks. This is especially important for personalization. If the onBeforeEventSend callback throws an error, personalization needs to un-hide the containers.

jonsnyder (changes requested):

src/core/createEventManager.js
jonsnyder: Additionally, if onBeforeEventSend callback returns false, the onRequestFailure component callbacks should be called.

src/core/createEvent.js
garyr: Nice catch, this was an artifact of working on the other `onLinkClick()` callback that we decided not to pursue

src/core/createEventManager.js
Aaronius: Ooooooh. This is very interesting and a good point. Let's talk about this one in today's check-in.

src/core/createEventManager.js
```diff
@@ -58,12 +58,28 @@ export default ({
           return consent.awaitConsent();
         })
         .then(() => {
-          // NOTE: this calls onBeforeEventSend callback (if configured)
-          event.finalize(onBeforeEventSend);
+          let shouldSend;
+          try {
+            // NOTE: this calls onBeforeEventSend callback (if configured)
+            event.finalize(onBeforeEventSend);
+            shouldSend = event.shouldSend();
```
Aaronius, line 65: This line doesn't need to be part of the of `try...catch`, so let's pull it out and then down where you have `if (!shouldSend) {` we can do `if (!event.shouldSend()) {`

test/unit/specs/core/createEventManager.spec.js
```diff
@@ -118,21 +119,21 @@ describe("createEventManager", () => {
...
     });
 
     it("does send event when event.shouldSend returns true", () => {
-      const eventToIgnore = jasmine.createSpyObj("event", {
+      event = jasmine.createSpyObj("event", {
         finalize: () => {},
         shouldSend: true
       });
-      return eventManager.sendEvent(eventToIgnore).then(() => {
+      return eventManager.sendEvent(event).then(() => {
```
Aaronius, line 136: You'll need to add tests/assertions regarding what happens when `event.finalize` throws an error and when `event.shouldSend` returns `false`. For example, you'll need to check that the various failure callbacks get called in both cases. In the case that `event.finalize` throws an error, we'll need to make sure that the promise returned from `eventManager.sendEvent` gets rejected. In the case that `event.shouldSend` returns `false`, we'll need to make sure the promise returned from `eventManager.sendEvent` gets resolved with `undefined`.

We also need functional tests for these things. Set up a scenario where you configure Alloy with `onBeforeEventSend` throwing an error and make sure no request goes out when you attempt to send an event and that the promise you receive as a customer gets rejected. Set up another scenario where you configure Alloy with `onBeforeEventSend` returning `false` and make sure no request goes out when you attempt to send an event and that the promise you receive as a customer gets resolved with an object (even though you resolve the promise being returned from eventManager with `undefined`, there's code higher up in Alloy that ensures that all promises returned from commands are resolved with an object, even if it's an empty object).

Aaronius (commented):
This looks great so far @garyr. Thanks! Now we need more tests!

test/unit/specs/core/createEvent.spec.js
```diff
@@ -268,7 +268,7 @@ describe("createEvent", () => {
       const subject = createEvent();
       subject.setUserXdm({ c: "3" });
       subject.setUserData({ d: "4" });
-      subject.setLastChanceCallback(callback);
+      subject.finalize(callback);
```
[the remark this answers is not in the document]
garyr: I added tests for these cases

jonsnyder (approved):

test/functional/specs/Data Collector/C1715149.js
```diff
@@ -0,0 +1,102 @@
...
+  return false;
+});
+
+const getOnBeforeEventSendCalled = ClientFunction(() => {
+  return window.onBeforeEventSendCalled === true;
+});
+
+const getSendEventPromiseResult = ClientFunction(() => {
+  return window.sendEventPromiseResult;
+});
```
Aaronius, line 39: The way you're configuring Alloy and sending events is "old school" as of a week or so ago. @jonsnyder did a bunch of work to make stuff like this easier. You'll need to merge the latest from `main` into your branch. Then check out how it's done in https://github.com/adobe/alloy/blob/main/test/functional/specs/Privacy/C2593.js#L24-L40

Also see Jon's documentation for `createAlloyProxy`: https://github.com/adobe/alloy/blob/d86b274002ab1c50d379b24b933bc938952de885/test/functional/helpers/createAlloyProxy.js#L2-L47

Aaronius (changes requested):

test/functional/specs/Data Collector/C1715149.js
```diff
@@ -0,0 +1,102 @@
...
+test("C1715149 Events should call onBeforeEventSend callback and still send event", async () => {
+  await configureAlloyInstance("alloy", {
+    ...orgMainConfigMain,
+    onBeforeEventSend
+  });
+
+  await triggerAlloyEvent();
+
+  await t.expect(getOnBeforeEventSendCalled()).eql(true);
+  await t.expect(destinationLogger.requests.length > 0).eql(true);
```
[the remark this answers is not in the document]
garyr: Okay, I will remove this (I think it came along with a copy and paste). I had thought about adding xdm data and checking it. I think that's a good suggestion

src/core/createEventManager.js
```diff
@@ -69,6 +58,28 @@ export default ({
...
+          try {
+            // NOTE: this calls onBeforeEventSend callback (if configured)
+            event.finalize(onBeforeEventSend);
+          } catch (error) {
+            const throwError = () => {
+              throw error;
+            };
+            onRequestFailureCallbackAggregator.add(lifecycle.onRequestFailure);
+            return onRequestFailureCallbackAggregator
+              .call()
```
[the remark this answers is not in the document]
garyr: There is no error that comes through. I will stop passing in the error

Aaronius:
@garyr and I worked together on the latest commit, which addresses my feedback.

Aaronius (approved):

PR #664, opened 2021-02-23
PDCL-1036 Use XDM mixin for personalization notifications instead of meta block.

Aaronius:
## Description
Previously, we were sending personalization notifications under `meta`. This PR changes that so that we're using the new proposition interaction XDM mixin. The proposition interaction XDM mixin is now bundled inside the AEP Web SDK ExperienceEvent Mixin. It looks like this:

image

Aditi Shete is working on updating the Target Konductor plugin to work with this new mixin and providing backward compatibility support for existing versions of the web SDK that are still sending the information using `meta`. @ninaceban, once Aditi gets the Target Konductor plugin working, I'd love your help in verifying that everything is working properly, specifically that (1) the Target Konductor plugin is working for the new Alloy (2) the Target Konductor plugin is still working for previous Alloy versions, and (3) that data properly flows into a platform dataset using the new mixin.

Some functional tests are failing, but they also seem to be failing in the `main` branch, so we'll work on fixing those tests there.

## Related Issue
https://jira.corp.adobe.com/browse/PDCL-1260

## Motivation and Context
This allows customers to report notifications for form-based offers (offers that are not automatically rendered). Offer decisioning also needs this information as XDM is a standard way of providing this data.

## Types of changes

## Checklist:


sandbox/src/Home.js
```diff
@@ -1,106 +1,42 @@
...
+      console.log(
+        "Sandbox: Get Identity command has completed.",
+        result.identity.ECID
+      );
+    } else {
+      console.log(
+        "Sandbox: Get Identity command has completed but no identity was provided in the result (possibly due to lack of consent)."
+      );
     }
-    previousPath = loc.pathname;
```
Aaronius, line 20: Sending the event within a `history.listen` didn't make a lot of sense, because it would only send an event with the user navigated and, I think, only if the user hit Home first (I forget). I changed it so that every view that wants to send an event when the view loads can leverage a `useSendPageViewEvent` module.

sandbox/public/index.html
```diff
@@ -64,10 +64,10 @@
             doesOptInApply: getUrlParameter("legacyOptIn") === "true"
           });
           // Alloy only looks for window.Visitor when it initially loads, so only load Alloy after Visitor loaded.
-          includeScript("alloy.js");
+          includeScript("/alloy.js");
```
Aaronius, line 67: We now have views that are multiple levels deep (`/personalization/products`), so we need to use an absolute path to reference alloy.js correctly in those instances.

src/components/Personalization/createCollect.js
```diff
@@ -10,12 +10,12 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-export default ({ eventManager, mergeMeta }) => {
-  return ({ meta }) => {
+export default ({ eventManager, mergeDecisionsMeta }) => {
+  // Called when a decision is auto-rendered for the __view__ scope (non-SPA view).
```
jfkhoury, line 14: Love those comments!

jfkhoury (approved):
Awesome work, and love the Sandbox changes!! ✌️ 🦅

ninaceban (approved):
great 👍

test/functional/specs/Personalization/C28755.js
```diff
@@ -34,10 +34,10 @@ test.meta({
   TEST_RUN: "Regression"
 });
 
-test.skip("Test C28755: A VEC offer for all visitors should return in every event if __view__ scope exist", async () => {
+test("Test C28755: A VEC offer for all visitors should return in every event if __view__ scope exist", async () => {
```
ninaceban, line 37: I think we should change the test case name to something like "The first sendEvent on the page should fetch Personalization VEC offers" or the test doesn't reflect the testcase name.
As we currently support SPA as well, I would validate in this testcase that the first sendEvent on the page would fetch Personalization by having the `__view__` scope in the request

ninaceban (approved):
thanks for fixing the functional tests 👍

PR #681, opened 2021-03-02
Add build to prod workflow so it will build dist/baseCode.js

jonsnyder:
The prod workflow tests use dist/baseCode.js in their tests, and are failing because it is not there. The build step builds this file.


jfkhoury (approved):
Do you think we need a build step to build this file only instead of the whole library?

Aaronius (approved):

jonsnyder:
@jfkhoury I'm working on a change to have more control about which files are built.

PR #683, opened 2021-03-02
Only build what's needed for each test. Use the published NPM package in prod tests.

jonsnyder:
## Description

* Only build what's needed for each test. Two new scripts: test:functional:build:int, and test:functional:build:prod which builds the necessary files for functional tests run with ALLOY_ENV=int and ALLOY_ENV=prod.
* Changing rollup environment parameters to be able to build minified and unminified files at the same time. This simplified some of the package.json scripts.
* Use the baseCode.min.js file for running functional tests.
* Test the published NPM package in prod tests.
* Add a "prepare" script in package.json to replace what was in "install" before. Install runs when you are installing the NPM package as a dependency. "prepare" is the best solution for now, but still runs before publish which we don't need. See https://npm.community/t/feature-request-prepare-script-that-does-not-run-before-publish/2498

## Related Issue

https://jira.corp.adobe.com/browse/PDCL-4316
https://jira.corp.adobe.com/browse/PDCL-4317

## Motivation and Context

Before we were building extra files to run tests that weren't needed.
Now that the library is published in NPM we can add a dev dependency and build the published npm package version when running prod functional tests.

## Types of changes

## Checklist:


shammowla (approved):

jfkhoury (approved):
I'd love to have Aaron look at this too but thank you so much for doing this @jonsnyder !

package.json
```diff
@@ -75,6 +77,7 @@
     "uuid": "^3.3.2"
   },
   "devDependencies": {
+    "@adobe/alloy": "^2.4.0-alpha.1",
```
Aaronius, line 80: Are we going no need to install the latest version after every npm release or something? It's not really clear to me. You might need to update the [release steps](https://github.com/adobe/alloy/wiki/Publishing-a-New-Release).

rollup.config.js
```diff
@@ -80,49 +75,79 @@ const buildPlugins = version => {
...
-    plugins: buildPlugins(NPM_PACKAGE_LOCAL)
-  });
-}
+    plugins
+  };
+};
+
+const config = [];
 
+const addConfig = version => {
```
Aaronius, line 133: `version` is a pretty overloaded term. Something like `variant` might be clearer.

Aaronius (approved):
Nice! This is a good improvement! Thanks.

package.json
```diff
@@ -18,17 +18,19 @@
...
+    "test:functional:build:prod": "rollup -c --environment BASE_CODE_MIN,NPM_PACKAGE_PROD",
     "test:functional:ci:predeploy": "ALLOY_ENV=\"int\" testcafe 'browserstack:chrome@87:Windows 10','browserstack:ie@11:Windows 10','browserstack:firefox@85:Windows 10','browserstack:safari@14.0:OS X Big Sur'",
     "test:functional:ci:dev": "EDGE_BASE_PATH=\"ee-pre-prd\" ALLOY_ENV=\"int\" testcafe 'browserstack:chrome@87:Windows 10','browserstack:ie@11:Windows 10','browserstack:firefox@85:Windows 10','browserstack:safari@14.0:OS X Big Sur'",
     "test:functional:ci:prod": "ALLOY_ENV=\"prod\" testcafe 'browserstack:chrome@87:Windows 10','browserstack:ie@11:Windows 10','browserstack:firefox@85:Windows 10','browserstack:safari@14.0:OS X Big Sur'",
     "test:functional:ci:edgepredeploy": "ALLOY_ENV=\"prod\" EDGE_BASE_PATH=\"ee-pre-prd\" testcafe 'browserstack:chrome@87:Windows 10','browserstack:ie@11:Windows 10','browserstack:firefox@84:Windows 10','browserstack:safari@14.0:OS X Big Sur'",
     "test:functional:api": "testcafe chrome:headless",
     "sandbox:build": "cd sandbox && npm run build",
```
Aaronius, line 30: I think this needs to change so that it builds the sandbox variant/version now, right?

package.json
```diff
@@ -18,17 +18,19 @@
...
     "dev": "concurrently --names build,sandbox \"rollup -c -w --environment SANDBOX\" \"cd sandbox && export REACT_APP_NONCE=321 && npm start\"",
```
Aaronius, line 31: It seems like we shouldn't have to build the standalone variant (which is done through the `build` npm script) and the sandbox variant (which is done on this line). I'm kind of losing track of things though.

Aaronius: I talked to @jonsnyder about this. I got confused with how concurrently works. My comment can be ignored.

jonsnyder: It isn't calling build. `--names build,sandbox` is just naming the two commands for logging.

package.json
jonsnyder: You are right that only full releases will need to update this. I will update the release steps.

PR #684, opened 2021-03-03
Properly handle undefined and null in merging methods.

Aaronius:
## Description
When something called `event.mergeXdm`, `event.mergeMeta`, or `event.mergeQuery` with `undefined` or `null` (this would happen in Personalization), it would wipe out the previous state. Instead, we want to make these method calls a no-op.

## Related Issue
None

## Motivation and Context
Ensure that we get the expected merge result.

## Types of changes

## Checklist:

### DIFF
--- a/src/core/createEvent.js
+++ b/src/core/createEvent.js
@@ -39,15 +39,21 @@ export default () => {
     },
     mergeXdm(xdm) {
       throwIfEventFinalized("mergeXdm");
-      deepAssign(content, { xdm });
+      if (xdm) {
+        deepAssign(content, { xdm });
+      }
     },
     mergeMeta(meta) {
       throwIfEventFinalized("mergeMeta");
-      deepAssign(content, { meta });
+      if (meta) {
+        deepAssign(content, { meta });
+      }
     },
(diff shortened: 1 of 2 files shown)


garyr:
Good catch and very simple change. How did you find this behavior? Manual testing?

garyr (approved):

jonsnyder (approved):

ninaceban (approved):
looks great, I just don't know if we will ever fall into this condition

Aaronius:
Yeah, manual testing. It may be that a functional test would have caught it, but I'm still working through updating the functional tests for activities we've set up on the new org.

As part of some cleanup, I changed the `notification-mixin` branch in a way that it's calling `event.mergeXdm` with a value of `undefined`. The old `createEvent` properly ignored our `undefined` value, but then I merged `main` (which contains the recent `createEvent` changes) into the `notification-mixin` branch, I started noticing that some personalization click collection XDM wasn't making it into the request and then I pinned it down to this. I don't know if anything in `main` is calling these merge methods with `undefined` or `null`, but it's good to handle that case regardless.

PR #689, opened 2021-03-09
Skipping tests that need to get refactored for stability.

Aaronius:
## Description
These test are unstable because of how the `sendBeacon` API works and some of our assertions. They need to get refactored. Skipping for now.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Improvement (non-breaking change which does not add functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] I have signed the [Adobe Open Source CLA](https://opensource.adobe.com/cla.html) or I'm an Adobe employee.
- [ ] I have made any necessary test changes and all tests pass.
- [ ] I have run the Sandbox successfully.

PR #690, opened 2021-03-09
Calculate prod version of Alloy

Aaronius:
## Description
This determines the production version of Alloy by looking at the latest non-draft, non-prelease github release, retrieving the package.json from the related tag, and reading the value of the `version` field. This is used when running the production github workflow where we're testing prod Alloy against prod Konductor.

## Motivation and Context
TestCafe's proxy was caching the production version of Alloy (which was in a `latest` directory on Akamai), which was causing test failures when the prod github workflow would run. While we could figure out why Akamai was caching Alloy, we decided instead to calculate the latest production version and use that in the Akamai path. This, I believe, also allows us to get rid of the `latest` directory on Akamai, since I don't think there's any other use case for it.

## Types of changes

## Checklist:


jfkhoury (approved):

jfkhoury:
Nice @Aaronius !!

test/functional/specs/Data Collector/C8118.js
```diff
@@ -49,7 +49,7 @@ const assertRequest = async request => {
   });
 };
 
-test("Test C8118: Verify link click sends a request to the collect endpoint when identity has been established, interact endpoint otherwise", async () => {
+test.skip("Test C8118: Verify link click sends a request to the collect endpoint when identity has been established, interact endpoint otherwise", async () => {
```
Aaronius, line 52: I didn't mean to include this in this commit, since it's in https://github.com/adobe/alloy/pull/689. I'll just make sure I merge https://github.com/adobe/alloy/pull/689 first.

test/functional/helpers/createFixture/clientScripts.js
```diff
@@ -16,9 +16,14 @@ import readCache from "read-cache";
 import { ClientFunction } from "testcafe";
 
 const alloyEnv = process.env.ALLOY_ENV || "int";
-
+const alloyProdVersion = process.env.ALLOY_PROD_VERSION;
```
Aaronius, line 19: While technically I could have retrieved the prod version from code somewhere inside here rather retrieving it in a separate script and passing it through as an environment variable, it would change `getFixtureClientScripts` from being synchronous to asynchronous, which means we would have to `await` calls to `getFixureClientScripts`. `createFixture` is the module that calls `getFixtureClientScripts`, which means it would also have to change to be asynchronous. Each spec file calls `createFixture`, so each spec file would have to `await` `createFixture`. And since each spec file is a top-level module, we'd then have to deal with this error: `Cannot use keyword 'await' outside an async function`, which means, I believe, that we would have to wrap the code in each spec file with an IIFE. 😢  So I decided to do it this way instead.

test/functional/helpers/createFixture/clientScripts.js
```diff
@@ -16,9 +16,14 @@ import readCache from "read-cache";
...
 // eslint-disable-next-line no-console
-console.log("ALLOY ENV:", alloyEnv);
+console.log(`ALLOY ENV: ${alloyEnv}`);
+
+if (alloyProdVersion) {
+  // eslint-disable-next-line no-console
+  console.log(`ALLOY PROD VERSION: ${alloyProdVersion}`);
+}
```
jonsnyder, line 26: Do you think we should throw an error If ALLOY_ENV is "prod" and alloyProdVersion is not specified? Or perhaps we can re-think these parameters. Having both ALLOY_ENV and ALLOY_PROD_VERSION seems redundant. Having the ALLOY_PROD_VERSION implies that we want ALLOY_ENV="prod".

jonsnyder (approved):

test/functional/helpers/createFixture/clientScripts.js
Aaronius: I kept the separate parameters because I think combining them would make it a bit messy, but I am now throwing an error if `ALLOY_ENV` is `prod` and `ALLOY_PROD_VERSION` is not specified.

PR #691, opened 2021-03-09
Run Functional test in batches using GitHub Actions Matrix

shammowla:
Functional tests idle when running when we run the full suite. Testcafe times out and other times Browserstack stalls. Broke down the suite to run by component instead. 

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/dev.yaml
+++ b/.github/workflows/dev.yaml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        browser: ["browserstack:chrome@87","browserstack:firefox@85","browserstack:ie@11","browserstack:safari@14.0"]
+        browser: ["'browserstack:chrome@87'","'browserstack:firefox@85'","'browserstack:ie@11'","'browserstack:safari@14.0'"]
     steps:
     - uses: actions/checkout@v2
     - name: Serialize workflow runs
@@ -49,7 +49,7 @@ jobs:
         npx testcafe ${{ matrix.browser }} test/functional/specs/Personalization/*.js
         npx testcafe ${{ matrix.browser }} test/functional/specs/Privacy/*.js
         npx testcafe ${{ matrix.browser }} test/functional/specs/Visitor/*.js
-    - name: Clear Browserstack Tunnels
-      run: |
-        chmod +x "${GITHUB_WORKSPACE}/scripts/clearBrowserStackJobs.sh"
-        "${GITHUB_WORKSPACE}/scripts/clearBrowserStackJobs.sh"
+    # - name: Clear Browserstack Tunnels
+    #   run: |
+    #     chmod +x "${GITHUB_WORKSPACE}/scripts/clearBrowserStackJobs.sh"
+    #     "${GITHUB_WORKSPACE}/scripts/clearBrowserStackJobs.sh"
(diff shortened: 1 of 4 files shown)


.github/workflows/dev.yaml
```diff
@@ -1,10 +1,22 @@
...
 
 jobs:
-  unit-test:
+  alloy-dev-test:
     name: "Run Unit & E2E Test"
     runs-on: macos-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        browser: ["browserstack:chrome@87","browserstack:firefox@85","browserstack:ie@11","browserstack:safari@14.0"]
```
jfkhoury, line 19: Can't we use `latest` ?

.github/workflows/dev.yaml
```diff
@@ -21,24 +33,23 @@ jobs:
...
-      uses: 'browserstack/github-actions/setup-local@master'
-      with:
-        local-testing: start
-        local-identifier: random
-    - name: Run Unit & Functional tests on BS
+    - name: Unit Test
+      run: npm run test:unit:browserstack:ci
+    - name: Functional e2e Tests
+      run: |
+        npx testcafe ${{ matrix.browser }} test/functional/specs/Audiences/*.js,
```
jfkhoury, line 40: The only danger I see her is if we add more tests, we need to remember to update this file. Any way around this?

.github/workflows/pre-deploy.yaml
```diff
@@ -36,8 +36,8 @@ jobs:
     - name: Run TestCafe Tests
```
jfkhoury, line 36: These workflows don't cause the same problem as the dev tests?

.github/workflows/dev.yaml
```diff
@@ -1,10 +1,22 @@
...
 
 jobs:
-  unit-test:
+  alloy-dev-test:
     name: "Run Unit & E2E Test"
     runs-on: macos-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        browser: ["'browserstack:chrome@87'","'browserstack:firefox@85'","'browserstack:ie@11'","'browserstack:safari@14.0'"]
```
jfkhoury, line 19: Can't we use the `latest` version?

.github/workflows/dev.yaml
Aaronius: @shammowla and I discussed this. I've recommended that we create a script that finds all spec files and then runs them in, say, batches of 10. After you left the release party, we decided to write that script later (@shammowla, please create a ticket for yourself so you don't forget).

jonsnyder (approved):

.github/workflows/pre-deploy.yaml
```diff
@@ -36,8 +36,8 @@ jobs:
...
       run: |
         chmod +x "${GITHUB_WORKSPACE}/scripts/clearBrowserStackJobs.sh"
-        "${GITHUB_WORKSPACE}/scripts/clearBrowserStackJobs.sh"
         npm run test:functional:ci:predeploy
+        "${GITHUB_WORKSPACE}/scripts/clearBrowserStackJobs.sh"
```
Aaronius, line 40: I'm curious to know why you switched this to be executed after running the tests instead of before. If tests fail, will it still run the `clearBrowserStackJobs` script?

Aaronius (approved):

.github/workflows/dev.yaml
jfkhoury: thanks @Aaronius

PR #692, opened 2021-03-15
Re-enabled sendBeacon tests after fixing them for Safari.

Aaronius:
## Description
The tests involving the collect endpoint + sendBeacon were skipped before the release because they needed some re-work to ensure they were asserting the things we were expecting and that they weren't intermittently failing. This PR addresses those problem. Hopefully the code comments explain the problem and why we do what we have to do.

## Related Issue
https://jira.corp.adobe.com/browse/PDCL-4617

## Motivation and Context
Re-enable tests for better coverage.

## Types of changes

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "test:unit:browserstack:ci": "karma start karma.browserstack.conf.js --single-run",
     "test:unit:coverage": "karma start --single-run --reporters spec,coverage",
     "test:functional": "rollup -c --environment BASE_CODE,NPM_PACKAGE_LOCAL && EDGE_BASE_PATH=\"ee-pre-prd\" ALLOY_ENV=\"int\" testcafe chrome",
-    "test:functional:watch": "EDGE_BASE_PATH=\"ee-pre-prd\" ALLOY_ENV=\"int\" ./scripts/watchFunctionalTests.js",
+    "test:functional:watch": "EDGE_BASE_PATH=\"ee-pre-prd\" ALLOY_ENV=\"int\" ./scripts/watchFunctionalTests.js --browsers chrome",
     "test:functional:ci:build": "rollup -c --environment BASE_CODE,NPM_PACKAGE_LOCAL",
     "test:functional:ci:predeploy": "ALLOY_ENV=\"int\" testcafe -c 1 'browserstack:chrome@87:Windows 10','browserstack:ie@11:Windows 10','browserstack:firefox@85:Windows 10','browserstack:safari@14.0:OS X Big Sur'",
     "test:functional:ci:dev": "EDGE_BASE_PATH=\"ee-pre-prd\" ALLOY_ENV=\"int\" testcafe -c 1 'browserstack:chrome@87:Windows 10','browserstack:ie@11:Windows 10','browserstack:firefox@85:Windows 10','browserstack:safari@14.0:OS X Big Sur'",
@@ -126,6 +126,7 @@
     "testcafe": "^1.10.0",
     "testcafe-browser-provider-browserstack": "^1.13.0-alpha.1",
     "ua-parser": "^0.3.5",
-    "url-parse": "^1.4.7"
+    "url-parse": "^1.4.7",
+    "yargs": "^16.2.0"
   }
 }
--- a/test/functional/helpers/createCollectEndpointAsserter.js
+++ b/test/functional/helpers/createCollectEndpointAsserter.js
@@ -36,13 +36,12 @@ export default async () => {
 
   const assertCollectCalled = async () => {
     if (isSendBeaconSupported()) {
-      // When sendBeacon is called, the browser doesn't send
-      // the request right away, but instead queues the request to be sent sometime
-      // later. Therefore, the sendEvent promise resolves before any network
-      // request may have actually been made. By calling the RequestLogger's count
-      // method, TestCafe will retry this assertion until it succeeds or until the
-      // timeout is reached. The parameter to count is a filter function. In this
-      // case we want to count all the requests.
+      // When sendBeacon is called, the browser doesn't send the request right away,
+      // but instead queues the request to be sent sometime later. Therefore, the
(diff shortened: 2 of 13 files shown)


package.json
```diff
@@ -19,7 +19,7 @@
     "test:unit:browserstack:ci": "karma start karma.browserstack.conf.js --single-run",
     "test:unit:coverage": "karma start --single-run --reporters spec,coverage",
     "test:functional": "rollup -c --environment BASE_CODE,NPM_PACKAGE_LOCAL && EDGE_BASE_PATH=\"ee-pre-prd\" ALLOY_ENV=\"int\" testcafe chrome",
-    "test:functional:watch": "EDGE_BASE_PATH=\"ee-pre-prd\" ALLOY_ENV=\"int\" ./scripts/watchFunctionalTests.js",
+    "test:functional:watch": "EDGE_BASE_PATH=\"ee-pre-prd\" ALLOY_ENV=\"int\" ./scripts/watchFunctionalTests.js --browsers chrome",
```
Aaronius, line 22: When we wanted to watch functional tests run in a certain browser, we would have to go edit `watchFunctionalTests.js`. Now we can just tweak this `browsers` argument.

test/functional/helpers/createCollectEndpointAsserter.js
```diff
@@ -0,0 +1,122 @@
...
+    async assertNeitherCollectNorInteractCalled() {
+      await assertCollectNotCalled();
+      await assertInteractNotCalled();
+    },
+    async reset() {
+      await networkLogger.clearLogs();
+      await sendBeaconMock.reset();
+    }
+  };
+};
```
Aaronius, line 122: This file is really the meat of the refactor. It's intended to abstract the complexities of the issue away from the tests themselves. While I could have broken these methods down more to something like:

* `assertCollectCalled`
* `assertCollectNotCalled`
* `assertInteractCalled`
* `assertInteractNotCalled`

it would require the test calling the methods to understand that sometimes the order in which they are called matters (see the comment on `assertCollectCalledAndNotInteract`). To make it hard for other developers to make mistakes, I decided to take the route I have taken.

test/functional/helpers/createFixture/destinationRequestMock.js
```diff
@@ -0,0 +1,23 @@
...
+import { RequestMock } from "testcafe";
+
+// The main edge configuration we're using for testing
+// has URL destinations enabled and a URL destination handle is sent
+// back as part of the response from each interact endpoint request.
+// So that we're not relying on a third-party server for the URL
+// destination to successfully execute, we'll mock the third-party
+// URL destination endpoint.
+export default RequestMock()
+  .onRequestTo("https://cataas.com/cat")
```
Aaronius, line 22: This endpoint was down last week, which was causing a bunch of tests to fail, so I mocked the endpoint, which we should probably should have done originally anyway.

Aaronius:
Also, github actions are currently down (https://www.githubstatus.com/) which I believe is why our workflows aren't running on this PR. https://www.githubstatus.com/ I'd like to see them pass on the CI servers before merging.

test/functional/helpers/createCollectEndpointAsserter.js
```diff
@@ -0,0 +1,122 @@
...
+  };
+
+  return {
+    async assertCollectCalledAndNotInteract() {
+      // The order of these matter because we need to make sure sendBeacon
+      // requests have actually been sent by the browser before checking
+      // that they weren't interact calls.
+      await assertCollectCalled();
+      await assertInteractNotCalled();
+      return networkLogger.edgeCollectEndpointLogs.requests[0];
```
jonsnyder, line 106: From the name of this function it is weird that it returns a value. You could create a `getInteractRequest` and `getCollectRequest` to return the request. You'd have to make sure that a call to getInteractRequest or getCollectRequest is done after an assertion. This could be done by throwing an exception if getXXXRequest is done before an assertion method.

```
  let collectRequestAsserted = false;
  let interactRequestAsserted = false;
  return {
     async assertCollectCalledAndNotInteract() {
       collectRequestAsserted = true;
       // The order of these matter because we need to make sure sendBeacon
       // requests have actually been sent by the browser before checking
       // that they weren't interact calls.
       await assertCollectCalled();
       await assertInteractNotCalled();
    },
     async assertInteractCalledAndNotCollect() {
       interactRequestAsserted = true;
       await assertCollectNotCalled();
       await assertInteractCalled();
     },
     async assertNeitherCollectNorInteractCalled() {
       await assertCollectNotCalled();
       await assertInteractNotCalled();
     },
     async reset() {
       interactRequestAsserted = false;
       collectRequestAsserted = false;
       await networkLogger.clearLogs();
       await sendBeaconMock.reset();
     },
     getInteractRequest() {
       if (!interactRequestAsserted) {
         throw new Error("You must call assertInteractCalledAndNotCollect before getting the interact request");
       }
       return networkLogger.edgeInteractEndpointLogs.requests[0];
     },
     getCollectRequest() {
       if (!collectRequestAsserted) {
         throw new Error("You must call assertCollectCalledAndNotInteract before getting the collect request");
       }
       return networkLogger.edgeCollectEndpointLogs.requests[0];
     }
 };
```

jonsnyder (approved):
Great work! This is a good way to abstract these tricky assertions.

test/functional/helpers/createCollectEndpointAsserter.js
jfkhoury, line 122: Thank you for the awesome comments.

test/functional/helpers/createFixture/destinationRequestMock.js
jfkhoury: Nice! I'm gonna that funny picture of the cat though.. 😸

jfkhoury (approved):
Let's stabilize those bad boys!!

Aaronius:
Thank you for the feedback! It's been addressed.

PR #704, opened 2021-03-24
PDCL-4717: Move CI test configurations from package.json to github workflow files

shammowla:
Moved CI test configurations out of package.json and into the Github workflow files.

## Types of changes

## Checklist:


package.json
```diff
@@ -22,10 +22,7 @@
...
-    "test:functional:ci:dev": "SAUCE_CAPABILITIES_OVERRIDES_PATH=\"sauceLabsCapabilities.json\" EDGE_BASE_PATH=\"ee-pre-prd\" ALLOY_ENV=\"int\" testcafe -q -c 5 'saucelabs:Chrome@latest-1:macOS 11.00','saucelabs:IE@11.285:Windows 10','saucelabs:Firefox@latest:Windows 10','saucelabs:Safari@latest:macOS 11.00'",
-    "test:functional:ci:prod": "SAUCE_CAPABILITIES_OVERRIDES_PATH=\"sauceLabsCapabilities.json\" ALLOY_PROD_VERSION=`scripts/getProdVersion.js` ALLOY_ENV=\"prod\" testcafe -q -c 5 'saucelabs:Chrome@latest-1:macOS 11.00','saucelabs:IE@11.285:Windows 10','saucelabs:Firefox@latest:Windows 10','saucelabs:Safari@latest:macOS 11.00'",
-    "test:functional:ci:edgepredeploy": "SAUCE_CAPABILITIES_OVERRIDES_PATH=\"sauceLabsCapabilities.json\" ALLOY_PROD_VERSION=`scripts/getProdVersion.js` ALLOY_ENV=\"prod\" EDGE_BASE_PATH=\"ee-pre-prd\" testcafe -q -c 5 'saucelabs:Chrome@latest-1:macOS 11.00','saucelabs:IE@11.285:Windows 10','saucelabs:Firefox@latest:Windows 10','saucelabs:Safari@latest:macOS 11.00'",
+    "test:functional:ci": "testcafe -q -c 5",
```
Aaronius, line 25: At this point I'm wondering if we should even get rid of `test:functional:ci`. If we move `testcafe -q -c 5` into the workflow files, then we can tweak those arguments for all the versions of Alloy under test without having to go update tags for previous releases. What do you think?

Aaronius: Also, do you think we should be doing something similar for unit tests? I think that would mean getting rid of karma.saucelabs.conf and moving the configuration into the workflow files.

shammowla: @Aaronius 

1) I wasn't able to use "getProdVersion.js" script on the prod workflow to retrieve the version number. Should I revert the changes and have a separate PR? I have a ticket specifically for this task - https://jira.corp.adobe.com/browse/PDCL-4714

2) I tried running the unit tests without the "karma.saucelabs.conf.js" with the test environments but was unable to with the karma runner. I removed the package script and moved it to the dev workflow to run the unit test directly on karma.

3) I enabled logging performance tests on saucelabs, should I disable it?

jonsnyder (approved):

PR #706, opened 2021-03-25
[Redirect offer support] - Adding scope details to the decision

ninaceban:
## Description

In case of a redirect from PAGE_A to PAGE_B we will have to notify Analytics that we are on PAGE_B because of a Target Redirect Offer.
In this case, we will have along `scope, id and items` another field called `scopeDetails` where we will store data like: `activity, experience, strategies and blob`.
`blob` will contain analytics related data, like: `tnta, state and eventToken`. 
The `scopeDetails` object will not be processed by alloy, it will be just cached and forwarded back to konductor.

We will use a special `eventType: "redirect"` to differentiate a notification call for a redirect offer being executed. 
When a notification for a redirect will be fired, on the Konductor side we will extract the `tnta` value and return it to alloy in the `state:store` payload, so that alloy could create a cookie. It is similar to how we process `target session ID`. This is why we will need to wait till the response is back and then we can do the magic with `window.location.replace` to the new URL. 

The next interact call from the PAGE_B will read the cookie, do the Analytics data exchange and then remove the cookie.

PS: this scenario will work in case customer decides to execute the redirect offer by themselves and send the `redirect` notification (granted they will provide the same `scopeDetails` )
## Types of changes

## Checklist:

### DIFF
--- a/sandbox/src/RedirectOffers.js
+++ b/sandbox/src/RedirectOffers.js
@@ -1,5 +1,5 @@
 import React, { useEffect } from "react";
-import ContentSecurityPolicy from "./ContentSecurityPolicy";
+import ContentSecurityPolicy from "./components/ContentSecurityPolicy";
 
 export default function RedirectOffers() {
   useEffect(() => {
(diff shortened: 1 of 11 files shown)


jonsnyder (approved):

ninaceban:
@jfkhoury Based on our discussion, we will be using the old good Query String Param Method of forwarding the **tnta** therefore Alloy doesn't need to know about **tnta**. I will close this PR.

PR #708, opened 2021-03-26
PDCL-4785 Log when an event has been canceled by onBeforeEventSend.

### LINKED ISSUE #700: Console logging for sendEvent return false

Aaronius:
## Description
Closes #700. When an event is canceled by an `onBeforeEventSend` callback returning `false`, the customer would like to know that the event has been canceled by seeing a log message in the console.

While working on this, I noticed we were using `logger.log` sometimes and `logger.info` other times, without any real pattern as to when/why we were using the separate APIs. I've removed the `logger.log` API entirely and converted all usages to `logger.info`.

## Related Issue
https://jira.corp.adobe.com/browse/PDCL-4785

## Motivation and Context
Improved transparency into what's happening with events.

## Types of changes

## Checklist:

### DIFF
--- a/src/core/createEventManager.js
+++ b/src/core/createEventManager.js
@@ -12,8 +12,12 @@ governing permissions and limitations under the License.
 
 import { createCallbackAggregator } from "../utils";
 
+const EVENT_CANCELLATION_MESSAGE =
+  "Event was canceled because the onBeforeEventSend callback returned false.";
+
 export default ({
   config,
+  logger,
   lifecycle,
   consent,
   createEvent,
@@ -74,7 +78,8 @@ export default ({
           // if the callback returns false, the event should not be sent
           if (!event.shouldSend()) {
             onRequestFailureCallbackAggregator.add(lifecycle.onRequestFailure);
-            const error = new Error("Event was cancelled.");
+            logger.info(EVENT_CANCELLATION_MESSAGE);
+            const error = new Error(EVENT_CANCELLATION_MESSAGE);
             return onRequestFailureCallbackAggregator
               .call({ error })
               .then(() => {
--- a/src/components/Audiences/injectProcessDestinations.js
+++ b/src/components/Audiences/injectProcessDestinations.js
@@ -25,7 +25,7 @@ const processUrls = (fireReferrerHideableImage, logger, destinations) => {
     urlDestinations.map(urlDestination => {
       return fireReferrerHideableImage(urlDestination.spec)
         .then(() => {
-          logger.log(createResultLogMessage(urlDestination, true));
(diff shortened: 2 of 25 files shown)


src/core/createEventManager.js
```diff
@@ -12,8 +12,12 @@ governing permissions and limitations under the License.
 
 import { createCallbackAggregator } from "../utils";
 
+const EVENT_CANCELLATION_MESSAGE =
+  "Event was canceled because the onBeforeEventSend callback returned false.";
+
 export default ({
   config,
+  logger,
```
Aaronius, line 20: `logger` was already being passed in.

src/core/createEventManager.js
```diff
@@ -74,7 +78,8 @@ export default ({
           // if the callback returns false, the event should not be sent
           if (!event.shouldSend()) {
             onRequestFailureCallbackAggregator.add(lifecycle.onRequestFailure);
-            const error = new Error("Event was cancelled.");
+            logger.info(EVENT_CANCELLATION_MESSAGE);
```
Aaronius, line 81: This is the meat of the PR where we actually log a message.

As you can see, I made the message say `Event was canceled because the onBeforeEventSend callback returned false.` rather than just `Event was canceled.` because I think that gives the customer a better idea of how the event was canceled.

With this in mind, I was on the fence as to whether we should log this message here in `createEventManager.js` or `createEvent.js` (where we actually call `onBeforeEventSendCallback`). From the perspective of the event manager, the event manager only knows that the event shouldn't be sent. It doesn't necessarily know for certain that the reason it shouldn't be sent is that `onBeforeEventSend` returned `false`. From the perspective of the event, the event knows that the `onBeforeEventSend` callback returned `false` and therefore the event makes its `shouldSend` method return `false`, but it technically doesn't know that the event will not actually be sent or that it's not going to be sent due to other potential reasons as well.

Anyway, I'm curious if you have opinions here.

jonsnyder: You could add a method on the event `getCancelReason` or something similar. That way the event could keep track of why the event shouldn't be sent and then the eventManager can decide what to do with that information. For now, though, there is only one possible reason that the event shouldn't be sent namely that false was returned. When an error is thrown, there is a different reason, but then the error is thrown from finalize. So maybe it doesn't make sense to add this method.

jonsnyder (approved):

src/core/createEventManager.js
Aaronius: Yeah, that's similar to what I was thinking as well. When the time comes that are more reasons for the event to not be sent, we can play with event's API.

jfkhoury (approved):

PR #718, opened 2021-04-26
Only allow positive values in viewportWidth/Height, use documentElement

jonsnyder:
## Description

Use window.document.documentElement.clientWidth/Height instead of window.innerWidth/Height. Add a check for negative values. If the values are negative or not present, don't include the data in the final xdm.

## Related Issue

https://jira.corp.adobe.com/browse/PDCL-4786

## Motivation and Context

Some customers are getting negative width or height values causing XDM validation errors. https://stackoverflow.com/questions/4629969/ios-return-bad-value-for-window-innerheight-width/39345914

## Types of changes

## Checklist:


src/components/Context/injectEnvironment.js
```diff
@@ -14,14 +14,24 @@ import { deepAssign } from "../../utils";
...
+
     const environment = {
-      type: "browser",
-      browserDetails: {
-        viewportWidth: innerWidth,
-        viewportHeight: innerHeight
-      }
+      type: "browser"
     };
+    const hasWidth = clientWidth && clientWidth >= 0;
```
Aaronius, line 24: This probably isn't what you're intending. If `clientWidth` is `0`, then `hasWidth` will end up being `0`, which means `environment.browserDetails.viewportWidth` won't be set. It's probably worth adding a unit test for this case.

Aaronius (changes requested):
Overall it's good. Just need to fix the `0` value issue.

src/components/Context/injectEnvironment.js
jonsnyder: Thanks for finding this!

src/components/Context/injectEnvironment.js
```diff
@@ -10,18 +10,28 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
 */
 
-import { deepAssign } from "../../utils";
+import { deepAssign, isNumber } from "../../utils";
 
 export default window => {
   return xdm => {
-    const { innerWidth, innerHeight } = window;
+    const {
+      document: { documentElement: { clientWidth, clientHeight } = {} }
```
liljenback, line 18: So AppMeasurement is using offsetWidth here. What made you pick clientWidth? There seems to be a little bit of a difference in that offsetWidth includes some additional browser elements etc.

src/components/Context/injectEnvironment.js
```diff
@@ -10,18 +10,28 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
-      type: "browser",
-      browserDetails: {
-        viewportWidth: innerWidth,
-        viewportHeight: innerHeight
-      }
+      type: "browser"
     };
+    const hasWidth = isNumber(clientWidth) && clientWidth >= 0;
+    const hasHeight = isNumber(clientHeight) && clientHeight >= 0;
+    if (hasWidth || hasHeight) {
```
liljenback, line 26: Is there a valid use-case for not doing && here?

src/components/Context/injectEnvironment.js
jonsnyder: I was just following the stack overflow answer linked above. I didn't realize appMeasurement was using documentElement as a fallback. offsetWidth and offsetHeight are probably better.

src/components/Context/injectEnvironment.js
jonsnyder: I can't think of one. I like it.

src/components/Context/injectEnvironment.js
liljenback: Using the property that will fluctuate the least (based on margin settings etc.) to enable for common resolutions to group well would probably be preferred.

jfkhoury (approved):

liljenback (approved):

PR #739, opened 2021-06-07
PDCL-5968 - resolve the sendEvent command with decisions with render status

ninaceban:
## Description

Resolve the sendEvent promise with `decisions` array.
We will have different use cases:

- `renderDecisions: true`

     1. Autorenderable decisions with 3 different `renderStatus`: 
         1. `RENDERED` - when all the items were successfully rendered
         2. `FAILED_TO_RENDER` - when all the items failed to render
         3. `PARTIALLY_RENDERED` - when at least one item has failed to render
         4.  items will also have a status: `SUCCEEDED_TO_RENDER` or `FAILED_TO_RENDER` 
     2. Form-Based composed decisions with `NON_RENDERABLE` status
     
- `renderDecisions: false` 
     1. Autorenderable decisions with `NOT_RENDERED` status
     2. Form-Based composed decisions with `NON_RENDERABLE` status

When `page-load` we will return decisions for:
1. `__view__` scope
2. `xdm.web.webPageDetails.viewName` view
3. `decisionScopes`

When viewChange we will return decisions for:
1.   `xdm.web.webPageDetails.viewName` view
    

---

## Types of changes

## Checklist:


ninaceban:
@Aaronius I'm working on the tests, but I would appreciate if you could review the overall approach I took into solving this issue. Thank you

src/components/Personalization/createViewChangeHandler.js
```diff
@@ -10,17 +10,34 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
   return ({ personalizationDetails, onResponse, onRequestFailure }) => {
     const viewName = personalizationDetails.getViewName();
-
+    const executedDecisionsResult = [];
     if (personalizationDetails.isRenderDecisions()) {
-      executeCachedViewDecisions({ viewName });
-      return;
+      executeCachedViewDecisions({ viewName }).then(result => {
+        executedDecisionsResult.push(result);
+      });
```
Aaronius, line 28: I don't really know what I'm talking about here, but should we be waiting for this promise? It seems like there might be a race condition where our `onResponse` callback reports `executedDecisionsResult` before `executedDecisionsResult.push` has had a chance to be called.

Aaronius (commented):
Thanks for this @ninaceban. The `NOT_RENDERED` status is an interesting one. This makes me wonder if we could simplify the number of constants we have since we're up to six at this point and they're a bit nuanced. What do you think if we just had status values of `succeeded`, `failed`, and `notAttempted`. So something like this:

* `renderDecisions: true`

  * Every auto-renderable decision will have one of the following `renderStatus` values:
    * `succeeded` - when all items were successfully rendered
    * `failed` - when _any_ item failed to render
  * Every item in every auto-renderable decision will have a `renderStatus` value of `succeeded` or `failed`.
  * Every form-based decision will have a `renderStatus` value of `notAttempted`.
  * Every item in every form-based decision will have a `renderStatus` values of `notAttempted`.

* `renderDecisions: false`

  * Every auto-renderable and form-based decision will have a `renderStatus` value of `notAttempted`.
  * Every item in every decision will have a `renderStatus` value of `notAttempted`.

I don't know for sure that this is the right way to go. I'd like your feedback.

I reviewed the code a bit, but it's hard for me to tell what's going on without putting a debugger in there and following it through, but we don't have sandbox examples that cover the various cases yet. I'll dig deeper once tests are implemented.

ninaceban:
> * `failed` - when _any_ item failed to render

We are still sending the `display` notification call for that decision even though an item have failed to render. I think we should keep the `PARTIALLY_RENDERED` status. 
I would still go with `NON_RENDERABLE` instead of `NOT_ATTEMPTED` because it is more descriptive for me, we can still discuss the naming in our check-in.
I will reduce the number of statuses to: `SUCCEEDED` , `FAILED`, `PARTIALLY_RENDERED` and `NON_RENDERABLE`. An Item will be  either`SUCCEEDED` or `FAILED`.

Aaronius:
Thanks, Nina. Keep going on the tests and we'll discuss during Check-in.

jfkhoury:
@ninaceban Do we close this PR?

Aaronius:
Closing in favor of https://github.com/adobe/alloy/pull/741.

PR #741, opened 2021-06-29
PDCL-5968 - Response Tokens: return all propositions with response tokens

ninaceban:
## Description
Response tokens is a feature that is used by our customers mainly to send data to third parties, like google analytics, facebook etc. Every item within a decision object that will be returned from `konductor` will contain a `meta` block which contains the response tokens.
As we decided to proceed with a simple solution we will add `render:true` to decisions that we attempted to render. So that the customer could identify which decision items were attempted to render, so that he could then send the response tokens further. We will return  a `propositions` array and we will still return `decisions` array for backward compatibility.

note: Only the `propositions` array will contain decisions with `render:true` flag.

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Personalization/constants/loggerMessages.js
+++ b/src/components/Personalization/constants/loggerMessages.js
@@ -1,5 +1,0 @@
-export const decisionsDeprecatedWarning =
-  "The decisions property is decisionsDeprecatedWarning. Please use the propositions property instead.";
-export const authoringEnabled = "Rendering is disabled, authoring mode.";
-export const redirectExecutionError =
-  "An error occurred while executing the redirect offer.";
--- a/src/components/Personalization/constants/loggerMessages.js
+++ b/src/components/Personalization/constants/loggerMessages.js
@@ -1,0 +1,5 @@
+export const decisionsDeprecatedWarning =
+  "The decisions property is decisionsDeprecatedWarning. Please use the propositions property instead.";
+export const authoringEnabled = "Rendering is disabled, authoring mode.";
+export const redirectExecutionError =
+  "An error occurred while executing the redirect offer.";
--- a/src/components/Personalization/createAutoRenderingHandler.js
+++ b/src/components/Personalization/createAutoRenderingHandler.js
@@ -11,7 +11,7 @@ governing permissions and limitations under the License.
 */
 
 import addRenderAttemptedToDecisions from "./utils/addRenderAttemptedToDecisions";
-import { decisionsDeprecatedWarning } from "./constants/loggerMessages";
+import { DECISIONS_DEPRECATED_WARNING } from "./constants/loggerMessage";
 
 export default ({
   viewCache,
@@ -32,7 +32,7 @@ export default ({
 
         return {
           get decisions() {
-            logger.warn(decisionsDeprecatedWarning);
+            logger.warn(DECISIONS_DEPRECATED_WARNING);
 
             return [...formBasedComposedDecisions];
           },
@@ -55,7 +55,7 @@ export default ({
 
     return {
       get decisions() {
-        logger.warn(decisionsDeprecatedWarning);
+        logger.warn(DECISIONS_DEPRECATED_WARNING);
 
         return [...formBasedComposedDecisions];
       },
--- a/src/components/Personalization/createAutoRenderingHandler.js
+++ b/src/components/Personalization/createAutoRenderingHandler.js
@@ -1,0 +1,39 @@
+import addRenderToExecutedDecisions from "./utils/addRenderToExecutedDecisions";
+
+export default ({
+  viewCache,
+  executeDecisions,
+  executeCachedViewDecisions,
+  showContainers
+}) => {
+  return ({ viewName, pageWideScopeDecisions, formBasedComposedDecisions }) => {
+    if (viewName) {
+      return viewCache.getView(viewName).then(currentViewDecisions => {
+        executeDecisions(pageWideScopeDecisions);
+        executeCachedViewDecisions({
+          viewName,
+          viewDecisions: currentViewDecisions
+        });
+        showContainers();
+
+        return {
+          decisions: [
+            ...addRenderToExecutedDecisions(pageWideScopeDecisions),
+            ...addRenderToExecutedDecisions(currentViewDecisions),
+            ...formBasedComposedDecisions
+          ]
+        };
+      });
+    }
+
+    executeDecisions(pageWideScopeDecisions);
+    showContainers();
+
+    return {
+      decisions: [
+        ...addRenderToExecutedDecisions(pageWideScopeDecisions),
+        ...formBasedComposedDecisions
+      ]
+    };
+  };
+};
--- a/src/components/Personalization/createAutoRenderingHandler.js
+++ b/src/components/Personalization/createAutoRenderingHandler.js
@@ -17,7 +17,8 @@ export default ({
         showContainers();
 
         return {
-          decisions: [
+          decisions: [...formBasedComposedDecisions],
+          propositions: [
             ...addRenderToExecutedDecisions(pageWideScopeDecisions),
             ...addRenderToExecutedDecisions(currentViewDecisions),
             ...formBasedComposedDecisions
@@ -30,7 +31,8 @@ export default ({
     showContainers();
 
     return {
-      decisions: [
+      decisions: [...formBasedComposedDecisions],
+      propositions: [
         ...addRenderToExecutedDecisions(pageWideScopeDecisions),
         ...formBasedComposedDecisions
       ]
--- a/src/components/Personalization/createAutoRenderingHandler.js
+++ b/src/components/Personalization/createAutoRenderingHandler.js
@@ -10,7 +10,8 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import addRenderToExecutedDecisions from "./utils/addRenderToExecutedDecisions";
+import addRenderAttemptedToDecisions from "./utils/addRenderAttemptedToDecisions";
+import { decisionsDeprecatedWarning } from "./constants/loggerMessages";
 
 export default ({
   viewCache,
@@ -31,14 +32,19 @@ export default ({
 
         return {
           get decisions() {
-            logger.warn("Decisions property is deprecated");
+            logger.warn(decisionsDeprecatedWarning);
 
             return [...formBasedComposedDecisions];
           },
           propositions: [
-            ...addRenderToExecutedDecisions(pageWideScopeDecisions),
-            ...addRenderToExecutedDecisions(currentViewDecisions),
-            ...formBasedComposedDecisions
+            ...addRenderAttemptedToDecisions({
+              decisions: [...pageWideScopeDecisions, ...currentViewDecisions],
+              renderAttempted: true
+            }),
+            ...addRenderAttemptedToDecisions({
+              decisions: formBasedComposedDecisions,
+              renderAttempted: false
+            })
           ]
         };
       });
@@ -49,13 +55,19 @@ export default ({
 
     return {
       get decisions() {
-        logger.warn("Decisions property is deprecated");
+        logger.warn(decisionsDeprecatedWarning);
 
         return [...formBasedComposedDecisions];
       },
       propositions: [
-        ...addRenderToExecutedDecisions(pageWideScopeDecisions),
-        ...formBasedComposedDecisions
+        ...addRenderAttemptedToDecisions({
+          decisions: pageWideScopeDecisions,
+          renderAttempted: true
+        }),
+        ...addRenderAttemptedToDecisions({
+          decisions: formBasedComposedDecisions,
+          renderAttempted: false
+        })
       ]
     };
   };
--- a/src/components/Personalization/createAutoRenderingHandler.js
+++ b/src/components/Personalization/createAutoRenderingHandler.js
@@ -1,10 +1,23 @@
+/*
+Copyright 2021 Adobe. All rights reserved.
+This file is licensed to you under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License. You may obtain a copy
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
 import addRenderToExecutedDecisions from "./utils/addRenderToExecutedDecisions";
 
 export default ({
   viewCache,
   executeDecisions,
   executeCachedViewDecisions,
-  showContainers
+  showContainers,
+  logger
 }) => {
   return ({ viewName, pageWideScopeDecisions, formBasedComposedDecisions }) => {
     if (viewName) {
@@ -17,7 +30,11 @@ export default ({
         showContainers();
 
         return {
-          decisions: [...formBasedComposedDecisions],
+          get decisions() {
+            logger.warn("Decisions property is deprecated");
+
+            return [...formBasedComposedDecisions];
+          },
           propositions: [
             ...addRenderToExecutedDecisions(pageWideScopeDecisions),
             ...addRenderToExecutedDecisions(currentViewDecisions),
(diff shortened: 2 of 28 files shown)


src/components/Personalization/createAutoRenderingHandler.js
```diff
@@ -0,0 +1,39 @@
+import addRenderToExecutedDecisions from "./utils/addRenderToExecutedDecisions";
+
+export default ({
+  viewCache,
+  executeDecisions,
+  executeCachedViewDecisions,
+  showContainers
+}) => {
+  return ({ viewName, pageWideScopeDecisions, formBasedComposedDecisions }) => {
```
jfkhoury, line 9: I recommend we rename those to something more generic that would apply to ODE, AJO...

`formBasedComposedDecisions` & `pageWideScopeDecisions`

Maybe something like: `renderableDecisions` & `nonRenderableDecisions`.

Also, aren't we renaming `decisions` to `propositions`?

ninaceban: Thank you, Joe. I will rename those. 
Yes we return `decisions` the way we did earlier and newly `propositions` with decisions with rendered flag.

Aaronius:
@ninaceban how do we test the non-rendering propositions in the sandbox? Are there any examples of that set up?

Aaronius:
Oh, I think that's the `decisionScopes: ["sandbox-personalization-page"]` you've commented out. Maybe we should request the scope on the personalization page or something like that and "manually" render the proposition to the UI in some form so we can see that it's working?

src/components/Personalization/decisionsExtractor.js
```diff
@@ -75,11 +77,35 @@ const extractDecisionsByScope = (decisions, scope) => {
...
+};
+
 export default {
-  groupDecisionsBySchema({ decisions, schema }) {
-    return splitDecisions(decisions, schema);
-  },
-  groupDecisionsByScope({ decisions, scope }) {
-    return extractDecisionsByScope(decisions, scope);
+  groupDecisions(decisions) {
+    return getGroupedDecisions(decisions);
```
Aaronius, line 109: I really like this change. I think it cleaned up a lot of the code that depended on it. Consider exporting this function as the default and renaming the module accordingly.

test/unit/specs/components/Personalization/createNonRenderingHandler.spec.js
```diff
@@ -0,0 +1,71 @@
+/*
+Copyright 2020 Adobe. All rights reserved.
```
Aaronius, line 2: 2021

test/unit/specs/components/Personalization/utils/addRenderToExecutedDecisions.spec.js
```diff
@@ -0,0 +1,29 @@
...
+      {
+        blah: "123"
+      },
+      {
+        blah: "345"
+      }
+    ];
+    const result = addRenderToExecutedDecisions(decisions);
+    expect(result[0].rendered).toEqual(true);
+    expect(result[1].rendered).toEqual(true);
```
Aaronius, line 27: I think it's important to assert that `decisions[0].rendered` and `decisions[1].rendered` are still `undefined`.

Aaronius (commented):
This is looking good, @ninaceban! Thanks!

A few remaining items:

* We need functional tests.

* In the JIRA issue, I wrote this regarding the `decisions` property on the response object:

> For backward-compatibility, this needs to stay the same (still returning only form-based decisions), but when the property is accessed, a warning needs to be logged to the console explaining that the decisions property has been deprecated and that the customer should start using a new propositions property.

I still think we should do this. That means creating a getter for the `decisions` property.

* Let's finalize the name for the property currently named `rendered`. I suggested `isRenderAttempted` (to be clear that we've attempted to render it but not promising that it will end up succeeding). Artur preferred `autoRenderingEnabled`, which I'm also fine with, though I'd shorten it to `autoRenderEnabled`. @jonsnyder what's your take?

src/components/Personalization/createOnResponseHandler.js
```diff
@@ -9,87 +9,60 @@ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTA
...
-    if (viewName && nonPageWideScopeDecisions[viewName]) {
-      decisionsToBeReturned.push(...nonPageWideScopeDecisions[viewName]);
+      return autoRenderingHandler({
+        viewName,
+        pageWideScopeDecisions,
+        formBasedComposedDecisions
+      });
     }
-
-    return { decisions: decisionsToBeReturned };
```
Aaronius, line 93: This file is cleaning up well.

src/components/Personalization/createNonRenderingHandler.js
```diff
@@ -0,0 +1,40 @@
...
+        ...pageWideScopeDecisions,
+        ...formBasedComposedDecisions
+      ],
+      propositions: [
+        ...redirectDecisions,
+        ...pageWideScopeDecisions,
+        ...formBasedComposedDecisions
+      ]
+    };
+  };
```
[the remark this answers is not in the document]
ninaceban: thank you Aaron!

test/unit/specs/components/Personalization/utils/addRenderToExecutedDecisions.spec.js
ninaceban: thank you

src/components/Personalization/createOnResponseHandler.js
ninaceban: thank you, yes, that was my intention

test/unit/specs/components/Personalization/createNonRenderingHandler.spec.js
ninaceban: thank you

src/components/Personalization/decisionsExtractor.js
ninaceban: thank you :)

src/components/Personalization/createExecuteCachedViewDecisions.js
```diff
@@ -12,17 +12,15 @@ governing permissions and limitations under the License.
...
-      if (isNonEmptyArray(viewDecisions)) {
-        executeViewDecisions(viewDecisions);
-        return;
-      }
-      const xdm = { web: { webPageDetails: { viewName } } };
+export default ({ executeViewDecisions, collect }) => {
+  return ({ viewName, viewDecisions }) => {
+    if (isNonEmptyArray(viewDecisions)) {
+      executeViewDecisions(viewDecisions);
+      return;
```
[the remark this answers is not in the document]
ninaceban: sure thing!

src/components/Personalization/createNonRenderingHandler.js
```diff
@@ -0,0 +1,29 @@
+export default ({ viewCache }) => {
```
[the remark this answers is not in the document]
ninaceban: thank you

src/components/Personalization/createAutoRenderingHandler.js
```diff
@@ -0,0 +1,62 @@
...
+        executeDecisions(pageWideScopeDecisions);
+        executeCachedViewDecisions({
+          viewName,
+          viewDecisions: currentViewDecisions
+        });
+        showContainers();
+
+        return {
+          get decisions() {
+            logger.warn("Decisions property is deprecated");
```
jfkhoury, line 34: Awesome. Should we add: `. Please use "propositions" instead.` ?

src/components/Personalization/createAutoRenderingHandler.js
jfkhoury: @ninaceban are we going to rename those props?

test/unit/specs/components/Personalization/createAutoRenderingHandler.spec.js
```diff
@@ -0,0 +1,97 @@
...
+    expect(executeDecisions).toHaveBeenCalledWith(pageWideScopeDecisions);
+    expect(executeCachedViewDecisions).toHaveBeenCalledWith({
+      viewName,
+      viewDecisions: CART_VIEW_DECISIONS
+    });
+    expect(showContainers).toHaveBeenCalled();
+
+    result.decisions.forEach(decision => {
+      if (decision.scope === "__view__" || decision.scope === viewName) {
+        expect(decision.rendered).toEqual(true);
```
jfkhoury, line 71: @ninaceban I agree with @Aaronius , I think we should change `rendered` to something more generic, like `hasAttemptedRendering` or whatever he was suggesting :)

src/components/Personalization/createAutoRenderingHandler.js
Aaronius: I like that idea. Let's use complete sentences as well, please. Something like `The decisions property is deprecated. Please use the propositions property instead.`

src/components/Personalization/createExecuteCachedViewDecisions.js
```diff
@@ -14,10 +14,12 @@ import isNonEmptyArray from "../../utils/isNonEmptyArray";
 
 export default ({ executeViewDecisions, collect }) => {
   return ({ viewName, viewDecisions }) => {
+    // if there are viewDecisions for current view we will execute them and then send the collect call
     if (isNonEmptyArray(viewDecisions)) {
       executeViewDecisions(viewDecisions);
-      return;
+      return; // return here is to avoid the following code to be executed, that one is meant for the condition when viewDecisions is empty
```
Aaronius, line 20: Hey @ninaceban, I was hoping for more of the _why_ and less of the _what_. I think the _why_ should answer the question, "_Why_ don't we want to send an event when there are view decisions?" For example, is it because we're sending events inside of `executeViewDecisions` in that case so we don't need to send an event here?

src/components/Personalization/createNonRenderingHandler.js
```diff
@@ -1,40 +1,54 @@
...
-        ...redirectDecisions,
-        ...pageWideScopeDecisions,
-        ...formBasedComposedDecisions
-      ]
-    };
+    return Promise.resolve(propositions)
+      .then(items =>
+        getViewPropositions({ viewCache, viewName, propositions: items })
+      )
+      .then(items => buildFinalResult({ logger, propositions: items }));
```
Aaronius, line 52: Can we shorten this to:

```
return getViewPropositions({ viewCache, viewName, propositions })
  .then(items => buildFinalResult({ logger, propositions: items }));
```

src/components/Personalization/createAutoRenderingHandler.js
Aaronius: Also, let's de-dupe this string into a shared constant as well, since it's used in a three different places and doesn't minify.

src/components/Personalization/groupDecisions.js
```diff
@@ -104,8 +104,6 @@ const getGroupedDecisions = unprocessedDecisions => {
   };
 };
 
-export default {
-  groupDecisions(decisions) {
-    return getGroupedDecisions(decisions);
-  }
+export default decisions => {
+  return getGroupedDecisions(decisions);
 };
```
Aaronius, line 109: Preferably rename `getGroupedDecisions` to `groupDecisions` and then just shorten this to `export default groupDecisions;`.

src/components/Personalization/utils/addRenderToExecutedDecisions.js
```diff
@@ -0,0 +1,17 @@
...
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+import { assign } from "../../../utils";
+
+export default decisions => {
+  return decisions.map(decision => assign({ rendered: true }, decision));
+};
```
Aaronius, line 17: This module should probably be named consistently with whatever the property ends up being named. Right now the property is named `rendered`, which means the module should probably be named `addRenderedToExecutedDecisions` rather than `addRenderToExecutedDecisions`.

Aaronius:
@ninaceban, for consistency, I think the `rendered` property (or whatever it ends up being named) should be on every proposition returned to the customer. It will have a value of `true` for some propositions and `false` for others. Currently, the property only exists on the proposition if the value is `true`.

ninaceban:
> @ninaceban, for consistency, I think the `rendered` property (or whatever it ends up being named) should be on every proposition returned to the customer. It will have a value of `true` for some propositions and `false` for others. Currently, the property only exists on the proposition if the value is `true`.

@Aaronius  I was under the impression that we should add the `rendered` flag only when `renderDecisions` is set to `true`. If PMs think that we should set this flag all the time to all type of decisions we can address it later.

src/components/Personalization/utils/addRenderToExecutedDecisions.js
ninaceban: Agree, but let's decide on the name first!

src/components/Personalization/groupDecisions.js
ninaceban: ok, I will rename it. Thanks

src/components/Personalization/createExecuteCachedViewDecisions.js
ninaceban: Ok, I will change the comment

test/unit/specs/components/Personalization/createAutoRenderingHandler.spec.js
ninaceban: ok, let's decide during our checkin meeting.

src/components/Personalization/createAutoRenderingHandler.js
ninaceban: ok, will address the changes

Aaronius:
After discussion amongst the team during Tech check-in, we've made some additional decisions/clarifications:

- The boolean property on each proposition that indicates whether the web SDK has attempted to render the proposition should be renamed from rendered to renderAttempted. This more appropriately reflects that the web SDK has attempted to render the proposition, but not that it's necessarily complete or whether it will succeed or fail.
- For propositions within the propositions array returned from a sendEvent command, each proposition will always have a renderAttempted property. Sometimes the value may be true and other times it will be false, but it will always exist. This provides a more predictable API since the property can always be relied on existing.

src/components/Personalization/decisionsExtractor.js
```diff
@@ -75,11 +77,35 @@ const extractDecisionsByScope = (decisions, scope) => {
...
   }
```
jonsnyder: Since there is only one function that you are returning, you could change this filename to groupDecisions.js, and just export the one function.

src/components/Personalization/createAutoRenderingHandler.js
jonsnyder: I agree. I now know that "formBased" => non auto renderable and "domAction" => auto renderable. I think it makes things more clear to use renderable and nonRenderable.

jonsnyder (approved):

src/components/Personalization/decisionsExtractor.js
ninaceban: yes, it was already renamed to groupDecisions.

src/components/Personalization/constants/loggerMessages.js
```diff
@@ -0,0 +1,5 @@
+export const decisionsDeprecatedWarning =
+  "The decisions property is decisionsDeprecatedWarning. Please use the propositions property instead.";
```
Aaronius, line 2: I don't think you ended to have `decisionsDeprecatedWarning` in this text.

src/components/Personalization/constants/loggerMessages.js
```diff
@@ -0,0 +1,5 @@
...
+export const authoringEnabled = "Rendering is disabled, authoring mode.";
```
Aaronius, line 3: Maybe `Rendering is disabled for authoring mode.` instead.

src/components/Personalization/constants/loggerMessages.js
```diff
@@ -0,0 +1,5 @@
...
+export const redirectExecutionError =
+  "An error occurred while executing the redirect offer.";
```
Aaronius, line 5: For consistency, follow [these conventions](https://github.com/adobe/alloy/wiki/Conventions#enumerations), specifically using all caps for the constant names and changing the filename to `loggerMessage.js`

Aaronius (approved):

PR #749, opened 2021-07-28
PDCL-6554: guard against spreading non iterable values

jfkhoury:
## Description

If `currentViewDecisions` is undefined, trying to spread it will throw an exception.

## Types of changes

## Checklist:


ninaceban:
@jfkhoury looks good to me, but I would fix the cause. The cause is because we returned in viewCache `viewStorage[viewName]` which means if there are no decisions for a specific view we would get `undefined`. I propose to return a empty array. Here is the PR: https://github.com/adobe/alloy/pull/750

ninaceban (approved):

PR #750, opened 2021-07-29
PDCL-6554 - return empty array when no view decisions for a specific view

ninaceban:
## Description
Because the viewCache promise gets resolved with `undefined`.  `undefined` would've been passed to handlers, in some handlers we would `currentViewDecisions || []`.  I would make `viewCache` return empty array when there are no decisions for a specific view, rather than returning `undefined` and then have to check for `undefined` everywhere we trigger `viewCache.getView`.

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Personalization/createViewCacheManager.js
+++ b/src/components/Personalization/createViewCacheManager.js
@@ -35,7 +35,7 @@ export default () => {
   };
 
   const getView = viewName => {
-    return viewStorageDeferred.promise.then(() => viewStorage[viewName]);
+    return viewStorageDeferred.promise.then(() => viewStorage[viewName] || []);
   };
 
   const isInitialized = () => {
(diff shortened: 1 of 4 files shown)


jfkhoury (approved):
👏🏻 
Nina, is there a test we can add? Can we also search for similar situations in the library and make sure they are covered?

ninaceban:
> 👏🏻
> Nina, is there a test we can add? Can we also search for similar situations in the library and make sure they are covered?

yes, I have added some checks in the current tests and have removed another guard from `createViewChangeHandler.js`.

jfkhoury (approved):

Aaronius (approved):

shammowla (approved):

jonsnyder (approved):
Let's make sure we add a functional test to make sure we don't run into this bug again.

PR #760, opened 2021-08-12
PDCL-4508 Resolve promise with empty object from configure command.

Aaronius:
## Description
Previously, when a user executed the `configure` command, the returned promise would be resolved with the component registry. It should have been resolved with an empty object, since the component registry isn't intended to be exposed.

## Related Issue
https://jira.corp.adobe.com/browse/PDCL-4508

## Motivation and Context
We shouldn't be exposing internal details to the user.

## Types of changes

## Checklist:

### DIFF
--- a/test/unit/specs/core/injectExecuteCommand.spec.js
+++ b/test/unit/specs/core/injectExecuteCommand.spec.js
@@ -210,9 +210,12 @@ describe("injectExecuteCommand", () => {
   });
 
   it("executes the core commands", () => {
+    const componentRegistry = {
+      getCommand() {}
+    };
     const configureCommand = jasmine
       .createSpy()
-      .and.returnValue(Promise.resolve("configureResult"));
+      .and.returnValue(Promise.resolve(componentRegistry));
     const setDebugCommand = jasmine.createSpy();
     const executeCommand = injectExecuteCommand({
       logger,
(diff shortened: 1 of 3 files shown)


test/unit/specs/core/injectExecuteCommand.spec.js
```diff
@@ -210,9 +210,12 @@ describe("injectExecuteCommand", () => {
   });
 
   it("executes the core commands", () => {
+    const componentRegistry = {
+      getCommand() {}
+    };
     const configureCommand = jasmine
       .createSpy()
-      .and.returnValue(Promise.resolve("configureResult"));
+      .and.returnValue(Promise.resolve(componentRegistry));
```
Aaronius, line 218: This test was erroneously passing previously because this mock was resolving the promise with a string rather than an object. We have logic in Alloy that says if the command provides an object, return it, but if the command provides something else, replace it with an empty object: https://github.com/adobe/alloy/blob/c43cdd3be1637f8671f9cf660c1da71249d23494/src/core/injectExecuteCommand.js#L107.

We needed to more closely align our mock with reality in order for the faulty behavior to be exhibited.

jonsnyder (approved):

jfkhoury (approved):
Clean up clean up 🧹

PR #770, opened 2021-09-09
PDCL-6797: Return edge destinations in sendEvent.

jfkhoury:
PLEASE DON'T MERGE THIS PR!

## Description

We are going to return the handles of type `activation:pull` in the sendEvent promise. This design might not be final, but we need it for a beta program. Please don't merge since we will be having more design conversations around returning additional handles from the response in the sendEvent command, without changing the SDK.

## Related Issue

https://jira.corp.adobe.com/browse/PDCL-6797

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Audiences/index.js
+++ b/src/components/Audiences/index.js
@@ -12,19 +12,19 @@ governing permissions and limitations under the License.
 
 import { fireReferrerHideableImage } from "../../utils";
 import injectProcessDestinations from "./injectProcessDestinations";
+import injectProcessResponse from "./injectProcessResponse";
 
 const createAudiences = ({ logger }) => {
   const processDestinations = injectProcessDestinations({
     fireReferrerHideableImage,
     logger
   });
-  const processDestinationsFromResponse = ({ response }) => {
-    const destinations = response.getPayloadsByType("activation:push");
-    return processDestinations(destinations);
-  };
+
+  const processResponse = injectProcessResponse({ processDestinations });
+
   return {
     lifecycle: {
-      onResponse: processDestinationsFromResponse
+      onResponse: processResponse
     },
     commands: {}
   };
--- a/src/components/Audiences/injectProcessResponse.js
+++ b/src/components/Audiences/injectProcessResponse.js
@@ -28,7 +28,7 @@ export default ({ processDestinations }) => {
     const resolveValue = isNonEmptyArray(destinations)
       ? { destinations }
       : undefined;
-    return Promise.resolve(resolveValue);
+    return resolveValue;
   };
 
   return ({ response }) => {
--- a/src/components/Audiences/injectProcessResponse.js
+++ b/src/components/Audiences/injectProcessResponse.js
@@ -10,25 +10,16 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import { isNonEmptyArray } from "../../utils";
-
 export default ({ processDestinations }) => {
   const processPushDestinations = ({ response }) => {
     const destinations = response.getPayloadsByType("activation:push");
     return processDestinations(destinations);
   };
 
-  // IMPORTANT: We are breaking our own interface here, and not returning
-  // a handle if it's not available in the response. We are doing this because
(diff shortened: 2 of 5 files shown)


src/components/Audiences/index.js
```diff
@@ -18,13 +18,27 @@ const createAudiences = ({ logger }) => {
...
+    );
+  };
+
   return {
     lifecycle: {
-      onResponse: processDestinationsFromResponse
+      onResponse: processResponse
     },
     commands: {}
   };
```
jfkhoury, line 44: I'm starting to think we should rename this Component to `Activation` instead of `Audiences` since it's broader.

jfkhoury:
I will update the Sandbox as soon as our ORGs have been provisioned for Edge Destinations.

test/unit/specs/components/Audiences/index.spec.js
```diff
@@ -0,0 +1,46 @@
+/*
```
jonsnyder, line 1: Generally, the problem with writing unit tests for the index.js files is that its impossible to inject mocks for dependencies. The index files do the injecting. In this case you aren't able to mock the processDestinations function, and indeed it is getting imported and called. The index.js files should just be wiring up all the dependencies together and shouldn't be doing any work.  I recommend creating a "injectProcessResponse.js" file like this:

```
export default ({ processDestinations }) => {

  const processDestinationsFromResponse = ({ response }) => {
     const destinations = response.getPayloadsByType("activation:push");
     return processDestinations(destinations);
   };

   const retrievePullDestinations = ({ response }) => {
     return Promise.resolve({
       destinations: response.getPayloadsByType("activation:pull")
     });
   };

   return ({ response }) => {
     return processDestinationsFromResponse({ response }).then(() =>
       retrievePullDestinations({ response })
     );
   };
};
```

Then index.js looks like this:

```
import { fireReferrerHideableImage } from "../../utils";
import injectProcessDestinations from "./injectProcessDestinations";
import injectProcessResponse from "./injectProcessResponse";

const createAudiences = ({ logger }) => {
  const processDestinations = injectProcessDestinations({
     fireReferrerHideableImage,
     logger
   });

   const processResponse = injectProcessResponse({ processDestinations });

   return {
     lifecycle: {
       onResponse: processResponse
     },
     commands: {}
   };
};

```

Then you can move this file to injectProcessResponse.spec.js and create a mock for processDestinations. You can ensure that processDestinations is called correctly, and that "activation:pull" handles are returned.

jonsnyder (approved):

jonsnyder:
It would be great to add a functional test too.

test/unit/specs/components/Audiences/index.spec.js
jfkhoury: That's how you know I'm terribly rusty :) I'm usually obsessed with dependency injection; not anymore! :) :)

src/components/Audiences/index.js
jfkhoury: I will create a ticket for this

src/components/Audiences/injectProcessResponse.js
```diff
@@ -0,0 +1,39 @@
...
+  // a handle if it's not available in the response. We are doing this because
+  // a small group of customers need the edge destinations for a beta program.
+  // This behavior along with the `destinations` variable name will most probably
+  // change in the near future.
+  const retrievePullDestinations = ({ response }) => {
+    const destinations = response.getPayloadsByType("activation:pull");
+    const resolveValue = isNonEmptyArray(destinations)
+      ? { destinations }
+      : undefined;
+    return Promise.resolve(resolveValue);
```
jonsnyder, line 31: No need to wrap this in a Promise.resolve. You can just return resolveValue.

jonsnyder (approved):

liljenback (approved):

jonsnyder:
This was released as https://cdn1.adoberesources.net/alloy/edge-destinations-beta/2.7.0-beta.0/alloy.js

Further changes necessary for the beta should be added as PRs against the edge-destinations-beta branch.

PR #773, opened 2021-10-11
TNT-42028 Support Shadow DOM selectors

XDex:
## Description

This PR adds support for shadow DOM selectors in delivered Target offers, for proper rendering of Target offers on custom elements with shadow roots.

## Related Issue

https://jira.corp.adobe.com/browse/TNT-42028

## Motivation and Context

This PR adds support for shadow DOM selectors in delivered Target offers, for proper rendering of Target offers on custom elements with shadow roots.
More details in Shadow DOM spec: https://wiki.corp.adobe.com/x/xVOOj

## Types of changes

## Checklist:


src/components/Personalization/dom-actions/dom/selectNodesWithShadow.js
```diff
@@ -0,0 +1,37 @@
...
+
+  if (parts.length < 2) {
+    return querySelectorAll(context, selector);
+  }
+
+  // split the selector into parts separated by :shadow pseudo-selectors
+  // find each subselector element based on the previously selected node's shadowRoot
+  let parent = context;
+  for (let i = 0; i < parts.length; i += 1) {
+    const part = cleanupPrefix(parts[i]);
```
Aaronius, line 27: I suspect I may be misunderstanding something here, but couldn't this result in invalid matches? For example, given the selector `BUY-NOW-BUTTON:nth-of-type(2):shadow > .logo`, it seems stripping the `>` would result in us looking for `.logo` at any descendent level within the buy now button rather than a direct descendent.

src/components/Personalization/dom-actions/dom/helperForShadow.js
```diff
@@ -0,0 +1,25 @@
...
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+const SHADOW_SEPARATOR = ":shadow";
+
+export const isNotShadowSelector = str => str.indexOf(SHADOW_SEPARATOR) === -1;
```
Aaronius, line 15: I recommend not using a negative. Instead, make it a positive (`isShadowSelector`) and let the consuming module determine whether to negate it.

XDex: Here I was also guided by existing code in `helperForEq.js` where we have `isNotEqSelector`. I'll fix it in both places then, per your recommendation..

src/components/Personalization/dom-actions/dom/selectNodesWithShadow.js
XDex: The polyfill I mentioned above won't work, as there's no way to set an attribute on a `shadowRoot`. Anyhow, turns out custom elements and Shadow DOM aren't supported in IE in any case, so there's no need to also add support for `:scope`, since pages with custom elements and Shadow DOM will be broken in IE anyway.

src/components/Personalization/dom-actions/dom/helperForShadow.js
Aaronius: Thank you!

XDex:
@Aaronius I've pushed some fixes, please take another look when you get a chance, thank you!

src/components/Personalization/dom-actions/dom/helperForShadow.js
```diff
@@ -0,0 +1,25 @@
...
+
+export const splitWithShadow = selector => {
+  return selector.split(SHADOW_SEPARATOR);
+};
+
+export const cleanupPrefix = selector => {
+  const result = selector.trim();
+  const subselIdx = result.indexOf(">");
+  return subselIdx === 0 ? result.substring(1).trim() : result;
+};
```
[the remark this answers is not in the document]
Aaronius: Sorry, I wasn't aware of the `helperForEq.js` precedent. Moving these particular helpers back to `selectNodesWithShadow` is good though IMO. At this point, I would recommend not exporting or unit testing `splitWithShadow` or `transformPrefix` directly, since those are implementation details that should remain private to `selectNodesWithShadow.js`. Instead, just test the default export of `selectNodesWithShadow.js` in a way that you are sure `splitWithShadow` and `transformPrefix` are correctly performing their job. By not exporting or testing `splitWithShadow` or `transformPrefix` directly, our tests become more resilient to change. In other words, we'll spend less time updating our tests if we need to refactor those functions down the road.

`isShadowSelector` is the oddball here, since that function is reasonably used both internally within `selectNodesWithShadow.js` and in `selectNodes.js`. I didn't realize this when I suggesting moving the functions into `selectNodesWithShadow.js`. It's probably fine where you have it.

src/components/Personalization/dom-actions/dom/selectNodesWithShadow.js
```diff
@@ -27,6 +27,11 @@ export const transformPrefix = selector => {
 export const isShadowSelector = str => str.indexOf(SHADOW_SEPARATOR) !== -1;
 
 export default (context, selector) => {
+  // Shadow DOM is not supported in IE
+  if (window.document.documentMode) {
```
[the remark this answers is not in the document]
jfkhoury: @XDex first of all, great work; and thank you for your detailed answers and patience.  Is shadow dom supported in the VEC? If yes, can we add a functional test to this PR? If no, why are we making this change before we support it in the VEC?

thank you!

Aaronius (approved):
Thanks for the good work on this, @XDex, and, as @jfkhoury said, for your patience with our feedback. The last thing I would still suggest is not exporting `transformPrefix` or `splitWithShadow` since they are implementation details of the `selectNodesWithShadow` function. Instead, just test that their logic is behaving properly when testing the actual `selectNodesWithShadow` function.

Also, if you plan on contributing in the future, I'd recommend requesting write access to the repo. You can then do the work on a branch of the main repo rather than having to do the work on your fork (we encourage you to work on the main repo if you're a regular contributor). Instructions for requesting write access is documented [here](https://github.com/adobe/alloy/wiki#requesting-write-access-adobe-employees).

Thank you!

XDex:
@jfkhoury Alexandru Dan Popa from the UI team just let me know the Shadow DOM mods in VEC didn't make into today's UI release and will go out in a week, on Oct 26/27th, would you be ok with merging this PR now? I'll raise another one with an e2e functional test after the release happens next week..

XDex:
@Aaronius @jonsnyder Could you please help with reviewing #780 and uploading the updated alloyio.com sandbox to Azure (as I don't have access)? I need this deployed, before I can add a functional test.
Thanks!

src/components/Personalization/dom-actions/dom/helperForEq.js
```diff
@@ -15,7 +15,7 @@ import { isNonEmptyString } from "../../../../utils";
 const EQ_START = ":eq(";
 const EQ_PATTERN = /:eq\((\d+)\)/g;
 
-export const isNotEqSelector = str => str.indexOf(EQ_START) === -1;
+export const isEqSelector = str => str.indexOf(EQ_START) !== -1;
```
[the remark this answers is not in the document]
XDex: I tend to agree with Aaron on this, as keeping boolean names affirmative is a good practice:
 - "Use positive boolean variable names" -- Code Complete, page 269
 - https://softwareengineering.stackexchange.com/a/196837
 - https://www.serendipidata.com/posts/naming-guidelines-for-boolean-variables
However, as discussed during the last sync up with the team, I'll revert only the `isEqSelector` change, which is not directly related to this PR..

XDex:
@jfkhoury @jonsnyder I've added the functional test, please take another look. Thank you!

package.json
```diff
@@ -19,6 +19,7 @@
     "test:unit:coverage": "karma start --single-run --reporters spec,coverage",
     "test:functional": "EDGE_BASE_PATH=\"ee-pre-prd\" ALLOY_ENV=\"int\" testcafe chrome",
     "test:functional:watch": "EDGE_BASE_PATH=\"ee-pre-prd\" ALLOY_ENV=\"int\" ./scripts/watchFunctionalTests.js --browsers chrome",
+    "test:functional:debug": "EDGE_BASE_PATH=\"ee-pre-prd\" ALLOY_ENV=\"int\" testcafe --inspect-brk chrome -T \"Test C28758.*\"",
```
ninaceban, line 22: I guess this is a non-intentional change. You could also change inside functional test `test.only` and run `test:functional:watch`, this will run only that specific test in watch mode.

ninaceban (changes requested):
@XDex thank you! I have only one small change request otherwise looks good!

package.json
XDex: @ninaceban No, this is actually an intentional change, as with the Testcafe Live Runner used in the `watch` script, there's no way to start a proper Inspector debug server, so that the test can be debugged directly in the IDE.

XDex: @ninaceban I have also moved the test focusing to this external `debug` script, instead of using `test.only`, as ESLint warns you not to do it: `ESLint: don't focus tests(ban/ban)` as it's easier to forget to remove it once you set it in the test, which is not an issue when you have a separate debug script.

ninaceban: As far as I can understand this is a debug configuration only for one test case. You could have this configured locally without this being committed to the repo.

test/functional/specs/Personalization/C28758.js
```diff
@@ -0,0 +1,144 @@
...
+  ID: "C28758",
+  SEVERITY: "P0",
+  TEST_RUN: "Regression"
+});
+
+const shadowDomScript = `
+    const buyNowContent = \`
+    <div>
+      <input type="radio" id="buy" name="buy_btn" value="Buy NOW">
+      <label for="buy">Buy Now</label><br>
```
ninaceban, line 37: Could you please extract these to a `/Personalization/constants` or in `functional/helpers/constants/`, to keep only the test code here.

test/functional/helpers/createFixture/clientScripts.js
```diff
@@ -74,7 +74,7 @@ const getProdNpmLibraryCode = () => {
   return readCache.sync(prodNpmLibraryPath, "utf8");
 };
 
-const injectInlineScript = ClientFunction(code => {
+export const injectInlineScript = ClientFunction(code => {
```
ninaceban, line 77: should we extract this explicitly in a test util function? 
cc: @jonsnyder

package.json
XDex: In addition to the reasons I mentioned above, the idea is to commit it to the repo so that there's an explicit script for debugging a single functional test, and any new contributors don't have to rely on tacit knowledge on how to debug a single test, losing time for looking up how to focus a single test in our particular testing framework, but can instead do it quickly just taking a look at the script and adjusting it to their needs.

test/functional/helpers/createFixture/clientScripts.js
XDex: I don't think extracting it to a separate test util is warranted, since it's already used in other `clientScripts.js` functions, like `injectAlloyDuringTestForInt()`

jonsnyder: Good question Nina. I probably would have added a parameter to "createFixture" for extra scripts to include. I think this is fine though. There are already two exports for clientScripts.js.

jonsnyder (approved):

package.json
XDex: @jonsnyder Fixed as suggested and added instructions to the Wiki

ninaceban (approved):

PR #777, opened 2021-10-22
qa mode persistence

jasonwaters:
## Description

In at.js QA mode begins when certain url params are present ( `at_preview_token`, `at_preview_index`, `at_preview_listed_activities_only`).  These values are included in the request to target edge and tell target to override some decision making for QA testing.  at.js also sets a cookie (`at_qa_mode`) that persists these values for a short time.  This makes it possible for customers to navigate their site while testing  even if the url params are lost during site navigation.  

alloy.js does not persist the qa params in this way, making it frustrating for customers testing a site using alloy and navigating around a bit.  

This PR closes the gap by setting a cookie that matches the one at.js uses.  There are benefits in matching the cookie at.js reads/sets because it means that customers migrating from at.js will be able to rely on qa mode persisting across their (same-domain) web properties regardless of whether at.js or alloy.js is in use.  However, if there is objection to the approach, we can rename the cookie or persist it in some other way.

Related PR: [Changes to target-upstream to support qa mode in personalization query](https://git.corp.adobe.com/TnT/target-upstream/pull/55)

## Related Issue

- [TNT-42144 QA URLs not working with AEP Web SDK ](https://jira.corp.adobe.com/browse/TNT-42144)

## Types of changes

- [X] Improvement (non-breaking change which does not add functionality)

## Checklist:


jfkhoury:
Hey @jasonwaters , I will go ahead and close this PR per our last discussion, and look forward to adding this logic into the Debugger. Any plans on making this change anytime soon? Thanks.

jasonwaters:
@jfkhoury I'm not sure how this work will be prioritized with our other target initiatives.  I'm sure David will let us know!

PR #783, opened 2021-11-16
TNT-43002 Fix duplicate inline scripts added to DOM

XDex:
## Description

When HTML offers with inline SCRIPT tags are rendered, these are added twice to page DOM, even though the scripts are actually executed just once on the page. We should fix it, so that no unexpected inline SCRIPT tags are added to page DOM after Personalization offers are applied.

## Related Issue

https://jira.corp.adobe.com/browse/TNT-43002

## Motivation and Context

When HTML offers with inline SCRIPT tags are rendered, these are added twice to page DOM, even though the scripts are actually executed just once on the page. We should fix it, so that no unexpected inline SCRIPT tags are added to page DOM after Personalization offers are applied.

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Personalization/dom-actions/scripts.js
+++ b/src/components/Personalization/dom-actions/scripts.js
@@ -80,8 +80,11 @@ export const getRemoteScriptsUrls = fragment => {
   return result;
 };
 
-export const executeInlineScripts = (container, scripts, func) => {
-  scripts.forEach(script => func(container, script));
+export const executeInlineScripts = (parent, scripts) => {
+  scripts.forEach(script => {
+    parent.appendChild(script);
+    parent.removeChild(script);
+  });
 };
 
 export const executeRemoteScripts = urls => {
--- a/test/unit/specs/components/Personalization/dom-actions/setHtml.spec.js
+++ b/test/unit/specs/components/Personalization/dom-actions/setHtml.spec.js
@@ -2,6 +2,8 @@ import { appendNode, createNode } from "../../../../../../src/utils/dom";
 import { initDomActionsModules } from "../../../../../../src/components/Personalization/dom-actions";
 import cleanUpDomChanges from "../../../../helpers/cleanUpDomChanges";
 
+const sleep = ms => new Promise(resolve => setTimeout(resolve, ms));
+
 describe("Personalization::actions::setHtml", () => {
   beforeEach(() => {
     cleanUpDomChanges("setHtml");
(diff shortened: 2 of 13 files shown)


src/components/Personalization/dom-actions/scripts.js
```diff
@@ -80,8 +80,11 @@ export const getRemoteScriptsUrls = fragment => {
   return result;
 };
 
-export const executeInlineScripts = (container, scripts, func) => {
-  scripts.forEach(script => func(container, script));
+export const executeInlineScripts = (parent, scripts) => {
+  scripts.forEach(script => {
+    parent.appendChild(script);
+    parent.removeChild(script);
```
jonsnyder, line 86: How can you be sure the inline script is done executing? Do scripts that attach event handlers still work?

test/unit/specs/components/Personalization/dom-actions/setHtml.spec.js
```diff
@@ -49,13 +49,16 @@ describe("Personalization::actions::setHtml", () => {
...
       meta
     };
     const event = { elements };
 
     return setHtml(settings, event).then(() => {
       expect(window.someEvar123).toEqual(1);
+
+      const scriptElements = document.querySelectorAll("#evar123");
+      expect(scriptElements.length).toEqual(1);
     });
```
jonsnyder, line 62: Can you add a test to see if an event handler defined in the inlineScript still works? And what if you have a setTimeout in the inline script?

jonsnyder (changes requested):
Can you add an integration test?

XDex:
> Can you add an integration test?

@jonsnyder I don't think an integration test is necessary, as we're only changing the rendering part here, which can be fully tested via unit tests, which I've added.

Note that I've also added some changes to unit test runner config, so async/await is now supported in unit tests.

jfkhoury:
@XDex Thank you did the changes to the unit tests and for making this change; a cleaner approach even if we weren’t double executing those script tags! 

I’m not sure how the unit test is verifying that the script is being removed. I might have missed it. 

One last thing, If @jonsnyder believes an integration test is helpful, let’s add one please; being extra safe never hurts, especially with those complicated Personalization features. No rush on this one since we can’t release anything anyway. 

Thanks again, it’s a pleasure to have you contribute to the project.

XDex:
@jfkhoury Thanks! I'll add a functional test as you suggest..
> I’m not sure how the unit test is verifying that the script is being removed. I might have missed it.

This is checked at line 62 in the test: `    expect(scriptElements.length).toEqual(1);`

ninaceban:
@XDex we have discussed about Alloy vs at.js in the past. Just because at.js does something in particular way, it doesn't mean that Alloy must follow suit.
I would argue that having scripts being part of the DOM both during `innerHTML` operation and after `appendChild` allows us to clearly see which DOM operation contained scripts and must be "treated" a little bit differently and it helps with debugging.

I would let @jfkhoury and @jonsnyder share their thoughts, but I think just documenting the behavior should be enough. We don't need to add more complexity to the existing DOM manipulation code.

XDex:
@ninaceban I have provided the motivation for this change in the linked [Jira ticket](https://jira.corp.adobe.com/browse/TNT-43002) comments, and I think @jfkhoury agrees that we should have this fixed.
@jfkhoury @jonsnyder Please take another look, I've also added an integration test.. Thank you!

jonsnyder (approved):
Thanks for adding the unit tests!

jfkhoury (approved):
Thank you for the tests.

PR #785, opened 2021-11-22
TNT-42028 Support Shadow DOM selectors (updated)

XDex:
## Description

Previously merged TNT-42028 Support Shadow DOM selectors #773 was reverted as the tests were failing in Firefox and Safari.
This PR fixes the Shadow DOM support implementation, so it works across all browsers (except IE, which is expected).

## Related Issue

https://jira.corp.adobe.com/browse/TNT-42028
https://jira.corp.adobe.com/browse/TNT-43043

## Motivation and Context

Previously merged TNT-42028 Support Shadow DOM selectors #773 was reverted as the tests were failing in Firefox and Safari.

Applying of offers with Shadow DOM :shadow selectors only worked in Chrome/Edge, and didn't work in Firefox/Safari - although the `:scope` pseudo-class which we're using in the implementation is supposed to be supported in FF/Safari - https://caniuse.com/mdn-css_selectors_scope_dom_api it isn't actually working as expected in FF and Safari, not matching any child nodes when `querySelectorAll` is called on a shadow root.

I have opened a  bug report with WebKit - https://bugs.webkit.org/show_bug.cgi?id=233380 and reopened an older bug in Firefox(Gecko) - https://bugzilla.mozilla.org/show_bug.cgi?id=1689893

Based on the latest discussion there, it turns out that `:scope` is not supposed to be working when called on a shadow tree, instead it is the Chrome/Edge support of this that is unexpected, and thus a bug was opened with Chromium - https://bugs.chromium.org/p/chromium/issues/detail?id=1272434

Thankfully, there's an alternate solution for selecting direct descendants of the shadow root using `:host` pseudo-class, which I've confirmed to be working across all browsers (except IE, which is expected).

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Personalization/dom-actions/dom/selectNodesWithShadow.js
+++ b/src/components/Personalization/dom-actions/dom/selectNodesWithShadow.js
@@ -1,0 +1,70 @@
+/*
+Copyright 2021 Adobe. All rights reserved.
+This file is licensed to you under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License. You may obtain a copy
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+import querySelectorAll from "./querySelectorAll";
+import { startsWith } from "../../../../utils";
+
+const SHADOW_SEPARATOR = ":shadow";
+
+const splitWithShadow = selector => {
+  return selector.split(SHADOW_SEPARATOR);
+};
+
+const transformPrefix = (parent, selector) => {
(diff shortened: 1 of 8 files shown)


jonsnyder (approved):

src/components/Personalization/dom-actions/dom/selectNodesWithShadow.js
```diff
@@ -0,0 +1,70 @@
...
+
+  // IE doesn't support :scope
+  if (window.document.documentMode) {
+    return result.substring(1).trim();
+  }
+
+  const prefix =
+    parent instanceof Element || parent instanceof HTMLDocument
+      ? ":scope"
+      : ":host"; // see https://bugs.webkit.org/show_bug.cgi?id=233380
```
jfkhoury, line 37: Thank you for the inline comments!

jfkhoury (approved):
Great work. @XDex Do we need to update our docs to mention this feature?

XDex:
@jfkhoury I think it's ok to just have this mentioned in release notes, as it's mainly a Target UI feature and we're just adding support for it in client SDKs.

PR #794, opened 2022-01-18
TNT-43421 Fix circular dependency in Web SDK

XDex:
## Description

Fix circular dependency introduced after ShadowDOM implementation.

## Related Issue

https://jira.corp.adobe.com/browse/TNT-43421

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Improvement (non-breaking change which does not add functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have signed the [Adobe Open Source CLA](https://opensource.adobe.com/cla.html) or I'm an Adobe employee.
- [x] I have made any necessary test changes and all tests pass.
- [x] I have run the Sandbox successfully.

PR #795, opened 2022-01-19
TNT-43421 Fix circular dependency in Web SDK (alternate implementation)

jonsnyder:
## Description

Fix circular dependency introduced after ShadowDOM implementation. This is an alternate implementation to #794.

The issue is that src/utils/dom/selectNodes.js was importing querySelectorAll and selectNodesWithShadow from the personalization component. Utils should not import files from components. I've added a lint rule to make sure imports are not done from src/utils to src/components.

I considered going the other way and moving selectNodes into the personalization components, but selectNodes is used by awaitSelector, and awaitSelector is used by fireReferrerHideableImage, which is used by Audiences and Identity.

I also separated isShadowSelector so that each util is only exporting one thing.

## Motivation and Context

https://jira.corp.adobe.com/browse/TNT-43421

I'd like to include this in the 2.8.0 release because it would throw warnings for customers using the NPM version.

## Types of changes

## Checklist:

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -25,7 +25,45 @@ module.exports = {
     // want to disallow importing extraneous dependencies.
     "import/no-extraneous-dependencies": "off",
     "prefer-destructuring": "off",
-    "import/prefer-default-export": "off"
+    "import/prefer-default-export": "off",
+    // Make rules about importing between the top level folders
+    // core can import from components, utils, and constants
+    // components can import from utils, and constants
+    // utils can import from constants
+    "import/no-restricted-paths": [
+      "error",
+      {
+        zones: [
+          // TODO: Remove dependencies from components into core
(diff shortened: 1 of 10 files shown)


.eslintrc.js
```diff
@@ -25,7 +25,45 @@ module.exports = {
...
+    "import/prefer-default-export": "off",
+    // Make rules about importing between the top level folders
+    // core can import from components, utils, and constants
+    // components can import from utils, and constants
+    // utils can import from constants
+    "import/no-restricted-paths": [
+      "error",
+      {
+        zones: [
+          // TODO: Remove dependencies from components into core
```
jonsnyder, line 37: https://jira.corp.adobe.com/browse/PDCL-7517

jfkhoury (approved):

PR #797, opened 2022-01-24
Sandbox app: A4T client side logging example 

ninaceban:
I have changed the `organisationTwo` edge configuration to have Analytics disabled, which means A4T Logging is client -side.

## Types of changes

## Checklist:

### DIFF
--- a/sandbox/public/index.html
+++ b/sandbox/public/index.html
@@ -12,7 +12,7 @@
     <script src="https://cdn.jsdelivr.net/npm/promise-polyfill@8/dist/polyfill.min.js"></script>
 
     <title>Mock website hosting Alloy</title>
-    <script src="perf/js/onBeforeEventSentHelper.js" />
+    <script src="onBeforeEventSentHelper.js" />
 
     <!-- prettier-ignore -->
     <script nonce="%REACT_APP_NONCE%">
--- a/sandbox/public/index.html
+++ b/sandbox/public/index.html
@@ -12,8 +12,7 @@
     <script src="https://cdn.jsdelivr.net/npm/promise-polyfill@8/dist/polyfill.min.js"></script>
 
     <title>Mock website hosting Alloy</title>
-    <script src="perf/js/visitorapi.min.js"></script>
-    <script src="perf/js/AppMeasurement.js"></script>
+    <script src="perf/js/onBeforeEventSentHelper.js" />
 
     <!-- prettier-ignore -->
     <script nonce="%REACT_APP_NONCE%">
@@ -107,7 +106,27 @@
         edgeConfigId: "7984963a-6609-4e84-98d5-4e2ff8c0dd5e:prod",
         orgId: "97D1F3F459CE0AD80A495CBE@AdobeOrg",
         debugEnabled: true,
-        clickCollectionEnabled: false
+        clickCollectionEnabled: false,
+        onBeforeEventSend: function(options) {
+          const xdm = options.xdm;
+          const eventType = xdm.eventType;
+          if (eventType === "decisioning.propositionInteract") {
+            const analyticsPayloads = new Set();
+            const propositions = xdm._experience.decisioning.propositions;
+
+            for (var i = 0; i < propositions.length; i++) {
+              var proposition = propositions[i];
+              analyticsPayloads.add(getAnalyticsPayload(proposition));
+            }
+
+            getECID("organizationTwo").then(visitorID => {
+              const analyticsPayload = concatinateAnalyticsPayloads(
+                analyticsPayloads
+              );
+              triggerAnalyticsHit({ analyticsPayload, visitorID });
+            });
+          }
+        }
       });
     </script>
   </head>
--- a/sandbox/public/index.html
+++ b/sandbox/public/index.html
@@ -12,7 +12,18 @@
     <script src="https://cdn.jsdelivr.net/npm/promise-polyfill@8/dist/polyfill.min.js"></script>
 
     <title>Mock website hosting Alloy</title>
+    <script src="perf/js/visitorapi.min.js"></script>
+    <script src="perf/js/AppMeasurement.js"></script>
 
+    <script>
+      var visitor = Visitor.getInstance("97D1F3F459CE0AD80A495CBE@AdobeOrg", {
+        disableIdSyncs: true
+      });
+      var s_account = "ujslecommerce";
+      var s = s_gi(s_account);
+      // Make sure data is sent to the correct location
+      s.trackingServer = "ujsl.sc.omtrdc.net";
+    </script>
     <!-- prettier-ignore -->
     <script nonce="%REACT_APP_NONCE%">
       !function(e,a,n,t){var i=e.head;if(i){
@@ -102,7 +113,7 @@
       // For Testing multiple instances.
       // We use a different orgId and edgeConfigId.
       organizationTwo("configure", {
-        edgeConfigId: "9999999",
+        edgeConfigId: "7984963a-6609-4e84-98d5-4e2ff8c0dd5e:prod",
         orgId: "97D1F3F459CE0AD80A495CBE@AdobeOrg",
         debugEnabled: true,
         clickCollectionEnabled: false
--- a/sandbox/public/index.html
+++ b/sandbox/public/index.html
@@ -120,10 +120,10 @@
             }
 
             getECID("organizationTwo").then(visitorID => {
-              const analyticsPayload = concatinateAnalyticsPayloads(
+              const analyticsPayload = concatenateAnalyticsPayloads(
                 analyticsPayloads
               );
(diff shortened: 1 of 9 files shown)


sandbox/public/index.html
```diff
@@ -12,7 +12,18 @@
...
 
+    <script>
+      var visitor = Visitor.getInstance("97D1F3F459CE0AD80A495CBE@AdobeOrg", {
+        disableIdSyncs: true
+      });
+      var s_account = "ujslecommerce";
+      var s = s_gi(s_account);
+      // Make sure data is sent to the correct location
+      s.trackingServer = "ujsl.sc.omtrdc.net";
+    </script>
```
jonsnyder, line 26: Why did you need Visitor on the page? For AppMeasurement to use Visitor, you need to do `window.s.visitor = visitor`. This is working now because you do `window.s.marketingCloudVisitorID = identity.identity.ECID;` which sets up the ECID for AppMeasurement.

To be clear there are two ways for AppMeasurement to get the same visitor id as Alloy:
1. Set `s.marketingCloudVisitorID` with the result from Alloy's getIdentity call. If you do it this way there is no need to put Visitor on the page.
2. Set `s.visitor = Visitor.getInstance(...)`, and update Alloy config to have idMigrationEnabled=true.

Of course, if we can't use appMeasurement anyways because we need `tnta` to work, we'll have to generate a fetch call and we won't need appMeasurement at all.

jonsnyder (approved):

sandbox/public/index.html
ninaceban: Good to know! thank you, I'll remove all the dependencies on AppMeadurement.js.

jfkhoury (approved):
Thank you @ninaceban!! Are you going to add the Fetch call before merging this PR?

sandbox/src/DataInsertionAPI.js
```diff
@@ -0,0 +1,14 @@
+const trackingServer = "ujsl.sc.omtrdc.net";
+const reportSuite = "ujslecommerce";
+
+export function triggerAnalyticsHit({ analyticsPayload, visitorID }) {
+  const url = `https://${trackingServer}/b/ss/${reportSuite}/0?g=${window.location}&r=${document.referrer}&vid=${visitorID}&tnta=${analyticsPayload}`;
```
jonsnyder, line 5: This should be "mid" not "vid" for the visitor id. "mid" is for the ECID, and "vid" is for the AID

jonsnyder (approved):
Nice work!

sandbox/src/DataInsertionAPI.js
jonsnyder: Also, should you include "pe=tnt"?

sandbox/public/perf/js/onBeforeEventSentHelper.js
```diff
@@ -0,0 +1,38 @@
+// code to collect tnta and trigger the Analytics Hit
+const trackingServer = "ujsl.sc.omtrdc.net";
+const reportSuite = "ujslecommerce";
+
+function triggerAnalyticsHit({ analyticsPayload, visitorID }) {
+  const url = `https://${trackingServer}/b/ss/${reportSuite}/0?g=${window.location}&r=${document.referrer}&vid=${visitorID}&tnta=${analyticsPayload}`;
```
jfkhoury, line 6: I think you need to apply @jonsnyder's comment here: vid -> mid. And do we need `pe:tnt`?

sandbox/src/DataInsertionAPI.js
ninaceban: Thank you Jon! I'll change that. Related to `pe`:  I'm not sure if it is required, but I will double check with Ogden.

sandbox/public/perf/js/onBeforeEventSentHelper.js
```diff
@@ -0,0 +1,38 @@
+// code to collect tnta and trigger the Analytics Hit
+const trackingServer = "ujsl.sc.omtrdc.net";
+const reportSuite = "ujslecommerce";
+
+function triggerAnalyticsHit({ analyticsPayload, visitorID }) {
```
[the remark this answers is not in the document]
ninaceban: yes!

sandbox/public/onBeforeEventSentHelper.js
```diff
@@ -0,0 +1,40 @@
...
+const reportSuite = "ujslecommerce";
+
+function triggerAnalyticsHit({ analyticsPayload, visitorID }) {
+  const url = `https://${trackingServer}/b/ss/${reportSuite}/0?g=${window.location}&r=${document.referrer}&mid=${visitorID}&tnta=${analyticsPayload}`;
+
+  return fetch(url)
+    .then(success => {
+      console.log("success", success);
+    })
+    .catch(error => {
```
liljenback, line 12: Should this trigger and fail the whole alloy request as well?

ninaceban: nope, we shouldn't change any alloy requests, just retrieve the information needed from the event.

sandbox/src/PersonalizationAnalyticsClientSide.js
```diff
@@ -0,0 +1,245 @@
...
+      });
+    });
+  }, []);
+
+  const match = useRouteMatch();
+
+  return (
+    <div>
+      <ContentSecurityPolicy />
+      <h1>Personalization with A4T client side logging</h1>
```
liljenback, line 193: Maybe we should add a bit more information here about this specific use-case (using the fetch call to the analytics end-point etc.). And link to documentation that describes it.

sandbox/src/PersonalizationAnalyticsClientSide.js
```diff
@@ -0,0 +1,245 @@
+import React, { useEffect } from "react";
```
liljenback, line 1: There's a lot going on in this file, maybe break out some of the logic into separate files so it's easier to follow.

sandbox/src/components/ContentSecurityPolicy.js
```diff
@@ -23,7 +23,7 @@ export default function ContentSecurityPolicy() {
               script-src 'self' 'nonce-${process.env.REACT_APP_NONCE}' cdn.jsdelivr.net assets.adobedtm.com cdn.tt.omtrdc.net;
               style-src 'self' 'unsafe-inline';
               img-src * data:;
-              connect-src 'self' *.alloyio.com *.adobedc.net *.demdex.net`}
+              connect-src 'self' *.alloyio.com *.adobedc.net *.demdex.net *.sc.omtrdc.net`}
```
liljenback, line 26: How about we add a comment on top (similar to the cdn.tt.omtrdc.net) about why we need the omtrdc.net domain added to the CSP?

ninaceban: yes! Thank you!

sandbox/src/PersonalizationAnalyticsClientSide.js
ninaceban: Agree!

sandbox/public/index.html
```diff
@@ -102,10 +103,30 @@
...
+            for (var i = 0; i < propositions.length; i++) {
+              var proposition = propositions[i];
+              analyticsPayloads.add(getAnalyticsPayload(proposition));
+            }
+
+            getECID("organizationTwo").then(visitorID => {
+              const analyticsPayload = concatinateAnalyticsPayloads(
+                analyticsPayloads
+              );
+              triggerAnalyticsHit({ analyticsPayload, visitorID });
```
liljenback, line 126: Maybe call this something like sendAnalyticsPayload instead perhaps.

sandbox/public/onBeforeEventSentHelper.js
```diff
@@ -0,0 +1,40 @@
+// code to collect tnta and trigger the Analytics Hit
+const trackingServer = "ujsl.sc.omtrdc.net";
+const reportSuite = "ujslecommerce";
+
+function triggerAnalyticsHit({ analyticsPayload, visitorID }) {
```
liljenback, line 5: Is this duplicated in DataInsertionAPI.js or did you forget to delete it after breaking it out?

liljenback (approved):

sandbox/public/onBeforeEventSentHelper.js
ninaceban: yes, it is duplicated from DataInsertionApi. I can't import it here because it is outside the modules.

sandbox/public/index.html
ninaceban: ok, thank you

ninaceban: Named it "triggerAnalyticsHit" so that is clear that this is the Analytics hit (like the one triggered by AppMeasurement), but contains the Target details as well.

PR #802, opened 2022-02-04
TNT-43012 Return default content items

XDex:
Note: DO NOT MERGE until target-upstream v0.0.56 with support for `characteristics -> scopeType` is released to production.
## Description

Support default content proposition items delivered for control experiences with no offer content.

## Related Issue

https://jira.corp.adobe.com/browse/TNT-43012

## Motivation and Context

Support default content propositions in the response, so that customers know when a control experience is selected.

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Personalization/createPersonalizationDetails.js
+++ b/src/components/Personalization/createPersonalizationDetails.js
@@ -13,6 +13,7 @@ governing permissions and limitations under the License.
 import { includes } from "../../utils";
 import PAGE_WIDE_SCOPE from "./constants/scope";
 import {
+  DEFAULT_CONTENT_ITEM,
   DOM_ACTION,
   HTML_CONTENT_ITEM,
   JSON_CONTENT_ITEM,
@@ -41,7 +42,12 @@ export default ({ renderDecisions, decisionScopes, event, viewCache }) => {
         scopes.push(PAGE_WIDE_SCOPE);
       }
 
-      const schemas = [HTML_CONTENT_ITEM, JSON_CONTENT_ITEM, REDIRECT_ITEM];
+      const schemas = [
+        DEFAULT_CONTENT_ITEM,
+        HTML_CONTENT_ITEM,
+        JSON_CONTENT_ITEM,
+        REDIRECT_ITEM
+      ];
 
       if (includes(scopes, PAGE_WIDE_SCOPE)) {
         schemas.push(DOM_ACTION);
--- a/src/components/Personalization/groupDecisions.js
+++ b/src/components/Personalization/groupDecisions.js
@@ -10,16 +10,21 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import { isNonEmptyArray } from "../../utils";
-import { DOM_ACTION, REDIRECT_ITEM } from "./constants/schema";
+import { isNonEmptyArray, includes } from "../../utils";
+import {
+  DOM_ACTION,
+  REDIRECT_ITEM,
+  DEFAULT_CONTENT_ITEM
+} from "./constants/schema";
+import { VIEW_SCOPE_TYPE } from "./constants/scopeType";
 import PAGE_WIDE_SCOPE from "./constants/scope";
 
-const splitItems = (items, schema) => {
+const splitItems = (items, schemas) => {
   const matched = [];
   const nonMatched = [];
 
   items.forEach(item => {
-    if (item.schema === schema) {
+    if (includes(schemas, item.schema)) {
       matched.push(item);
     } else {
       nonMatched.push(item);
@@ -38,13 +43,13 @@ const createDecision = (decision, items) => {
   };
 };
 
-const splitDecisions = (decisions, schema) => {
+const splitDecisions = (decisions, ...schemas) => {
   const matchedDecisions = [];
   const unmatchedDecisions = [];
 
   decisions.forEach(decision => {
     const { items = [] } = decision;
-    const [matchedItems, nonMatchedItems] = splitItems(items, schema);
+    const [matchedItems, nonMatchedItems] = splitItems(items, schemas);
 
     if (isNonEmptyArray(matchedItems)) {
       matchedDecisions.push(createDecision(decision, matchedItems));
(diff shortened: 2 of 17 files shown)


src/components/Personalization/createPersonalizationDetails.js
```diff
@@ -41,7 +42,12 @@ export default ({ renderDecisions, decisionScopes, event, viewCache }) => {
         scopes.push(PAGE_WIDE_SCOPE);
       }
 
-      const schemas = [HTML_CONTENT_ITEM, JSON_CONTENT_ITEM, REDIRECT_ITEM];
+      const schemas = [
+        DEFAULT_CONTENT_ITEM,
```
jfkhoury, line 46: I like the explicit nature of this PR.

I wish there was a way for Target Upstream to figure out which schemas to act on without a change to the Web SDK  which enable us to add more schemas in the future without a change to the SDKs..

jfkhoury (approved):

src/components/Personalization/createPersonalizationDetails.js
XDex: @jfkhoury We can make an optimization in this regard in case when content for all available schemas should be retrieved by not specifying `schemas` array at all in the request - in this case content for all available schemas will be returned implicitly from Target Upstream.
Thus, only when the page wide scope is not included in the request would we need to set the schemas explicitly to all values other than DOM_ACTION. Let me know if this would be desired and I'll open a ticket for it..

5amfung:
@XDex Your PR title is referring to the wrong TNT number.

ninaceban (changes requested):
@XDex great job!
I saw you are checking that `renderAttempted=false` although there was nothing to render, and `renderAttempted` is confusing in this situation, it is an experience that was executed on the page because of the nature of `__view__` scope. The customers are chaining the `sendEvent` promise to filter the executed experiences, collect the response tokens and send them to third parties. 
I would say that in case `renderDecisions=true` we should add the `renderAttempted=true` flag and send a display notification.

XDex:
@ninaceban As we discussed today, I'll wait for your A4T PR to get merged before adding the mods to mark default content items as renderable actions

test/functional/specs/Personalization/C28761.js
```diff
@@ -0,0 +1,81 @@
...
+  await t
+    .expect(defaultContentItem.schema)
+    .eql("https://ns.adobe.com/personalization/default-content-item");
+  await t
+    .expect(defaultContentItem.meta["activity.name"])
+    .eql("Functional: C28761 AB");
+  await t.expect(defaultContentItem.meta["offer.name"]).eql("Default Content");
+  await t.expect(defaultContentItem.data).eql(undefined);
+
+  await t.expect(eventResult.propositions[0].renderAttempted).eql(false);
```
jonsnyder, line 80: I'm wondering if renderAttempted should be true for default-content-items. In the documentation @ninaceban wrote for client side logging (https://git.corp.adobe.com/AdobeDocs/experience-platform.en/pull/2470), she recommends using the renderAttempted flag to determine which TNTAs/analyticsPayloads to record. If renderAttempted is false for default-content-items, client side logging would miss these.

src/components/Personalization/createPersonalizationDetails.js
jonsnyder: @XDex the optimization you suggested would still require a code change when a new schema was added. Target upstream should know that only when the "__view__" scope is requested, it should return DOM_ACTION schemas. If this were the case the Web SDK would only need to be updated when new schemas needed to be handled by the SDK. Currently only redirect, and dom action schemas are handled. The rest are passed through. 

If a new pass-through schema were added, this code would need to be updated even though the SDK isn't doing anything with it.

test/functional/specs/Personalization/C28761.js
jonsnyder: Oh, I see Nina already brought up this concern.

src/components/Personalization/createPersonalizationDetails.js
XDex: Right, let's just leave it as-is then for now, the schemas being always explicitly specified in the request.

XDex:
@ninaceban @jonsnyder I've implemented support for default-content items when `renderDecisions=true`, with proper grouping  to view/mbox scopes, please take another look when you get a chance.
Note - This PR requires the support of `characteristics -> scopeType` to be released in Target upstream before being merged, see https://jira.corp.adobe.com/browse/TNT-43696
Thank you!

test/functional/specs/Personalization/C28761.js
XDex: Yes, this has been addressed in the latest changes

jonsnyder (approved):
These changes look good. I would like Nina to review before we merge though.

src/components/Personalization/groupDecisions.js
```diff
@@ -57,50 +62,68 @@ const splitDecisions = (decisions, schema) => {
...
 
   return {
     redirectDecisions: decisionsGroupedByRedirectItemSchema.matchedDecisions,
     pageWideScopeDecisions,
-    viewDecisions: nonPageWideScopeDecisions,
-    nonAutoRenderableDecisions:
-      decisionsGroupedByDomActionSchema.unmatchedDecisions
+    viewDecisions: viewScopeDecisions,
+    nonAutoRenderableDecisions: decisionsGroupedByRenderableSchemas.unmatchedDecisions.concat(
+      nonPageWideScopeDecisions
```
ninaceban, line 125: I would suggest to keep the old order of extracting the decisions:
1. Extract the `redirect` decisions by Redirect Schema(no changes needed).
2. Extract from remaining decisions from step 1 the `renderable` decisions (by DOM_ACTION, DEFAULT_CONTENT_ITEM schema and `scopeType: view`).
3. Extract `page-wide scope` decisions and `viewDecisions` from the `renderableDecisions` from step 2, by `__view__` scope.
4. Remaining decisions from step 2 are the form-based composed decisions.

Otherwise, it becomes extremely hard to understand why we group then concat at the end the decisions, then also maintain the code.

ninaceban (changes requested):
@XDex Thank you Andrei. I see you have added a functional test,  could you also add unit tests that assert a proper grouping of the decisions based on the use-cases:
- default content item is returned for a view 
- default content item is returned for a decision scope
- default content item is returned for `__view__` scope

PR #813, opened 2022-02-14
Fix broken link to User Docs

littlebutty:
## Description

Just fixed a broken link to the user docs.

## Motivation and Context

Broken link

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Improvement (non-breaking change which does not add functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have signed the [Adobe Open Source CLA](https://opensource.adobe.com/cla.html) or I'm an Adobe employee.
- [ ] I have made any necessary test changes and all tests pass.
- [ ] I have run the Sandbox successfully.

Issue #808, opened 2022-02-11
Update the docs link in the readme


yrodrigez:
### Expected Behaviour
Redirect to the documentation

### Actual Behaviour
Error 404: Page not found

jfkhoury:
Thank you @yrodrigez !

carterworks:
It looks like there was some change at developer.adobe.com that broke our documentation links. `https://developer.adobe.com/apis/experienceplatform` now either 404's or redirects to `https://developer.adobe.com/experience-platform-apis`, but that only covers endpoints and doesn't have any Web SDK documentation. 

I've found 4 total links in our code that need fixing. I don't know where the _should_ go, but probably to the documentation that's on Experience League.

- [ ] [The README link to the documentation](https://github.com/adobe/alloy/blob/b11decc00b7687b573a83e4556a81c1fc4749f61/README.md?plain=1#L7): `https://adobe.ly/36dGGp6` → `https://developer.adobe.com/apis/experienceplatform/home/services/web-sdk.html`
- [ ] [`COMMAND_DOC_URI` in `validateCommandOptions.js`](https://github.com/adobe/alloy/blob/ecf056ae6aab6e61831f23ecbd8f8cc669f893a2/src/core/validateCommandOptions.js#L13): `https://adobe.ly/2UH0qO7` → `https://developer.adobe.com/apis/experienceplatform/home/services/web-sdk.html#!api-specification/markdown/narrative/edge/fundamentals/executing-commands.md`
- [ ] [The `documentationLink` for the `sendEvent` command in the `DataCollection` module](https://github.com/adobe/alloy/blob/d6f4cff7b8f23f4a173113e3d93c3648c63fa021/src/components/DataCollector/index.js#L19): `https://adobe.ly/2r0uUjh` → `https://developer.adobe.com/apis/experienceplatform/home/services/web-sdk.html#!api-specification/markdown/narrative/edge/fundamentals/tracking-events.md`
- [ ] [`COMMAND_DOC_URI` in `buildAndValidateConfig.js`](https://github.com/adobe/alloy/blob/ecf056ae6aab6e61831f23ecbd8f8cc669f893a2/src/core/buildAndValidateConfig.js#L15): `https://adobe.ly/2M4ErNE` → `https://developer.adobe.com/apis/experienceplatform/home/services/web-sdk.html#!api-specification/markdown/narrative/edge/fundamentals/configuring-the-sdk.md`

carterworks:
#813

PR #815, opened 2022-02-14
Fix broken documentation links in code

carterworks:
## Description

I fixed 3 broken documentation links.

## Related Issue

#808 first made us aware of the issue.

#813 fixes the issue in the README, this fixes a few of the issues throughout the code, in error messages and comments.

## Motivation and Context

https://developer.adobe.com/apis/experience-platform was deprecated, which all our short links pointed to. The new home for Alloy user documentation is now at https://experienceleague.adobe.com/docs/experience-platform/edge/

## Types of changes

## Checklist:

### DIFF
--- a/src/components/DataCollector/index.js
+++ b/src/components/DataCollector/index.js
@@ -16,7 +16,7 @@ const createDataCollector = ({ eventManager }) => {
   return {
     commands: {
       sendEvent: {
-        documentationUri: "https://adobe.ly/2r0uUjh",
+        documentationUri: "https://adobe.ly/3GQ3Q7t",
         optionsValidator: options => {
           return validateUserEventOptions({ options });
         },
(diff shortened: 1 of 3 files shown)


carterworks:
It looks like the unit tests failed because of a SauceLabs thing (maybe rate limiting?), not because of any actual test failures.

jonsnyder (changes requested):
The idea with the bitly links is that we can update the URLs without having to change the code. This is so that people with old versions of the SDK can follow these links and be redirected to the correct URL. We should update the bitly links to point to the correct place. Then a code change isn't necessary. We should do this with #813 as well. Maybe we could add some documentation to the GIT wiki on how to do this.

carterworks:
I've talked with Martin, and I don't think we can just update the target of the shortlinks like that without changing the hash. To point to a new location, I'm 85% sure we need to change the link itself.

liljenback (approved):

jonsnyder (approved):
If we don't hear back from corporate on possibly redirecting the old links, I'm fine with changing these.

jfkhoury:
@liljenback I thought there's a tool somewhere that allows updating those URL without changing them, is that not the case?

PR #816, opened 2022-02-16
Set sameSite attributes on Konductor cookies

jonsnyder:
## Description

Konductor is adding attributes `sameSite` and `secure` to cookies returned in the state:store handle. This change honors those attributes. Anytime sameSite is set to none, the secure flag is added because that is required by browsers for the cookie to work.

I attempted to create a failing functional test to show how this fixes an issue with cross-domain iframe, but the testCafe hammerhead proxy doesn't accurately reflect browser behavior for cross-domain iframe. I've committed this change [here](https://github.com/adobe/alloy/commit/4f4c822b27eabbc467ce6e5c421f78bd2d6b46f5), but not added it to this PR because it doesn't actually fail before this fix. Also it would require changes to our alloyio.com site to include an additional domain name and cert. See https://github.com/adobe/alloy/wiki/Running-the-sandbox-locally-over-HTTPS-on-Mac-OS for instructions on how to test cross-domain requests manually.

## Related Issue

https://jira.corp.adobe.com/browse/PDCL-7396

## Types of changes

## Checklist:


sandbox/public/iframed.html
```diff
@@ -0,0 +1,20 @@
...
+      iframe {
+        height: 100%;
+      }
+      iframe {
+        width: 100%;
+      }
+    </style>
+  </head>
+  <body>
+    <iframe id="iframe" src="https://alloyio.com:3000" />
```
jfkhoury, line 18: Will this work once we host this file on `alloyio.com`? Are you trying to create a cross domain situation?

src/core/createCookieTransfer.js
```diff
@@ -67,11 +67,18 @@ export default ({ cookieJar, orgId, apexDomain }) => {
...
+        const sameSite = stateItem.sameSite && stateItem.sameSite.toLowerCase();
 
         if (stateItem.maxAge !== undefined) {
           // cookieJar expects "expires" in days
           options.expires = convertTimes(SECOND, DAY, stateItem.maxAge);
         }
+        if (sameSite !== undefined) {
+          options.sameSite = sameSite;
+        }
+        if (sameSite === "none" || stateItem.secure === true) {
```
jfkhoury, line 79: Should we add a comment that we don't really expect the Edge to set the `secure` attribute, so that's why we're setting it? Just for future eyes :)

jfkhoury (approved):
Sweet!!

Should we add a functional test now that Konductor has released this feature?

ninaceban (approved):
👍

jonsnyder:
There's no way to add a functional test because you cannot inspect what attributes were set for a cookie.

sandbox/public/iframed.html
jonsnyder: This is meant to be used locally. Yes, I'm trying to create a cross domain iframe. https://github.com/adobe/alloy/wiki/Running-the-sandbox-locally-over-HTTPS-on-Mac-OS

ninaceban: Maybe we could use a dummy GitHub page in the iFrame, that would work for both local and hosted on alloyio.com?

jonsnyder: I removed the port number. I'm still able to test by hosting alloyio.com locally on port 443. I added instructions to the wiki: https://github.com/adobe/alloy/wiki/Running-the-sandbox-locally-over-HTTPS-on-Mac-OS

src/components/Audiences/injectProcessDestinations.js
```diff
@@ -10,48 +10,59 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
-    dest => dest.type === "cookie"
-  );
+export default ({
+  fireReferrerHideableImage,
+  logger,
+  cookieJar,
+  isPageSsl
+}) => {
+  const extraCookieOptions = isPageSsl
+    ? { sameSite: "none", secure: true }
```
ninaceban, line 28: Aren't these options returned by the Konductor if it's https?

jonsnyder: They are for "state:store" handles, but not "activation:push" handles.

jonsnyder:
We found issues with identity cookies when the user visits a secure page and then a non-secure page. We need to add a functional test with this use case, and we need to modify the behavior of Experience Edge before releasing this.

jonsnyder:
> > We found issues with identity cookies when the user visits a secure page and then a non-secure page. We need to add a functional test with this use case, and we need to modify the behavior of Experience Edge before releasing this.
> 
> @jonsnyder What's the latest on these issues?

We decided to go ahead with this change. We should encourage customers to host all their pages securely.

PR #817, opened 2022-02-17
Enable Alloy sourcemaps for sandbox builds

carterworks:
## Description

Modify the `rollup.config.js` to enable sourcemaps when building sandboxes (i.e. on alloyio.com and on the local development sandboxes). It makes rollup generate an extra `alloy.js.map` file on build, then the browser requests that automatically when opening up the devtools. That map file provides a mapping between alloy.js line numbers and file structure. If we don’t want to host an extra file, there is an option to include that same information inside automatically generated comments on the js file, using `"inline"` instead of `true`

## Motivation and Context

This should make in-browser debugging easier, as it allows you to set breakpoints in familiar file names instead of inside the bundled file. It also makes the origin of console messages more clear.

[MDN documentation on how to use sourcemaps](https://developer.mozilla.org/en-US/docs/Tools/Debugger/How_to/Use_a_source_map)
[Rollup.js config documentation on the `output.sourcemap` option](https://rollupjs.org/guide/en/#outputsourcemap)

## Types of changes

## Checklist:


shammowla (approved):

jfkhoury (approved):
Nice!

ninaceban (approved):

PR #818, opened 2022-02-17
PDCL-6476 - Include propositions on every request response

carterworks:
## Description

With this change, Personalization includes a lifecycle method for `onResponse` that provides `propositions: []` if none existed before.

The test to verify this fix is found in the functional test [`C5298194.js`](https://github.com/adobe/alloy/blob/ba7ea6a31e0a6e6ef4a0e1f2906450af919af1e2/test/functional/specs/Personalization/C5298194.js), where I have added some assertions. This test can be run with the following command:

```
npm run test:functional -- "--test-grep C5298194"
```

## Related Issue

[PDCL-6476: Resolved send event promise does not include propositions key on collect calls](https://jira.corp.adobe.com/browse/PDCL-6476)

## Motivation and Context

From [the Jira ticket](https://jira.corp.adobe.com/browse/PDCL-6476)

> Currently, an object with decisions and propositions are resolved in the sendEvent promise. However, when send event chooses to use "collect" instead of "interact", those keys are not returned. A customer building a javascript handler to do something with these objects would need to check if the key was defined before iterating on the array.
> 
> We should include "propositions: []" as part of the response when collect is used. I don't think we need to include decisions as that is a deprecated api now.
> 
> This also happens when consent is declined, and the previous send event calls are resolved.

## Types of changes

## Checklist:

### DIFF
--- a/test/functional/specs/Personalization/C5298194.js
+++ b/test/functional/specs/Personalization/C5298194.js
@@ -3,6 +3,7 @@ import createFixture from "../../helpers/createFixture";
 import { orgMainConfigMain } from "../../helpers/constants/configParts";
 import createAlloyProxy from "../../helpers/createAlloyProxy";
 import createCollectEndpointAsserter from "../../helpers/createCollectEndpointAsserter";
+import { TEST_PAGE as TEST_PAGE_URL } from "../../helpers/constants/url";
 
 createFixture({
   title: "C5298194: Include propositions on every request"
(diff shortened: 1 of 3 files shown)


carterworks:
I have the pull request in draft because I’m only like 45% confident that the way to solve the issue is to put a onRequest lifecycle handler and putting propositions: []  on every request. That feels like too much, but I’m unsure what else to do at this point. [The branch in `createOnResponseHandler.js` that includes the stub](https://github.com/adobe/alloy/blob/52703e1c84b9b372350efda6e95cf98082c47522/src/components/Personalization/createOnResponseHandler.js#L28) only gets called [when making data requests or when cached data is being used](https://github.com/adobe/alloy/blob/4e1cae4c41a5de731f54538fe791e7eb6d99c825/src/components/Personalization/createComponent.js#L50), which isn’t all the time.

jonsnyder (changes requested):
I added a commit with the way I think it should be done. So if we aren't fetching data or using the cached data, it returns an empty propositions array. I think this is better than having to do the JSON parse stuff you had before.

I'm requesting changes because the functional test should be a part of the Personalization test suite. You can add a new test in testrail, and then copy this test to the personalization folder. Then remove all the other assertions to just leave the ones about the propositions property.

jonsnyder:
Looking at it again, maybe we should be returning proposition: [] when authoring mode is enabled too. Let's discuss in standup tomorrow. Or if @ninaceban sees this before standup she can chime in.

test/functional/specs/Personalization/C5298194.js
```diff
@@ -0,0 +1,36 @@
...
+
+  let response = await alloy.sendEvent({ documentUnloading: true });
+  await collectEndpointAsserter.assertInteractCalledAndNotCollect();
+  await t
+    .expect("propositions" in response)
+    .ok("The 'sendEvent()' response was missing the 'propositions' property");
+  response = null;
+  await collectEndpointAsserter.reset();
+
+  response = await alloy.sendEvent();
```
jonsnyder, line 30: We probably want to call "collect" at some point in this test. Normally setting documentUnloading to true will trigger a collect call, but if the identity isn't established, it will call interact. If you move the `{ documentUnloading: true }` parameter to this line, it will call collect here.

test/functional/specs/Personalization/C5298194.js
```diff
@@ -0,0 +1,79 @@
...
+    .expect("propositions" in response)
+    .ok("The 'sendEvent()' response was missing the 'propositions' property");
+  await t
+    .expect(Array.isArray(response.propositions))
+    .ok("response.propositions is not an array");
+  response = null;
+  await collectEndpointAsserter.reset();
+});
+
+test.page(`${TEST_PAGE_URL}?adobe_authoring_enabled=true`)(
```
jonsnyder, line 48: If both these tests are the same, you can extract them into a function.

```
const runTest = async () => {
  ...
};
test("Test C5298194: Include propositions on every request", runTest);
test.page(`${TEST_PAGE_URL}?adobe_authoring_enabled=true`)(
  "Test C5298194: Include propositions on every request in authoring mode",
  runTest
);
```

jonsnyder (approved):

ninaceban (approved):
Looks great! 👍

jonsnyder (approved):
Thanks!

PR #827, opened 2022-03-07
PDCL-7878 - fix clicks for spa view

ninaceban:
## Description

Quick fix for the autorenderable click metrics for SPA. We will be adding the `viewName` to the `INTERACT` notification call so that the conversions are properly handled. 
Autorenderable click metrics are either for `__view__` scope or an SPA view.

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Personalization/createOnClickHandler.js
+++ b/src/components/Personalization/createOnClickHandler.js
@@ -12,6 +12,7 @@ governing permissions and limitations under the License.
 
 import { isNonEmptyArray } from "../../utils";
 import { INTERACT } from "./constants/eventType";
+import PAGE_WIDE_SCOPE from "./constants/scope";
 
 export default ({
   mergeDecisionsMeta,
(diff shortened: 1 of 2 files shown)


src/components/Personalization/createOnClickHandler.js
```diff
@@ -30,7 +31,18 @@ export default ({
       );
 
       if (isNonEmptyArray(decisionsMeta)) {
-        event.mergeXdm({ eventType: INTERACT });
+        const xdm = { eventType: INTERACT };
+        const scope = decisionsMeta[0].scope;
+
+        if (scope !== PAGE_WIDE_SCOPE) {
+          xdm.web = {
```
jfkhoury, line 38: Wouldn't this replace fully replace `xdm.web`? Should we be merging in `viewName: scope` instead?

Let's add this test case maybe, where we add `viewName` to an existing `web.webPageDetails`.

jfkhoury (commented):
I added one quick comment about assigning viewName instead of replacing `webpageDetails`

jonsnyder (approved):
I was thinking that we would have to remember the view from the last view-change event and use that, but this is much easier.

src/components/Personalization/createOnClickHandler.js
ninaceban: `event.mergeXDM` is using `deepAssign`, so it is safe to do so.

jfkhoury (approved):
Nice & Simple!!

src/components/Personalization/createOnClickHandler.js
```diff
@@ -30,7 +31,18 @@ export default ({
       );
 
       if (isNonEmptyArray(decisionsMeta)) {
-        event.mergeXdm({ eventType: INTERACT });
+        const xdm = { eventType: INTERACT };
+        const scope = decisionsMeta[0].scope;
```
liljenback, line 35: The scope needed will always be the first element?

liljenback (approved):

PR #829, opened 2022-03-09
Added "applyPropositions" command to enable re-rendering of personalized content.

dcottingham:
Proposal:  https://wiki.corp.adobe.com/pages/viewpage.action?spaceKey=Target&title=Alloy+SPA+Re-Render+Support
I've added a solution #4 to the wiki based on your comments @ninaceban @jonsnyder 

## Types of changes

## Checklist:


ninaceban (changes requested):
Don it seems that we are exposing multiple capabilities which will solve different use-cases:
1.  `applyPropositions` with viewName, propositions and `notificationEnabled: true` will behave like `sendEvent` for a view-change.
2.  `applyPropositions` with `viewName`,  propositions and `notificationEnabled: false` will behave like`triggerView` with `page: false`
3.  `applyPropositions` with propositions seems like a utility to render the propositions that the customer have got from an event triggered earlier.

I thought that `applyPropositions` should take care of only re-rendering propositions for a view.
For example:
```
alloy("applyPropositions", {viewName: "cart" });
```
would re-render the cached propositions for `cart` view and not send the notification. As the initial version of this command, `viewName` will be the only supported option.

src/components/Personalization/createApplyPropositions.js
```diff
@@ -0,0 +1,47 @@
...
+          collect({ decisionsMeta, viewName });
+        }
+        collect({ decisionsMeta });
+      }
+      showContainers();
+    });
+  };
+
+  return ({ propositions, viewName, notificationsEnabled = false }) => {
+    if (isNonEmptyArray(propositions)) {
```
dcottingham, line 33: This would be the equivalent of applyOffers in at.js - and can be used to render previously fetched propositions.

src/components/Personalization/createApplyPropositions.js
```diff
@@ -0,0 +1,47 @@
...
+      return applyPropositions({ propositions, notificationsEnabled });
+    }
+    if (viewName) {
```
dcottingham, line 36: This re-renders view decisions from the cache and is only used for SPAs.

jonsnyder:
@ninaceban, display and notification events have always been handled automatically. They have never required the user to call sendEvent explicitly again. Said another way, why would they have to call sendEvent after an applyPropositions if they don't have to do that after a sendEvent?

Having said that, I'm wondering if we need the notificationsEnabled flag at all. I'm curious to know the use case where you would want applyPropositions to send display notifications. Is there a way to get renderable propositions from sendEvent without having them render?

src/components/Personalization/validateApplyPropositionsOptions.js
```diff
@@ -0,0 +1,40 @@
...
+            prehidingSelector: string()
+          })
+        })
+      ),
+      scopeDetails: objectOf(anything())
+    })
+  ).nonEmpty(),
+  viewName: string(),
+  notificationsEnabled: boolean()
+}).required();
```
jonsnyder, line 40: I'm wondering if we need to provide such strict validation here when users will simply take the propositions from the sendEvent call and send them to applyPropositions. For example, when we support ODE offers, they may not be in the same format.

src/components/Personalization/createApplyPropositions.js
```diff
@@ -0,0 +1,38 @@
...
+      return Promise.resolve(applyPropositions({ propositions }));
+    }
+    if (isNonEmptyString(viewName)) {
+      return viewCache.getView(viewName).then(viewPropositions =>
+        applyPropositions({
+          propositions: viewPropositions
+        })
+      );
+    }
+    return Promise.resolve();
```
jonsnyder, line 36: This should return an empty propositions array so customers don't have to check to see if it is defined.

src/components/Personalization/createApplyPropositions.js
```diff
@@ -0,0 +1,38 @@
...
+      return composePersonalizationResultingObject(propositions, true);
+    });
+  };
+
+  return ({ propositions, viewName }) => {
+    if (isNonEmptyArray(propositions)) {
+      return Promise.resolve(applyPropositions({ propositions }));
+    }
+    if (isNonEmptyString(viewName)) {
+      return viewCache.getView(viewName).then(viewPropositions =>
```
jonsnyder, line 30: Consider the case where a customer calls applyPropositions before calling sendEvent. The promise from getView would only resolve after the sendEvent returns and the viewCache is initialized. What should happen in this case? Should we throw an error, or is it appropriate to wait?

src/components/Personalization/createApplyPropositions.js
```diff
@@ -0,0 +1,38 @@
...
+
+export default ({ viewCache, executeDecisions, showContainers }) => {
+  const applyPropositions = ({ propositions }) => {
+    return executeDecisions(propositions).then(() => {
+      showContainers();
+      return composePersonalizationResultingObject(propositions, true);
+    });
+  };
+
+  return ({ propositions, viewName }) => {
```
jonsnyder, line 25: Consider when propositions and viewName are both specified. As written here, only the propositions would be rendered. Is this the desired behavior? If so we should log a warning to explain that only one should be specified and that only the propositions in the array will be applied.

dcottingham: I don't think there is a valid use case where a user should send both viewName and propositions.  SPAs should only specify viewName and get the propositions from the view cache.  Non-SPAs should only specify propositions that should get rendered.  We'll definitely need to document this.

src/components/Personalization/createApplyPropositions.js
dcottingham: @ninaceban, what are your thoughts on this use case?

dcottingham:
@ninaceban, @jonsnyder - I've updated this PR with the latest from our discussions, including updates to the sandbox app.  Please review again.

src/components/Personalization/createApplyPropositions.js
```diff
@@ -0,0 +1,73 @@
...
+  type: "appendHtml"
+};
+
+export default ({ executeDecisions, showContainers }) => {
+  const updatePropositions = ({ proposition, metadataForScope }) => {
+    proposition.items.forEach(item => {
+      if (item.schema.includes(HTML_CONTENT_ITEM)) {
+        item.data.selector =
+          metadataForScope.selector || DEFAULT_METADATA.selector;
+        item.data.type = metadataForScope.actionType || DEFAULT_METADATA.type;
```
jonsnyder, line 31: Should we be making changes to the propositions the user is sending in? This could leak across calls to applyPropositions.

src/components/Personalization/createApplyPropositions.js
```diff
@@ -0,0 +1,73 @@
...
+  const preparePropositions = ({ propositions, metadata }) => {
+    return Promise.resolve(
+      propositions.map(proposition => {
+        const completeProposition = { ...proposition };
+        if (isNonEmptyArray(completeProposition.items)) {
+          const metadataForScope =
+            !isEmptyObject(metadata) &&
+            isObject(metadata[completeProposition.scope])
+              ? metadata[completeProposition.scope]
+              : DEFAULT_METADATA;
```
jonsnyder, line 46: There is a mismatch between updatePropositions expecting actionType, and DEFAULT_METADATA containing type. Really since you use the DEFAULT_METADATA in the updatePropositions, you could just pass {}. And then no need for the isEmptyObject check.

src/components/Personalization/createApplyPropositions.js
```diff
@@ -0,0 +1,73 @@
...
+            metadataForScope
+          });
+        }
+        return completeProposition;
+      })
+    );
+  };
+
+  const applyPropositions = ({ propositions, metadata }) => {
+    return preparePropositions({ propositions, metadata })
```
jonsnyder, line 58: There is no need for preparePropositions to return a promise. You can just do:
```
return executeDecisions(preparePropositions({ propositions, metadata }))
.then(...
```

src/components/Personalization/createApplyPropositions.js
```diff
@@ -0,0 +1,73 @@
...
+      .then(completePropositions => executeDecisions(completePropositions))
+      .then(() => {
+        showContainers();
+        propositions.forEach(proposition => delete proposition.renderAttempted);
+        return composePersonalizationResultingObject(propositions, true);
+      });
```
jonsnyder, line 64: Is it possible for executeDecisions to throw an exception? Should we include a catch(() => showContainers())?

src/components/Personalization/createApplyPropositions.js
```diff
@@ -0,0 +1,73 @@
...
+  };
+
+  return ({ propositions, metadata = {} }) => {
+    if (isNonEmptyArray(propositions)) {
+      return Promise.resolve(applyPropositions({ propositions, metadata }));
```
jonsnyder, line 69: applyPropositions already returns a promise, no need to wrap it.

jonsnyder (changes requested):

ninaceban (changes requested):
Looks great, Don!

I would avoid appending to HEAD by default if no metadata is present for a HTML_CONTENT_ITEM proposition item.

Let's say we have an event that fetches the propositions for `scope1, scope2, scope3` then it applies the propositions for `tab1 - scope1`, `tab2 - scope2` and so on.
Having metadata passed to the `applyPropositions` command gives the customer flexibility to append to HEAD explicitly. 

It would be great to have functional tests for both SPA and form-based composer based activities.

ninaceban:
Another case that needs attention is that during the first event on the page WEB SDK allways adds `__view__` scope to the query to create SPA cache (even when `renderDecisions: false`). 

In case the event that fetches `scope1, scope2` is the first event on the page, the propositions will contain propositions for `scope1, scope2, __view__`. 
The `__view__` propositions, will be rendered twice if first event had `renderDecisions: true` and the `sendEvent` propositions are sent to `applyPropositions`, this will lead to flickering.

src/components/Personalization/createApplyPropositions.js
dcottingham: On line 40 I copy the proposition before modifying it:
`completeProposition = { ...proposition }`
But maybe to make it more clear I can use `Array.from(propositions)` in the first line of applyPropositions?

jonsnyder: No, I don't think you need to change anything. I just didn't look close enough.

src/components/Personalization/createApplyPropositions.js
```diff
@@ -12,61 +12,78 @@ governing permissions and limitations under the License.
...
-    proposition.items.forEach(item => {
-      if (item.schema.includes(HTML_CONTENT_ITEM)) {
-        item.data.selector =
-          metadataForScope.selector || DEFAULT_METADATA.selector;
-        item.data.type = metadataForScope.actionType || DEFAULT_METADATA.type;
-      }
-    });
+export default ({ executeDecisions }) => {
+  const filterItemsPredicate = item => {
+    if (SUPPORTED_SCHEMAS.includes(item.schema)) {
```
jonsnyder, line 24: Can you simplify these predicates (filterItemsPredicate and filterPropositionsPredicate) to just return the boolean in the if statement? i.e.
```
return SUPPORTED_SCHEMAS.includes(item.schema);
```

src/components/Personalization/createApplyPropositions.js
```diff
@@ -12,61 +12,78 @@ governing permissions and limitations under the License.
...
+    return false;
+  };
+
+  const updatePropositionItems = ({ proposition, metadataForScope }) => {
+    proposition.items = proposition.items
+      .filter(filterItemsPredicate)
+      .map(item => {
+        if (item.schema === HTML_CONTENT_ITEM && isObject(metadataForScope)) {
+          item.data.selector = metadataForScope.selector;
+          item.data.type = metadataForScope.actionType;
```
jonsnyder, line 36: This mutates the original item. Inside the if you could use this to return a new object. Also you can simplify things if you just return the new items.

```
const updatePropositionItems = ({ items, metadataForScope }) => {
  return items
    .filter(filterItemsPredicate)
    .map(item => {
      if (item schema === HTML_CONTENT_ITEM && isObject(metadataForScope)) {
        return {
          ...item,
          data: {
            ...item.data,
            selector: metadataForScope.selector,
            type: metadataForScope.actionType
          }
        };
      }
      return item;
    });
};
```

src/components/Personalization/createApplyPropositions.js
```diff
@@ -12,61 +12,78 @@ governing permissions and limitations under the License.
...
+    return true;
+  };
+
   const preparePropositions = ({ propositions, metadata }) => {
-    return Promise.resolve(
-      propositions.map(proposition => {
-        const completeProposition = { ...proposition };
+    return propositions
+      .filter(filterPropositionsPredicate)
+      .map(proposition => {
```
jonsnyder, line 54: You can simplify this map function with this code (assuming you've include my other suggestion for updatePropositionItems):

```
.map(proposition => {
  if (isNonEmptyArray(proposition.items)) {
    return {
      ...proposition,
      items: updatePropositionItems({
        items: proposition.items,
        metadataForScope: metadata[completeProposition.scope]
      })
    };
  }
  return proposition;
});
```

src/components/Personalization/createApplyPropositions.js
```diff
@@ -12,61 +12,78 @@ governing permissions and limitations under the License.
...
+    });
+    return executeDecisions(propositionsToExecute).then(() => {
+      return Promise.resolve(
+        composePersonalizationResultingObject(propositionsToExecute, true)
+      );
+    });
   };
 
   return ({ propositions, metadata = {} }) => {
     if (isNonEmptyArray(propositions)) {
```
jonsnyder, line 85: Is this check necessary? The parameter validation should make sure that propositions is an array. If it is an empty array, then executeDecisions won't do anything anyways. Then you could just return the applyPropositions function.

src/components/Personalization/createApplyPropositions.js
```diff
@@ -12,61 +12,78 @@ governing permissions and limitations under the License.
 
 import composePersonalizationResultingObject from "./utils/composePersonalizationResultingObject";
 import isNonEmptyArray from "../../utils/isNonEmptyArray";
-import isEmptyObject from "../../utils/isEmptyObject";
 import isObject from "../../utils/isObject";
```
jonsnyder, line 15:
```
import { isNonEmptyArray, isObject } from "../../utils";
```

src/components/Personalization/createApplyPropositions.js
```diff
@@ -12,61 +12,78 @@ governing permissions and limitations under the License.
...
-        showContainers();
-        propositions.forEach(proposition => delete proposition.renderAttempted);
-        return composePersonalizationResultingObject(propositions, true);
-      });
+    const propositionsToExecute = preparePropositions({
+      propositions,
+      metadata
+    });
+    return executeDecisions(propositionsToExecute).then(() => {
+      return Promise.resolve(
```
jonsnyder, line 78: No need to wrap this in Promise.resolve. `executeDecisions` already returns a promise.

test/unit/specs/components/Personalization/createApplyPropositions.spec.js
```diff
@@ -154,96 +196,75 @@ describe("Personalization::createApplyPropositions", () => {
...
+      const executedPropositions = executeDecisions.calls.all()[0].args[0];
+      expect(executedPropositions.length).toEqual(2);
+      executedPropositions.forEach(proposition => {
+        expect(proposition.items.length).toEqual(1);
+        proposition.items.forEach(item => {
+          expect(expectedItemIds.includes(item.id));
+        });
       });
     });
   });
```
jonsnyder, line 271: Can you add a test to make sure that the original propositions array or underlying proposition objects aren't mutated when the metadata scopes are applied?

jonsnyder (commented):
Great work. I think you are really close here. I like that you removed the default selector and type. That will match up with the sendEvent behavior. I just have a few suggestions here to remove mutations and simplify the code.

src/components/Personalization/createApplyPropositions.js
dcottingham: @jonsnyder Tests fail with the error "applyPropositions(...).then is not a function" if I remove the Promise.resolve().  I didn't think I would need Promise.resolve() here either, but I think it's because we aren't using the promise value returned by executeDecisions.

jonsnyder: Ah, I see. This is because the mock object returned from executeDecisions in the test is not a real promise. i.e:

```
diff --git a/test/unit/specs/components/Personalization/createApplyPropositions.spec.js b/test/unit/specs/components/Personalization/createApplyPropositions.spec.js
index be6e5dc5..43b605d8 100644
--- a/test/unit/specs/components/Personalization/createApplyPropositions.spec.js
+++ b/test/unit/specs/components/Personalization/createApplyPropositions.spec.js
@@ -116,11 +116,7 @@ describe("Personalization::createApplyPropositions", () => {
   });
 
   it("it should return an empty propositions promise if propositions is empty array", () => {
-    const executeDecisionsPromise = {
-      then: callback => callback(PAGE_WIDE_SCOPE_DECISIONS)
-    };
-
-    executeDecisions.and.returnValue(executeDecisionsPromise);
+    executeDecisions.and.returnValue(Promise.resolve(PAGE_WIDE_SCOPE_DECISIONS));
 
     const applyPropositions = createApplyPropositions({
       executeDecisions
```

ninaceban (changes requested):
Hello @dcottingham , I have tested the use case when we would use `HTML_CONTENT_ITEM` propositions but no metadata for its scope and it fails at execution time, see error below:

Screen Shot 2022-05-30 at 7 44 18 PM

test/unit/specs/components/Personalization/createApplyPropositions.spec.js
```diff
@@ -0,0 +1,283 @@
...
+    const applyPropositions = createApplyPropositions({
+      executeDecisions
+    });
+
+    return applyPropositions({
+      propositions: []
+    }).then(result => {
+      expect(result).toEqual({ propositions: [] });
+      expect(executeDecisions).toHaveBeenCalledTimes(1);
+    });
```
ninaceban, line 132: In my opinion if `propositions` is an empty array we should short-circuit and the code wouldn't reach `executeDecisions`.

dcottingham: That is my preference as well.  @jonsnyder, what do you think?

jonsnyder: I don't think it matters either way. You could remove the check that executeDecisions was called once.

jonsnyder (approved):

test/unit/specs/components/Personalization/createApplyPropositions.spec.js
ninaceban: I would still recommend that we avoid calling `executeDecisions` in case the propositions array is empty to be less error prone.

src/components/Personalization/createApplyPropositions.js
```diff
@@ -0,0 +1,86 @@
...
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+import composePersonalizationResultingObject from "./utils/composePersonalizationResultingObject";
+import { isNonEmptyArray, isObject } from "../../utils";
+import { DOM_ACTION, HTML_CONTENT_ITEM } from "./constants/schema";
+import PAGE_WIDE_SCOPE from "./constants/scope";
+
+export const EMPTY_PROPOSITIONS = { propositions: [] };
```
ninaceban, line 18: since this is used only in `validateApplyPropositionsOptions` can we move it there as well?

ninaceban (approved):
Great job Don! I have left a few minor comments.

PR #848, opened 2022-04-06
Use a persistent iframe for all ID syncs and segment destinations

carterworks:
## Description

This change keeps a local closure variable of the `<iframe>` so that only a single one gets created, no matter how many ID syncs and segment destinations appear in the payload. It also changes `fireReferrerHideableImage` into a function provided by a factory/injector function so that all the DOM-related functions are injected so that they can be mocked for unit tests. Unit tests around the function have also been created.

After this change, the `<iframe>` persists indefinitely on the page after syncs and contains all the referrer images. It is only cleared when a sync fails. Alternatives to this are: 

- go back to creating an `<iframe>` per sync, but use `.finally()` instead of `.then()` to make sure it always goes away.
    - However, the customer thought that creating so many `<iframe>s` was giving their site performance issues, so that doesn't resolve this problem
- Clear out the referrer image after a sync instead of deleting the entire `<iframe>`
    - It's not hard, but it is extra work. If there is risk from persistant referrer images in the DOM, I would prefer this option.

## Related Issue

[PDCL-8050–Cleanup iFrame after ID syncing](https://jira.corp.adobe.com/browse/PDCL-8050)

> We currently generate anonymous iFrames to trigger ID syncs: basically we generate an iFrame with no src, inject it with an image and set the source of the image to a specific URL that we receive from Konductor.
>
> The problems:
>
> If the sync fails (Meaning if the promise is rejected), we are not removing the iFrame (https://github.com/adobe/alloy/blob/main/src/utils/fireReferrerHideableImage.js#L42)
> It seems like we might not reusing iFrames as well, so if there was a sync later on in the page lifecycle, we create a new iFrame, which is causing performance issues.
> The Ask:
> 
> Remove the iFrame on rejection and reuse one iFrame for all ID syncs in the same session.

## Motivation and Context

[The original customer ticket](https://jira.corp.adobe.com/browse/PLATIR-19466) mentions a high-volume customer that has ≥12 ID syncs and segment destinations for new browsers, and anywhere from 1-4 of them would fail. We creating all these `<iframe>s`, and the code used a `.then()` to clear them, so they weren't getting cleaned up on failure.

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Audiences/index.js
+++ b/src/components/Audiences/index.js
@@ -10,12 +10,10 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import { injectFireReferrerHideableImage } from "../../utils";
 import injectProcessDestinations from "./injectProcessDestinations";
 import injectProcessResponse from "./injectProcessResponse";
 
-const createAudiences = ({ logger }) => {
-  const fireReferrerHideableImage = injectFireReferrerHideableImage();
+const createAudiences = ({ logger, fireReferrerHideableImage }) => {
   const processDestinations = injectProcessDestinations({
     fireReferrerHideableImage,
     logger
--- a/src/components/Audiences/index.js
+++ b/src/components/Audiences/index.js
@@ -10,11 +10,12 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import { fireReferrerHideableImage } from "../../utils";
+import { injectFireReferrerHideableImage } from "../../utils";
 import injectProcessDestinations from "./injectProcessDestinations";
 import injectProcessResponse from "./injectProcessResponse";
 
 const createAudiences = ({ logger }) => {
+  const fireReferrerHideableImage = injectFireReferrerHideableImage();
   const processDestinations = injectProcessDestinations({
     fireReferrerHideableImage,
     logger
--- a/src/components/Identity/index.js
+++ b/src/components/Identity/index.js
@@ -11,7 +11,7 @@ governing permissions and limitations under the License.
 */
 
 import {
-  fireReferrerHideableImage,
+  injectFireReferrerHideableImage,
   areThirdPartyCookiesSupportedByDefault,
   injectDoesIdentityCookieExist
 } from "../../utils";
@@ -78,6 +78,7 @@ const createIdentity = ({
     awaitIdentityCookie,
     logger
   });
+  const fireReferrerHideableImage = injectFireReferrerHideableImage();
   const processIdSyncs = injectProcessIdSyncs({
     fireReferrerHideableImage,
     logger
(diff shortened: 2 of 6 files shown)


src/utils/injectFireReferrerHideableImage.js
```diff
@@ -28,23 +31,39 @@ const IFRAME_PROPS = {
...
   };
 
   const fireInIframe = ({ src }) => {
     return createIframe().then(iframe => {
       const currentDocument = iframe.contentWindow.document;
-      return fireImage({ src, currentDocument }).then(() => {
+      return fireImage({ src, currentDocument }).catch(() => {
+        hiddenIframe = undefined;
         removeNode(iframe);
       });
```
jonsnyder, line 61: Why do we need to remove the iframe? There is a race condition here where the iframe is removed between another idSync's createIframe and fireImage, so that by the time fireImage is called, the iframe is no longer part of the document.

src/components/Identity/index.js
```diff
@@ -78,6 +78,7 @@ const createIdentity = ({
     awaitIdentityCookie,
     logger
   });
+  const fireReferrerHideableImage = injectFireReferrerHideableImage();
```
jonsnyder, line 81: This will potentially create 2 iframes. One for Audiences, and one for Identity. We could move this into core and then just pass `fireReferrerHideableImage` as a component tool.

jonsnyder (changes requested):
I'm worried about the race condition. Let's talk in standup about what the requirements are for this ticket.

src/core/index.js
```diff
@@ -134,10 +140,12 @@ export const createExecuteCommand = ({
       componentRegistry,
       getImmediatelyAvailableTools(componentName) {
         const componentLogger = createComponentLogger(componentName);
+        const fireReferrerHideableImage = injectFireReferrerHideableImage();
```
jonsnyder, line 143: If you put this above the `createExecuteCommand` export, we can share the iframe across instances of Web SDK.

jonsnyder (approved):

jonsnyder (approved):

src/components/Audiences/index.js
```diff
@@ -10,11 +10,10 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import { fireReferrerHideableImage } from "../../utils";
 import injectProcessDestinations from "./injectProcessDestinations";
 import injectProcessResponse from "./injectProcessResponse";
 
-const createAudiences = ({ logger }) => {
+const createAudiences = ({ logger, fireReferrerHideableImage }) => {
```
jfkhoury, line 16: Nice!

jfkhoury (approved):
Thank for the details in the PR description. Good work!

ninaceban (approved):
I like the re-using the iFrame approach! thank you Carter!
Can we test this scenario in a functional test?

PR #851, opened 2022-04-10
Fixed an issue where personalized content with a child selector directly following an "eq" pseudo selector selects the wrong node.

### LINKED ISSUE #850: Child selectors not applied correctly in selectNodesWithEq
When `selectNodesWithEq` emulates `eq()`, the child selectors adjacent to the `eq()`s are stripped off, causing incorrect selection.

### Given
HTML:
```html
<body>
  <div id="target">
    <div>
      <div>
        <span id="wrong"></span>
      </div>
      <span id="right"></span>
    </div>
  </div>
</body>
```

Selector:
```
#target > div:eq(0) > span
```

### Expected Behaviour
When `selectNodesWithEq` is called with the above selector and markup in the document, it should select `<span id="right"></span>` as it is the only element that matches `#target > div > span`.

### Actual Behaviour
When `selectNodesWithEq` is called with the above selector and markup in the document, it actually selects `<span id="wrong"></span>`

#### Platform and Version
Tested in Chrome (version 98+), but presumed to be an issue in all browsers

alloy.js info
```json
{
  "libraryInfo": {
    "version": "2.9.0"
  }
}
```

#### Sample Code that illustrates the problem
Add the following test case to `selectNodesWithEq.spec.js`
```javascript
  it("should respect child selectors", () => {
    const content = `
      <div>
        <div>
            <span id="wrong"></span>
        </div>
        <span id="right"></span>
      </div>
    `;

    const node = createNode(
      "DIV",
      { id: "abc", class: "eq" },
      { innerHTML: content }
    );

    appendNode(document.body, node);

    const result = selectNodesWithEq("#abc > div:eq(0) > span");

    expect(result[0].tagName).toEqual("SPAN");
    expect(result[0].id).toEqual("right");
  });
```

paul-bjorkstrand:
## Description

When a child selector (`>`) immediately follows an `eq()` pseudo selector, prefix the child portion of the split selector with `:scope` to preserve the direct-child relationship when `querySelectorAll()` is executed.

### Note

This fix is applied for non-IE browsers only.  Per the [caniuse.com listing for `:scope`](https://caniuse.com/?search=%3Ascope), this not possible in any version of IE. Given IE will completely lose support in [newer version of Windows starting June 15, 2022](https://docs.microsoft.com/en-us/lifecycle/products/internet-explorer-11), this seems like a minimal risk. That said, this PR should have sufficient guards to ensure IE functions as it did before, and tests should still pass on IE.

## Related Issue

#850

## Motivation and Context

Summary: with certain markup situations, removing the child selector causes incorrect node selection. The principal side effect of this incorrect selection is that target tests can be inserted into the wrong location in markup. See #850 for more details.

While this is a relatively simple & straight forward bug fix, it may have impact on existing code and should be treated as a breaking change.

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Personalization/dom-actions/dom/selectNodesWithEq.js
+++ b/src/components/Personalization/dom-actions/dom/selectNodesWithEq.js
@@ -13,13 +13,23 @@ governing permissions and limitations under the License.
 import escape from "css.escape";
 import { selectNodes } from "../../../../utils/dom";
 import { isNotEqSelector, splitWithEq } from "./helperForEq";
+import startsWith from "../../../../utils/startsWith";
 
 // Trying to match ID or CSS class
 const CSS_IDENTIFIER_PATTERN = /(#|\.)(-?\w+)/g;
-// This is required to remove leading " > " from parsed pieces
-const SIBLING_PATTERN = /^\s*>?\s*/;
 
-const cleanUp = str => str.replace(SIBLING_PATTERN, "").trim();
+const cleanUp = selector => {
+  if (!startsWith(selector, ">")) {
+    return selector;
+  }
+
+  // IE doesn't support :scope
+  if (window.document.documentMode) {
+    return selector.substring(1).trim();
+  }
+
+  return `:scope ${selector}`;
+};
 
 // Here we use CSS.escape() to make sure we get
 // correct values for ID and CSS class
@@ -39,7 +49,7 @@ export const parseSelector = rawSelector => {
   let i = 0;
 
   while (i < length) {
-    const sel = cleanUp(parts[i]);
+    const sel = cleanUp(parts[i].trim());
     const eq = parts[i + 1];
 
     if (eq) {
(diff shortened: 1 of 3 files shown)


paul-bjorkstrand:
In addition to a simple unit test, and some adjustments to one existing test, in `selectNodesWithEq.spec.js`, I also included changes to the unit tests in `selectNodesWithShadow.spec.js` to demonstrate that the addition of the `:scope` pseudo selector also worked correctly across the shadow-dom barrier.

jonsnyder:
We need to verify that these modified selectors will work with the existing VEC editor, and see what at.js does with selectors like this.

paul-bjorkstrand:
As far as I know, at.js is (somehow) properly emulating the `:eq()` with direct child selectors. This issue was found by one of my project's target users where a test that was functioning correctly before swapping from at.js to WebSDK, was suddenly misbehaving after our projects move to WebSDK. I found the underlying issue (incorrect emulation of child selectors) after doing some deep debugging into the alloy part of the WebSDK.

Hopefully that helps give you some context around the issue while you are digging into at.js.

Thanks for the update @jonsnyder

ninaceban:
Thanks a lot for your help!
We will follow the same approach that is used in At.js, here is the draft PR:  https://github.com/adobe/alloy/pull/858

PR #858, opened 2022-04-22
Fixing the select with eq issue

ninaceban:
## Description
This is a draft PR for the issue opened by @paul-bjorkstrand. 

In At.js we are using Zepto to query DOM. In Web SDK we have partially transferred the functionality and have introduced a bug. Currently Web SDK doesn't correctly select children if `eq()` pseudo selector is involved. 
I have looked into Zepto library and I have followed the same approach.

## Related Issue
https://github.com/adobe/alloy/issues/850

## Types of changes

## Checklist:

### DIFF
--- a/src/utils/dom/querySelectorAll.js
+++ b/src/utils/dom/querySelectorAll.js
@@ -21,6 +21,8 @@ export default (context, selector) => {
 
   const tag = `alloy-${Date.now()}`;
 
+  // We could use a :scope selector here, but we want to be IE compliant
+  // so we add a dummy css class to be able to select the children
   try {
     context.classList.add(tag);
 
--- a/src/utils/dom/querySelectorAll.js
+++ b/src/utils/dom/querySelectorAll.js
@@ -19,15 +19,13 @@ export default (context, selector) => {
     return toArray(context.querySelectorAll(selector));
   }
 
-  const tag = `alloy${+new Date()}`;
+  const tag = `alloy-${Date.now()}`;
 
-  try { // adding a dummy css class to be able to select the children
-    context.classList.add(tag); // needs to be IE compliant
+  try {
+    context.classList.add(tag);
 
     return toArray(context.querySelectorAll(`.${tag} ${selector}`));
-  } catch (e) {
-    throw e;
   } finally {
-    context.classList.remove(tag); // needs to be IE compliant
+    context.classList.remove(tag);
   }
 };
--- a/src/utils/dom/querySelectorAll.js
+++ b/src/utils/dom/querySelectorAll.js
@@ -12,5 +12,22 @@ governing permissions and limitations under the License.
 
 import toArray from "../toArray";
 
-export default (context, selector) =>
-  toArray(context.querySelectorAll(selector));
+const SIBLING_PATTERN = /^\s*>/;
+
+export default (context, selector) => {
+  if (!SIBLING_PATTERN.test(selector)) {
+    return toArray(context.querySelectorAll(selector));
+  }
+
+  const tag = `alloy${+new Date()}`;
+
+  try { // adding a dummy css class to be able to select the children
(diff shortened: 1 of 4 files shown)


src/utils/dom/querySelectorAll.js
```diff
@@ -12,5 +12,22 @@ governing permissions and limitations under the License.
...
-export default (context, selector) =>
-  toArray(context.querySelectorAll(selector));
+const SIBLING_PATTERN = /^\s*>/;
+
+export default (context, selector) => {
+  if (!SIBLING_PATTERN.test(selector)) {
+    return toArray(context.querySelectorAll(selector));
+  }
+
+  const tag = `alloy${+new Date()}`;
```
jonsnyder, line 22: Although equivalent, it is more clear to change this to `alloy${new Date().getTime()}`.

src/utils/dom/querySelectorAll.js
```diff
@@ -12,5 +12,22 @@ governing permissions and limitations under the License.
...
+
+  try { // adding a dummy css class to be able to select the children
```
jonsnyder, line 24: // We could use a :scope selector here, but we want to be IE compliant
// so we add a dummy css class to be able to select the children

src/utils/dom/querySelectorAll.js
```diff
@@ -12,5 +12,22 @@ governing permissions and limitations under the License.
...
+    context.classList.add(tag); // needs to be IE compliant
```
jonsnyder, line 25: I'm not sure what you mean by this. From what I can tell classList.add/remove is supported in IE 10 and above.

jonsnyder (approved):

test/unit/specs/utils/dom/selectNodesWithShadow.spec.js
```diff
@@ -136,4 +136,24 @@ describe("Utils::DOM::selectNodesWithShadow", () => {
...
+    const content = `
+      <div>
+        <div>
+          <span id="wrong"></span>
+        </div>
+        <span id="right"></span>
+      </div>
+    `;
+
+    const node = createNode("DIV", { id: "target" }, { innerHTML: content });
```
paul-bjorkstrand, line 150: I think this needs an `eq` class to allow the `afterEach` on [lines 75-77][1] to automatically clean up the markup that was added (so other tests function correctly).

Perhaps, this entire file of needs a refactor to make the purpose of the `eq` class more apparent (perhaps by renaming it to `auto-cleanup`, or change the `afterEach` to just empty the `body`, to ensure tests are executed in isolation?

[1]: https://github.com/paul-bjorkstrand/alloy/blob/main/test/unit/specs/components/Personalization/dom-actions/dom/selectNodesWithEq.spec.js#L75...L77

src/utils/dom/querySelectorAll.js
ninaceban: @jonsnyder I have replaced it with `Date.now()` as far as I saw it is supported in IE 9+ and all other modern browsers.

src/utils/dom/querySelectorAll.js
ninaceban: @jonsnyder it was a placeholder for me to double check IE compatibility. According to MDN it is supported in IE 10+ and modern browsers. I have removed the comments.

test/unit/specs/utils/dom/selectNodesWithShadow.spec.js
ninaceban: @paul-bjorkstrand thanks for suggestion. Adjusted both `selectNodesWithEq.spec.js` and `selectNodesWithShadow.spec.js` to use `cleanup` CSS class to ensure `afterEach` get rids of all the elements that have been added during test run.

carterworks (approved):
Looks good to me. We can wait for @XDex to give the final word if you want.

PR #859, opened 2022-04-26
Fixed an issue where multiple persistent iframes were created for id syncs and segment destinations

carterworks:
## Description

When firing off a referrer-hideable image for destinations and ID syncs, if there is not an `iframe` in memory, look in the `document` for one before creating a new one.

## Related Issue

[PDCL-8050 comment from Ian Ramsey](https://jira.corp.adobe.com/browse/PDCL-8050?focusedCommentId=30000443&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-30000443). 

Version 2.10.0 and PR #848 was supposed to restrict the Web SDK from creating one `iframe` per ID sync/segment destination, but the customer deployed it to [their beta site](https://beta-sports.bwin.com/en/sports) (connect from a UK VPN location) and the issue was still there–many `iframe[name="Adobe Alloy"]` elements appeared.

## Motivation and Context

I am at a loss for why #848 did not solve the issue, but here we are.

<details>
    <summary>Screenshot of persistent issue on 2.10.0</summary>
    Screenshot of persistent issue on 2.10.0
</details>

## Types of changes

## Checklist:

### DIFF
--- a/src/utils/injectFireReferrerHideableImage.js
+++ b/src/utils/injectFireReferrerHideableImage.js
@@ -14,12 +14,14 @@ import fireImageInDocument from "./fireImage";
 import {
   appendNode as appendNodeToDocument,
   awaitSelector as awaitSelectorInDocument,
-  createNode as createNodeInDocument
+  createNode as createNodeInDocument,
+  querySelectorAll as querySelectorAllInDocument
 } from "./dom";
 import { BODY, IFRAME } from "../constants/tagName";
 
 const IFRAME_ATTRS = {
-  name: "Adobe Alloy"
+  name: "Adobe Alloy",
+  id: "alloy-sync-frame"
 };
 
 const IFRAME_PROPS = {
@@ -34,7 +36,8 @@ export default ({
   appendNode = appendNodeToDocument,
   awaitSelector = awaitSelectorInDocument,
   createNode = createNodeInDocument,
-  fireImage = fireImageInDocument
+  fireImage = fireImageInDocument,
+  querySelectorAll = querySelectorAllInDocument
 } = {}) => {
   const fireOnPage = fireImage;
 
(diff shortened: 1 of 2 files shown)


src/utils/injectFireReferrerHideableImage.js
```diff
@@ -44,6 +47,11 @@ export default ({
     if (hiddenIframe) {
       return Promise.resolve(hiddenIframe);
     }
+    const potentialHiddenFrames = selectNodes(`#${IFRAME_ATTRS.id}`);
+    if (potentialHiddenFrames.length > 0) {
+      hiddenIframe = potentialHiddenFrames[0];
+      return Promise.resolve(hiddenIframe);
+    }
     return awaitSelector(BODY).then(([body]) => {
       hiddenIframe = createNode(IFRAME, IFRAME_ATTRS, IFRAME_PROPS);
```
jonsnyder, line 56: I wonder if you need to move the `if (hiddenIframe)` to be inside this then. There may be multiple requests for firing an image before the `awaitSelector(BODY)` resolves. I'd like to avoid adding an ID to the page.

jonsnyder (changes requested):

src/utils/injectFireReferrerHideableImage.js
carterworks: > I wonder if you need to move the if (hiddenIframe) to be inside this then.

Are you suggesting something like this?

```js
return awaitSelector(BODY).then(([body]) => {
  hiddenIframe =
    hiddenIframe ?? createNode(IFRAME, IFRAME_ATTRS, IFRAME_PROPS);
  return appendNode(body, hiddenIframe);
});
```

> There may be multiple requests for firing an image before the `awaitSelector(BODY)` resolves.

That's an excellent point. I didn't think about that, but multiple requests getting to the `awaitSelector(BODY)` before the `iframe` is created makes a lot of sense.

> I'd like to avoid adding an ID to the page.

Why is that? I created an ID because I thought that selecting by ID would be a lot faster than querying by `iframe[name="Adobe Alloy"]`.

carterworks: You can see the error if you visit https://beta-sports.bwin.com/en/sports and search for the `iframe[name="Adobe Alloy"]` selector.

carterworks: The sandbox has it as well.

carterworks: I've tested this most recent version on the sandbox and it works. 
I guess I was getting a false positive race condition when I was testing #848 in the sandbox 😬

jonsnyder (approved):

shammowla (approved):

PR #862, opened 2022-05-10
Added support for ECIDs passed via the adobe_mc query string parameter. Added appendIdentityToUrl command to enable cross-domain identity sharing. 

jonsnyder:
## Description

Support legacy solution for passing ECID via the query string parameter adobe_mc. https://experienceleague.adobe.com/docs/id-service/using/id-service-api/methods/appendvisitorid.html?lang=en

- Query string param: "adobe_mc"
- Format: pipe delimited key value pairs. (All the keys below are required, extra keys are ignored)
- Key - TS: timestamp in seconds of query string parameter generation. TTL = 5 minutes
- Key - MCORGID: the marketing cloud org id. The configured orgId must match.
- Key - MCMID: the ECID in decimal.
- Identity Map Namespace: "ECID"
- Priority for namespace "ECID":
  - provided value in identityMap parameter of sendEvent
  - value from query string
  - value from AMCV cookie (if idMigrationEnabled and no Konductor identity cookie)
  - value from Visitor (if idMigrationEnabled and no Konductor identity cookie)

For the use case of navigating from one domain to another, the customer can use the "appendIdentityToUrl" command to generate the "adobe_mc" query string parameter. The command accepts an object with one property, "url", and returns an object with the property "url".
If consent has not been given, the url is returned unchanged. This command does not wait for consent.
If an ECID has not yet been established, "/acquire" endpoint will be hit to generate an ECID.

Added idOverwriteEnabled config to control whether the query string identity overwrites the cookie or not.

## Related Issue

https://jira.corp.adobe.com/browse/PDCL-7572

## Types of changes

## Checklist:


src/components/Identity/configValidators.js
```diff
@@ -14,7 +14,8 @@ import { boolean } from "../../utils/validation";
 
 const configValidators = {
   thirdPartyCookiesEnabled: boolean().default(true),
-  idMigrationEnabled: boolean().default(true)
+  idMigrationEnabled: boolean().default(true),
+  idOverwriteEnabled: boolean().default(true)
```
jonsnyder, line 18: Do you think true is the right default?

test/functional/specs/ID Migration/C5752639.js
```diff
@@ -0,0 +1,73 @@
...
+
+test.meta({
+  ID: "C5752639",
+  SEVERITY: "P0",
+  TEST_RUN: "Regression"
+});
+
+const getDocumentCookie = ClientFunction(() => document.cookie);
+
+test.skip("C5752639: Identity can be changed via the adobe_mc query string parameter when id_migration is enabled", async () => {
```
jonsnyder, line 51: This test still fails. Waiting on Konductor changes to work.

src/components/Identity/configValidators.js
jfkhoury: I think false might be a safer bet since that’s how it behaves today in Visitor.js

jfkhoury: Let’s ask Mitch though

carterworks (approved):

PR #863, opened 2022-05-11
Hybrid mode POC

jasonwaters:
## Description

This PR contains code for a POC that implements hybrid mode in alloy.  It's purpose is to illustrate one way that hybrid mode could work.  I would love any and all feedback the team wants to provide.  First around the general approach, second about implementation details.    Ideally this can serve as a starting point to iterate on.

I introduced a new command called `applyHandles`.  This command is similar to `sendEvent` in concept.  sendEvent constructs a konductor request and fulfills it client side.  But `applyHandles` simply accepts an object with an array of `handles`.  The user is simply informing alloy to apply the handles from a response.  By doing so any request to konductor could be made outside of alloy and then it's response handles applied to it -- not just personalization.

Details of Hybrid mode and the approach taken for this PR can be found on [this wiki page](https://wiki.corp.adobe.com/pages/viewpage.action?pageId=2615646420). 

Here is a related PR with an updated demo that makes a konductor request server side and supplies it to alloy using `applyHandles` 👉  https://github.com/jasonwaters/target-nodejs-sdk-samples/pull/4/files

Here is an example of using the `applyHandles` command:

```js
alloy("applyHandles", {
  renderDecisions: true,
  handles: [
    {
      payload: [
        {
          id: "50229536553757344003388121363781298146",
          namespace: {
            code: "ECID",
          },
        },
      ],
      type: "identity:result",
    },
    {
      payload: [
        {
          id: "AT:eyJhY3Rpdml0eUlkIjoiNTYzMTY5IiwiZXhwZXJpZW5jZUlkIjoiMCJ9",
          scope: "__view__",
          scopeDetails: {
            decisionProvider: "TGT",
            activity: {
              id: "563169",
            },
            experience: {
              id: "0",
            },
            strategies: [
              {
                algorithmID: "0",
                trafficType: "0",
              },
            ],
            characteristics: {
              eventToken:
                "4crrKZDKsZPb+Oi/3KLSF2qipfsIHvVzTQxHolz2IpTMromRrB5ztP5VMxjHbs7c6qPG9UF4rvQTJZniWgqbOw==",
            },
          },
          items: [
            {
              id: "0",
              schema: "https://ns.adobe.com/personalization/dom-action",
              data: {
                type: "insertAfter",
                format: "application/vnd.adobe.target.dom-action",
                content: "<div>hi</div>",
                selector: "HTML > BODY > H3:nth-of-type(1)",
                prehidingSelector: "HTML > BODY > H3:nth-of-type(1)",
              },
            },
          ],
        },
        {
          id: "AT:eyJhY3Rpdml0eUlkIjoiNTYzMTcwIiwiZXhwZXJpZW5jZUlkIjoiMCJ9",
          scope: "superfluous",
          scopeDetails: {
            decisionProvider: "TGT",
            activity: {
              id: "563170",
            },
            experience: {
              id: "0",
            },
            strategies: [
              {
                algorithmID: "0",
                trafficType: "0",
              },
            ],
            characteristics: {
              eventToken:
                "3V418lEVL5DtdKGx6oE4WWqipfsIHvVzTQxHolz2IpTMromRrB5ztP5VMxjHbs7c6qPG9UF4rvQTJZniWgqbOw==",
            },
          },
          items: [
            {
              id: "780724",
              schema: "https://ns.adobe.com/personalization/html-content-item",
              data: {
                id: "780724",
                format: "text/html",
                content: "<div>welcome</div>",
              },
            },
          ],
        },
      ],
      type: "personalization:decisions",
      eventIndex: 0,
    },
  ],
}).then((result) => {
  const { decisions = [] } = result;

  decisions.forEach((decision) => {
    const { id, scope, items = [], scopeDetails = {} } = decision;
    sendDisplayEvent(id, scope, scopeDetails);
  });
});

```

## Related Issue

https://jira.corp.adobe.com/browse/TNT-44492

## Motivation and Context

Staples and other target customers rely on hybrid mode in at.js today.  But they want to migrate to alloy/AEP.


jfkhoury:
Thank you @jasonwaters !

I have 2 high level comments:

1. Let's think about having `applyEvent` call `applyPropositions` behind the scenes which would simplify the changes made to the Web SDK. This might require adding a feature to `applyPropositions` to support sending notifications which would make it easier for form based or ODE customers.
2. `applyEvent` should return the response handles the same way as `sendEvent`, to support the use case of customers requesting multiple things on the server, not only personalization propositions.

The immediate things we need to get feedback on from @jonsnyder, @ninaceban & @XDex :

- How do we feel about adding a new `applyEvent` command? If we like it, what's the best naming.
- Do we need the Request & Response as part of the serverState or is that overkill? Maybe only return the Request if debugging is turned on in the Debugger?

Thank you!

jonsnyder:
I met with Jason today to discuss the applyEvents PR. We came up with a number of cons to his approach:

- Including the entire request would include the XDM from the original event which may include data the customer doesn’t want visible on the client
- His PR calls the onBeforeEventSend callback, but changing the XDM there wouldn’t have any affect on the event
- There may be some Server-side edge / Konductor v2 API incompatibilities. (i.e. Web SDK would need to parse a v2 request/response)
- Extra page weight including the request/response interactions
- Cookie lifetime. One of the benefits of calling edge from the server is that it would be able to write the Konductor cookies first-party using a DNS A-record. This means the cookies will last longer than ones set from javascript or from Konductor directly. If the entire request/response interaction was processed by Web SDK, it would re-write the cookies from the state:store handles, possibly shortening the cookie lifetimes. (I'm not sure how browsers handle doubly written cookies).

Joe had the idea of using applyPropositions to handle hybrid use cases. This may work as a short-term solution for some customers, but there are limitations to this approach:

- applyPropositions does not populate the view cache. With this PR, a customer could request SPA offers on the server and then call sendEvent with a viewName to trigger SPA offer rendering.
- destinations, identity, machine learning, and other handles would not be processed.

My proposal is to have an "applyHandles" command that would process the handles from Konductor. One concern with this approach is making sure that all the Web SDK components process handles the same way they process event responses. Jason is going to try to use a lifecycle method to process handles. Currently we have per-event onResponse handlers that process handles.



* Consent: the server may have called set-consent. If that happened

dancingcactus:
So is the idea that you would pass the entire Handle array into the WebSDK? From the developer experience I want to make sure that it is super simple for the developer. Would it make sense to pass back a blob of stuff that we want the SDK to handle so that all the developer has to do is worry about that one variable?

jonsnyder:
Yes, that's what we're thinking. The dead simple way to do it would be to pass the handle array to the webSDK. The complicated part of this is the cookie transfer. The server-side developer needs to forward the cookies to experience edge. And if the customer wants the cookies to live longer, they need to process the handles for the state:store type ones to write the cookies.

ninaceban:
This is a good start Jason, thank you for detailing the functionality.

Although this follows same approach as `serverState` in at.js, I was thinking about a more granular approach.
The customer will call `alloy("configure", {...})` command when configuring the library, but they might call `alloy("setServerState", {...})` command multiple times, depending on their implementation.

In the hybrid approach we expect the customer to make the first call on the server side and be able to render the propositions on the client side to avoid flickering. Currently in WEB SDK we have the `sendEvent` command that sends an event, `applyPropositions` - in progress, which will be responsible just for rendering. We are missing a command that would handle the content retrieved on the server side.

I would suggest adding a new command something like `alloy("setServerState")` that would receive as parameters Konductor handles, just like @jonsnyder mentioned, plus other metadata like cookies, location hint or whatever else might be required. This way we will ensure that consent, identity, personalization, SPA view cache, etc is properly initialized.

This is not necessarily needed here, but since our customers that are using `Hybrid mode` are using A4T as a reporting source, we will need to take that into account.  Currently A4T data exchange for page-wide scope and decision scopes happens during the fetch call, which means that the fetch call will be considered as a `page-view`. To receive analytics payload client side the customer has to enable the `Analytics Client Side logging`, by disabling Analytics.

jasonwaters:
Thank you for your feedback!

I have updated the PR (and the description) based on the feedback/ideas I have heard from @jonsnyder and @ninaceban .  I changed the command from `applyEvents` to `applyHandles`.  The `applyHandles` command takes an array of handles.  Any propositions will be rendered by alloy if the `renderDecisions` flag is true (the same way sendEvent does).  The `applyHandles` command is also similar to sendEvent in that it returns the responses to the `then()` handler which can be used.

In terms of A4T, I think the intent is to make requests for propositions be `prefetch` by default rather than `execute`.  This means that customers will need to send a display event on their own for metric reporting.  This works well for hybrid approach because no display metrics would be automatically captured in the ee request made server side.  Instead the customer implementation would need to handle it.  For instance if the customer server-renders offers, they'd need to send a second request to ee with the display event.  Or if they rely on on alloy to render the propositions client side, alloy could have an option to automatically send the display event at that stage.  Or it could be done manually using `sendEvent` as well.  But so long as the appropriate eventToken and analytics details are in the `characteristics` object, reporting and A4T will work the way is already does.  The change is simply to change when and how that display event is sent.  

The thinking around A4T ☝️ is a WIP, but wanted to share.  Please feel free to share your opinion and ideas as well.

I have also updated [my other PR related to this one](https://github.com/jasonwaters/target-nodejs-sdk-samples/pull/4/files) so that it uses experience edge v2 api and the new `applyHandles` convention.  It also has examples of both a SPA and a simple server-rendered page.

src/components/DataCollector/index.js
```diff
@@ -67,38 +67,25 @@ const createDataCollector = ({ eventManager }) => {
...
-                xdm = {},
-                query = { personalization: { decisionScopes: [] } },
-                data = {}
-              } = fulfilledEvent;
-
-              const decisionScopes = query.personalization.decisionScopes;
+        run: options => {
+          const {
+            renderDecisions = false,
+            decisionScopes,
```
jonsnyder, line 78: Why is decisionScopes necessary? Is there a way we can remove it so that the customer doesn't have to duplicate the data from the server to the client call?

src/core/createEventManager.js
```diff
@@ -93,8 +95,38 @@ export default ({
...
+      const onResponseCallbackAggregator = createCallbackAggregator();
+      const onRequestFailureCallbackAggregator = createCallbackAggregator();
+
+      return lifecycle
+        .onBeforeEvent({
+          event,
+          renderDecisions,
+          decisionScopes,
+          onResponse: onResponseCallbackAggregator.add,
+          onRequestFailure: onRequestFailureCallbackAggregator.add
```
jonsnyder, line 120: There is no way the request could fail because it is not going to be made. This could be a no-op function

src/core/edgeNetwork/injectApplyEdgeResponseHandles.js
```diff
@@ -0,0 +1,39 @@
...
+
+    const onRequestFailureCallbackAggregator = createCallbackAggregator();
+    onRequestFailureCallbackAggregator.add(lifecycle.onRequestFailure);
+    onRequestFailureCallbackAggregator.add(runOnRequestFailureCallbacks);
+
+    return lifecycle
+      .onBeforeRequest({
+        request,
+        onResponse: onResponseCallbackAggregator.add,
+        onRequestFailure: onRequestFailureCallbackAggregator.add
```
jonsnyder, line 23: This can also be a no-op.

jonsnyder (commented):
Identity query - If the response does not include an "identity:result" handle, then there may be problems processing the response. Can we update the error message that occurs in this case?

getIdentity and appendIdentityToUrls - These commands wait for consent before checking if the ECID is available. With Hybrid mode, there is a possibility that the user has not given consent, yet we have the ECID. Should these command operate with the identity or wait for consent?

jasonwaters:
> Identity query - If the response does not include an "identity:result" handle, then there may be problems processing the response. Can we update the error message that occurs in this case?
> 
> getIdentity and appendIdentityToUrls - These commands wait for consent before checking if the ECID is available. With Hybrid mode, there is a possibility that the user has not given consent, yet we have the ECID. Should these command operate with the identity or wait for consent?

@jonsnyder I'm not familiar with the consent workflow.  Is there a sample app where I can see this behavior?  It seems like an ECID is always obtained by alloy without consent.  How can i configure it to require consent?  Just trying to understand the use case so i can test it.

I think the question around requiring consent after an ECID is already obtained is probably for someone else to answer.  Who could best answer it for us?

src/core/createEventManager.js
```diff
@@ -116,6 +116,9 @@ export default ({
         })
         .then(() => {
           payload.addEvent(event);
+          return consent.awaitConsent();
```
jonsnyder, line 119: This is a tricky issue. The problem with this change is that it could lead to multiple ECIDs. Consider this scenario:

1. New user visits page configured with defaultConsent = "pending"
2. Server side call occurs. Konductor generates ECID1.
3. applyHandles is called, but is not processed because consent is pending.
4. setConsent is called. Konductor generates ECID2.
5. Web SDK now processes handles from applyHandles call.

jonsnyder: I think you should go back to not waiting for consent to process the handles.

jasonwaters:
hey @jonsnyder is there a github repo for alloy code samples?  The code sample i put together for this POC was added to the https://github.com/adobe/target-nodejs-sdk-samples repo. But I think it is misplaced there and I'd rather put it along side other AEP / alloy code samples when it's done.  Please advise.

I'll share what we have done in target with the https://github.com/adobe/target-nodejs-sdk-samples.  Basically there are a dozen or so sample apps with documentation for customers and/or consultants to look at and run in order to learn about various implementation strategies.

To illustrate hybrid I think there are several code samples that might help:

1. Sample app using just alloy to display personalization
2. Sample app using server side calls alone to display personalization (server side rendering)
3. Sample app using server side personalization requests and using alloy to render with `applyHandles`

Please let me know if you have more ideas or feedback.

jasonwaters:
This pr is replaced by https://github.com/adobe/alloy/pull/876

PR #864, opened 2022-05-12
TNT-44322 TNT-44442 Support AJO surfaces and propositions

XDex:
## Description

Support Surfaces in personalization query requests and AJO proposition rendering
The first AJO Beta release will only support rendering of MPA page-wide propositions (SPA support will be added in subsequent releases)
Functional e2e tests and test sandbox page added in https://jira.corp.adobe.com/browse/TNT-45510

## Related Issues

https://jira.corp.adobe.com/browse/TNT-44322
https://jira.corp.adobe.com/browse/TNT-44442
https://jira.corp.adobe.com/browse/TNT-45303
https://jira.corp.adobe.com/browse/TNT-45510

## Motivation and Context

Support Surfaces in personalization query requests and AJO proposition rendering
The first AJO Beta release will only support rendering of MPA page-wide propositions (SPA support will be added in subsequent releases)

## Types of changes

## Checklist:

### DIFF
--- a/sandbox/public/perf/personalization.html
+++ b/sandbox/public/perf/personalization.html
@@ -34,7 +34,7 @@
       const onSendEvent = () => {
         alloy("sendEvent", {
           renderDecisions: true,
-          decisionScopes: ["alloy-location-1", "alloy-location-2"],
+          decisionScopes: ["alloy-location-1", "alloy-location-2"], // Note: this option will soon be deprecated, please use personalization.decisionScopes instead
           xdm: {
             device: {
               screenHeight: 1
--- a/sandbox/src/PersonalizationFormBased.js
+++ b/sandbox/src/PersonalizationFormBased.js
@@ -13,7 +13,10 @@ const metadata = {
 
 const usePropositions = () => {
   const [propositions, setPropositions] = useState(undefined);
-  useSendPageViewEvent({ setPropositions, decisionScopes: SCOPES_FOR_PAGE });
+  useSendPageViewEvent({
+    setPropositions,
+    decisionScopes: SCOPES_FOR_PAGE // Note: this option will soon be deprecated, please use personalization.decisionScopes instead
+  });
   useEffect(() => {
     if (propositions) {
       window.alloy("applyPropositions", {
--- a/sandbox/src/personalizationAnalyticsClientSideHelper.js
+++ b/sandbox/src/personalizationAnalyticsClientSideHelper.js
@@ -63,7 +63,7 @@ const sendEvent = ({
 
   return window[instanceName]("sendEvent", {
     renderDecisions,
-    decisionScopes,
+    decisionScopes, // Note: this option will soon be deprecated, please use personalization.decisionScopes instead
     xdm
   });
 };
@@ -71,7 +71,7 @@ const sendEvent = ({
 export const getFormBasedOffer = () => {
   sendEvent({
     eventType: "form-based-offer",
-    decisionScopes: ["a4t-test"]
+    decisionScopes: ["a4t-test"] // Note: this option will soon be deprecated, please use personalization.decisionScopes instead
   }).then(result => {
     if (!result.propositions) {
       return;
--- a/sandbox/src/useSendPageViewEvent.js
+++ b/sandbox/src/useSendPageViewEvent.js
@@ -35,7 +35,7 @@ export default ({
 
     window[instanceName]("sendEvent", {
       renderDecisions: true,
-      decisionScopes,
+      decisionScopes, // Note: this option will soon be deprecated, please use personalization.decisionScopes instead
       xdm,
       data
     }).then(res => {
--- a/src/components/DataCollector/index.js
+++ b/src/components/DataCollector/index.js
@@ -29,7 +29,8 @@ const createDataCollector = ({ eventManager }) => {
             type,
             mergeId,
             renderDecisions = false,
-            decisionScopes = [],
+            decisionScopes = [], // Note: this option will soon be deprecated, please use personalization.decisionScopes instead
+            personalization = {},
             datasetId
           } = options;
           const event = eventManager.createEvent();
@@ -63,7 +64,8 @@ const createDataCollector = ({ eventManager }) => {
 
           return eventManager.sendEvent(event, {
             renderDecisions,
-            decisionScopes
+            decisionScopes,
+            personalization
           });
         }
       },
--- a/src/components/DataCollector/validateUserEventOptions.js
+++ b/src/components/DataCollector/validateUserEventOptions.js
@@ -28,6 +28,10 @@ export default ({ options }) => {
     documentUnloading: boolean(),
     renderDecisions: boolean(),
     decisionScopes: arrayOf(string()).uniqueItems(),
+    personalization: objectOf({
+      decisionScopes: arrayOf(string()).uniqueItems(),
+      surfaces: arrayOf(string()).uniqueItems()
+    }),
     datasetId: string(),
     mergeId: string()
   })
--- a/src/components/Personalization/createGetPageLocation.js
+++ b/src/components/Personalization/createGetPageLocation.js
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 Adobe. All rights reserved.
+Copyright 2022 Adobe. All rights reserved.
 This file is licensed to you under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License. You may obtain a copy
 of the License at http://www.apache.org/licenses/LICENSE-2.0
@@ -10,9 +10,6 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-/**
- * Returns true when the value is null.
- * @param {*} value
- * @returns {boolean}
- */
-export default value => value == null;
+export default ({ window }) => () => {
+  return window.location;
+};
--- a/src/components/Personalization/createPersonalizationDetails.js
+++ b/src/components/Personalization/createPersonalizationDetails.js
@@ -10,7 +10,8 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import { includes } from "../../utils";
+import { includes, isNonEmptyString, isNonEmptyArray } from "../../utils";
+import { buildPageSurface, normalizeSurfaces } from "./utils/surfaceUtils";
 import PAGE_WIDE_SCOPE from "./constants/scope";
 import {
   DEFAULT_CONTENT_ITEM,
@@ -19,9 +20,29 @@ import {
   JSON_CONTENT_ITEM,
   REDIRECT_ITEM
 } from "./constants/schema";
-import isNonEmptyString from "../../utils/isNonEmptyString";
 
-export default ({ renderDecisions, decisionScopes, event, viewCache }) => {
+const addPageWideScope = scopes => {
+  if (!includes(scopes, PAGE_WIDE_SCOPE)) {
+    scopes.push(PAGE_WIDE_SCOPE);
+  }
+};
+
+const addPageSurface = (surfaces, getPageLocation) => {
+  const pageSurface = buildPageSurface(getPageLocation);
+  if (!includes(surfaces, pageSurface)) {
+    surfaces.push(pageSurface);
+  }
+};
+
+export default ({
+  getPageLocation,
+  renderDecisions,
+  decisionScopes,
+  personalization,
+  event,
+  viewCache,
+  logger
+}) => {
   const viewName = event.getViewName();
   return {
     isRenderDecisions() {
@@ -31,15 +52,31 @@ export default ({ renderDecisions, decisionScopes, event, viewCache }) => {
       return viewName;
     },
     hasScopes() {
-      return decisionScopes.length > 0;
+      return (
+        decisionScopes.length > 0 ||
+        isNonEmptyArray(personalization.decisionScopes)
+      );
+    },
+    hasSurfaces() {
+      return isNonEmptyArray(personalization.surfaces);
     },
     hasViewName() {
       return isNonEmptyString(viewName);
     },
     createQueryDetails() {
       const scopes = [...decisionScopes];
-      if (!this.isCacheInitialized() && !includes(scopes, PAGE_WIDE_SCOPE)) {
-        scopes.push(PAGE_WIDE_SCOPE);
+      if (isNonEmptyArray(personalization.decisionScopes)) {
+        scopes.push(...personalization.decisionScopes);
+      }
+      const eventSurfaces = normalizeSurfaces(
+        personalization.surfaces,
+        getPageLocation,
+        logger
+      );
+
+      if (!this.isCacheInitialized()) {
+        addPageWideScope(scopes);
+        addPageSurface(eventSurfaces, getPageLocation);
       }
 
       const schemas = [
@@ -55,14 +92,17 @@ export default ({ renderDecisions, decisionScopes, event, viewCache }) => {
 
       return {
         schemas,
-        decisionScopes: scopes
+        decisionScopes: [...new Set(scopes)],
+        surfaces: [...new Set(eventSurfaces)]
       };
     },
     isCacheInitialized() {
       return viewCache.isInitialized();
     },
     shouldFetchData() {
-      return this.hasScopes() || !this.isCacheInitialized();
+      return (
+        this.hasScopes() || this.hasSurfaces() || !this.isCacheInitialized()
+      );
     },
     shouldUseCachedData() {
       return this.hasViewName() && this.isCacheInitialized();
--- a/src/components/Personalization/groupDecisions.js
+++ b/src/components/Personalization/groupDecisions.js
@@ -11,6 +11,7 @@ governing permissions and limitations under the License.
 */
 
 import { isNonEmptyArray, includes } from "../../utils";
+import isPageWideScope from "./utils/isPageWideScope";
 import {
   DOM_ACTION,
   REDIRECT_ITEM,
@@ -18,7 +19,6 @@ import {
   MEASUREMENT_SCHEMA
 } from "./constants/schema";
 import { VIEW_SCOPE_TYPE } from "./constants/scopeType";
-import PAGE_WIDE_SCOPE from "./constants/scope";
 
 const splitItems = (items, schemas) => {
   const matched = [];
@@ -93,7 +93,7 @@ const extractDecisionsByScope = decisions => {
 
   if (isNonEmptyArray(decisions)) {
     decisions.forEach(decision => {
-      if (decision.scope === PAGE_WIDE_SCOPE) {
+      if (isPageWideScope(decision.scope)) {
         pageWideScopeDecisions.push(decision);
       } else if (isViewScope(decision.scopeDetails)) {
         appendScopeDecision(viewScopeDecisions, decision);
--- a/src/components/Personalization/utils/surfaceUtils.js
+++ b/src/components/Personalization/utils/surfaceUtils.js
@@ -1,0 +1,118 @@
+/*
+Copyright 2022 Adobe. All rights reserved.
+This file is licensed to you under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License. You may obtain a copy
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+import {
+  WEB,
+  WEBAPP,
+  SURFACE_TYPE_DELIMITER,
+  FRAGMENT_DELIMITER
+} from "../constants/surface";
+import { startsWith, isNil, isNonEmptyString, includes } from "../../../utils";
+
+const SURFACE_REGEX = /^(\w+):\/\/([^/#]+)(\/[^#]*)?(#.*)?$/;
+const AUTHORITY_REGEX = /^(?:.*@)?(?:[a-z\d\u00a1-\uffff.-]+|\[[a-f\d:]+])(?::\d+)?$/;
+const PATH_REGEX = /^\/(?:[/\w\u00a1-\uffff-.~]|%[a-fA-F\d]{2})*$/;
+const FRAGMENT_REGEX = /^#(?:[/\w\u00a1-\uffff-.~]|%[a-fA-F\d]{2})+$/;
+
+const normalizePath = (path = "/") => {
+  let end = path.length;
+  while (end > 0 && "/".indexOf(path.charAt(end - 1)) !== -1) {
+    end -= 1;
+  }
+  return path.substring(0, end) || "/";
+};
+
+const getSurfaceType = surfaceTypeMatch =>
+  isNonEmptyString(surfaceTypeMatch) ? surfaceTypeMatch.toLowerCase() : "";
+
+const getAuthority = authorityMatch =>
+  isNonEmptyString(authorityMatch) ? authorityMatch.toLowerCase() : "";
+
+const getPath = pathMatch =>
+  isNonEmptyString(pathMatch) ? normalizePath(pathMatch) : "/";
+
+const parseSurface = surfaceString => {
+  const matched = surfaceString.match(SURFACE_REGEX);
+  return matched
+    ? {
+        surfaceType: getSurfaceType(matched[1]),
+        authority: getAuthority(matched[2]),
+        path: getPath(matched[3]),
+        fragment: matched[4]
+      }
+    : null;
+};
+
+const stringifySurface = surface =>
+  `${surface.surfaceType}${SURFACE_TYPE_DELIMITER}${
+    surface.authority
+  }${surface.path || ""}${surface.fragment || ""}`;
+
+export const buildPageSurface = getPageLocation => {
+  const location = getPageLocation();
+  const host = location.host.toLowerCase();
+  const path = location.pathname;
+  return WEB + SURFACE_TYPE_DELIMITER + host + normalizePath(path);
+};
+
+const expandFragmentSurface = (surface, getPageLocation) =>
+  startsWith(surface, FRAGMENT_DELIMITER)
+    ? buildPageSurface(getPageLocation) + surface
+    : surface;
+
+const validateSurface = (surface, getPageLocation, logger) => {
+  const invalidateSurface = validationError => {
+    logger.warn(validationError);
+    return null;
+  };
+
+  if (!isNonEmptyString(surface)) {
+    return invalidateSurface(`Invalid surface: ${surface}`);
+  }
+  const expanded = expandFragmentSurface(surface, getPageLocation);
+  const parsed = parseSurface(expanded);
+  if (parsed === null) {
+    return invalidateSurface(`Invalid surface: ${surface}`);
+  }
+  if (!includes([WEB, WEBAPP], parsed.surfaceType)) {
+    return invalidateSurface(
+      `Unsupported surface type ${parsed.surfaceType} in surface: ${surface}`
+    );
+  }
+  if (!parsed.authority || !AUTHORITY_REGEX.test(parsed.authority)) {
+    return invalidateSurface(
+      `Invalid authority ${parsed.authority} in surface: ${surface}`
+    );
+  }
+  if (parsed.path && !PATH_REGEX.test(parsed.path)) {
+    return invalidateSurface(
+      `Invalid path ${parsed.path} in surface: ${surface}`
+    );
+  }
+  if (parsed.fragment && !FRAGMENT_REGEX.test(parsed.fragment)) {
+    return invalidateSurface(
+      `Invalid fragment ${parsed.fragment} in surface: ${surface}`
+    );
+  }
+  return parsed;
+};
+
+export const isPageWideSurface = scope =>
+  !!scope &&
+  scope.indexOf(WEB + SURFACE_TYPE_DELIMITER) === 0 &&
+  scope.indexOf(FRAGMENT_DELIMITER) === -1;
+
+export const normalizeSurfaces = (surfaces = [], getPageLocation, logger) =>
+  surfaces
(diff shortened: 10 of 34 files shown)


src/components/Personalization/createPersonalizationDetails.js
```diff
@@ -20,8 +20,19 @@ import {
   REDIRECT_ITEM
 } from "./constants/schema";
 import isNonEmptyString from "../../utils/isNonEmptyString";
+import { WEB_SURFACE } from "./constants/surface";
+import getPageLocation from "./utils/getPageLocation";
 
-export default ({ renderDecisions, decisionScopes, event, viewCache }) => {
+const buildPageSurface = window => WEB_SURFACE + getPageLocation(window);
```
jonsnyder, line 26: Rather than passing window everywhere, you could pass this function that was already injected with window. i.e.
`components/Personalization/createBuildPageSurface.js`

```
import { WEB_SURFACE } from "./constants/surface";

export default ({ window }) => () => {
  const url = win.location.href;
  return WEB_SURFACE + url.substring(url.indexOf(":") + 3);
};
```

src/components/Personalization/groupDecisions.js
```diff
@@ -86,14 +88,21 @@ const isViewScope = scopeDetails =>
   scopeDetails.characteristics.scopeType &&
   scopeDetails.characteristics.scopeType === VIEW_SCOPE_TYPE;
 
-const extractDecisionsByScope = decisions => {
+const isSurfaceScope = scope => scope.indexOf(WEB_SURFACE) !== -1;
+
+// only exactly matching page surfaces supported in initial AJO Beta release
+const matchesPage = (win, surface) => surface === WEB_SURFACE + getPageLocation(win);
+
+const isMatchingSurfaceScope = (win, scope) => isSurfaceScope(scope) && matchesPage(win, scope);
```
jonsnyder, line 96: The hash part of the page url could have changed in the time between request and processing. Would that be okay?

jonsnyder (commented):
As we start supporting AJO and Target, I'd like to keep things better separated and more modular. This code will turn into a huge mess if we don't. We could even separate VEC and form-based offers. Certainly there is code that can be shared between AJO and Target, but lets try to make it so the differences are separated.

Here is one idea on how to make this more modular. Let me know what you think.

Build a chain of renderers or processors that handle the returned propositions. ie
- Redirect handler
- __view__ scope Form-based handler
- __view__ scope VEC handler
- SPA scope handler
- AJO surface handler

We could also have a set of functions that modify the request which could add the surface.

src/components/Personalization/groupDecisions.js
XDex: > The hash part of the page url could have changed in the time between request and processing. Would that be okay?

@jonsnyder As the initial AJO Beta release will only support MPA surfaces, for SPA use cases content will be properly rendered only for the first SPA page matching the surface URI. There won't be any re-rendering on subsequent client-side SPA view changes.

ninaceban (changes requested):
Thank you, Andrei!
Here are a few suggestions that I have:
- Surface should be part of the `scopeDetails` xdm schema, so that it gets collected in Platform
- Even if the returned proposition is coming from AJO, it should have `__view__` scope so that Alloy doesn't have to know about AJO. It will continue to auto render the page-wide scope, DOM_ACTION schema propositions that are coming from AJO (in future we could support SPA view scope )

src/components/DataCollector/index.js
```diff
@@ -62,7 +63,8 @@ const createDataCollector = ({ eventManager }) => {
 
           return eventManager.sendEvent(event, {
             renderDecisions,
-            decisionScopes
+            decisionScopes,
+            surfaces
```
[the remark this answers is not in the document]
XDex: @ninaceban Yes, Erik asked for us to support this from the get-go.

src/components/Personalization/createGetPageLocation.js
```diff
@@ -0,0 +1,16 @@
...
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+export default ({ window }) => () => {
+  const url = window.location.href;
+  return url.substring(url.indexOf(":") + 3);
```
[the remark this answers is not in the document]
XDex: @ninaceban This is rather an edge-case, since very few if any customers would be running their prod sites on non-standard ports, but yes, I think the port should be part of surface URI for the web channel.

src/components/Personalization/createPersonalizationDetails.js
XDex: Fixed

XDex:
@ninaceban Thank you for the review!
Regarding Artur's suggestion on having a separate `surface` proposition field, and have `scope` set to `__view__` for page-wide surfaces, I have detailed it [in the wiki](https://wiki.corp.adobe.com/pages/viewpage.action?pageId=2624577900), but we still need to approve this approach with Erik and Pete. Will implement the mods, once these are green-lit.

src/components/Personalization/createGetPageLocation.js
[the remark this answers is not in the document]
XDex: port is part of the surface id, as per [Surfaces pattern matching spec](https://wiki.corp.adobe.com/display/CJM/Surfaces+pattern-matching+implementation) 
Also added a test for this

XDex:
@artur-ciocanu @ninaceban @jonsnyder I've consolidated and updated the PR according to the latest requirements documented in [AJO Web SDK integration wiki](https://wiki.corp.adobe.com/display/CJM/AJO+Web+Channel+-+Web+SDK+integration#AJOWebChannelWebSDKintegration-Conclusion). Please take another look when you have some spare time. Thanks!

5amfung:
@XDex Is this PR in good state that @jonsnyder can approve it in the comment here?  I know this won't be merged as we'll have a separate build for alpha testing.  But it would be great to have some status to reflect that we're complete.

XDex:
@jonsnyder @ninaceban I have implemented the latest requested mods to sendEvent API, as per Nina's wiki and [TNT-45303](https://jira.corp.adobe.com/browse/TNT-45303), please take another look when you get a chance. Thanks!
Note: ES/EDC/AJO backend should be finally available in a staging env soon, so I'm hoping to finally be able to also add a functional test sometime this week..
cc: @5amfung @jfkhoury

carterworks:
What is the difference between the top-level `decisionScopes` and `personalization.decisionScopes`? Is the top-level for Target and the other for AJO? Are they identical?

sandbox/public/perf/personalization.html
```diff
@@ -34,7 +34,9 @@
       const onSendEvent = () => {
         alloy("sendEvent", {
           renderDecisions: true,
```
ninaceban, line 36: I would create a separate page and leave this one as is.

sandbox/src/PersonalizationFormBased.js
```diff
@@ -13,7 +13,10 @@ const metadata = {
 
 const usePropositions = () => {
   const [propositions, setPropositions] = useState(undefined);
```
ninaceban, line 15: Same here, I would have a dedicated page for the new `personalization -> surface` functionality. We could deprecate them in a separate PR.

sandbox/src/personalizationAnalyticsClientSideHelper.js
```diff
@@ -64,14 +65,17 @@ const sendEvent = ({
...
     renderDecisions,
     decisionScopes,
+    personalization,
     xdm
   });
 };
 
 export const getFormBasedOffer = () => {
   sendEvent({
     eventType: "form-based-offer",
```
ninaceban, line 75: Same here, let's leave everything as is.

sandbox/src/useSendPageViewEvent.js
```diff
@@ -19,7 +19,9 @@ export default ({
   viewName,
   data = {},
   xdm = {},
```
ninaceban, line 21: Same here.

src/components/DataCollector/validateUserEventOptions.js
```diff
@@ -27,6 +27,10 @@ export default ({ options }) => {
     data: objectOf({}),
     renderDecisions: boolean(),
     decisionScopes: arrayOf(string()).uniqueItems(),
```
ninaceban, line 29: I am wondering if we shouldn't make `decisionScopes` and `personalization` mutually exclusive

src/components/Personalization/createPersonalizationDetails.js
```diff
@@ -10,7 +10,8 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import { includes } from "../../utils";
+import { includes, isNonEmptyString, isNonEmptyArray } from "../../utils";
```
ninaceban, line 13: Any reason we can't have `buildPageSurface, normalizeSurfaces` directly in `./utils`? You can still have a separate file, but it is easier to control everything through a top level file/module.

src/components/Personalization/groupDecisions.js
```diff
@@ -11,6 +11,7 @@ governing permissions and limitations under the License.
 */
 
 import { isNonEmptyArray, includes } from "../../utils";
```
ninaceban, line 13: Same here, any reason we can't have `isPageWideSurface` in `./utils`.

src/components/Personalization/groupDecisions.js
```diff
@@ -86,14 +87,17 @@ const isViewScope = scopeDetails =>
   scopeDetails.characteristics.scopeType &&
   scopeDetails.characteristics.scopeType === VIEW_SCOPE_TYPE;
```
ninaceban, line 89: We could move this to `./utils`

src/components/Personalization/utils/surfaceUtils.js
```diff
@@ -0,0 +1,113 @@
...
+const normalizePath = (path = "/") => {
+  let end = path.length;
+  while (end > 0 && "/".indexOf(path.charAt(end - 1)) !== -1) {
+    end -= 1;
+  }
+  return path.substring(0, end) || "/";
+};
+
+const parseSurface = surfaceString => {
+  const matched = surfaceString.match(SURFACE_REGEX);
```
ninaceban, line 35: This looks really, really, really bad. Could we use `if/else` and maybe extract some code to helper functions. Otherwise this code is pretty cryptic.

ninaceban: Also instead of calling `toLowerCase()` you could probably, make sure the initial value is lower cased.

src/components/Personalization/utils/surfaceUtils.js
```diff
@@ -0,0 +1,113 @@
...
+    );
+  }
+  return parsed;
+};
+
+export const isPageWideSurface = scope =>
+  !!scope &&
+  scope.indexOf(WEB + SURFACE_TYPE_DELIMITER) === 0 &&
+  scope.indexOf(FRAGMENT_DELIMITER) === -1;
+
```
ninaceban, line 108: I am not following, why we need this dependency injection here. Can't we send just the `surfaces` and page location and return a complex object like {surfaces, validation errors}. The caller of this function can decide to log or do something else. Otherwise `normalizeSurfaces` looks like an utility, but does a lot of stuff including logging. Testing all this via mocking etc, should be pretty painful.

test/functional/specs/Personalization/C28756.js
```diff
@@ -32,7 +32,9 @@ test("Test C28756: A form based offer should return if event command contains it
   const alloy = createAlloyProxy();
   await alloy.configure(config);
   const result = await alloy.sendEvent({
```
ninaceban, line 34: Revert, let's not touch existing tests in this PR.

test/functional/specs/Personalization/C44363.js
```diff
@@ -27,7 +27,9 @@ test("Test C44363: Return proposition when QA mode set up with token of experien
 
   const result = await alloy.sendEvent({
     renderDecisions: true,
```
ninaceban, line 29: Same here.

test/functional/specs/Personalization/C44363.js
```diff
@@ -46,7 +48,9 @@ test("Test C44363: Return proposition  when QA mode set up with token of experie
   });
   const result = await alloy.sendEvent({
     renderDecisions: true,
```
ninaceban, line 50: And here.

test/functional/specs/Personalization/C5805676.js
```diff
@@ -54,7 +54,9 @@ test("Test C5805676: Merged metric propositions should be delivered", async () =
   const alloy = createAlloyProxy();
   await alloy.configure(config);
   const eventResult = await alloy.sendEvent({
```
ninaceban, line 56: And here

test/functional/specs/Personalization/C6364799.js
```diff
@@ -52,7 +52,9 @@ test("Test C6364799: applyPropositions should render html-content-item schemas w
   const alloy = createAlloyProxy();
   await alloy.configure(config);
   const result = await alloy.sendEvent({
```
ninaceban, line 54: And here

test/functional/specs/Personalization/C782718.js
```diff
@@ -146,7 +146,6 @@ const simulateViewChange = async (alloy, personalizationPayload) => {
   // sendEvent at a view change, this shouldn't request any target data, it should use the existing cache
   const resultingObject = await alloy.sendEvent({
     renderDecisions: true,
```
ninaceban, line 148: Please revert.

test/functional/specs/Personalization/C782719.js
```diff
@@ -88,7 +88,6 @@ const simulateViewChange = async alloy => {
   // sendEvent at a view change, this shouldn't request any target data, it should use the existing cache
   const resultingObject = await alloy.sendEvent({
     renderDecisions: false,
```
ninaceban, line 90: Please revert

test/unit/specs/components/DataCollector/index.spec.js
```diff
@@ -60,7 +60,8 @@ describe("Event Command", () => {
       expect(event.setUserData).toHaveBeenCalledWith(data);
       expect(eventManager.sendEvent).toHaveBeenCalledWith(event, {
         renderDecisions: true,
```
ninaceban, line 62: Reevert

test/unit/specs/components/DataCollector/index.spec.js
```diff
@@ -69,13 +70,39 @@ describe("Event Command", () => {
   it("sends event with decisionScopes parameter when decisionScopes is not empty", () => {
     const options = {
       renderDecisions: true,
```
ninaceban, line 72: Revert

test/unit/specs/components/DataCollector/index.spec.js
```diff
@@ -69,13 +70,39 @@ describe("Event Command", () => {
...
-      decisionScopes: ["Foo1", "Foo2"]
+      decisionScopes: ["Foo1"],
+      personalization: {
+        decisionScopes: ["Foo2"]
+      }
     };
 
     return sendEventCommand.run(options).then(result => {
       expect(eventManager.sendEvent).toHaveBeenCalledWith(event, {
         renderDecisions: true,
```
ninaceban, line 81: Revert

test/unit/specs/components/DataCollector/index.spec.js
```diff
@@ -91,7 +118,8 @@ describe("Event Command", () => {
     return sendEventCommand.run({}).then(() => {
       expect(eventManager.sendEvent).toHaveBeenCalledWith(event, {
         renderDecisions: false,
```
ninaceban, line 120: Revert

test/unit/specs/components/Personalization/createApplyPropositions.spec.js
```diff
@@ -53,13 +53,13 @@ describe("Personalization::createApplyPropositions", () => {
     executeDecisions.and.returnValue(
       Promise.resolve(PAGE_WIDE_SCOPE_DECISIONS)
     );
```
ninaceban, line 55: Please changes in this file. It just adds noise. If you want to fix it add a separate PR.

test/unit/specs/components/Personalization/createAutoRenderingHandler.spec.js
```diff
@@ -12,10 +12,14 @@ governing permissions and limitations under the License.
 
 import {
   CART_VIEW_DECISIONS,
-  PAGE_WIDE_SCOPE_DECISIONS_WITH_DOM_ACTION_SCHEMA_ITEMS,
+  PAGE_WIDE_DECISIONS_WITH_DOM_ACTION_SCHEMA_ITEMS,
   SCOPES_FOO1_FOO2_DECISIONS
 } from "./responsesMock/eventResponses";
 import createAutoRenderingHandler from "../../../../../src/components/Personalization/createAutoRenderingHandler";
+import { isPageWideSurface } from "../../../../../src/components/Personalization/utils/surfaceUtils";
```
ninaceban, line 19: As mentioned ensure that everything is imported form `./utils`

test/unit/specs/components/Personalization/createAutoRenderingHandler.spec.js
```diff
@@ -12,10 +12,14 @@ governing permissions and limitations under the License.
...
+
```
ninaceban, line 20: Why we have this function redeclared here, when a similar is present in prod code.

test/unit/specs/components/Personalization/utils/surfaceUtils.spec.js
```diff
@@ -0,0 +1,256 @@
...
+    expect(isPageWideSurface("__view__")).toBeFalse();
+    expect(isPageWideSurface("name")).toBeFalse();
+    expect(isPageWideSurface("web://domain.com")).toBeTrue();
+    expect(isPageWideSurface("web://domain.com/path/page.html")).toBeTrue();
+    expect(isPageWideSurface("web://domain.com#fragment")).toBeFalse();
+    expect(isPageWideSurface("webapp://domain.com")).toBeFalse();
+    expect(isPageWideSurface("webapp://domain.com#view")).toBeFalse();
+  });
+
+  it("expands fragment surfaces", () => {
```
ninaceban, line 69: As mentioned earlier `normalizeSurfaces` could be much simpler if we do not use dep injection.

ninaceban (changes requested):
@XDex I have checked your PR. While functionally it does the right thing there are a lot of changes that are not relevant to adding `personalization -> surfaces`. Let's keep the PR focused.

There are a lot of files and hard to follow, what is relevant and what is additional refactoring. I like the idea of improving code, but let's have separate PRs focused on that.

Also let's cover with tests the functionality you have added. Since adding `personalization -> surfaces` and `personalization -> decisionScopes` is backward compatible, most of the existing test should work without any changes. You could add a separate functional test with `personalization -> surfaces & decisionScopes` just to prove that this combination also works.

sandbox/public/perf/personalization.html
XDex: @ninaceban I've made changes to the sandbox and the tests to prioritize the new `personalization` config, as in the meeting we had on this change it was decided that going forward it is now the recommended way to set `decisionScopes`, and the top-level `decisionScopes` config is now a deprecated one. So, anyone looking through the code/test/sandbox should be aware of this.
@jfkhoury please let me know if my understanding is wrong and we should revert all the mods relating to this.

src/components/Personalization/createPersonalizationDetails.js
XDex: @ninaceban Yes, in a previous comment @jonsnyder requested that we have all surface-related utils encapsulated in a separate module.

src/components/Personalization/utils/surfaceUtils.js
XDex: Sure, I've extracted helper functions
Regarding `toLowerCase()` - I can't make the whole initial value lower case, since path and fragment should still remain case-sensitive

src/components/Personalization/utils/surfaceUtils.js
XDex: This is so we're not leaking any surface normalization-related errors to the caller. Do we really need to handle surface related valdation error logging in `createPersonalizationDetails.createQueryDetails`? At that level, we're only interested in getting valid and normalized surfaces and shouldn't be handling any validation-specific errors.
Testing is also not a problem, as we can just mock `logger.warn` as I've done in the tests

test/unit/specs/components/Personalization/createAutoRenderingHandler.spec.js
XDex: I figured it doesn't make sense to export internal `isPageWideScope` from `groupDecisions` if it's just used in a test and not in actual code.

sandbox/src/PersonalizationFormBased.js
XDex: As I mentioned in the previous comment reply, we do need to also modify all existing sandbox pages to prioritize the new `personalization -> decisionScopes` format. However, as part of AJO tests, I will also be adding a dedicated sandbox page for AJO

test/unit/specs/components/Personalization/utils/surfaceUtils.spec.js
XDex: Replied in the earlier comment

sandbox/public/perf/personalization.html
ninaceban: In the current PR we should only add the `personalization` config to support surfaces, in a following PR we will address the deprecation strategy. Even though we are recommending that the new way of querying personalized content is by sending the `decisionScopes` in the `personalization` config. We still need to be able to validate that the existing functionality works properly. That is why it is so important that we leave the sandbox app and functional tests as are, and add new page/functional tests to test the new behavior.
cc: @jfkhoury

src/components/Personalization/utils/surfaceUtils.js
ninaceban: is it the fragment case-sensitive at authoring mode too?

src/components/Personalization/utils/surfaceUtils.js
```diff
@@ -0,0 +1,113 @@
...
+          ? matched[1].toLowerCase()
+          : "",
+        authority: isNonEmptyString(matched[2]) ? matched[2].toLowerCase() : "",
+        path: isNonEmptyString(matched[3]) ? normalizePath(matched[3]) : "/",
+        fragment: matched[4]
+      }
+    : null;
+};
+
+const stringifySurface = surface =>
```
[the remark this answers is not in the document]
ninaceban: @XDex in this repo we favor string templates, even if other `stackoverflow` users might not share this view.

test/unit/specs/components/Personalization/createAutoRenderingHandler.spec.js
ninaceban: add it in `Personalization/utils` and reuse it in both places.

src/components/Personalization/utils/surfaceUtils.js
```diff
@@ -0,0 +1,118 @@
...
+    end -= 1;
+  }
+  return path.substring(0, end) || "/";
+};
+
+const getSurfaceType = surfaceTypeMatch =>
+  isNonEmptyString(surfaceTypeMatch) ? surfaceTypeMatch.toLowerCase() : "";
+const getAuthority = authorityMatch =>
+  isNonEmptyString(authorityMatch) ? authorityMatch.toLowerCase() : "";
+const getPath = pathMatch =>
```
ninaceban, line 38: and here as well

sandbox/public/perf/personalization.html
jfkhoury: This comment addresses all the `Sandbox` related comments:

We need to tackled to goals here:

1. Test the new `personalization` option.
2. Confirm that the `decisionsScopes` option still works, until we decide to deprecate it. (By the way we need a ticket to do so in v 3.0)

So I suggest we do the following:

1. I agree with @ninaceban, let's leave these pages alone for now; maybe add an inline comment that this `decisionScopes` option will be deprecated in v 3.0.
2. Let's add a brand new AJO Surfaces/Web Channel Sandbox page that showcases the surfaces support and use the new `personalization` option.
3. I would like to see a test/sandbox page/something that showcases the usage of `decisionsScopes` and `personalization.decisionsScopes` and see what happens.

Thank you @XDex !!

XDex: @jfkhoury It's probably a misunderstanding on my end of the decisions taken in the meeting we had on nesting `decisionScopes` in `personalization` - I thought we'll be deprecating the top-level `decisionScopes` in my PR and not a separate one. Let me revert the sandbox mods and add inline comments to the API as you suggest.
Regarding p.2 - I've already added two e2e functional tests for AJO and a dedicated page to the sandbox, please take a look.
On p.3 - I've also previously added a test where we're checking that the top-level `decisionScopes` option is working as expected, either standalone or together with the new `personalization.decisionScopes`

sandbox/src/PersonalizationFormBased.js
XDex: Reverted

sandbox/src/personalizationAnalyticsClientSideHelper.js
XDex: Reverted

sandbox/src/useSendPageViewEvent.js
XDex: Reverted

src/components/DataCollector/validateUserEventOptions.js
XDex: I don't think we should do that, since we want to maintain backwards compatibility at least initially. If we do decide to enforce mutual exclusivity, it should be done as part of a future task for deprecating the top-level `decisionScopes` option.

src/components/Personalization/groupDecisions.js
XDex: Yes, see previous comment.

src/components/Personalization/groupDecisions.js
XDex: Yes, but Jon previously suggested we keep all surface-related utils in a single module

src/components/Personalization/utils/surfaceUtils.js
```diff
@@ -0,0 +1,118 @@
...
+  let end = path.length;
+  while (end > 0 && "/".indexOf(path.charAt(end - 1)) !== -1) {
+    end -= 1;
+  }
+  return path.substring(0, end) || "/";
+};
+
+const getSurfaceType = surfaceTypeMatch =>
+  isNonEmptyString(surfaceTypeMatch) ? surfaceTypeMatch.toLowerCase() : "";
+const getAuthority = authorityMatch =>
```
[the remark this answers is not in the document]
XDex: Fixed

src/components/Personalization/utils/surfaceUtils.js
XDex: Fixed

src/components/Personalization/utils/surfaceUtils.js
XDex: Fragments are not yet supported in the UI, but they are expected to be case-sensitive, yes

src/components/Personalization/utils/surfaceUtils.js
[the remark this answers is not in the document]
XDex: @ninaceban Hmm, I see that in `utils/injectStorage` for example, string concatenation is used.. But let me replace it with string templates as you suggest.

test/functional/specs/Personalization/C28756.js
XDex: Reverted

test/functional/specs/Personalization/C44363.js
XDex: Reverted

test/functional/specs/Personalization/C44363.js
XDex: Done

test/functional/specs/Personalization/C5805676.js
XDex: Done

test/functional/specs/Personalization/C6364799.js
XDex: Done

test/functional/specs/Personalization/C782718.js
XDex: Done

test/functional/specs/Personalization/C782719.js
XDex: Done

test/unit/specs/components/DataCollector/index.spec.js
XDex: I have reverted the changes related to the new option from the API and the sandbox, but as @jfkhoury mentioned in a previous comment, we need these to be present in tests, in order to confirm that both the top-level `decisionScopes` and the new `personalization.decisionScopes` options are working as expected.

test/unit/specs/components/Personalization/createApplyPropositions.spec.js
XDex: @ninaceban It's just a minor fix, can we please leave it?

test/unit/specs/components/Personalization/createAutoRenderingHandler.spec.js
XDex: See previous comment on this

test/unit/specs/components/Personalization/createAutoRenderingHandler.spec.js
XDex: Fixed

XDex:
@ninaceban @jonsnyder @jfkhoury I've addressed the review comments and rebased on latest `main`. Please take another look when you get a chance. Thank you!

src/components/DataCollector/index.js
```diff
@@ -30,7 +30,7 @@ const createDataCollector = ({ eventManager }) => {
             mergeId,
             renderDecisions = false,
             decisionScopes = [],
```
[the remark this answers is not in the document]
XDex: Done

jonsnyder:
Closing in favor of #920

PR #869, opened 2022-05-19
Fixed an issue where localTimezoneOffset could fail Adobe Experience Platform validation.

jonsnyder:
## Description

This coerces the localTimezoneOffset into an integer. If it isn't a number then the property is added to the XDM.

## Related Issue

https://jira.corp.adobe.com/browse/PDCL-8468

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Context/injectPlaceContext.js
+++ b/src/components/Context/injectPlaceContext.js
@@ -9,8 +9,7 @@ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTA
 OF ANY KIND, either express or implied. See the License for the specific language
 governing permissions and limitations under the License.
 */
-import { deepAssign, toISOStringLocal } from "../../utils";
-import toNumber from "../../utils/toNumber";
+import { deepAssign, toISOStringLocal, toInteger } from "../../utils";
 
 export default dateProvider => {
   return xdm => {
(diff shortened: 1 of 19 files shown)


carterworks (approved):

src/components/Context/injectPlaceContext.js
```diff
@@ -9,15 +9,20 @@ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTA
...
     const date = dateProvider();
+
     const placeContext = {
-      localTime: toISOStringLocal(date),
-      localTimezoneOffset: date.getTimezoneOffset()
+      localTime: toISOStringLocal(date)
     };
+
+    const localTimezoneOffset = toInteger(date.getTimezoneOffset());
+    if (localTimezoneOffset !== undefined) {
```
jfkhoury, line 23: Should we at least log a warning if the offset is undefined, or is that too much noise and too little of a detail for the customer to care about?

jfkhoury (approved):
Thank you for the quick solution!!

src/components/Context/injectPlaceContext.js
jonsnyder: I think mostly you would see this in the data in platform. A customer would be hard pressed to reproduce this with standard browsers.

PR #873, opened 2022-06-06
Added command & config information to getLibraryInfo

adityaguruprasad:
## Description
Adding command and config information to getLibraryInfo method.

## Related Issue
getLibraryInfo call needs to display command and configuration information [(PDCL-1761)](https://jira.corp.adobe.com/browse/PDCL-1761)

## Motivation and Context
Allows for further detail provided to users upon request. getLibraryInfo command call would previously resolve to only returning library version information. Command / Config information is now also included alongside the original.

## Screenshots 
debugger_log

## Types of changes

## Checklist:


src/components/LibraryInfo/index.js
```diff
@@ -12,15 +12,17 @@ governing permissions and limitations under the License.
 
 import libraryVersion from "../../constants/libraryVersion";
 
-const createLibraryInfo = () => {
+const createLibraryInfo = tools => {
```
carterworks, line 15: Use [object destructuing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#unpacking_properties_from_objects_passed_as_a_function_parameter) for the parameters instead of a single parameter. [This matches the pattern of other component creators](https://github.com/adobe/alloy/blob/main/src/components/DataCollector/index.js#L15).

For example,

```js
// Instead of this…
const createCommand = tools => {
  return { foo: tools.foo + tools.bar };
}

// …do this.
const createCommand2 = ({ foo, bar }) => {
  return { foo: foo + bar };
};

File: src/components/LibraryInfo/index.js [hunk 2]
```diff
@@ -12,15 +12,17 @@ governing permissions and limitations under the License.
...
     commands: {
       getLibraryInfo: {
         run: () => {
           return {
             libraryInfo: {
               version: libraryVersion
-            }
+            },
+            configInfo: tools.config,
+            commandInfo: tools.componentRegistry.getCommandNames()
```
carterworks, line 25: It would be nice to sort the list of command names.

carterworks (changes requested):

src/components/LibraryInfo/index.js
jonsnyder, line 25: "configure" is a command name that is not in the componentRegistry, but is a core command.

jonsnyder (changes requested):
I'd like to see a functional test to make sure this all works together.

src/components/LibraryInfo/index.js
```diff
@@ -12,14 +12,16 @@ governing permissions and limitations under the License.
...
   return {
     commands: {
       getLibraryInfo: {
         run: () => {
           return {
             libraryInfo: {
-              version: libraryVersion
+              version: libraryVersion,
+              configs: config,
+              commands: componentRegistry.getCommandNames().sort()
```
jfkhoury, line 24: Does this log the `configure` command as well?

jfkhoury (commented):
Looks better! I suggest we add a functional test to make sure it outputs the results we are expecting.

src/components/LibraryInfo/index.js
```diff
@@ -12,14 +12,18 @@ governing permissions and limitations under the License.
 
 import libraryVersion from "../../constants/libraryVersion";
 
-const createLibraryInfo = () => {
+const createLibraryInfo = ({ config, componentRegistry }) => {
+  const coreCommandsArr = componentRegistry.getCommandNames();
+  coreCommandsArr.push("configure", "setDebug");
```
carterworks, line 17: Instead of using locally-defined strings for `"configure"` and `"setDebug"`, take [the constants from `src/core/injectExecuteCommand.js`](https://github.com/adobe/alloy/blob/main/src/core/injectExecuteCommand.js#L24), put them into a shared constants file [in `src/constants/`](https://github.com/adobe/alloy/blob/main/src/constants), and then use those new constants both here and in `injectExecuteCommand`. This will ensure consistency between command list definition and execution of those commands.

carterworks (changes requested):

src/core/injectExecuteCommand.js
```diff
@@ -11,10 +11,11 @@ governing permissions and limitations under the License.
 */
 
 import { isFunction, isObject, values } from "../utils";
+import { CONFIGURE, SET_DEBUG } from "../constants/libraryInfo";
 
 const coreCommands = {
```
carterworks, line 16: This variable is no longer required

carterworks (changes requested):

jonsnyder (approved):

carterworks (approved):

PR #875, opened 2022-06-09
 applyPropositions test should use array filter instead of find

dcottingham:
To avoid using a polyfill for IE I switched a test case from using Array.find() to Array.filter().

## Types of changes

## Checklist:

### DIFF
--- a/test/unit/specs/components/Personalization/createApplyPropositions.spec.js
+++ b/test/unit/specs/components/Personalization/createApplyPropositions.spec.js
@@ -237,7 +237,7 @@ describe("Personalization::createApplyPropositions", () => {
       result.propositions.forEach(proposition => {
         const original = originalPropositions.filter(
           originalProposition => originalProposition.id === proposition.id
-        );
+        )[0];
         if (original) {
           numReturnedPropositions += 1;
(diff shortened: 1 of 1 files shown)


ninaceban (approved):

test/unit/specs/components/Personalization/createApplyPropositions.spec.js
```diff
@@ -235,7 +235,7 @@ describe("Personalization::createApplyPropositions", () => {
       let numReturnedPropositions = 0;
       expect(originalPropositions).toEqual(MIXED_PROPOSITIONS);
       result.propositions.forEach(proposition => {
-        const original = originalPropositions.find(
+        const original = originalPropositions.filter(
```
carterworks, line 238: This should be `const [original] = originalPropositions.filter(…)` because `filter()` returns an array of matching elements, where `find()` returns a single matching element. As written, the assertion on line 243 will always pass because an instance of a proposition will never be the same as an array containing a single instance of a proposition.

carterworks (changes requested):

carterworks (approved):

PR #883, opened 2022-07-01
Enhanced link click collection to include link name and region, also included a new link click callback.

ninaceban:
## Description

This PR adds logic to determine the linkName, linkRegion. 

The configuration `clickCollectionEnabled` will enable the link click collection. If it is `false` then the Activity Collector component will not trigger the collect call with the link details in `xdm.web.webInteraction`.  Also if there is a Personalization metric set on a link that qualify as a link, it will not add the `xdm.web.webInteraction` details.

The customer can filter which link clicks he wants to collect by providing a callback function `onBeforeLinkClickSend` in the configure command, similar to the onBeforeEventSend. 

```
onBeforeLinkClickSend: function(options) {
    const xdm = options.xdm;
    const data = options.data;

    if (
        xdm &&
        xdm.web &&
        xdm.web.webInteraction &&
        xdm.web.webInteraction.name ===
        "Link Click with Personalization metric"
    ) {
        xdm.web.webInteraction.name = "Add to cart link";
        return true;
    }
    data.custom = "5";

    return false;
}
```

The callback function parameter `options` is an object that contains:
```
xdm - the event xdm, which can be augmented
data - the event data which can be augmented
clickedElement - the DOM element node that was clicked

```

If the callback returns true, the WEB SDK augments the event and triggers it. If it returns false or undefined it will cancel that event. If the callback function is not provided but `clickCollectionEnabled` is `true` the link click event is triggered.

The second part of this PR is exposing a function that can be called by the extension to deliver the customer an Activity Map.
The function is available thru the monitoring hook `onInstanceConfigured`.  It can be accessed and used like this:
```
window.__alloyMonitors = window.__alloyMonitors || [];
window.__alloyMonitors.push({
    onInstanceConfigured(data) {
        console.log("Alloy Instance Configured");
        console.log(JSON.stringify(data.config, null, 2));
        const { getLinkDetails } = data;

        const listOfLinks = document.links;

        for (var i = 0; i < listOfLinks.length; i++) {
            const getDetails = getLinkDetails(listOfLinks[i]);
            console.log("link details", getDetails);
        }
    }
});
```

## Types of changes

## Checklist:


src/core/createGetLinkElementDetails.js
```diff
@@ -0,0 +1,50 @@
+// eslint-disable-next-line import/no-named-as-default
+import createGetLinkDetails from "../components/ActivityCollector/createGetLinkDetails";
+
+export default (window, config, logger) => {
```
jonsnyder, line 4: I'd prefer to pass what's needed directly (onBeforeLinkClickSend) instead of the whole config object.

src/core/createGetLinkElementDetails.js
```diff
@@ -0,0 +1,50 @@
...
+  const { onBeforeLinkClickSent } = config;
+  return element => {
+    if (!element) {
+      return undefined;
+    }
+    const getLinkDetails = createGetLinkDetails(window, config);
```
jonsnyder, line 10: Pass this into the function directly and create it outside of this function. i.e.
```
const getLinkDetails = createGetLinkDetails({ window, config });
const getLinkElementDetails = createGetLinkElementDetails({ getLinkDetails, onBeforeLinkClickSend, logger });
```
In general, pass the object into the function instead of importing the create function. Avoiding the import function lets us test each file on its own. The exception to this guideline is to import simple util functions directly.

I notice that you do this in a few places. We want to separate the code we use to do dependency injection (which is only tested with functional tests), and the code we implement business logic (which can be tested with unit tests). When you import a file directly you can no longer write a unit test without also testing the code in the imported file. It's better to inject the function so that you can stub it out in the unit test.

jonsnyder (commented):
Looking good. There are a few tricky parts with dependency injection, and component structure.

src/core/index.js
```diff
@@ -51,6 +51,8 @@ import injectProcessWarningsAndErrors from "./edgeNetwork/injectProcessWarningsA
 import injectGetLocationHint from "./edgeNetwork/injectGetLocationHint";
 import isRequestRetryable from "./network/isRequestRetryable";
 import getRequestRetryDelay from "./network/getRequestRetryDelay";
+import createGetAnchorElementDetails from "./createGetAnchorElementDetails";
+import createGetLinkDetails from "../components/ActivityCollector/createGetLinkDetails";
```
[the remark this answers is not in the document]
jonsnyder: I took a stab at what I'm talking about: https://github.com/adobe/alloy/commit/3746edc75a4d2f36f171a62d40f7663d6452c54a What do you think?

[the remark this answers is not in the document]
ninaceban: Thank you, Jon. I have merged that branch.

src/core/createEvent.js
```diff
@@ -58,6 +58,20 @@ export default () => {
     documentMayUnload() {
       documentMayUnload = true;
     },
+    augmentEvent(callback, extraOptions) {
```
jonsnyder, line 61: Since augmentEvent is never called outside of this file, you can defined it locally and not expose it.

jonsnyder (approved):
This is looking great. I'd like to see a few more functional tests:
* onBeforeLinkClickSend can cancel a link
* onBeforeLinkClickSend can update the xdm
* The activity map extension can get details for links on the page.

src/core/createEvent.js
ninaceban: ok, do you think we should revert this at all?

jonsnyder: either way. You could go back to how it was before or define this as a local function.

src/components/ActivityCollector/getLinkRegion.js
```diff
@@ -0,0 +1,70 @@
...
+ * 6. Clicked node INPUT descendant value attribute.
+ *    Whichever is found first.
+ * 7. Clicked node IMG descendant src attribute.
+ *    Whichever is found first.
+ *
+ * @param {*} node The node to find link region for.
+ * @returns {string} link-region or BODY if no link-region is found.
+ */
+export default targetElement => {
+  let node = targetElement.parentNode;
```
jfkhoury, line 51: I would rename `node` to `parentNode` or `linkParentNode`

jfkhoury: We need unit and functional tests for the `name` and `region` functionalities.

src/components/ActivityCollector/createGetLinkDetails.js
```diff
@@ -0,0 +1,71 @@
...
+  window,
+  getLinkName,
+  getLinkRegion,
+  getAbsoluteUrlFromAnchorElement,
+  findSupportedAnchorElement,
+  determineLinkType
+}) => {
+  return ({ targetElement, config }) => {
+    // Search parent elements for an anchor element
+    // TODO: Replace with generic DOM tool that can fetch configured properties
```
jfkhoury, line 23: What does this TODO mean, and why not use a generic DOM tool now?

src/components/ActivityCollector/createGetLinkDetails.js
```diff
@@ -0,0 +1,71 @@
...
+      return undefined;
+    }
+
+    const linkUrl = getAbsoluteUrlFromAnchorElement(window, anchorElement);
+    if (!linkUrl) {
+      return undefined;
+    }
+
+    const linkType = determineLinkType(window, config, linkUrl, anchorElement);
+    // TODO: The user provided link click function needs to be called here
```
jfkhoury, line 35: What is this TODO for?

src/components/ActivityCollector/getLinkName.js
```diff
@@ -0,0 +1,130 @@
...
+            getNodeAttributeValue(supportedNode, "src", "IMG")
+          );
+      }
+      if (supportedNode.nodeValue) {
+        texts.push(supportedNode.nodeValue);
+      }
+    });
+    nodeText = truncateWhiteSpace(texts.join(""));
+    if (!nodeText) {
+      nodeText = truncateWhiteSpace(alt || title || inputValue || imgSrc || "");
```
jfkhoury, line 126: Why not call `truncateWhiteSpace` in this one spot only?

jfkhoury (changes requested):
Need to make some tweaks and add functional tests for `name` and `region`.

src/components/ActivityCollector/utils.js
```diff
@@ -63,10 +63,49 @@ const isExitLink = (window, linkUrl) => {
...
 };
 
+/**
+ * Reduces repeated whitespace within a string. Whitespace surrounding the string
+ * is trimmed and any occurrence of whitespace within the string is replaced with
+ * a single space.
+ * @param {string} str String to be formatted.
+ * @returns {string} Formatted string.
+ */
+const truncateWhiteSpace = str => {
```
mkoopmanATadobe, line 73: The current AppMeasurement_Module_ActivityMap uses this string to trim white space

`var wchar='[\\s\\n\\f\\r\\t\u0009-\u000d\u0020\u00a0\u1680\u180e\u2000-\u200a\u2028\u2029\u205f\u3000\ufeff]';
`

A number of other text editors like Word and languages can use more than just " " for white space.

src/components/ActivityCollector/createGetLinkDetails.js
```diff
@@ -0,0 +1,71 @@
...
+        }
+      },
+      data: {},
+      clickedElement: targetElement
+    };
+    if (!onBeforeLinkClickSend) {
+      return options;
+    }
+
+    const shouldEventBeTracked = onBeforeLinkClickSend(options);
```
jfkhoury, line 63: This check should be more specific. The use case is `return false to stop sending this event.`. This logic will stop sending the events if the callback returns any falsy values.

jfkhoury: We also need to handle exceptions thrown by `onBeforeLinkClickSend`. We need to try catch this here, or higher level when we call `getLinkDetails`.

ninaceban: we would not handle exceptions here, similar on how we're doing in `onBeforeEventSend`. If the code from the callback throws an exception then we cancel the event.

src/components/ActivityCollector/createGetLinkDetails.js
ninaceban: This was meant for the callback function to be called, which is not appropriate anymore. @liljenback is that right?

src/components/ActivityCollector/getLinkName.js
ninaceban: we'd like to avoid collecting values with spaces only

src/components/ActivityCollector/createGetLinkDetails.js
ninaceban: this does not apply anymore

jonsnyder (approved):

sandbox/src/PersonalizationSpa.js
```diff
@@ -55,13 +55,13 @@ const Promotion = () => {
...
-  useSendPageViewEvent({ viewName, setPropositions });
+  //useSendPageViewEvent({ viewName, setPropositions });
   useEffect(() => {
-    if (propositions) {
+    /*if (propositions) {
       window.alloy("applyPropositions", {
         propositions
       });
-    }
+    }*/
```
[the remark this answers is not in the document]
ninaceban: missed this, it caused a loop that is why I have commented it

src/components/ActivityCollector/createGetLinkDetails.js
```diff
@@ -0,0 +1,77 @@
...
+
+    if (shouldEventBeTracked !== false) {
+      return options;
+    }
+    logger.info(
+      "This link click event is not triggered because the callback function canceled it."
```
jfkhoury, line 73: I would replace `the callback function ...` with `because it was canceled in onBeforeLinkClickSend`

carterworks (approved):

src/components/ActivityCollector/createGetLinkDetails.js
ninaceban: thank you

src/components/ActivityCollector/configValidators.js
```diff
@@ -10,13 +10,14 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
 
 export default {
   clickCollectionEnabled: boolean().default(true),
   downloadLinkQualifier: string()
     .regexp()
     .default(
       "\\.(exe|zip|wav|mp3|mov|mpg|avi|wmv|pdf|doc|docx|xls|xlsx|ppt|pptx)$"
-    )
+    ),
+  onBeforeLinkClickSend: callback()
```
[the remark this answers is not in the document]
ninaceban: I have created a ticket, will be working on it after the release: PDCL-10198

src/components/ActivityCollector/utils.js
ninaceban: we have an assertion checking multiple spaces with the ` truncateWhiteSpace` method

PR #886, opened 2022-07-14
Re-enable Personalization click metrics

ninaceban:
## Description

The customer will be able to disable or enable the link click tracking, this configuration will not affect Personalization click tracking metrics.

I have added `event.documentMayUnload()` so that in case the link click or metric is set on a exit link the request is triggered before navigation.
## Types of changes

## Checklist:

### DIFF
--- a/src/components/ActivityCollector/attachClickActivityCollector.js
+++ b/src/components/ActivityCollector/attachClickActivityCollector.js
@@ -17,6 +17,7 @@ const createClickHandler = ({ eventManager, lifecycle, handleError }) => {
     // TODO: Consider safeguarding from the same object being clicked multiple times in rapid succession?
     const clickedElement = clickEvent.target;
     const event = eventManager.createEvent();
+    // this is to make sure a exit link personalization metric use send beacon
     event.documentMayUnload();
     return (
       lifecycle
--- a/src/components/ActivityCollector/attachClickActivityCollector.js
+++ b/src/components/ActivityCollector/attachClickActivityCollector.js
@@ -17,6 +17,7 @@ const createClickHandler = ({ eventManager, lifecycle, handleError }) => {
     // TODO: Consider safeguarding from the same object being clicked multiple times in rapid succession?
     const clickedElement = clickEvent.target;
     const event = eventManager.createEvent();
+    event.documentMayUnload();
     return (
       lifecycle
         .onClick({ event, clickedElement })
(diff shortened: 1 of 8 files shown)


src/components/Personalization/createOnClickHandler.js
```diff
@@ -41,6 +41,8 @@ export default ({
             }
           };
         }
+        // TODO: move this in attachClickActivityCollector get link details here and determine whether should set use beacon
+        event.documentMayUnload(); // this is needed in case it is an exit link and but clickCollection is disabled
```
jonsnyder, line 45: I think its okay to just use /collect no matter what type of link it is.

src/components/ActivityCollector/configValidators.js
```diff
@@ -13,7 +13,7 @@ governing permissions and limitations under the License.
 import { string, boolean } from "../../utils/validation";
 
 export default {
-  clickCollectionEnabled: boolean().default(true),
+  clickCollectionEnabled: boolean().default(false),
```
jonsnyder, line 16: I don't think we want to change the default yet. Changing the default is a breaking change. I documented a strategy for changing the default here: https://wiki.corp.adobe.com/display/DMSArchitecture/Migration+plan+to+collect+less+by+default
I think making the changes from the wiki should be a separate ticket/PR.

jonsnyder (commented):
I was expecting an updated functional test. It looks like we don't have a test for an interact with a proposition. You should add one.

src/components/Personalization/createOnClickHandler.js
ninaceban: ok, thank you

src/components/ActivityCollector/configValidators.js
ninaceban: I see, ok, I'll do that in a separate PR. Thank you Jon for referencing the Wiki

src/components/ActivityCollector/attachClickActivityCollector.js
```diff
@@ -17,6 +17,7 @@ const createClickHandler = ({ eventManager, lifecycle, handleError }) => {
     // TODO: Consider safeguarding from the same object being clicked multiple times in rapid succession?
     const clickedElement = clickEvent.target;
     const event = eventManager.createEvent();
+    event.documentMayUnload();
```
jfkhoury, line 20: Can we add an inline comment explaining why we have added this expression?

jfkhoury (approved):
Can we please update the description of this PR to remove the disabling of the config?

good work Nina!

src/components/ActivityCollector/attachClickActivityCollector.js
ninaceban: sure!

jonsnyder (approved):

PR #893, opened 2022-07-25
Legacy location hints

jonsnyder:
## Description

Read and extract the legacy Target location hint from cookies if there isn't already a cluster cookie. I also added a test to make sure Konductor can understand the location hint.

## Related Issue

https://jira.corp.adobe.com/browse/PDCL-8160

## Motivation and Context

https://wiki.corp.adobe.com/display/DMSArchitecture/Legacy+location+hints+for+experience+edge

## Types of changes

## Checklist:

### DIFF
--- a/sandbox/public/legacy_target.html
+++ b/sandbox/public/legacy_target.html
@@ -1,0 +1,68 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <link rel="shortcut icon" href="%PUBLIC_URL%/favicon.ico" />
+    <link rel="stylesheet" href="index.css" />
+    <meta
+      name="viewport"
+      content="width=device-width, initial-scale=1, shrink-to-fit=no"
+    />
+    <meta name="theme-color" content="#000000" />
+    <meta
+      http-equiv="Content-Security-Policy"
+      content="default-src 'self' *.demdex.net;
+            script-src 'self' 'unsafe-inline' *.amazonaws.com github.com *.githubusercontent.com;
+            style-src 'self' 'unsafe-inline';
+            img-src * data:;
+            connect-src 'self' *.adobedc.net *.demdex.net *.tt.omtrdc.net"
+    />
+    <script src="/at.js"></script>
+    <title>Mock website hosting Alloy</title>
+  </head>
+  <body>
+    <div>
+      <ul>
+        <li>
+          <a href="/">Home</a>
+        </li>
+        <li>
+          <a href="/consent">Consent</a>
+        </li>
+        <li>
+          <a href="/personalization">Personalization</a>
+        </li>
+        <li>
+          <a href="/personalizationSpa">Personalization - SPA</a>
+        </li>
+        <li>
+          <a href="/personalizationProfile">Personalization - Profile</a>
+        </li>
+        <li>
+          <a href="/links">Links</a>
+        </li>
+        <li>
+          <a href="/eventMerge">Event-Merge</a>
+        </li>
+        <li>
+          <a href="/largePayload">Large Payload</a>
+        </li>
+        <li>
+          <a href="/orgTwo">Multiple Orgs</a>
+        </li>
+        <li>
+          <a href="/dualTag">DualTag</a>
+        </li>
+        <li>
+          <a href="/legacy.html">Legacy Visitor ID</a>
+        </li>
+      </ul>
+    </div>
+    <hr />
+    <div>
+      <h1>
+        Only the legacy at.js is loaded on this page.
+      </h1>
+    </div>
+  </body>
+</html>
--- a/src/core/edgeNetwork/injectGetLocationHint.js
+++ b/src/core/edgeNetwork/injectGetLocationHint.js
@@ -12,9 +12,20 @@ governing permissions and limitations under the License.
 
 import { getNamespacedCookieName } from "../../utils";
 import { CLUSTER } from "../../constants/cookieNameKey";
+import { MBOX_EDGE_CLUSTER } from "../../constants/legacyCookies";
 
 export default ({ orgId, cookieJar }) => {
   const clusterCookieName = getNamespacedCookieName(orgId, CLUSTER);
+  const fromClusterCookie = () => cookieJar.get(clusterCookieName);
+  const fromTarget = () => {
+    const mboxEdgeCluster = cookieJar.get(MBOX_EDGE_CLUSTER);
+    if (mboxEdgeCluster) {
+      return `t${mboxEdgeCluster}`;
+    }
+    return undefined;
+  };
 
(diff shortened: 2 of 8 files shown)


sandbox/public/legacy_target.html
```diff
@@ -0,0 +1,68 @@
...
+    <meta name="theme-color" content="#000000" />
+    <meta
+      http-equiv="Content-Security-Policy"
+      content="default-src 'self' *.demdex.net;
+            script-src 'self' 'unsafe-inline' *.amazonaws.com github.com *.githubusercontent.com;
+            style-src 'self' 'unsafe-inline';
+            img-src * data:;
+            connect-src 'self' *.adobedc.net *.demdex.net *.tt.omtrdc.net"
+    />
+    <script src="/at.js"></script>
```
jfkhoury, line 20: I think we need Visitor.js as well so that all profile fragments are based on ECID right?

src/core/edgeNetwork/injectGetLocationHint.js
```diff
@@ -12,9 +12,20 @@ governing permissions and limitations under the License.
...
 import { CLUSTER } from "../../constants/cookieNameKey";
+import { MBOX_EDGE_CLUSTER } from "../../constants/legacyCookies";
 
 export default ({ orgId, cookieJar }) => {
   const clusterCookieName = getNamespacedCookieName(orgId, CLUSTER);
+  const fromClusterCookie = () => cookieJar.get(clusterCookieName);
+  const fromTarget = () => {
+    const mboxEdgeCluster = cookieJar.get(MBOX_EDGE_CLUSTER);
+    if (mboxEdgeCluster) {
+      return `t${mboxEdgeCluster}`;
```
jfkhoury, line 23: What's the `t` for?

jfkhoury (approved):
Nice! Hey @jonsnyder where is the Target `session` handled? Is it based on the ECID? Does it have a separate cookie?

src/core/edgeNetwork/injectGetLocationHint.js
jonsnyder: t is for target.

jonsnyder:
This PR is just for using the mboxEdgeCluster cookie.

sandbox/public/legacy_target.html
jonsnyder: Yes. The sandbox needs work to be able to test mixed mode. I'll do that in a separate PR.

jasonwaters (approved):
LGTM

carterworks (approved):
Send it.

For future reference (to myself, mostly)

Question:
> …where is the Target session handled? Is it based on the ECID? Does it have a separate cookie?

[Answer](https://jira.corp.adobe.com/browse/PDCL-8237#comment-30340110): 

> Target uses the session id at both the load balancer level (nginx ingress), to ensure requests for a session are routed to the same pod, and within Target upstream / Delivery API, to find the user context. In the case of calls from Konductor, the load balancer uses the ECID passed in a header to the upstream as the session persistence key. But within Target upstream, it doesn't use the ECID directly as the session id, it generates and caches a unique session id based on the ECID.
> 
> This makes it a little more complicated to use an override sessionId in the case of mixed implementation. I think the LB will want to have access to the override session id in a header. So the target feature in Konductor might need to pull the sessionId out of meta and populate a custom header for the Target LB to access.
> 
> Also, for the case where Konductor is the first to be called in a session ([PDCL-8236](https://jira.corp.adobe.com/browse/PDCL-8236)), it's not enough to simply save the ECID as the sessionId in the Target cookie. Target upstream will probably need to return the actual session id used (different than the ECID) in the response in a way that Web SDK can pick up and store in the Target cookie.

TL;DR: The Target session ID is not the ECID, but is the ECID _and_ a 6 digit random number.

ninaceban (commented):
@jonsnyder I was looking through implementation and it occurred to me that handling `mboxEdgeCluster` is not enough.

Here are a few details:
- `mboxEdgeCluster` cookie has a TTL of 31 minutes
- `mbox` cookie has a TTL of 2 years, just like AMCV cookie

So my recommendation would be to first check if `mboxEdgeCluster` exists and use it for cluster routing, if it is not present then use `mbox` cookie to extract the `PC` cookie fragment and parse the location hint from the fragment. If both of the above fail, use the default "any" cluster routing.

ninaceban:
> Send it.
> 
> For future reference (to myself, mostly)
> 
> Question:
> 
> > …where is the Target session handled? Is it based on the ECID? Does it have a separate cookie?
> 
> [Answer](https://jira.corp.adobe.com/browse/PDCL-8237#comment-30340110):
> 
> > Target uses the session id at both the load balancer level (nginx ingress), to ensure requests for a session are routed to the same pod, and within Target upstream / Delivery API, to find the user context. In the case of calls from Konductor, the load balancer uses the ECID passed in a header to the upstream as the session persistence key. But within Target upstream, it doesn't use the ECID directly as the session id, it generates and caches a unique session id based on the ECID.
> > This makes it a little more complicated to use an override sessionId in the case of mixed implementation. I think the LB will want to have access to the override session id in a header. So the target feature in Konductor might need to pull the sessionId out of meta and populate a custom header for the Target LB to access.
> > Also, for the case where Konductor is the first to be called in a session ([PDCL-8236](https://jira.corp.adobe.com/browse/PDCL-8236)), it's not enough to simply save the ECID as the sessionId in the Target cookie. Target upstream will probably need to return the actual session id used (different than the ECID) in the response in a way that Web SDK can pick up and store in the Target cookie.
> 
> TL;DR: The Target session ID is not the ECID, but is the ECID _and_ a 6 digit random number.

@carterworks it is a little bit more nuanced. When at.js is involved, Target session ID is an UUID v4 generated ID. When a data stream has Target enabled, then Konductor talks to Target Edge and ONLY in this case Target session ID will be ECID + a random alphanumeric string of 6 characters.

jasonwaters:
@ninaceban @jfkhoury Is anyone tending to this PR?  I think we need to get it finished and merged for mixed mode support

src/core/edgeNetwork/injectGetLocationHint.js
ninaceban: @dcottingham we'll have to remove the `t` part from `mboxEdgeCluster` cookie.

jfkhoury:
> @ninaceban @jfkhoury Is anyone tending to this PR? I think we need to get it finished and merged for mixed mode support

@jasonwaters Jon is back tomorrow and should be able to wrap this one up. 

I'd like us to address @ninaceban's  comments around the `mboxEdgeCluster` cookie to make sure we have all our bases covered, maybe on the check-in meeting today.

dcottingham:
I'll make the necessary updates here tomorrow.  Was working on the backend changes to support this today.  https://git.corp.adobe.com/TnT/target-upstream/pull/105

PR #901, opened 2022-09-02
PDCL-9192 - add propositionEventType to experience event for better proposition interaction handling

dcottingham:
## Description

- Description of the problem this PR solves is in [this wiki](https://wiki.corp.adobe.com/pages/viewpage.action?spaceKey=CJM&title=Proposition+Event+Types#PropositionEventTypes-Problem)
- https://jira.corp.adobe.com/browse/PDCL-9192

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Personalization/constants/eventType.js
+++ b/src/components/Personalization/constants/eventType.js
@@ -12,3 +12,7 @@ governing permissions and limitations under the License.
 
 export const DISPLAY = "decisioning.propositionDisplay";
 export const INTERACT = "decisioning.propositionInteract";
+export const EventType = {
+  DISPLAY: "display",
+  INTERACT: "interact"
+};
--- a/src/components/Personalization/createCollect.js
+++ b/src/components/Personalization/createCollect.js
@@ -9,7 +9,8 @@ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTA
 OF ANY KIND, either express or implied. See the License for the specific language
 governing permissions and limitations under the License.
 */
-import { DISPLAY, EventType } from "./constants/eventType";
+import { DISPLAY } from "./constants/eventType";
+import { PropositionEventType } from "./constants/propositionEventType";
 import { isNonEmptyArray } from "../../utils";
 
 export default ({ eventManager, mergeDecisionsMeta }) => {
@@ -24,7 +25,7 @@ export default ({ eventManager, mergeDecisionsMeta }) => {
       };
     }
     if (isNonEmptyArray(decisionsMeta)) {
-      mergeDecisionsMeta(event, decisionsMeta, EventType.DISPLAY);
+      mergeDecisionsMeta(event, decisionsMeta, PropositionEventType.DISPLAY);
     }
 
     event.mergeXdm(data);
--- a/src/components/Personalization/createCollect.js
+++ b/src/components/Personalization/createCollect.js
@@ -9,7 +9,7 @@ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTA
 OF ANY KIND, either express or implied. See the License for the specific language
 governing permissions and limitations under the License.
 */
-import { DISPLAY } from "./constants/eventType";
+import { DISPLAY, EventType } from "./constants/eventType";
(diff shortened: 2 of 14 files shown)


src/components/Personalization/constants/eventType.js
```diff
@@ -12,3 +12,7 @@ governing permissions and limitations under the License.
 
 export const DISPLAY = "decisioning.propositionDisplay";
 export const INTERACT = "decisioning.propositionInteract";
+export const EventType = {
+  DISPLAY: "display",
```
jfkhoury, line 16: Why not use the full event types `decisioning.proposition...`?

dcottingham: @jfkhoury, using the existing eventType field with `decisioning.proposition...` isn't flexible enough for all customer use cases, since the existing `eventType` lives on the event and not the proposition.  [This wiki ](https://wiki.corp.adobe.com/pages/viewpage.action?spaceKey=CJM&title=Proposition+Event+Types#PropositionEventTypes-Problem) describes the problem and why both Target and AJO need something like propositionEventType.

carterworks (approved):
I think I understand the use case. If I wanted to include something in the XDM decisions object for interactions, this is how I would do it.

dcottingham:
Backend changes for this are due next sprint (Sept 14-28).  I've tested this change locally in sandbox app and verified it is backward compatible.
Screen Shot 2022-09-12 at 10 12 50 AM

src/components/Personalization/createCollect.js
```diff
@@ -24,7 +24,7 @@ export default ({ eventManager, mergeDecisionsMeta }) => {
       };
     }
     if (isNonEmptyArray(decisionsMeta)) {
-      mergeDecisionsMeta(event, decisionsMeta);
+      mergeDecisionsMeta(event, decisionsMeta, EventType.DISPLAY);
     }
 
     event.mergeXdm(data);
```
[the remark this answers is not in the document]
ninaceban: @carterworks There are 2 types of events: `display` and `interact`. The `xdm.eventType` is optional, and customer can set any eventType they want to collect . The logic in Target Edge depends on the event type from `xdm.eventType`.  This was adding some complexity because was forcing alloy or the customer to set a specific event type so that the backend could count impressions/conversions properly. 
This is why we needed to add the propositionEventType at the `_experience.decisioning` level that will control the logic in Target Edge. And leave the customer to set any `xdm.eventType` when sending manually the notifications. 

We will not remove the `xdm.evenType` since it is a field related to the event itself.

ninaceban (approved):
@dcottingham looks good to me!  Thank you!

I would add only one general comment, ideally when we augment the existing functionality and it should be backward compatible, we should add a new functional test that would test the new behavior and leave the existing ones test the old behavior.
cc: @jonsnyder @carterworks

ninaceban:
@dcottingham do we plan to update the documentation with the new way of sending `display` and `click` notifications? This is related to the notifications that are sent by the customer.

dcottingham:
> @dcottingham looks good to me! Thank you!
> 
> I would add only one general comment, ideally when we augment the existing functionality and it should be backward compatible, we should add a new functional test that would test the new behavior and leave the existing ones test the old behavior. cc: @jonsnyder @carterworks

@ninaceban Do you want me to revert all the updates I made to the functional test cases and create some new ones then?  I only added new assertions to the existing test cases.  I hadn't removed or modified existing assertions.  If I create new tests they will be functionally the same as the ones I updated, just different assertions.  But happy to do that if that's the preferred approach.

src/components/Personalization/constants/eventType.js
```diff
@@ -12,3 +12,7 @@ governing permissions and limitations under the License.
 
 export const DISPLAY = "decisioning.propositionDisplay";
 export const INTERACT = "decisioning.propositionInteract";
+export const EventType = {
```
jonsnyder, line 15: This is confusing. Perhaps call this proposition event type? Maybe in a separate file.

jonsnyder (approved):

jonsnyder:
I think using the existing functional test is fine in this case. You are not adding functionality, but updating the format of the current functionality.

PR #911, opened 2022-09-22
Balance prod workflow test runs to avoid memory leaks caused by concurrent tests runs within in job

shammowla:
Currently the prod workflow runs jobs are based on the Alloy release versions supported. Each job represents a single runner VM environment which handles logging the output from 5 different Saucelabs test browsers. By separating the test matrix into multiple jobs based on the Saucelabs test environment and Alloy base version, it's able to run 5 concurrent tests with faster test runs.

Example test run: 
https://github.com/adobe/alloy/actions/runs/3102462002 

## Types of changes

## Checklist:


carterworks (approved):

carterworks (approved):

.github/workflows/prod.yaml
```diff
@@ -63,7 +63,139 @@ jobs:
...
+        run: npx testcafe -q -c 10 "saucelabs:Chrome@latest:macOS 11.00",
+        env:
+          ALLOY_PROD_VERSION: ${{ steps.package-version.outputs.current-version }}
+      - uses: craftech-io/slack-action@v1
+        with:
+          slack_webhook_url: ${{ secrets.SLACK_WEBHOOK_URL }}
+          status: failure
+        if: failure()
+
+  safari_workflow:
```
jonsnyder, line 75: The GitHub actions documentation says you can have a multi-dimensional matrix and it will do every combination. https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs

This could avoid a lot of the duplication here.

jonsnyder (approved):

.github/workflows/prod.yaml
shammowla: I tried this first with no luck, the jobs are handled differently and ends up running the version with the test browser in a queue.

PR #914, opened 2022-09-27
Fixed an issue with applyResponse

jonsnyder:
## Description

Fixed an issue with applyResponse where it would not apply personalized content if it was called after a sendEvent. There is a condition here: https://github.com/adobe/alloy/blob/6194f1b0cf4e5f4ae16bf18d21c3f5e2896550d4/src/components/Personalization/createComponent.js#L62 which is defined here: https://github.com/adobe/alloy/blob/6194f1b0cf4e5f4ae16bf18d21c3f5e2896550d4/src/components/Personalization/createPersonalizationDetails.js#L65

which makes it so that the personalized content from an applyResponse command is only processed if the cache is not initialized.  This is because for applyResponse commands, the scopes are always empty. In this PR, I add the page wide scope (`__view__`) to the applyResponse calls so that "shouldFetchData()" is always true for applyResponse commands.

## Types of changes

## Checklist:


carterworks (approved):

src/core/createEventManager.js
```diff
@@ -113,7 +114,7 @@ export default ({
         .onBeforeEvent({
           event,
           renderDecisions,
-          decisionScopes: [],
+          decisionScopes: [PAGE_WIDE_SCOPE],
```
jasonwaters, line 117: I see how this solves the problem, but it's not technically accurate.  I suppose an empty array of decision scopes was also inaccurate though.... so 🤷‍♂️ 

Since we have the `responseBody` here, you could instead iterate over the `payload` array of the `personalization:decisions` handle to construct a Set of `scope`s found in the list.  then provide the list in the decisionScopes array here.

jasonwaters (approved):

src/core/createEventManager.js
ninaceban: In my opinion `renderDecisions` and `decisionScopes` should be passed by the customer in the `applyResponses` command as options. If they need the `__view__` decisions to be rendered, then they would pass them in the command options. Otherwise, we make the assumption that the customer have fetched them and wants them to be rendered. 

The special scope `__view__` is attached to `decisionScopes` by Web SDK. In cases of server-side call, it is up to the customer to fetch the __view__ decisions or not.

ninaceban: Probably as a quick fix, @jasonwaters idea to extract the scopes from handles will work. Although with @XDex PR related to surfaces, we should think of how we should handle `personalization` object here as well.

jonsnyder: We will go ahead and merge this, and work on testing it more thoroughly later. If we find a bug we can patch later.

PR #918, opened 2022-10-04
Fixed an issue where visitor migration does not work if `window.Visitor` is defined after configure

jonsnyder:
## Description

This bug occurs when ID migration is turned on, and `window.Visitor` is defined after the Web SDK configure command. In most base library initializations, this bug will not be a problem. This is because the visitor and Web SDK libraries will be loaded at the top of the page, then Web SDK will be initialized. `window.Visitor` is defined when the visitor.js script is loaded. This bug is a problem especially in Adobe Tags where each extension is initialized one at a time. If the Web SDK extension is initialized first, configure will be called before the Experience Cloud ID Service is initialized.

## Related Issue

https://jira.corp.adobe.com/browse/PDCL-9421

## Types of changes

## Checklist:

### DIFF
--- a/test/unit/specs/components/Identity/visitorService/injectGetEcidFromVisitor.spec.js
+++ b/test/unit/specs/components/Identity/visitorService/injectGetEcidFromVisitor.spec.js
@@ -74,4 +74,18 @@ describe("getEcidFromVisitor", () => {
       return expectAsync(getEcidFromVisitor()).toBeResolvedTo(undefined);
     });
   });
+
+  it("should find Visitor if it was defined after Web SDK initialization.", () => {
+    const awaitVisitorOptIn = () => {
+      return Promise.resolve();
+    };
+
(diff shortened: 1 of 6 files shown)


test/unit/specs/components/Identity/visitorService/injectGetEcidFromVisitor.spec.js
```diff
@@ -74,4 +74,18 @@ describe("getEcidFromVisitor", () => {
       return expectAsync(getEcidFromVisitor()).toBeResolvedTo(undefined);
     });
   });
+
+  it("should find Visitor if it was defined after Web SDK initialization.", () => {
```
jfkhoury, line 78: Awesome!

jfkhoury (approved):
Awesome work, love the tests!

carterworks (approved):

PR #926, opened 2022-10-10
Update Unit Test Sauce Labs Compatibility with Firefox GeckoDriver 

shammowla:
Due to a recent change made by Mozilla, Sauce Labs will update Firefox GeckoDriver to version 0.31.0 on October 4, 2022 for Firefox versions 90 and later. 

Use GeckoDriver 0.27.0 with saucelabs for Unit test

## Types of changes

## Checklist:

### DIFF
--- a/karma.saucelabs.conf.js
+++ b/karma.saucelabs.conf.js
@@ -1,5 +1,4 @@
 const karmaSauceLauncher = require("karma-sauce-launcher");
-
 const karmaConfig = require("./karma.conf");
 
 module.exports = config => {
@@ -9,17 +8,7 @@ module.exports = config => {
       base: "SauceLabs",
       browserName: "chrome",
       browserVersion: "latest",
-      "sauce:options": {
-        tags: ["w3c-chrome"]
-      }
-    },
-    sl_firefoxW3C: {
-      base: "SauceLabs",
-      browserName: "firefox",
-      browserVersion: "latest",
-      "sauce:options": {
-        tags: ["w3c-firefox"]
-      }
+      platform: "Windows 11"
     },
     sl_safariW3C: {
       base: "SauceLabs",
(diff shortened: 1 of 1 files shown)


karma.saucelabs.conf.js
```diff
@@ -9,36 +8,41 @@ module.exports = config => {
...
+      platform: "macOS 11.00"
     },
     sl_firefoxW3C: {
       base: "SauceLabs",
       browserName: "firefox",
+      platformName: "Windows 11",
       browserVersion: "latest",
       "sauce:options": {
-        tags: ["w3c-firefox"]
+        geckodriverVersion: "0.27.0"
```
carterworks, line 25: Why peg the version to 0.27.0 the `geckodriverVersion` to an old version, instead of doing `"moz:debuggerAddress": true`?

shammowla: I tired "moz:debuggerAddress": true but the config isn't working. I added it back so you can take a look at the log

carterworks: Weird! I'm looking at this test run: https://app.saucelabs.com/tests/73a033f397a34d33a334d404724b6a19#0
I checked the source code for `geckodriver`. We have `moz:debuggerAddress` in the right place. I wonder if the latest geckodriver version is not getting pulled into the test? It needs to be version 0.29.1 or higher.
What I'm trying to avoid is pinning `geckodriver` to an old version and therefore losing compatibility with newer versions of Firefox. We probably want to be able to test on those newer versions of Firefox.

shammowla: Using geckodriver 0.27.0 we can still run tests against the latest version of firefox. Test run: https://github.com/adobe/alloy/actions/runs/3230313121/jobs/5288621845

carterworks: Huh. As long as future versions of Firefox don't remove the `–remote-debugging-port` flag, I guess we can stick with v0.27.0 of geckodriver. I looked through the geckodriver changelog and we aren't missing anything major by not being on the latest 0.31.0. The biggest change is support for getting an element's shadow root , which we don't do and I don't see us ever doing.
*Approved*.

carterworks (approved):
As long as `--remote-debugging-port` isn't removed by future versions of Firefox, I'm fine with this solution.

Issue #936, opened 2022-11-15
Hard-coded Referrer Causes Fetch Error in Worker


solaris007:
### Expected Behaviour

When alloy is served via a worker (`WorkerGlobalScope`), the tracking request sent via `injectSendFetchRequest` should be successfully passed to the network stack.

### Actual Behavior

The tracking request is aborted due to:
```
Uncaught (in promise) TypeError: [alloy] [DataCollector] An error occurred while executing the sendEvent command.
Caused by: Network request failed.
Caused by: Failed to execute 'fetch' on 'WorkerGlobalScope': Referrer 'client' is not a valid URL.
    at fetch (787bcdb0-6c08-4442-9830-5c5151bd24a0:2:18168)
    at injectSendFetchRequest (alloy.js:8694:14)
    at executeRequest (alloy.js:3182:16)
    at injectSendNetworkRequest (alloy.js:3224:14)
    at eval (alloy.js:9224:16)
```

It seems that the Fetch API performs Referrer validation (valid URL) when called via a web worker. The relevant possible `TypeError` is listed on the Fetch API doc: https://developer.mozilla.org/en-US/docs/Web/API/fetch#exceptions

### Reproduce Scenario (including but not limited to)

Embed alloy in a worker via `partytown` (https://partytown.builder.io) and send test event:
```
<script>
  partytown = {
    lib: '/scripts/',
    forward: [
      '__alloyMonitors',
      '__alloyNS',
      'alloy',
      'alloy_all',
      'alloy_click',
      'alloy_last_event',
      'alloy_load',
      'alloy_unload',
      'adobe',
    ]
  };
</script>
<script src="/scripts/partytown.js"></script>

<script type="text/partytown">
  !function(n,o){o.forEach(function(o){n[o]||((n.__alloyNS=n.__alloyNS||
    []).push(o),n[o]=function(){var u=arguments;return new Promise(
    function(i,l){n[o].q.push([i,l,u])})},n[o].q=[])})}
  (window,["alloy"]);
</script>
<script type="text/partytown" src="/scripts/alloy.js" async></script>
<script type="text/partytown">
    alloy("configure", {
      "edgeConfigId": "<some-dataset-id>",
      "orgId":"<some-org-id>"
    });
    alloy("sendEvent", {
      data: {
        myNameSpace: {
            init: true,
        }
      }
    });

</script>
```

#### Steps to Reproduce
- Load Page
- Observe dev tools JS console

#### Platform and Version
[...]

PR #937, opened 2022-11-15
fix: set referrer policy to allow for webworker inclusion (#936)

solaris007:
## Description

set hard-coded referrer "client" to `window.document.referrer`

## Related Issue

#936 

## Motivation and Context

#936 

## Types of changes

## Checklist:

### DIFF
--- a/src/core/network/requestMethods/injectSendFetchRequest.js
+++ b/src/core/network/requestMethods/injectSendFetchRequest.js
@@ -19,7 +19,7 @@ export default ({ fetch }) => {
       headers: {
         "Content-Type": "text/plain; charset=UTF-8"
       },
-      referrer: window.document.referrer,
+      referrerPolicy: "no-referrer-when-downgrade",
       body
     }).then(response => {
       return response.text().then(responseBody => ({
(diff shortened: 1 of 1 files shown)


src/core/network/requestMethods/injectSendFetchRequest.js
```diff
@@ -19,7 +19,7 @@ export default ({ fetch }) => {
       headers: {
         "Content-Type": "text/plain; charset=UTF-8"
       },
-      referrer: "client",
+      referrer: window.document.referrer,
```
jonsnyder, line 22: The referrer option in fetch should be the current page url. The proposed code here would send the previous page url. It is important that experience edge get the full url of the current page. For certain features to work the experience edge needs to be able to read the query string parameters.

I'm wondering if we should be just setting the referrerPolicy to "no-referrer-when-downgrade" instead of the referrer. Does this work in a web-worker?

jonsnyder (changes requested):

src/core/network/requestMethods/injectSendFetchRequest.js
solaris007: `no-referrer-when-downgrade` as referrer policy seems to work in the web-worker.

solaris007: @jonsnyder i updated the PR accordingly

solaris007:
@jonsnyder any update?

jonsnyder (commented):
Thanks for looking into this. We should add a functional test to make sure that the referrer is set properly for all browsers we support.

jonsnyder (approved):

jonsnyder:
Closed in favor of #951

PR #939, opened 2022-11-18
Fixed an issue with Adobe Target VEC custom code actions where the code was injected into an alternate location than with at.js.

jasonwaters:
## Description

There is a problem with the way customCode actions are handled in alloy, compared to at.js.  When customCode actions are created with VEC in target, the selector (for non-HEAD) is always `BODY > *:eq(0)`.  

In at.js customCode offers are always appended to the parent element of the element the selector points to.  In this case it would alway append it to the `BODY` element.

But in alloy, customCode offers are appended to the element targeted by the selector.  So every (non HEAD) offer is injected to the first element within the BODY tag.  

Many customers may not have even noticed this difference, because injecting to the first element could have been enough.  But one customer is having an issue because the first element inside the body of their page is a `script` tag.  So the HTML of the custom code offer is appended to the script tag.  Which obviously causes problems.  

This PR fixes this problem by remapping custom code offers that target `BODY > *:eq(0)` to `BODY`.  

Although I think customCode actions are a target-specific thing, I wanted to preserve the functionality of injecting code directly into the element (just in case it is used or will be used in other personalization solutions).  For the target use case, this code remaps the known selector to the appropriate (parent) one instead.

## Related Issue

https://jira.corp.adobe.com/browse/TNT-46186

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Personalization/dom-actions/executeActions.js
+++ b/src/components/Personalization/dom-actions/executeActions.js
@@ -10,7 +10,9 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import preprocess from "./remapHeadOffers";
+import remapHeadOffers from "./remapHeadOffers";
+import { assign } from "../../../utils";
+import remapCustomCodeOffers from "./remapCustomCodeOffers";
 
 const logActionError = (logger, action, error) => {
   if (logger.enabled) {
@@ -42,6 +44,14 @@ const executeAction = (logger, modules, type, args) => {
   }
   return execute(...args);
 };
+
+const PREPROCESSORS = [remapHeadOffers, remapCustomCodeOffers];
+
+const preprocess = action =>
+  PREPROCESSORS.reduce(
+    (processed, fn) => assign(processed, fn(processed)),
+    action
+  );
 
 export default (actions, modules, logger) => {
   const actionPromises = actions.map(action => {
--- a/src/components/Personalization/dom-actions/remapCustomCodeOffers.js
+++ b/src/components/Personalization/dom-actions/remapCustomCodeOffers.js
@@ -1,0 +1,34 @@
+/*
+Copyright 2021 Adobe. All rights reserved.
+This file is licensed to you under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License. You may obtain a copy
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+/*
+ * Preprocess customCode actions before rendering, so that offer selectors are remapped appropriately for
+ * target offers, to align with the way it works in at.js.
+ */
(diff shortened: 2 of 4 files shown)


jasonwaters:
@jfkhoury at present, CustomCode only applies to target

jonsnyder:
I'm wondering if we should just mirror Target's functionality exactly and inject the custom code into the parent of the selected element instead of just solving for the body use-case. Although this could change existing functionality, I think its better to have more compatibility with current target activities.

jasonwaters:
@jonsnyder I can do that, but I really think this solution is the better one.  when using custom code to inject html into a page, VEC will only ever use the `BODY > *:eq(0)` selector.  There is never a possibility for that selector to be different.  There are tests in `customCode.spec.js` that validate custom code can be injected in an arbitrary element with className `customCode`, but that is not a real world use case in target.  The selector for custom code (outside of the head) can never be anything other than `BODY > *:eq(0)`.  

So while this code seems specific to that selector, it's because it's the only one that will ever happen in target.  This fix allows the tests to keep working so that if the `customCode` dom action is ever implemented in other personalization tools like Offer Decisioning or AJO it will work in a logical way --- to inject custom HTML code into a specific selector.  In my view, the fact that target uses the `BODY > *:eq(0)` is not ideal in the first place.  it should have been just `BODY`.  But that ship has sailed.

If i change it so that instead of transforming `BODY > *:eq(0)` to `BODY` and instead get the parent element every time, that will work for the `BODY > *:eq(0)` use case.  But it could cause confusion if any other personalization tools decide to implement `customCode` with a custom selector.  Consider this example:

```html
<body>
  <div class="container">
  </div>
</body>
```

if a customCode action is created with selector `.container`, with the intent to insert custom code into that div -- it will instead be injected into the `<body>` tag.  So the selector would instead need to be ... actually there is no selector that would really work to inject code into `.container`.  Because it is an empty div with no children.  `.container > :first-child` would not work, because it is empty.  `.container  > *:eq(0)` would not work either, for the same reason.

So I guess if we want to keep customCode in a state that it only ever will support the `BODY > *:eq(0)` tag, we could change it to always just get the parent node.  And in that case never support targeting any other selectors with empty children.  But there is another bug (that admittedly would probably never happen) where if a customer tries to insert custom code into an empty `<body></body>` tag, the `BODY > *:eq(0)` selector would fail, because there is no first child to get the parent of.  

Changing the selector from `BODY > *:eq(0)` to just `BODY` resolves the 👆 problems mentioned allowing custom code to be injected into the body every time regardless of how many children it has.  And it also will allow personalization tools to leverage customCode in a logical way in the future if need be.

Thoughts?

jonsnyder (approved):
Thanks for the background info. I didn't realize that Target would only ever send that selector. This looks good.

ninaceban (approved):
@jasonwaters as I mentioned the fix looks great but let me know if you have considered pushing this logic to server side.

jasonwaters:
That an excellent observation, I agree it would be nice to do the change on target-upstream instead.  I have created a CE trace ticket for that [here](https://jira.corp.adobe.com/browse/TNT-46337).  

I will ping CE on the issue to see if there is any push back to changing plans on this ticket.

jfkhoury:
I love @ninaceban's suggestion so much!

Thank you for the fix @jasonwaters , I hope we can move it server side.

jonsnyder:
I will merge this now so that customers don't have to deal with this problem. Ideally, we will fix this on the server and then we can remove this hack.

PR #966, opened 2023-02-22
Remove console error for failed URL destination.

carterworks:
## Description

When a URL destination fails for any reason, do not show an error. It is up to the customer to make sure that they are entering in valid URL destinations and they will need to monitor their audience performance.

## Related Issue

[PDCL-10013: Web SDK fails to inject Facebook Pixel URL destination](https://jira.corp.adobe.com/browse/PDCL-10013)

## Motivation and Context

Facebook Pixel URL destinations return an empty 200 response. This causes Alloy to show a false console error, reporting that the URL destination failed. 

[Silent failure matches the behavior of VisitorJS.](https://git.corp.adobe.com/mc-visitor/visitor-js-client/blob/0a43fd23937eca42ffb55589dbf5984e68baa3ab/lib/units/makeDestinationPublishing.js#L454)

## Alternatives

- Tell the customer to ignore the error, since it only appears in Alloy debug mode.
- Try and `fetch()` the failed URL destination to see the response code.
  - This may run into CORS issues and would make firing a referrer take longer.

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Audiences/injectProcessDestinations.js
+++ b/src/components/Audiences/injectProcessDestinations.js
@@ -55,7 +55,6 @@ export default ({
             // We intentionally do not throw an error if destinations fail. We
             // consider it a non-critical failure and therefore do not want it to
             // reject the promise handed back to the customer.
-            logger.error(createResultLogMessage(urlDestination, false));
           });
       })
     ).then(noop);
--- a/test/unit/specs/components/Audiences/injectProcessDestinations.spec.js
+++ b/test/unit/specs/components/Audiences/injectProcessDestinations.spec.js
@@ -144,9 +144,6 @@ describe("Audiences::injectProcessDestinations", () => {
       expect(logger.info).toHaveBeenCalledWith(
         "URL destination succeeded: http://test.zyx"
       );
-      expect(logger.error).toHaveBeenCalledWith(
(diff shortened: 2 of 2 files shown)


src/components/Audiences/injectProcessDestinations.js
```diff
@@ -55,7 +55,6 @@ export default ({
             // We intentionally do not throw an error if destinations fail. We
             // consider it a non-critical failure and therefore do not want it to
             // reject the promise handed back to the customer.
-            logger.error(createResultLogMessage(urlDestination, false));
```
ninaceban, line 58: maybe we log a warning, or an error only when debug is enabled?

ninaceban (approved):

src/components/Audiences/injectProcessDestinations.js
carterworks: An error in debug mode is the current behavior. However, the error message reads "URL destination failed." We don't have a way to know if it failed because the server returned a 400 response code or an empty 200 response code (empty 200 response codes were what prompted this ticket).

jonsnyder (approved):
If there is an error thrown the browser should show that error anyways, so I'm good with removing this.

PR #968, opened 2023-02-24
CJM-42163 Support AJO click-tracking

XDex:
## Description

Support AJO click-tracking

In cases when for the same click we have multiple AJO click-tracking selectors on nested DOM elements matching, we need to include just the label of the innermost matching click-track in the outgoing interact event, i.e. the one closest to the clicked element in the DOM tree.
As in the current click collection implementation we're interating through **all** selectors from the local click-tracking cache trying to match each of them to the clicked element DOM path, we need a way of determining the innermost matching click-track. In order to accomplish this, we're assigning a weight to each matched AJO click-track with a label, which then enables us to determine the label to be included in the event as the one with the lowest weight (i.e. closest to the clicked-element).

## Related Issue

https://jira.corp.adobe.com/browse/CJM-42163

## Motivation and Context

Support AJO click-tracking, see more details [in the Wiki](https://wiki.corp.adobe.com/pages/viewpage.action?spaceKey=CJM&title=AJO+Web+Channel+in+Campaign+-+Click+Tracking+Support)

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Personalization/createExecuteDecisions.js
+++ b/src/components/Personalization/createExecuteDecisions.js
@@ -17,15 +17,22 @@ const DEFAULT_ACTION_TYPE = "defaultContent";
 const identity = item => item;
 
 const buildActions = decision => {
-  const meta = {
+  const decisionMeta = {
     id: decision.id,
     scope: decision.scope,
     scopeDetails: decision.scopeDetails
   };
 
-  return decision.items.map(item =>
-    assign({ type: DEFAULT_ACTION_TYPE }, item.data, { meta })
-  );
+  return decision.items.map(item => {
+    const meta =
+      item.characteristics && item.characteristics.trackingLabel
+        ? assign(
+            { trackingLabel: item.characteristics.trackingLabel },
+            decisionMeta
+          )
+        : decisionMeta;
+    return assign({ type: DEFAULT_ACTION_TYPE }, item.data, { meta });
+  });
 };
 
 const processMetas = (logger, actionResults) => {
--- a/src/components/Personalization/createExecuteDecisions.js
+++ b/src/components/Personalization/createExecuteDecisions.js
@@ -16,6 +16,14 @@ const DEFAULT_ACTION_TYPE = "defaultContent";
 
 const identity = item => item;
 
+const getItemMeta = (item, decisionMeta) =>
+  item.characteristics && item.characteristics.trackingLabel
+    ? assign(
+        { trackingLabel: item.characteristics.trackingLabel },
+        decisionMeta
+      )
+    : decisionMeta;
+
 const buildActions = decision => {
   const decisionMeta = {
     id: decision.id,
@@ -23,16 +31,11 @@ const buildActions = decision => {
     scopeDetails: decision.scopeDetails
   };
 
-  return decision.items.map(item => {
-    const meta =
-      item.characteristics && item.characteristics.trackingLabel
-        ? assign(
-            { trackingLabel: item.characteristics.trackingLabel },
-            decisionMeta
-          )
-        : decisionMeta;
-    return assign({ type: DEFAULT_ACTION_TYPE }, item.data, { meta });
-  });
+  return decision.items.map(item =>
+    assign({ type: DEFAULT_ACTION_TYPE }, item.data, {
+      meta: getItemMeta(item, decisionMeta)
+    })
+  );
 };
 
 const processMetas = (logger, actionResults) => {
--- a/src/components/Personalization/dom-actions/clicks/collectClicks.js
+++ b/src/components/Personalization/dom-actions/clicks/collectClicks.js
@@ -19,31 +19,76 @@ const getMetasIfMatches = (
 ) => {
   const { documentElement } = document;
   let element = clickedElement;
+  let i = 0;
 
   while (element && element !== documentElement) {
     if (matchesSelectorWithEq(selector, element)) {
-      return getClickMetasBySelector(selector);
+      const matchedMetas = getClickMetasBySelector(selector);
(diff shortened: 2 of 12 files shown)


src/components/Personalization/createExecuteDecisions.js
```diff
@@ -23,9 +23,12 @@ const buildActions = decision => {
...
   };
 
-  return decision.items.map(item =>
-    assign({ type: DEFAULT_ACTION_TYPE }, item.data, { meta })
-  );
+  return decision.items.map(item => {
+    if (item.characteristics && item.characteristics.trackingLabel) {
+      meta.trackingLabel = item.characteristics.trackingLabel;
+    }
+    return assign({ type: DEFAULT_ACTION_TYPE }, item.data, { meta });
```
carterworks, line 30: If you want each item in the decision to have the exact same assigned tracking label, then this is fine. Otherwise, you have a bug.

Run this code and examine the result
```js
const DEFAULT_ACTION_TYPE = "default_action_type";

const assign = (...args) => Object.assign({}, ...args);

const buildActions = decision => {
  const meta = {
    id: decision.id,
    scope: decision.scope,
    scopeDetails: decision.scopeDetails
  };

  return decision.items.map(item => {
    if (item.characteristics && item.characteristics.trackingLabel) {
      meta.trackingLabel = item.characteristics.trackingLabel;
    }
    return assign({ type: DEFAULT_ACTION_TYPE }, item.data, { meta });
  });
};

const decision = {
  id: "decision-id",
  scope: "decision-scope",
  scopeDetails: "decision-scope-details",
  items: [
    {
      characteristics: {
        trackingLabel: "tracking-label-1"
      },
      data: { value: "lorem" }
    },
    {
      characteristics: {
        trackingLabel: "tracking-label-2"
      },
      data: { value: "ipsum" }
    },
    {
      characteristics: {
        trackingLabel: "tracking-label-3"
      },
      data: { value: "sit dolor" }
    }
  ]
};

console.log(buildActions(decision));
```
All the items have the `meta.trackingLabel` of the final `item` because of the shared `meta` object. Line 28 edits the reference and then line 30 adds a reference to final object. If you want to have each item have a different label, then you need to create a new `meta` object in each iteration.

XDex: Good catch, thanks @carterworks !

XDex: Fixed

carterworks: Nicely done.

carterworks (approved):

src/components/Personalization/dom-actions/clicks/collectClicks.js
```diff
@@ -19,31 +19,76 @@ const getMetasIfMatches = (
 ) => {
   const { documentElement } = document;
   let element = clickedElement;
+  let i = 0;
 
   while (element && element !== documentElement) {
     if (matchesSelectorWithEq(selector, element)) {
-      return getClickMetasBySelector(selector);
+      const matchedMetas = getClickMetasBySelector(selector);
```
ninaceban, line 26: Can we just extract the first label found? 
As far as I saw in the design wiki we should pick up the first label from one of the matching selectors.

ninaceban (changes requested):
Thank you for your contribution @XDex .
I have looked over the PR and there is a bit of disconnect between the design wiki and the code. The implementation could be simplified.
It would help if you could add some use case scenarios to PR description, so it's easier to follow why the implementation diverges from proposed design.

src/components/Personalization/dom-actions/clicks/collectClicks.js
XDex: We can't just extract the first label, since we don't know for sure whether it belongs to the innermost DOM element, or whether it's higher up the DOM tree. This is mentioned explicitly in the design wiki:
> In case of multiple click-tracking selector matches, all matching propositions will be included in a single interact event, with the event propositionAction → label set to the label of the first click-tracking proposition having a label, **when traversing the DOM tree from the innermost to the outermost DOM element**.

src/components/Personalization/createExecuteDecisions.js
```diff
@@ -17,15 +17,22 @@ const DEFAULT_ACTION_TYPE = "defaultContent";
...
   };
 
-  return decision.items.map(item =>
-    assign({ type: DEFAULT_ACTION_TYPE }, item.data, { meta })
-  );
+  return decision.items.map(item => {
+    const meta =
+      item.characteristics && item.characteristics.trackingLabel
+        ? assign(
+            { trackingLabel: item.characteristics.trackingLabel },
```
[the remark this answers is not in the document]
XDex: Fixed

XDex:
@ninaceban I've pushed some fixes, please take another look. Thank you!
> I have looked over the PR and there is a bit of disconnect between the design wiki and the code.
> ...why the implementation diverges from proposed design.

Could you please clarify what exactly is diverging from the design in the implementation?

jonsnyder (approved):

ninaceban:
> @ninaceban I've pushed some fixes, please take another look. Thank you!
> 
> > I have looked over the PR and there is a bit of disconnect between the design wiki and the code.
> > ...why the implementation diverges from proposed design.
> 
> Could you please clarify what exactly is diverging from the design in the implementation?

My concern was the logic of extracting all the labels and setting a weight. We discussed offline about the reasoning behind the need of extracting all vs extracting first.
I'd like to recommend that when there is a specific logic outlined in the PR, it would be much easier for the reviewer to review the code when there is a description on why such steps were taken, or why this was needed, some use cases/edge cases.

ninaceban (approved):

PR #978, opened 2023-04-13
PDCL-10238 - encoding the cookie value for destinations

ninaceban:
## Description

The legacy behavior of cookie destinations was to use encodeURIComponent on all the values. The cookie library Alloy uses has a different behavior. Some customers are complaining that their cookie destinations are not the same from Audience Manager. 

This was most seen when including an "=" in the cookie value. i.e:
* Legacy behavior: setCookie(“name”, “joe=thebest”) would result in the following cookie: ` joe%3Dthebest`
* Existing behavior: setCookie(“name”, joe=thebest”) would result in the following cookie: ` joe=thebest`

This PR updates the encoding method for cookie destinations to the legacy behavior. All other cookies in Web SDK are still written with the existing behavior.

## Related Issue

https://jira.corp.adobe.com/browse/PDCL-10238

## Types of changes

## Checklist:

### DIFF
--- a/src/utils/cookieJar.js
+++ b/src/utils/cookieJar.js
@@ -12,4 +12,9 @@ governing permissions and limitations under the License.
 
 import cookie from "@adobe/reactor-cookie";
 
-export default cookie;
+ export const cookieJarWithEncoding = cookie.withConverter({
+  write: (value) => {
+    return encodeURIComponent(value);
+  }
+});
+ export const cookieJar = cookie;
--- a/src/components/Audiences/index.js
+++ b/src/components/Audiences/index.js
@@ -10,12 +10,15 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import { createLoggingCookieJar, cookieJar } from "../../utils";
+import { createLoggingCookieJar, cookieJarWithEncoding } from "../../utils";
 import injectProcessDestinations from "./injectProcessDestinations";
 import injectProcessResponse from "./injectProcessResponse";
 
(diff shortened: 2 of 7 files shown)


src/utils/cookieJar.js
```diff
@@ -12,4 +12,9 @@ governing permissions and limitations under the License.
 
 import cookie from "@adobe/reactor-cookie";
 
-export default cookie;
+ export const cookieJarWithEncoding = cookie.withConverter({
+  write: (value) => {
+    return encodeURIComponent(value);
```
carterworks, line 17: I don't understand how this is different than the default `js-cookie` behavior, since they both use `encodeURIComponent()`. 
Here is the default converter from `js-cookie`, for reference

```js
function write (value) {
  return encodeURIComponent(value).replace(
    /%(2[346BF]|3[AC-F]|40|5[BDE]|60|7[BCD])/g,
    decodeURIComponent
  )
}
``` 
Is the difference that we DO need to encode special characters like # and $ (which is what that regex does-it matches the URI-encoded codes for special characters)?

carterworks: Ignore that comment. I read the ticket description. The answer is "yes, we do need to encode everything all of the time".

carterworks (approved):

src/utils/cookieJar.js
```diff
@@ -12,4 +12,9 @@ governing permissions and limitations under the License.
 
 import cookie from "@adobe/reactor-cookie";
```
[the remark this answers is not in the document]
ninaceban: Yes, that was the plan B.

PR #1000, opened 2023-06-06
Add deprecated validator with tests

jonsnyder:
## Description

This allows you to mark a field as deprecated which has these consequences:

1. Log warning when specifying a deprecated setting.
2. Throw error when both old and new settings are specified and they are different.
3. In the returned validated settings only return the new setting.

This requires a logger to be specified when calling the validator. This can be done by binding this in the validator.
```
myValidator.call({ logger }, value);
```

## Types of changes

## Checklist:


src/utils/validation/createObjectOfValidator.js
```diff
@@ -13,34 +13,59 @@ governing permissions and limitations under the License.
...
-    const subPath = path ? `${path}.${subKey}` : subKey;
-    try {
-      const validatedValue = subSchema(subValue, subPath);
-      if (validatedValue !== undefined) {
-        validatedObject[subKey] = validatedValue;
+    // copy over unknown properties first
+    const validatedObject = { ...value };
+    const errors = {};
+    const redoFields = {};
+    const parent = new Proxy(value, {
```
carterworks, line 24: Are we still striving for IE11 support or does this get polyfilled?

src/utils/validation/createDeprecatedValidator.js
```diff
@@ -0,0 +1,41 @@
...
+    } = value;
+    const validatedOldValue = oldSchema(oldValue, path);
+
+    if (validatedOldValue !== undefined) {
+      if (newValue !== undefined && newValue !== validatedOldValue) {
+        throw new Error(
+          `'${path}': The field '${oldField}' is deprecated. Use '${newField}' instead.`
+        );
+      } else if (this && this.logger) {
+        this.logger.warn(
```
carterworks, line 32: The deprecated validator will call this.logger.warn() if a deprecated option is used. However, the call to validate the config (line 62 of src/core/buildAndValidateConfig.js)  is before setDebugEnabled()is called (line 65). Therefore, if debugEnabled: true in the config, the logger will not print anything because debug mode is not enabled (yet).
Thoughts?

src/utils/validation/createDeprecatedValidator.js
```diff
@@ -0,0 +1,41 @@
...
+      [oldField]: oldValue,
+      [newField]: newValue,
+      ...otherValues
+    } = value;
+    const validatedOldValue = oldSchema(oldValue, path);
+
+    if (validatedOldValue !== undefined) {
+      if (newValue !== undefined && newValue !== validatedOldValue) {
+        throw new Error(
+          `'${path}': The field '${oldField}' is deprecated. Use '${newField}' instead.`
```
ninaceban, line 29: should we explicitly say in the log message that the customer should remove the old config when both are present?

src/utils/validation/index.js
```diff
@@ -10,14 +10,71 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+ * return value of that is sent to the rightValidator. For example, when calling
+ * `string().nonEmpty().required()` the following chain is built up:
+ * ```
+ *              *
+ *            /   \
+ *          *     required
+ *        /   \
+ *      *     nonEmpty
+ *    /   \
+ * base   string
```
ninaceban, line 40: love it 👍

ninaceban (approved):
looks good to me

src/utils/validation/index.js
```diff
@@ -10,14 +10,71 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+ * or undefined values). To accomplish this, the validator needs to have a check
+ * at the begining of the function, short circuiting the validation logic and
+ * returning value if value is null or undefined. `default` and `required` do
+ * not want this null check though. Indeed, `default` should return the default
+ * value if value is null, and `required` should throw an error if value is
+ * null.
+ *
+ * So to keep from having to have a null check in front of most validators,
+ * nullSafeChain allows you to chain a validator in a null-safe way.
+ */
```
Aaronius, line 69: 😍

PR #995, opened 2023-05-23
Support per-command overrides of the datastream ID

carterworks:
## Description

Support per-command overrides of the datastream ID.

```js
await alloy("configure", {
  orgId: "5BFE274A5F6980A50A495C08@AdobeOrg",
  edgeConfigId: "bc1a10e0-aee4-4e0e-ac5b-cdbb9abbec83"
});

const isDevMode = window.location.search.includes("alloy_debug");
const sendEventOptions = isDevMode ? {
  edgeConfigOverrides: {
    edgeConfigId: "bc1a10e0-aee4-4e0e-ac5b-cdbb9abbec83:dev"
  }
} : {};
await alloy("sendEvent", sendEventOptions);
```

It is supported on all the commands that support config overrides, except `configure`:

* `sendEvent`
* `setConsent`
* `getIdentity`
* `appendIdentityToUrl`

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Improvement (non-breaking change which does not add functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have signed the [Adobe Open Source CLA](https://opensource.adobe.com/cla.html) or I'm an Adobe employee.
- [x] I have made any necessary test changes and all tests pass.
- [x] I have run the Sandbox successfully.

PR #1001, opened 2023-06-09
Deprecate edgeConfigId in favor of datastreamId

carterworks:
## Description

As mentioned in #995, we should rename `edgeConfigId` to `datastreamId`. This adds the `deprecated()` validator (as created in #1000, so merge this PR after that one) to `edgeConfigId`. It adds a warning whenever `edgeConfigId` is used in the `configure` command.

I did not use this validator on `datasetId` because it is deprecated in favor of a deeply nested value, nor did I deprecate `eventMergeId` because it does not have a replacement.

## Related Issue

https://jira.corp.adobe.com/browse/PDCL-10587

## Types of changes

## Checklist:


jonsnyder (changes requested):
This looks great. Can you update the functional tests so that they use datastreamId, and then add a test to check that the deprecated message is logged when using edgeConfigId, and it uses that as the datastreamId?

test/functional/specs/Command Logic/C11634155.js
```diff
@@ -0,0 +1,57 @@
...
+
+test("Test C11634155: Deprecates options like edgeConfigId and warns with use", async t => {
+  const alloy = createAlloyProxy();
+  await t.debug();
+  const warning = await alloy.configureErrorMessage(config);
+
+  await t.expect(warning).match(/The field 'edgeConfigId' is deprecated./);
+});
+
+test.skip("Test C11634155: When specifying a deprecated option like edgeConfigId, it uses the specified alternative, datastreamId", async t => {
```
carterworks, line 47: See https://github.com/adobe/alloy/pull/1000/files#r1227217825

src/core/buildAndValidateConfig.js
```diff
@@ -57,6 +44,23 @@ const buildAllOnInstanceConfiguredExtraParams = (
   );
 };
 
+const wrapLoggerInQueue = logger => {
```
jonsnyder, line 47: This wrapping doesn't correctly handle the "enabled" flag on the logger. This is fine, lets just add a comment to the logger to explain this.

jonsnyder (approved):

shammowla (approved):

test/functional/specs/Command Logic/C11634155.js
```diff
@@ -0,0 +1,61 @@
...
+
+  const alloy = createAlloyProxy();
+  await alloy.configure(config);
+
+  await logger.warn.expectMessageMatching(
+    /The field 'edgeConfigId' is deprecated./
+  );
+});
+
+test("Test C11634155: When specifying a deprecated option like edgeConfigId, it uses the specified alternative, datastreamId", async t => {
```
ninaceban, line 51: can we add a test where both configs are specified and assert that we throw an error?

carterworks: Done.

carterworks:
I will merge this after #1000 is merged.

PR #1024, opened 2023-08-22
Dismiss Messages & url click

shandilya3:
## Types of changes

## Checklist:


sandbox/src/App.js
```diff
@@ -96,6 +97,9 @@ function BasicExample() {
             <li>
               <a href="/configOverrides">Config Overrides</a>
             </li>
+            <li>
+              <Link to="/decisioningEngine">Decisioning Engine</Link>
```
jasonwaters, line 101: I know decisioning engine is used here, but would it be better if this was renamed to "In-app Messages"?  Might make it more obvious what is being tested on that page.

sandbox/src/App.js
```diff
@@ -125,6 +129,7 @@ function BasicExample() {
           <Route path="/redirectedNewPage" component={RedirectedNewPage} />
           <Route path="/identity" component={Identity} />
           <Route path="/configOverrides" component={ConfigOverrides} />
+          <Route path="/decisioningEngine" component={DecisioningEngine} />
```
jasonwaters, line 132: If renamed above, this route should also change to something like `/inAppMessages`

sandbox/src/components/ContentSecurityPolicy.js
```diff
@@ -20,7 +20,7 @@ export default function ContentSecurityPolicy() {
         http-equiv="Content-Security-Policy"
         // cdn.tt.omtrdc.net is necessary for Target VEC to function properly.
         // *.sc.omtrdc.net is necessary for Analytics Data Insertion API to function properly
-        content={`default-src 'self';
+        content={`default-src 'self' blob:;
```
jasonwaters, line 23: What does `blob:` do here

sandbox/src/components/DecisioningEngineDemo/DecisioningEngine.js
```diff
@@ -0,0 +1,111 @@
+import React, { useState, useEffect } from "react";
+import ContentSecurityPolicy from "../ContentSecurityPolicy";
+import "./DecisionEngineStyle.css";
+
+export default function DecisionEngine() {
+  const [realResponse, setRealResponse] = useState(null);
+
+  async function transformPayload(json) {
```
jasonwaters, line 8: Please merge `consequenceAdapter` into your branch (if you haven't already).  And then you can remove this `transformPayload` method.  It's taken care of elsewhere.

src/components/Personalization/in-app-message-actions/actions/displayIframeContent.js
```diff
@@ -60,41 +95,61 @@ const createIframeClickHandler = (container, collect) => {
...
+        //   eventData: {
+        //     action,
+        //     interaction
+        //   }
+        // });
+        if (link && interaction === "clicked") {
+          link = decodeURIComponent(link);
+          setWindowLocationHref(link);
+        } else if (action === "dismiss") {
+          const containerElement = document.getElementById(ELEMENT_TAG_ID);
```
jasonwaters, line 124: I propose you extract the dismiss code into a separate function called `dismissMessage` and call it here.

the code could be simplified like so...

```js
import { removeElements } from "../utils";

const dismissMessage = () =>
  [OVERLAY_TAG_CLASSNAME, ELEMENT_TAG_CLASSNAME].forEach(removeElements);
```

src/components/Personalization/in-app-message-actions/actions/displayIframeContent.js
```diff
@@ -137,10 +179,30 @@ const createContainerElement = settings => {
...
+    opacity: backdropOpacity,
+    backgroundColor: backdropColor
+  });
+
+  return element;
+};
 
-  const { content, contentType } = settings;
+export const displayHTMLContentInIframe = (settings, collect) => {
+  const { content, contentType, mobileParameters } = settings;
```
jasonwaters, line 206: call `dismissMessage()` here.  Since any pre-existing message should be removed if one exists. (only a single message and overlay should be displayed at a time)

src/components/Personalization/in-app-message-actions/actions/displayIframeContent.js
```diff
@@ -150,12 +212,18 @@ const displayHTMLContentInIframe = (settings, collect) => {
...
+    createIframeClickHandler(container, collect, mobileParameters)
   );
 
   container.appendChild(iframe);
 
-  document.body.append(container);
+  document.body.appendChild(container);
+
+  if (mobileParameters.uiTakeover) {
+    const overlay = createOverlayElement(mobileParameters);
```
jasonwaters, line 223: It seems to me that the overlay element should be added to the DOM before the message `container`.  Then you can remove the zIndex style.  zIndex may not always work, depending on the customers's CSS.  So it's best if we can add elements to the dom in an order that doesn't require it.

jasonwaters (changes requested):
Looking good, left a few comments for ya.

sandbox/src/components/ContentSecurityPolicy.js
shandilya3: To support Blob URLs as intended in our code
https://github.com/adobe/alloy/blob/e142bdc9747e86a54cd9fefaef23b31aa257d079/src/components/Personalization/in-app-message-actions/actions/displayIframeContent.js#L142

src/components/Personalization/in-app-message-actions/actions/displayIframeContent.js
shandilya3: Done.

src/components/Personalization/in-app-message-actions/actions/displayIframeContent.js
shandilya3: Done

jasonwaters (approved):

PR #1031, opened 2023-09-07
[CJM-46521] Send interact events when in-app messages are clicked

jasonwaters:
## Description

This PR ensures `interact` events are sent to edge when a user interacts with an in-app message.  Like when they close the message, or click one of the buttons within it.

screenshot-2023-09-07 at 17 19 48@2x

👾 

screenshot-2023-09-07 at 17 21 04@2x

🎧 

screenshot-2023-09-07 at 17 22 22@2x

🐮 

## Related Issue

- https://jira.corp.adobe.com/browse/CJM-46521

### DIFF
--- a/sandbox/src/components/InAppMessagesDemo/InAppMessages.js
+++ b/sandbox/src/components/InAppMessagesDemo/InAppMessages.js
@@ -1,5 +1,5 @@
 /* eslint-disable no-bitwise */
-import React, { useState, useEffect } from "react";
+import React from "react";
 import ContentSecurityPolicy from "../ContentSecurityPolicy";
 import "./InAppMessagesStyle.css";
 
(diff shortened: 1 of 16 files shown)


src/components/Personalization/in-app-message-actions/utils.js
```diff
@@ -37,3 +39,38 @@ export const removeElementById = id => {
...
+
+export const parseAnchor = anchor => {
+  const nothing = {};
+
+  if (!anchor || anchor.tagName.toLowerCase() !== "a") {
+    return nothing;
+  }
+
+  const { href } = anchor;
+  if (!startsWith(href, "adbinapp://")) {
```
shandilya3, line 51: may be add one more check
`if (!href || !startsWith(href, "adbinapp://")) {
    return nothing;
}`

shandilya3 (approved):

sandbox/src/components/InAppMessagesDemo/InAppMessages.js
```diff
@@ -40,62 +40,64 @@ const uuidv4 = () => {
...
-      responseBody: realResponse
-    });
   };
 
   return (
     <div>
       <ContentSecurityPolicy />
       <h1>In App Messages For Web</h1>
-      <button onClick={renderDecisions}>Execute Decisions and Render</button>
+      <button onClick={() => renderDecisions()}>
```
shandilya3, line 98: Here, When the button is clicked, this anonymous function is created and then immediately invoked just curious why would we create a new function on every render? Won't it impact performance?
@jasonwaters

jasonwaters: no, this does not impact performance.

PR #1040, opened 2023-09-22
Fix show containers timing

jonsnyder:
## Description

The personalization refactor contained a change with when showContainers was called. Previously, it was waiting for all rendering to be complete before showing containers. Before the refactor, showContainers was called immediately after starting to render everything. (see https://github.com/adobe/alloy/blob/e1105f1fd43224b46f392798eca823ae8a99fc72/src/components/Personalization/createAutoRenderingHandler.js#L54) With the change in this PR, I've updated the fetchDataHandler to follow the same old behavior.

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Personalization/createFetchDataHandler.js
+++ b/src/components/Personalization/createFetchDataHandler.js
@@ -69,15 +69,12 @@ export default ({
           [...pagePropositions, ...currentViewPropositions],
           nonRenderedPropositions
         ));
-        render()
-          .then(decisionsMeta => {
-            showContainers();
-            handleNotifications(decisionsMeta);
-          })
-          .catch(e => {
-            showContainers();
-            throw e;
-          });
+        render().then(handleNotifications);
+
+        // Render could take a long time especially if one of the renders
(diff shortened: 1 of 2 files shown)


src/components/Personalization/createFetchDataHandler.js
```diff
@@ -69,15 +69,12 @@ export default ({
...
-        render()
-          .then(decisionsMeta => {
-            showContainers();
-            handleNotifications(decisionsMeta);
-          })
-          .catch(e => {
-            showContainers();
-            throw e;
-          });
+        render().then(handleNotifications);
```
jfkhoury, line 72: Should we keep the `catch` in here?

jonsnyder: The catch was unnecessary anyways because render shouldn't throw an exception. The rendering exceptions are caught and logged inside of render.

liljenback (approved):

carterworks (approved):

PR #1041, opened 2023-09-23
Fix empty view change issues

jonsnyder:
## Description

I found this bug while testing the alloy-samples with no SPA decisions. Previously it was not waiting for applyPropositions to finish rendering before creating the bottom of page event. Also if a viewChange event had renderDecisions: true and pending display notifications, only one of those proposition arrays would be included in the event.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Improvement (non-breaking change which does not add functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] I have signed the [Adobe Open Source CLA](https://opensource.adobe.com/cla.html) or I'm an Adobe employee.
- [ ] I have made any necessary test changes and all tests pass.
- [ ] I have run the Sandbox successfully.

PR #1042, opened 2023-09-23
DO NOT MERGE!!! All changes from last release for Nina

jonsnyder:
## Description

DO NOT MERGE!!! This is all the changes for the new release.
* Top and bottom events
* matching views in view cache with lowercase
* Includes bug fixes PRs #1040 and #1041 

## Types of changes

## Checklist:


src/components/Personalization/createFetchDataHandler.js
```diff
@@ -9,21 +9,87 @@ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTA
...
+  mergeQuery,
+  collect,
+  processPropositions,
+  createProposition,
+  pendingDisplayNotifications
 }) => {
-  return ({ decisionsDeferred, personalizationDetails, event, onResponse }) => {
+  return ({ cacheUpdate, personalizationDetails, event, onResponse }) => {
     if (personalizationDetails.isRenderDecisions()) {
       hideContainers(prehidingStyle);
```
ninaceban, line 28: add else clause here with `showContainers`

src/components/Personalization/createOnClickHandler.js
```diff
@@ -35,10 +35,10 @@ export default ({
         const xdm = { eventType: INTERACT };
         const scope = decisionsMeta[0].scope;
 
-        if (scope !== PAGE_WIDE_SCOPE) {
+        if (isNonEmptyString(scope) && scope !== PAGE_WIDE_SCOPE) {
           xdm.web = {
             webPageDetails: {
-              viewName: scope
+              viewName: scope.toLowerCase()
```
ninaceban, line 41: `scope` is stored lowercase, no need for this here.

src/components/Personalization/createViewCacheManager.js
```diff
@@ -10,39 +10,84 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+    const emptyViewProposition = createProposition(
+      {
+        scope: viewName,
+        scopeDetails: {
+          characteristics: {
+            scopeType: "view"
+          }
+        },
+        items: [
+          {
```
ninaceban, line 37: Can we remove the `Default Content Item` schema it will causes confusion. 
`Default Content Item` schema is used to notify client side that the user has entered a Target Activity but it doesn't have an HTML change to render.

src/components/Personalization/event.js
```diff
@@ -19,6 +19,11 @@ export const mergeDecisionsMeta = (
   eventType,
   eventLabel = ""
 ) => {
+  // Do not send a display notification with no decisions. Even empty view changes
+  // should include a proposition.
```
ninaceban, line 23: let's move this logic where the event is triggered instead, this functions are calling `event` API

src/components/Personalization/handlers/createProcessDomAction.js
```diff
@@ -0,0 +1,39 @@
...
+    if (!selector) {
+      logger.warn("Invalid DOM action data: missing selector.", item.getData());
+      return {};
+    }
+    storeClickMetrics({ selector, meta: item.getMeta() });
+    return { setRenderAttempted: true, includeInNotification: false };
+  }
+
+  if (!modules[type]) {
+    logger.warn("Invalid DOM action data: unknown type.", item.getData());
```
ninaceban, line 30: In such cases we should set `renderAttempted: false, includeInNotifications:false`

src/components/Personalization/handlers/createProcessPropositions.js
```diff
@@ -0,0 +1,202 @@
...
+        }
+        return true;
+      })
+      .catch(error => {
+        if (logger.enabled) {
+          const { message, stack } = error;
+          const errorMessage = `Failed to execute action ${item.toString()}. ${message} ${stack}`;
+          logger.error(errorMessage);
+        }
+        return false;
```
ninaceban, line 29: logger level should be warrning

src/components/Personalization/handlers/createProcessPropositions.js
```diff
@@ -0,0 +1,202 @@
...
+    items,
+    proposition
+  }) => {
+    let renderers = [...existingRenderers];
+    let returnedPropositions = [...existingReturnedPropositions];
+    let returnedDecisions = [...existingReturnedDecisions];
+    let renderedItems = [];
+    let nonRenderedItems = [];
+    let itemRenderers = [];
+    let atLeastOneWithNotification = false;
```
ninaceban, line 64: so many flags, a decision diagram would be very helpful for CE engineers debugging an issue.

src/components/Personalization/index.js
```diff
@@ -43,41 +49,45 @@ const createPersonalization = ({ config, logger, eventManager }) => {
...
+    [schema.DEFAULT_CONTENT_ITEM]: processDefaultContent,
+    [schema.DOM_ACTION]: createProcessDomAction({
+      modules,
+      logger,
+      storeClickMetrics
+    }),
+    [schema.HTML_CONTENT_ITEM]: createProcessHtmlContent({ modules, logger }),
+    [schema.REDIRECT_ITEM]: createProcessRedirect({
+      logger,
+      executeRedirect: url => window.location.replace(url),
```
ninaceban, line 71: extract this to a function

src/components/Personalization/handlers/createProcessRedirect.js
```diff
@@ -0,0 +1,28 @@
...
+  const { content } = item.getData() || {};
+
+  if (!content) {
+    logger.warn("Invalid Redirect data", item.getData());
+    return {};
+  }
+
+  const render = () => {
+    return collect({ decisionsMeta: [item.getMeta()] }).then(() => {
+      executeRedirect(content);
```
ninaceban, line 22: Render would be called only when `renderDecisions:true`, otherwise it should be just part of the array of propositions returned.
Collect should be called here with `documentMayUnload` argument.
You should try/catch errors here and log the error and showContainers.

src/components/Personalization/handlers/injectCreateProposition.js
```diff
@@ -0,0 +1,98 @@
...
+          return scope;
+        }
+        return scope;
+      },
+      getScopeType() {
+        if (scope === PAGE_WIDE_SCOPE || isPageWideSurface(scope)) {
+          return "page";
+        }
+        if (scopeType === "view") {
+          return "view";
```
ninaceban, line 63: as this gets part of the proposition, extract this into an enum of proposition types.

src/components/Personalization/validateApplyPropositionsOptions.js
```diff
@@ -9,14 +9,15 @@ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTA
...
-import { anything, objectOf, arrayOf } from "../../utils/validation";
+import { anything, objectOf, arrayOf, string } from "../../utils/validation";
 
 export const EMPTY_PROPOSITIONS = { propositions: [] };
 
 export default ({ logger, options }) => {
   const applyPropositionsOptionsValidator = objectOf({
-    propositions: arrayOf(objectOf(anything())).nonEmpty(),
-    metadata: objectOf(anything())
+    propositions: arrayOf(objectOf(anything())),
```
ninaceban, line 18: this validator should be `oneOf`:
- `viewName`
- `propositions` && `metadata`

ninaceban (changes requested):
Generally, I would recommend breaking this into a few smaller PRs, it touches one of the most impactful and feature rich components, Personalization. Mixing the whole component refactoring and adding a big feature on top of that becomes pretty complicated for the PR reviewer to understand the changes and how it will impact the backward compatibility. 
The `src/components/Personalization/handlers/createProcessPropositions.js` is pretty complex, would be good to break it into smaller descriptive pieces and comments. A design spec would also help CE rotation engineers that will potentially have to debug a related issue.

While testing Target use-cases, have found a few use-cases that needs to be fixed:
- when `renderDecisions: false` the redirect offer should not be executed, the proposition should be returned in the result
- the notification should be a `collect` instead of `interact` call
- when `renderDecisions:false` the prehiding snippet should be removed(`showContainers` should be called)
- when a click from a surface occurs the surface scope should not be  sent in a`viewName`, `viewName` will be populated only when the proposition `scopeType:view`
- when something fails, we log warnings, not errors

As improvements:
- the functional tests for AJO should be improved to have the assertions for rendering, click notifications `viewName` I have mentioned above.
- add test for `renderDecisions:false` and redirect offer being returned
- add a functional test to validate that prehiding style is removed:
              -- when no personalization payload is returned
              -- when `renderDecisions` is false
              -- when an exception while executing

jonsnyder:
Display notifications have been interact calls for a while now. I think we found collect calls have some problems and should only be used when the page is potentially unloading.

src/components/Personalization/createFetchDataHandler.js
jonsnyder: Fixed in #1058

src/components/Personalization/createOnClickHandler.js
```diff
@@ -35,10 +35,10 @@ export default ({
         const xdm = { eventType: INTERACT };
         const scope = decisionsMeta[0].scope;
 
-        if (scope !== PAGE_WIDE_SCOPE) {
+        if (isNonEmptyString(scope) && scope !== PAGE_WIDE_SCOPE) {
```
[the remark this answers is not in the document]
jonsnyder: Fixed in #1056

src/components/Personalization/createOnClickHandler.js
jonsnyder: Fixed in #1056

src/components/Personalization/handlers/createProcessDomAction.js
jonsnyder: Fixed in #1060

src/components/Personalization/handlers/createProcessPropositions.js
jonsnyder: Fixed in #1058

src/components/Personalization/handlers/createProcessRedirect.js
jonsnyder: If renderDecisions is false, processRedirect isn't even called, and the proposition is just returned as part of the returned propositions. See https://github.com/adobe/alloy/blob/bf813fa5da1da3011d29ddcf357a3a55f7316878/src/components/Personalization/createFetchDataHandler.js#L63.

Changing it to use collect is fixed in #1061 

The try/catch is done in https://github.com/adobe/alloy/blob/bf813fa5da1da3011d29ddcf357a3a55f7316878/src/components/Personalization/handlers/createProcessPropositions.js#L23
The showContainers is in https://github.com/adobe/alloy/blob/bf813fa5da1da3011d29ddcf357a3a55f7316878/src/components/Personalization/createFetchDataHandler.js#L77

src/components/Personalization/handlers/injectCreateProposition.js
jonsnyder: This is fixed in #1056

src/components/Personalization/validateApplyPropositionsOptions.js
jonsnyder: If you don't send in any propositions or a viewName, then this command does nothing. That's not an error. Additionally it is a valid use case to send in propositions without metadata. These could be page_wide_scope propositions that were not rendered automatically.

src/core/createEvent.js
```diff
@@ -115,7 +148,12 @@ export default () => {
       return shouldSendEvent;
     },
     getViewName() {
-      if (!userXdm || !userXdm.web || !userXdm.web.webPageDetails) {
+      if (
+        !userXdm ||
+        !userXdm.web ||
+        !userXdm.web.webPageDetails ||
+        !userXdm.web.webPageDetails.viewName
+      ) {
```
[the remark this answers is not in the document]
jonsnyder: Fixed in #1061

src/components/Personalization/createOnClickHandler.js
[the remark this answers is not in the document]
jonsnyder: Fixed in #1056

src/components/Personalization/event.js
jonsnyder: Fixed in #1061

src/components/Personalization/index.js
jonsnyder: Fixed in #1061

PR #1044, opened 2023-09-25
support "~timestampu" and  "~timestampz"

shandilya3:
~timestampu - the current device time in epoch format (seconds since epoch).
~timestampz - the current device time in ISO 8601 data elements and interchange format.
~sdkver - the current Adobe Experience Platform SDKs version string.

## Types of changes

## Checklist:

### DIFF
--- a/src/components/DecisioningEngine/createContextProvider.js
+++ b/src/components/DecisioningEngine/createContextProvider.js
@@ -47,7 +47,9 @@ export default ({ eventRegistry, window }) => {
       currentMinute: now.getMinutes(),
       currentMonth: now.getMonth(),
       currentYear: now.getFullYear(),
-      pageVisitDuration: currentTimestamp - pageLoadTimestamp
+      pageVisitDuration: currentTimestamp - pageLoadTimestamp,
+      "~timestampu": currentTimestamp / 1000,
+      "~timestampz": now.toISOString()
     };
   };
 
--- a/src/components/DecisioningEngine/createContextProvider.js
+++ b/src/components/DecisioningEngine/createContextProvider.js
@@ -12,6 +12,7 @@ governing permissions and limitations under the License.
 import getBrowser from "../../utils/getBrowser";
 import parseUrl from "../../utils/parseUrl";
 import flattenObject from "../../utils/flattenObject";
+import { version } from "../../../package.json";
 
 export default ({ eventRegistry, window }) => {
   const pageLoadTimestamp = new Date().getTime();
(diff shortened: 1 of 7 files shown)


src/components/DecisioningEngine/createContextProvider.js
```diff
@@ -74,7 +77,8 @@ export default ({ eventRegistry, window }) => {
     return {
       ...coreGlobalContext,
       ...getTimeContext(),
-      window: getWindowContext()
+      window: getWindowContext(),
+      "~sdkver": version
```
carterworks, line 81: The version can be imported from `src/constants/libraryVersion.js`. No need to add a dependency and plugin to the build.

carterworks (changes requested):

src/components/DecisioningEngine/createContextProvider.js
shandilya3: Thanks @carterworks  for bringing this up. I was also looking for existing use, guess didn't search with matching keyword.

test/unit/specs/components/DecisioningEngine/decisioningContext.sdkVersion.spec.js
```diff
@@ -0,0 +1,59 @@
...
+      })
+    );
+  });
+
+  it("does not satisfy rule due to unmatched dk version", () => {
+    setupResponseHandler(applyResponse, mockWindow({}), {
+      definition: {
+        key: "~sdkver",
+        matcher: "eq",
+        values: ["2.18.0-beta.0"]
```
jasonwaters, line 48: Guessing this test will fail every time the SDK version is updated.  I suggest mocking it instead.

src/components/DecisioningEngine/createContextProvider.js
jasonwaters: +1

test/unit/specs/components/DecisioningEngine/decisioningContext.sdkVersion.spec.js
shandilya3: Testing above with real libraryVersion, that will always match.
 This one is false test case.

carterworks (approved):

test/unit/specs/components/DecisioningEngine/decisioningContext.sdkVersion.spec.js
jasonwaters: 😆 ah that what i get for not reading it closely enough.

jasonwaters (approved):

PR #1045, opened 2023-09-28
In browser messages e2e

jasonwaters:
Updated IAM demo to use stage with a web IAM campaign

This PR updates this demo page 👇 
http://localhost:3000/inAppMessages

At the top of the `InAppMessages.js` file you can change `const configKey = "stage";`.  Doing so will use a standard sendEvent call to retrieve a real-life web campaign with IAM directly from konductor. 🎉 

Don't worry, the old method still works if you set `configKey` to one of the other options in prod.

The `consequenceAdapters` take care of translating consequences to work no matter the format they come in.

Here's an in app message coming from a web campaign:

screenshot-2023-09-28 at 14 16 49


jasonwaters:
hi @shandilya3 could you take a look at this PR when you have a minute?

shandilya3 (approved):
LGTM

PR #1048, opened 2023-10-05
Historical lookup indexed db

shandilya3:
## Description

IndexedDB for historical search
Wiki Doc:  https://wiki.corp.adobe.com/display/Target/Historical+Events+Storage+-IndexedDB+-+In+Progress

To run locally, NPM link to this PR branch https://github.com/adobe/aepsdk-rulesengine-typescript/pull/20

## Types of changes

## Checklist:


package.json
```diff
@@ -59,7 +59,7 @@
     }
   ],
   "dependencies": {
-    "@adobe/aep-rules-engine": "^2.0.1",
+    "@adobe/aep-rules-engine": "file:../aepsdk-rulesengine-typescript",
```
shandilya3, line 62: This is for local e2e test, will add dependency once rule engine changes are approved &  published.

src/components/DecisioningEngine/createDecisionHistory.js
```diff
@@ -10,15 +10,14 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-const QUALIFIED_EVENT_TYPE = "decisioning.propositionQualified";
+import { PropositionEventType } from "../Personalization/constants/propositionEventType";
```
carterworks, line 13: We don't allow importing between components in Alloy. If this needs to be shared between the two components, put it into `src/shared`.

src/components/DecisioningEngine/createOnResponseHandler.js
```diff
@@ -30,8 +30,11 @@ export default ({
       response.getPayloadsByType(PERSONALIZATION_DECISIONS_HANDLE)
     );
 
-    const propositions = decisionProvider.evaluate(context);
-
-    applyResponse({ viewName, renderDecisions, propositions });
+    return new Promise(resolve => {
+      decisionProvider.evaluate(context).then(propositions => {
+        applyResponse({ viewName, renderDecisions, propositions });
+        resolve();
```
carterworks, line 36: `new Promise()` and `resolve()` are unnecessary.

src/components/DecisioningEngine/index.js
```diff
@@ -9,27 +9,30 @@ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTA
...
-  const { orgId } = config;
-  const storage = createNamespacedStorage(
-    `${sanitizeOrgIdForCookieName(orgId)}.decisioning.`
-  );
-  const eventRegistry = createEventRegistry({ storage: storage.persistent });
+const createDecisioningEngine = () => {
+  const indexedDB = createIndexedDB();
+  indexedDB
+    .setupIndexedDB()
+    .then(() => console.log("IndexedDB setup is complete."));
```
carterworks, line 33: Please use the `logger` for logging messages.

src/components/DecisioningEngine/createIndexedDB.js
```diff
@@ -0,0 +1,128 @@
...
+          autoIncrement: true
+        });
+        objectStore.createIndex(
+          "iam_id_iam_eventType_index",
+          ["iam_id", "iam_eventType"],
+          { unique: false }
+        );
+      };
+
+      request.onsuccess = event => {
```
carterworks, line 47: [In the Identity component,](https://github.com/adobe/alloy/blob/main/src/components/Identity/injectAwaitIdentityCookie.js) we have a similar situation: we need to wait for the identity cookie before we firing off other requests. However, the `awaitIdentityCookie()` is implemented with Promises. 

I think if you use Promises here, it would make the implementation a neater and removes the need for `waitForDBInitialization()` and it's `setTimeout()`.

For example, if you write `setupIndexDB()` like so…

```js
export default () => {
  /** @type {Promise<IDBDatabase>} */
  let dbPromise;
  const setupIndexedDB = () => {
    dbPromise = new Promise((resolve, reject) => {
      // set up login
    });
    // instead of returning a promise that resolves to 
    // db and exposes it outside of this module, return
    // a cleanup function
    return dbPromise.then(db => () => db.close());
  };
}
```

Then you can use `dbPromise` like this…

```js
const addRecord = record => {
  return dbPromise.then(db => {
    const transaction = db.transaction("events", "readwrite");
    const objectStore = transaction.objectStore("events");
    const objectStoreRequest = objectStore.add(record);
    return new Promise((resolve, reject) => {
      objectStoreRequest.onsuccess = () => resolve(true);
      objectStoreRequest.onerror = (txEvent) => reject(txEvent.target.error);
    });
  })
};
```

And you don't need the `setTimeout` (which is kind of a code smell, in my opinion).

src/components/DecisioningEngine/createIndexedDB.js
```diff
@@ -0,0 +1,128 @@
...
+import { HISTORICAL_DATA_STORE } from "./constants";
+
+const dbName = HISTORICAL_DATA_STORE;
+
+export default () => {
+  let db;
+  const DB_INITIALIZATION_TIMEOUT = 200;
+
+  const setupIndexedDB = () => {
+    return new Promise((resolve, reject) => {
```
carterworks, line 21: [In the Identity component,](https://github.com/adobe/alloy/blob/main/src/components/Identity/injectAwaitIdentityCookie.js) we have a similar situation: we need to wait for the identity cookie before we firing off other requests. However, the `awaitIdentityCookie()` is implemented with Promises. 

I think if you use Promises here, it would make the implementation a little neater and removes the need for `waitForDBInitialization()` and it's `setTimeout()`.

For example, if you write `setupIndexDB()` like so…

```js
export default () => {
  let dbPromise;
  const setupIndexedDB = () => {
    dbPromise = new Promise((resolve, reject) => {
      // set up login
    });
    // no need to expose the DB outside of this module
    return dbPromise.then(() => {});
  };
}
```

Then you can use `dbPromise` like this…

```js
const addRecord = record => {
  return dbPromise.then(db => {
    const transaction = db.transaction("events", "readwrite");
    const objectStore = transaction.objectStore("events");
    const objectStoreRequest = objectStore.add(record);
    return new Promise((resolve, reject) => {
      objectStoreRequest.onsuccess = () => resolve(true);
      objectStoreRequest.onerror = (txEvent) => reject(txEvent.target.error);
    });
  })
};
```

This way, you don't need `waitForDBInitialization()` with it's `setTimeout()` or `getIndexDB()` (which are kind of a code smells, in my opinion).

carterworks (changes requested):

shandilya3:
We realized IndexedDB is more limited than expected. It's not as efficient as we had planned. Example, you can not write standard sequel statements, nor can you ask for results to be sorted (by timestamp for example). These limitations make the solution less performant than what we've already got with localStorage. With that in mind, it was decided that it would be best to stick with the existing implementation with the expectation with some changes to expect there will always be iam.id and iam.eventType in historical rulesets for IAM on the web.
We can still continue iterating on the IndexedDB approach, but we'll need to add a service worker for performance, and configuration options to the rule engine, among other things.  Since we will be adding a service worker for web push, **Jason** recommended revisiting IndexedDB at that time.

PR #1059, opened 2023-10-18
IAM: support multiple subscriptions

jasonwaters:
In this PR:

- added support for multiple subscriptions via the `subscribeRulesetItems` and `subscribeMessageFeed` commands.
- subscriptions now return an `unsubscribe` function that can be used to end the subscription.  An example use case  is in react apps when a component is rendered, it might create a subscription.  And when the component is removed, it can end the subscription.  An example of this is in the InAppMessage and MessageFeed sandbox.
- removed the debounce for writing local storage.
- only evaluate rulesets when an event includes a query.  This limits excessive ruleset evaluations for every `display` or `interact` event.
- removed `in-app-response-format` from personalization requests.


sandbox/src/components/InAppMessagesDemo/InAppMessages.js
```diff
@@ -66,59 +51,27 @@ if (alloyInstance !== window.alloy) {
...
+      }
+    }).then(({ unsubscribe }) => {
+      unsubscribeRulesetItems = () => unsubscribe();
+    });
+
+    return () => {
+      unsubscribeRulesetItems();
+    };
+  }, []);
+
```
jonsnyder, line 74: There is a race condition here if React tries to unsubscribe before "subscribeRulesetItems" completes. This might be a better pattern for documentation:

```
useEffect(() => {
  const unsubscribePromise = alloyInstance("subscribeRulesetItems", {
    callback: result => {
      console.log("subscribeRulessetItems", result);
    }
  });
  return () => {
    unsubscribePromise.then(unsubscribe => unsubscribe());
  }
,[]);
```

src/components/DecisioningEngine/createOnResponseHandler.js
```diff
@@ -30,6 +31,11 @@ export default ({
       response.getPayloadsByType(PERSONALIZATION_DECISIONS_HANDLE)
     );
 
+    // only evaluate events that include a personalization query
+    if (!eventContent.query) {
+      return;
+    }
+
```
jonsnyder, line 38: I think it would be better to include a method on event "hasQuery". I'm not sure if this test is unique to personalization requests though, I think we also add something to query for ECID.

src/components/DecisioningEngine/createSubscribeRulesetItems.js
```diff
@@ -27,20 +28,21 @@ const validateOptions = ({ options }) => {
...
-  let subscriptionHandler;
+  const subscription = createSubscription();
   let surfacesFilter;
   let schemasFilter;
 
   const run = ({ surfaces, schemas, callback }) => {
-    subscriptionHandler = callback;
+    const unsubscribe = subscription.add(callback);
     surfacesFilter = surfaces instanceof Array ? surfaces : undefined;
     schemasFilter = schemas instanceof Array ? schemas : undefined;
```
jonsnyder, line 38: You're using the same filters for all subscriptions. I'd expect the filters to be specific to this callback.

src/components/DecisioningEngine/utils.js
```diff
@@ -29,12 +28,11 @@ export const createRestoreStorage = (storage, storageKey) => {
...
   storageKey,
-  debounceDelay = 500,
   prepareFn = value => value
 ) => {
-  return debounce(value => {
+  return value => {
     storage.setItem(storageKey, JSON.stringify(prepareFn(value)));
-  }, debounceDelay);
+  };
 };
```
jonsnyder, line 36: :+1 getting rid of debounce

src/components/Personalization/createSubscribeMessageFeed.js
```diff
@@ -27,11 +28,12 @@ const validateOptions = ({ options }) => {
 };
 
 export default ({ collect }) => {
-  let subscriptionHandler;
+  const subscription = createSubscription();
   let surfaceIdentifier;
   const run = ({ surface, callback }) => {
-    subscriptionHandler = callback;
+    const unsubscribe = subscription.add(callback);
     surfaceIdentifier = surface;
```
jonsnyder, line 35: Same problem here. You are overwriting the surface instead of having one for each call.

jonsnyder (changes requested):

src/utils/createSubscription.js
```diff
@@ -0,0 +1,47 @@
...
+      return () => undefined;
+    }
+
+    counter += 1;
+    subscriptions[counter] = callback;
+    return createUnsubscribe(counter);
+  };
+
+  const emit = (...args) => {
+    Object.values(subscriptions).forEach(callback => {
```
shandilya3, line 33: Could use Alloy Utils methods.

sandbox/src/components/InAppMessagesDemo/InAppMessages.js
jasonwaters: good catch

src/components/DecisioningEngine/createSubscribeRulesetItems.js
jasonwaters: wow, glad you caught this. 🤦‍♂️

shandilya3 (approved):

PR #1063, opened 2023-10-23
Rename fetch flags

jonsnyder:
## Description

As recommended by Erik Wiener:

I see the use of "notification" to refer to proposition interactions or events in the docs and in the code, e.g. sendDisplayNotifications and includePendingDisplayNotifications.  I personally think it would be better to avoid the term "notification", which is a Target concept, and instead try to align around "proposition", "proposition interaction", and "event", which are solution-agnostic terms. Those two new parameters above stick out to me because I don't think we've tended to use the term "notification" in public docs related to personalization (though I could be wrong). I feel it would be more consistent to use names like sendDisplayEvent and includeRenderedPropositions.

I think of the whole event as the notification, and the proposition interactions inside _experience.decisioning.propositions as the experiences that are the subject of the event (i.e. what personalization experiences were interacted with as part of that event).

## Types of changes

## Checklist:

### DIFF
--- a/src/components/DataCollector/validateApplyResponse.js
+++ b/src/components/DataCollector/validateApplyResponse.js
@@ -31,8 +31,8 @@ export default ({ options }) => {
       ).required()
     }).required(),
     personalization: objectOf({
-      sendDisplayNotifications: boolean().default(true)
-    }).default({ sendDisplayNotifications: true })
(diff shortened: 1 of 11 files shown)


jfkhoury (approved):
I love this change! Do we need to update anything in the docs?

PR #1064, opened 2023-10-24
Consent & local storage config flag

shandilya3:
## Description

 

-    **personalizationStorageEnabled** is set to true by default. Local storage is only created when the flag is enabled.
-  If a customer sets **personalizationStorageEnabled** to false at some point, local storage  is cleared.
-    consent can have three states:  **in**, **out**, **pending** 
-    when personalizationStorageEnabled is true and consent is **in**, in that case local storage is setup once the consent is resolved.
- in the case of defaultConsent set to **out**,  await consent promise is not resolved & the promise falls into catch block, where we clear the local storage if it existed before.
- in the case of defaultConsent set to **pending**, sendEvents calls are queued up and processed only when opted-in

Future work :
- [ ] will add additional test with  functional tests in the same PR 

 
## Types of changes

## Checklist:


src/components/DecisioningEngine/index.js
```diff
@@ -48,6 +48,26 @@ const createDecisioningEngine = ({ config, createNamespacedStorage }) => {
       },
       onComponentsRegistered(tools) {
         applyResponse = createApplyResponse(tools.lifecycle);
+        if (personalizationStorageEnabled) {
```
carterworks, line 51: Why is the storaged created inside the `onComponentsRegistered()` callback?

src/components/DecisioningEngine/index.js
```diff
@@ -48,6 +48,26 @@ const createDecisioningEngine = ({ config, createNamespacedStorage }) => {
...
+            .then(() => {
+              const eventRegistry = createEventRegistry({
+                storage: storage.persistent
+              });
+              eventContext.setCurrentEventRegistry(eventRegistry);
+            })
+            .catch(() => {
+              // If consent is rejected, we want to clear the local storage.
+              if (storage) {
+                storage.persistent.clear();
```
jasonwaters, line 66: If a customer sets `personalizationStorageEnabled` to false at some point, local storage should also be cleared.  That way customer's won't disable it and wonder why they still see an entry in local storage.

src/components/DecisioningEngine/index.js
```diff
@@ -48,6 +48,26 @@ const createDecisioningEngine = ({ config, createNamespacedStorage }) => {
       },
       onComponentsRegistered(tools) {
         applyResponse = createApplyResponse(tools.lifecycle);
+        if (personalizationStorageEnabled) {
+          const storage = createNamespacedStorage(
+            `${sanitizeOrgIdForCookieName(orgId)}.decisioning.`
+          );
+          consent
```
jasonwaters, line 55: I think it's fine to have `.awaitConsent()` here in the `index.js` file.  But I wonder if it makes sense to move the listener within `eventRegistry` instead? 🤔 There are pros/cons.

jasonwaters (changes requested):

src/components/DecisioningEngine/createEventContext.js
```diff
@@ -0,0 +1,49 @@
...
+  const toJSON = () => {
+    return currentEventRegistry.toJSON();
+  };
+
+  const clear = () => {
+    currentEventRegistry.clear();
+  };
+
+  const setCurrentEventRegistry = eventRegistry => {
+    currentEventRegistry = eventRegistry;
```
carterworks, line 38: If a new event registry is set, should it copy over the events from the previous registry?

src/components/DecisioningEngine/createEventRegistry.js
```diff
@@ -132,5 +139,16 @@ export default ({ storage }) => {
     return events[eventType][eventId];
   };
 
-  return { addExperienceEdgeEvent, addEvent, getEvent, toJSON: () => events };
+  const setStorage = newStorage => {
```
jasonwaters, line 142: I think it's redundant to have both `updateStorage` and `setStorage` functions.

I recommend removing this version of `setStorage` and renaming the `updateStorage` method above to `setStorage`.  Implementation of `setStorage` can be modified to accept `newStorage`  for example...

```js
  let currentStorage;
  let restore;
  let save;
  let events;

  const setStorage = newStorage => {
    currentStorage = newStorage;

    restore = createRestoreStorage(currentStorage, STORAGE_KEY);
    save = createSaveStorage(
      currentStorage,
      STORAGE_KEY,
      createEventPruner(MAX_EVENT_RECORDS, RETENTION_PERIOD)
    );
    events = restore({});
  };

  setStorage(storage);
```

jasonwaters (commented):
looks great @shandilya3 !  Just one suggestion.

jasonwaters (approved):

PR #1070, opened 2023-10-27
querystring

shandilya3:
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Improvement (non-breaking change which does not add functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] I have signed the [Adobe Open Source CLA](https://opensource.adobe.com/cla.html) or I'm an Adobe employee.
- [ ] I have made any necessary test changes and all tests pass.
- [ ] I have run the Sandbox successfully.

PR #1071, opened 2023-10-27
IAM code fixes

jasonwaters:


src/components/Personalization/in-app-message-actions/actions/displayIframeContent.js
```diff
@@ -292,18 +286,18 @@ export default (settings, collect) => {
...
-        propositionEventTypes.add(action);
+        propositionEventTypes[action] = EVENT_TYPE_TRUE;
       }
 
       collect({
         decisionsMeta: [meta],
         propositionAction,
         eventType: INTERACT,
-        propositionEventTypes: toArray(propositionEventTypes)
+        propositionEventTypes: Object.keys(propositionEventTypes)
```
jasonwaters, line 300: we can't use `toArray` here.  it does not work with Sets.  so i switched to another convention.

src/components/Personalization/in-app-message-actions/utils.js
```diff
@@ -9,7 +9,7 @@ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTA
 OF ANY KIND, either express or implied. See the License for the specific language
 governing permissions and limitations under the License.
 */
-import { isNonEmptyArray, startsWith } from "../../../utils";
+import { isNonEmptyArray, queryString, startsWith } from "../../../utils";
```
carterworks, line 12: This will conflict with #1070 where both you and @shandilya3 made similar changes for the same thing

shandilya3 (approved):

src/components/Personalization/in-app-message-actions/utils.js
shandilya3: I will close https://github.com/adobe/alloy/pull/1070

PR #1077, opened 2023-10-30
back button bug fix

shandilya3:
## Description

`createRedirect` util method is using [replace()](https://developer.mozilla.org/en-US/docs/Web/API/Location/replace) that means  the current page will not be saved in session history, meaning the user won't be able to use the back button to navigate to the page from where it came.

Fix: 
added a second optional parameter named preserveHistory to the createRedirect method. That parameter value default to false. That way existing functionality is the same, but in the place where need to go back to the page where user came we can pass in true as the second parameter.

## Types of changes

## Checklist:


carterworks (approved):

src/components/Personalization/in-app-message-actions/actions/displayIframeContent.js
```diff
@@ -28,7 +28,7 @@ const dismissMessage = () =>
   [MESSAGING_CONTAINER_ID, OVERLAY_CONTAINER_ID].forEach(removeElementById);
 export const createIframeClickHandler = (
   interact,
-  navigateToUrl = createRedirect(window)
+  navigateToUrl = redirectWithHistory(window)
```
jasonwaters, line 31: I think it would be better not to add the `redirectWithHistory` function, and instead add a second optional parameter to the `createRedirect` method.  That parameter value could be named `preserveHistory` and default to false.  That way existing functionality is the same, but in this one place you can pass in `true` as the second parameter.

What do you think?

shandilya3: Now using `preserveHistory ` parameter set to false by default.

jasonwaters (approved):

PR #1078, opened 2023-10-30
CJM-56594 - IAM respects sendDisplayEvent and includeRenderedPropositions 

jasonwaters:
Make sure the In-App logic respects the Fetch use cases when it comes to preventing display events from going out

https://jira.corp.adobe.com/browse/CJM-56594

### DIFF
--- a/sandbox/src/components/InAppMessagesDemo/InAppMessages.js
+++ b/sandbox/src/components/InAppMessagesDemo/InAppMessages.js
@@ -54,6 +54,7 @@ if (alloyInstance !== window.alloy) {
     edgeDomain,
     thirdPartyCookiesEnabled: false,
     targetMigrationEnabled: false,
+    personalizationStorageEnabled: true,
     debugEnabled: true
   });
 }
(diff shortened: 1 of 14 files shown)


shandilya3 (approved):
lgtm

jonsnyder (approved):
Good work. I appreciate that you are also now returning the propositions that get rendered. 

To tackle after after the release, I see another problem. Consider this use case:

1. sendEvent called with sendDisplayEvent: true.
2. IAM and VEC experiences are returned.
3. Two display notifications are sent. One with VEC experiences, and one with IAM. Ideally, these would be combined.

The central problem here is that processPropositions is getting called twice, once via the onDecisionHandler and once via the fetchDataHandler. I have some ideas on how to fix this, but would require substantial changes to how these components are organized. We should discuss over Teams.

jfkhoury:
> Good work. I appreciate that you are also now returning the propositions that get rendered.
> 
> To tackle after after the release, I see another problem. Consider this use case:
> 
> 1. sendEvent called with sendDisplayEvent: true.
> 2. IAM and VEC experiences are returned.
> 3. Two display notifications are sent. One with VEC experiences, and one with IAM. Ideally, these would be combined.
> 
> The central problem here is that processPropositions is getting called twice, once via the onDecisionHandler and once via the fetchDataHandler. I have some ideas on how to fix this, but would require substantial changes to how these components are organized. We should discuss over Teams.

Let's discuss this and the `renderAttempted` flag next week, I will set something up.

PR #1079, opened 2023-10-31
renderAttempted

shandilya3:
## Description

propositions should include the "renderAttempted" flag, when returned via `subscribeRulesetItems`

## Types of changes

## Checklist:

### DIFF
--- a/src/components/DecisioningEngine/createSubscribeRulesetItems.js
+++ b/src/components/DecisioningEngine/createSubscribeRulesetItems.js
@@ -40,6 +40,7 @@ const emissionPreprocessor = (params, propositions) => {
         const { items = [] } = payload;
         return {
           ...payload,
+          renderAttempted: propositions.renderAttempted || true,
           items: items.filter(item =>
             schemasFilter ? includes(schemasFilter, item.schema) : true
           )
@@ -47,7 +48,6 @@ const emissionPreprocessor = (params, propositions) => {
       })
       .filter(payload => payload.items.length > 0)
   };
-
   return [result];
 };
 
--- a/test/unit/specs/components/DecisioningEngine/createSubscribeRulesetItems.spec.js
+++ b/test/unit/specs/components/DecisioningEngine/createSubscribeRulesetItems.spec.js
@@ -237,6 +237,7 @@ describe("DecisioningEngine:subscribeRulesetItems", () => {
       propositions: [
         {
           id: "1a3d874f-39ee-4310-bfa9-6559a10041a4",
+          renderAttempted: true,
           items: [
             {
               schema: "https://ns.adobe.com/personalization/message/feed-item",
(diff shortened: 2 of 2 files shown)


jasonwaters (changes requested):
The `renderAttempted` flag indicates that the personalization component attempted to render the proposition.  The DecisioningEngine should never set this flag, since it has no idea whether a proposition was attempted to be rendered or not.  Only the personalization component knows that.

In my opinion, this feature may need some refactoring of the sendEvent/request lifecycle and we probably need to schedule a meeting with @jonsnyder and others from the alloy team to discuss.

But let me try to describe the problem briefly.  When rulesets are evaluated, the resulting propositions are emitted using the `onDecision` lifecycle hook.  It's separate from the sendEvent/request lifecycle where `onResponse` receives all the propositions from a network request.  As a result, although you can trigger the rulesets to be evaluated using `alloy("sendEvent",{}).then(result =>{})` -- the result object in the `then` clause will not contain any of those propositions from the rulesets.  Instead, if a developer wants to get at those they will use the `subscribeRulesetItems` command.

I see two options for this ticket:

1) Make it so the personalization component follows a shared codepath when the `onDecision` lifecycle hook receives propositions -- so that `renderAttempted` is set on the proposition within the result for `alloy('subscribeRulesetItems',{}).then(result => {})` -- the same way it is for propositions that come directly from an edge response using sendEvent .

2) Refactor the sendEvent/request lifecycle so that the response object can be added to.  Today, when `sendEvent` is used, the propositions come from the edge response and from the edge response only.  And although any component can hook into the request/response lifecycle, they can't modify the response in any way.  I had an idea yesterday that may be worth exploring.  What if, as part of the `onResponse` lifecycle, a component can augment the response?  This would be useful in the case of DecisioningEngine because first, the lifecycle would make an edge network request and get propositions.  Then the DecisioningEngine could evaluate rulesets from that response and add the qualifying propositions to the response during the lifecycle.  Then subsequent components, like the personalization component, will have both the propositions from edge and the propositions from decisioning engine.  Then, the same code path could be leveraged AND the result object in a sendEvent call like this one `alloy("sendEvent",{}).then(result =>{})`, would also have the propositions that came from rulesets.  Along with the renderAttempted flag -- we'd get that for free since the same code path would be leveraged.

I think we at least need to do number 1 above, and possibly number 2 as well.  Something to discuss with the alloy team first.

jonsnyder (approved):

jonsnyder (changes requested):
Does't #1080 already do this?

jasonwaters:
> Does't #1078 already do this?

No, it doesn't.  That PR just ensures the display event is deferred properly.  We can discuss

jfkhoury:
@jonsnyder @shandilya3 What is the plan for this PR?

shandilya3:
Needs further discussion.

PR #1075, opened 2023-10-30
Move personalization flag to DecisioningEngine component

jonsnyder:
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Improvement (non-breaking change which does not add functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] I have signed the [Adobe Open Source CLA](https://opensource.adobe.com/cla.html) or I'm an Adobe employee.
- [ ] I have made any necessary test changes and all tests pass.
- [ ] I have run the Sandbox successfully.

PR #1080, opened 2023-10-31
[PDCL-11254] limit personalization event storage to only AJO campaigns

jasonwaters:
This PR changes the behavior of eventRegistry so that only experience edge events with a `decisionProvider` equal to `AJO` will be stored.

Related Jira: https://jira.corp.adobe.com/browse/PDCL-11254

### DIFF
--- a/src/components/DecisioningEngine/createEventRegistry.js
+++ b/src/components/DecisioningEngine/createEventRegistry.js
@@ -14,9 +14,11 @@ import {
   createSaveStorage,
   getExpirationDate,
   getActivityId,
-  hasExperienceData
+  hasExperienceData,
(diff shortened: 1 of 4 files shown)


shandilya3 (approved):

jonsnyder (approved):
So the storage option still defaults to true, but now only AJO events are stored there.

jfkhoury (changes requested):
Let's default to `false` as discussed, even though it's limited to AJO. Talked to Don and Mitch about this, and we will improve this logic/solution in the next couple of weeks.

jonsnyder:
default to false added to #1075

PR #1082, opened 2023-11-08
Fix propositions return value

jonsnyder:
## Description

While writing documentation wiki pages for the apply propositions, I noticed that the send event call was not returning anything in propositions. It was always an empty array. After investigation, I found that this was caused by the DecisioningEngine also returning propositions from its onBeforeEvent lifecycle method. This PR changes how the lifecycle return values are merged together. If the merged values are arrays, then the arrays are concat'ed together instead of overwritten.

Additionally, I found a few more tests failing because experience edge is now returning default-content-items when before it wasn't. I've updated the tests to be more lenient on their expectations regarding the number of propositions returned.

Lastly, there are two tests for applyResponse that I skipped because I noticed there are a bunch of new propositions returned from experience edge that are messing up the expectations. See https://jira.corp.adobe.com/browse/PDCL-11302

## Types of changes

## Checklist:

### DIFF
--- a/src/core/edgeNetwork/mergeLifecycleResponses.js
+++ b/src/core/edgeNetwork/mergeLifecycleResponses.js
@@ -10,7 +10,7 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import assignConcatArrayValues from "../../utils/assignConcatArrayValues";
+import { assignConcatArrayValues } from "../../utils";
 
 export default returnValues => {
   // Merges all returned objects from all `onResponse` callbacks into
--- a/src/core/edgeNetwork/mergeLifecycleResponses.js
+++ b/src/core/edgeNetwork/mergeLifecycleResponses.js
@@ -9,7 +9,8 @@ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTA
 OF ANY KIND, either express or implied. See the License for the specific language
 governing permissions and limitations under the License.
 */
-import { assign } from "../../utils";
+
+import { isObject } from "../../utils";
 
 export default returnValues => {
   // Merges all returned objects from all `onResponse` callbacks into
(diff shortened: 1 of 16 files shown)


test/unit/specs/core/edgeNetwork/mergeLifecycleResponses.spec.js
```diff
@@ -70,6 +70,9 @@ describe("mergeLifecycleResponses", () => {
                 ]
               }
             ]
+          },
+          {
+            propositions: []
```
carterworks, line 75: A unit test that coves the "merge two arrays" code path would be good.

carterworks (approved):

src/core/edgeNetwork/mergeLifecycleResponses.js
```diff
@@ -18,10 +19,25 @@ export default returnValues => {
...
   const lifecycleOnBeforeRequestReturnValues = returnValues;
 
-  return assign(
-    {},
+  return [
     ...lifecycleOnResponseReturnValues,
     ...consumerOnResponseReturnValues,
     ...lifecycleOnBeforeRequestReturnValues
-  );
+  ].reduce((accumulator, currentValue) => {
```
jasonwaters, line 26: This is the defect I was going to tell you about in our meeting the other day.  much improved 🏆

jasonwaters: i would really like this `accumulation` of results to be available as the lifecycle progresses.  I know we discussed it possibly being less performant, but might it be worth trying it out?  Changing from `promise.all` to a sequential evaluation where the responses are accumulated.  that still seems like an ideal solution for the decisioning engine propositions to be added.

jfkhoury, line 26: Would it help to extract this code into a separate util and unit test that alone?

jfkhoury (commented):
Love the fix, and all the test updates!

carterworks (approved):

src/utils/assignConcatArrayValues.js
```diff
@@ -0,0 +1,26 @@
+import assign from "./assign";
```
jfkhoury, line 12: Look at that beauty :)

test/unit/specs/utils/assignConcatArrayValues.spec.js
```diff
@@ -0,0 +1,65 @@
+import assignConcatArrayValues from "../../../../src/utils/assignConcatArrayValues";
```
jfkhoury, line 1: More beauty!!!

jfkhoury (approved):
Looks great @jonsnyder ! I think @jasonwaters 's comment is worth discussing, but it can be after the release.

PR #1101, opened 2024-03-04
FEATURE:  Automatic proposition interaction tracking

jasonwaters:
## Description
This PR is to support automatic proposition click tracking as described [here](https://wiki.corp.adobe.com/pages/viewpage.action?spaceKey=CJM&title=Automating+Proposition+Click+Events+in+Web+SDK).

In depth technical details about this feature can be read [here](https://wiki.corp.adobe.com/display/CJM/Automatic+Click+Tracking)

This change does not modify the way element clicks are tracked.  Any element designated for click tracking (whether from AJO or Target) is tracked by iterating all known click selectors and checking if the element clicked and it's ancestors match.  **This functionality remains untouched.**

This PR adds new functionality that tracks interactions (clicks) for proposition item(s) rendered to the DOM.
It does so using data-attributes.  When alloy renders proposition items, any time the DOM is modified (either by inserting elements or mutating an existing DOM element) alloy adds a data attribute, `data-aep-interact-id` to the element, which is then used for interaction tracking.  Interact ids have a one to many relationship with propositions and proposition-items.  Since there can be a plethora of campaigns targeting the same element.  For this reason, the interact id is used as a key for alloy to look up an in-memory mapping to see which propositions and items are relevant.

Additionally, a label and token may be added to an HTML proposition received from AJO using a data attribute.

### Data attributes

HTML data attributes are used to store these details.

| Name        | Data Attribute     | Source                                                       |
| ----------- | ------------------ | ------------------------------------------------------------ |
| Label       | `data-aep-click-label` | Manually added to the html offer by customer during campaign creation. |
| Token       | `data-aep-click-token` | Manually added to the html offer by customer during campaign creation. |
| Interact ID | `data-aep-interact-id`    | A unique ID automatically added to container elements by alloy when rendering DOM actions.  Each `interact-id` maps to at least one proposition, but it can map to more than one if multiple propositions reference the same element. |

ℹ️ These 👆 names are _tentative_ and subject to change.  Please offer your feedback for naming if you have an opinion. ℹ️

### Determining proposition, label & token when the user clicks within a proposition element

Beginning with the clicked element, each element's data attributes for `click-label`, `click-token` and `interact-id` are read.  If a value is found on the element, it is used for the interact event.  If not, the parent element is inspected for the data attributes.  If not found there, each parent element thereafter will be inspected until the root element has been reached. Note that there could exist nested labels and tokens within elements.  In that case, only the first label and token values found while walking up the DOM tree will be used.  

### New config option

A new _optional_ configuration option has been added called `autoCollectPropositionInteractions`.  The value is a map of decision providers, each with value that indicates how automatic proposition interactions should be handled.

By default, automatic proposition interactions are always collected for AJO, and never collected for Target.  The default value of `autoCollectPropositionInteractions` is:
```json
{
  "AJO": "always",
  "TGT": "never"
}
```

Possible configuration values for a decision provider are:

| Value                 | Description of behavior                                      |
| --------------------- | ------------------------------------------------------------ |
| `always`                | Alloy will always automatically emit `interact` events for any elements associated with a proposition. |
| `never`                | Alloy will never automatically emit `interact` events for DOM elements associated with a proposition. |
| `decoratedElementsOnly` | Alloy will automatically emit `interact` events for elements associated with a proposition but ONLY if the element has also have been "decorated" with data attributes specifying a label or token. |

### enhanced `applyPropositions` command.

The `applyPropositions` command now allows users to associate a `json-content-item` proposition item with a DOM element.  This comes in handy for code based JSON content propositions.

For example, imagine a customer has a `json-content-item` they used to generate a dom element or widget.  They want to take advantage of automatic click tracking for that element.  So they use the `applyPropositions` command to associate the proposition with the element selector.  Sample code...

```js
alloy("sendEvent", {
  renderDecisions: true,
}).then((result) => {
  const { propositions = [] } = result;
  const proposition = propositions.find(
    (proposition) => proposition.scope === "web://aepdemo.com/#woof"
  );

  if (proposition) {
    renderMyPropositionToDomInSuperCoolWay(proposition);

    alloy("applyPropositions", {
      propositions: [proposition],
      metadata: {
        "web://aepdemo.com/#woof": {
          selector: "div.woof-widget",
          actionType: "collectInteractions",
        },
      },
    });
  }
});

```

Alloy adds an `interact-id` to the element, and associates it with the provided proposition.  Now automatic click tracking will work the same way it does for other automatically rendered propositions.

## Types of changes

- [X] Improvement (non-breaking change which does not add functionality)

## Checklist:


src/components/Personalization/handlers/createProcessPropositions.js
```diff
@@ -10,6 +10,13 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
+const executeSequentially = promiseFunctions =>
+  promiseFunctions.reduce((sequence, promiseFunction) => {
+    return sequence.then(result =>
+      promiseFunction().then(nextResult => [...result, nextResult])
+    );
+  }, Promise.resolve([]));
+
```
jasonwaters, line 19: `executeSequentially` acts similar to `Promise.all`.  But instead of running all promises simultaneously and waiting for them all to finish, each promise is resolved in sequence one after the next.

This is necessary when rendering visual propositions that build on top of each other.  Consider a campaign that adds a new hero element to the page.  During authoring the customer revisits the hero element and changes the image within it.  Then later updates the text.  These three changes will be captured a 3 separate proposition items.  Each one needs the one proior to it to be completed first.

src/components/Personalization/handlers/injectCreateProposition.js
```diff
@@ -69,11 +78,23 @@ export default ({ preprocess, isPageWideSurface }) => {
...
+        const first = [];
+        const last = [];
+        items.forEach(rawItem => {
+          const item = createItem(rawItem, this);
+          Array.prototype.push.call(
+            item.getSchemaType() !== DOM_ACTION_CLICK ? first : last,
+            item
+          );
+        });
+        return [...first, ...last];
```
jasonwaters, line 90: As you can see, the `getItems` method was modified to ensure any `click` proposition items are at the end of the list.  This goes hand in hand with `executeSequentially`.  During authoring, a user can define a `click` to track.  Those click proposition items can appear anywhere within that array depending on when the user made the change.  The problem with that is that we need all the DOM elements to be rendered prior to the click handler.  That way every click handler will be sure to have the element it relies on already rendered to the DOM. 

This wasn't an issue for target, because click items always appeared within a second proposition for the activity.  Not the case in AJO.

src/utils/dom/awaitSelector.js
```diff
@@ -42,16 +42,21 @@ export const awaitUsingMutationObserver = (
...
   return createPromise((resolve, reject) => {
+    let timer;
+
     const mutationObserver = new win[MUTATION_OBSERVER](() => {
       const nodes = selectFunc(selector);
 
       if (isNonEmptyArray(nodes)) {
         mutationObserver.disconnect();
+        if (timer) {
+          clearTimeout(timer);
```
jasonwaters, line 53: doing a little garbage collection here

jasonwaters:
@jfkhoury 

1. We intend to add an optional configuration property called `autoTrackPropositionInteractions`.  Customers can use this value to configure which decision providers proposition interactions should be automatically tracked for.   The value is a string array of decision providers.  By default it will be set to `["AJO"]`.  So only propositions coming from AJO will have automatic click tracking.  But customers could configure the option to `["AJO", "TGT"]` in order to enable automatic click tracking in target.  This option will allow us to deliver this functionality for AJO only without changing behavior for target customers.  
2. I was about to say no changes are necessary for the tags extention, but in light of the prior bullet point I think we probably need to add the config option for `autoTrackPropositionInteractions` to the tags extension config UI.

src/components/Personalization/handlers/createProcessPropositions.js
```diff
@@ -193,7 +200,7 @@ export default ({ schemaProcessors, logger }) => {
       );
     });
     const render = () => {
-      return Promise.all(renderers.map(renderer => renderer())).then(metas =>
+      return executeSequentially(renderers).then(metas =>
```
jonsnyder, line 203: Why does this need to be sequential too? I understand that the items within a proposition need to be rendered sequentially, but how about different propositions? I didn't think their order was important.

src/components/Personalization/handlers/createProcessPropositions.js
```diff
@@ -10,6 +10,13 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
+const executeSequentially = promiseFunctions =>
+  promiseFunctions.reduce((sequence, promiseFunction) => {
+    return sequence.then(result =>
+      promiseFunction().then(nextResult => [...result, nextResult])
```
jonsnyder, line 16: If you change this line, then you don't have to make sure everything is a promise:
```
return sequence.then(result => promiseFunction()).then(nextResult => [...result, nextResult]);
```
.then() always returns a promise

src/utils/dom/removeNode.js
```diff
@@ -11,11 +11,13 @@ governing permissions and limitations under the License.
 */
 
 export default node => {
-  const parent = node.parentNode;
+  return new Promise(resolve => {
```
jonsnyder, line 14: If you change executeSequentially, then this doesn't have to be a promise.

jonsnyder (commented):
Looks good! Of course we'll need functional tests. I'm wondering what the use case is for changing the "always", "never" fields. It will be good to review all the configuration parameters.

src/components/Personalization/createTrackProposition.js
```diff
@@ -0,0 +1,124 @@
...
+      if (!Object.hasOwnProperty.call(items[i], REQUIRED_FIELDS_ITEM[j])) {
+        return new Error(
+          `Proposition item is missing "${REQUIRED_FIELDS_ITEM[j]}" field`
+        );
+      }
+    }
+  }
+
+  return undefined;
+};
```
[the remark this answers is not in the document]
jasonwaters: good point, will use those instead.

src/components/Personalization/handlers/createProcessPropositions.js
jasonwaters: This code is invalid, since `result` is out of scope in the second `then` clause.
```js
return sequence
  .then((result) => promiseFunction())
  .then((nextResult) => [...result, nextResult]);
```

what problem are you trying to solve?

src/components/Personalization/handlers/createProcessPropositions.js
jasonwaters: in target, activities can be prioritized which results in the edge response propositions being ordered according to that priority.  in this case, order matters.

test/functional/specs/LibraryInfo/C2589.js
```diff
@@ -61,7 +70,11 @@ test("C2589: getLibraryInfo command returns library information.", async () => {
     orgId: "5BFE274A5F6980A50A495C08@AdobeOrg",
     thirdPartyCookiesEnabled: true,
     targetMigrationEnabled: false,
-    personalizationStorageEnabled: false
+    personalizationStorageEnabled: false,
+    autoTrackPropositionInteractions: {
+      [ADOBE_JOURNEY_OPTIMIZER]: ALWAYS,
```
liljenback, line 75: Maybe we should try and follow the `propertyEnabled` syntax for these as well too. I can see in this case that there are three states and it might be a bit more dynamic in the way these are added, but something to think about.

src/components/Personalization/dom-actions/setStyles.js
```diff
@@ -12,10 +12,15 @@ governing permissions and limitations under the License.
...
-export default (container, styles) => {
-  const { priority, ...style } = styles;
+export default (container, styles, decorateProposition) => {
+  return new Promise((resolve) => {
+    const { priority, ...style } = styles;
 
-  Object.keys(style).forEach((key) => {
-    setStyle(container, key, style[key], priority);
+    Object.keys(style).forEach((key) => {
+      decorateProposition(container);
```
ninaceban, line 20: The container is the same, no need to re-decorate it.

ninaceban (commented):
Hi Jason, the PR looks good!

The feature for automatically collecting clicks on personalized content, as per Erik's Wiki, is intended solely for HTML rendered content (either VEC or html-content-item). This feature is only supported by AJO, with Target only providing support for explicit click-tracking.

Upon reviewing the wiki, I found no requirement for automatic click tracking on json-content-item offers, leading to the introduction of a new action type, `collectInteractions`. Unless there's a specific customer request or use-case, I would advise against adding a new capability to Alloy.

src/components/Personalization/dom-actions/setStyles.js
jasonwaters: good catch, thank you! 🙏

test/functional/specs/Personalization/C17409728.js
```diff
@@ -0,0 +1,371 @@
...
+  // TODO: Testcafe no longer captures the request body for sendBeacon requests.
+  // const notification = JSON.parse(edgeEndpointLogs.requests[0].request.body);
+  //
+  // await t.expect(notification.events[0].xdm.eventType).eql(INTERACT);
+  //
+  // await t
+  //   .expect(
+  //     notification.events[0].xdm._experience.decisioning.propositions[0].id,
+  //   )
+  //   .eql(propositionId);
```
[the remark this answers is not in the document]
jasonwaters: I suggest keeping them and un-commenting them when the `TODO: Testcafe no longer captures the request body for sendBeacon requests.` is resolved.

After merging `main` to my branch, I noticed testcafe was upgraded to the next major version.  This upgrade broke the test.  Version 3 of test cafe correctly captured request body for sendBeacon requests.  I'm not sure if this is a regression on their part or an intentional change.  But it definitely made it so we can't test as deeply as before.

test/functional/specs/Personalization/C17409728.js
```diff
@@ -0,0 +1,371 @@
...
+  await responseStatus(edgeEndpointLogs.requests, [200, 204]);
+  await t.expect(edgeEndpointLogs.count(() => true)).eql(1);
+
+  // TODO: Testcafe no longer captures the request body for sendBeacon requests.
+  // const notification = JSON.parse(edgeEndpointLogs.requests[0].request.body);
+  //
+  // await t.expect(notification.events[0].xdm.eventType).eql(INTERACT);
+  //
+  // await t
+  //   .expect(
```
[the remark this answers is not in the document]
jasonwaters: Please let me know if there is a way to force `fetch` instead of `sendBeacon` for `collect` calls.  I tried to find a way but didn't see a clear path.

src/utils/dom/removeNode.js
```diff
@@ -11,11 +11,13 @@ governing permissions and limitations under the License.
 */
 
 export default (node) => {
-  const parent = node.parentNode;
+  return new Promise((resolve) => {
```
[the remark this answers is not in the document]
jasonwaters: i think this was necessary to guarantee propositions are applied in sequential order.  but @jonsnyder asked me to revert that bug fix.  So i'll roll back this change as well.

jasonwaters:
> Hi Jason, the PR looks good!
> 
> The feature for automatically collecting clicks on personalized content, as per Erik's Wiki, is intended solely for HTML rendered content (either VEC or html-content-item). This feature is only supported by AJO, with Target only providing support for explicit click-tracking.
> 
> Upon reviewing the wiki, I found no requirement for automatic click tracking on json-content-item offers, leading to the introduction of a new action type, `collectInteractions`. Unless there's a specific customer request or use-case, I would advise against adding a new capability to Alloy.

Thanks @ninaceban !  You're right, Erik's wiki doesn't mention tracking `json-content-item` experiences.  It was a case he brought up with me during development.  If a customer uses `applyPropositions` with an `html-content-item`, an interact-id will be applied to that DOM element automatically for interaction tracking.  But the same was not true for `json-content-item`.  So the question was: How can customers benefit from automatic proposition interaction tracking in a simple way when they render something to the DOM based on a `json-content-item` code-based experience?  To support that case, I added a new command: `trackProposition`.  However, @jonsnyder made an excellent recommendation to remove that new command and instead enhance `applyPropositions` to suport json-content-item with a new `actionType` of `collectInteractions`.  I made that change, and now customers can also easily track DOM elements they manually render based on a `json-content-item`.  What are the down sides and risks for keeping this functionality?  

BTW, you're right about target services not fully benefiting from automatic interaction tracking yet.  I'm not sure the product roadmap priorities for target, but I believe it is on the backlog.

ninaceban:
> BTW, you're right about target services not fully benefiting from automatic interaction tracking yet. I'm not sure the product roadmap priorities for target, but I believe it is on the backlog.

Given the uncertainty surrounding Target's plans to adopt the automatic click tracking feature, it would be advisable to avoid adding it now. The same applies to json-content-item items.

It's always easier to add features later than to remove them.

src/components/Personalization/dom-actions/insertHtmlAfter.js
```diff
@@ -29,8 +29,12 @@ export default (container, html) => {
 
   loadImages(fragment);
 
+  let insertionPoint = container;
+
   elements.forEach((element) => {
-    insertAfter(container, element);
+    decorateProposition(element);
```
ninaceban, line 35: here you'd probably want to decorate the HTML fragment rather than each element

src/components/Personalization/dom-actions/rearrangeChildren.js
```diff
@@ -12,20 +12,28 @@ governing permissions and limitations under the License.
...
+    if (from < to) {
+      insertAfter(elementTo, elementFrom);
+    } else {
+      insertBefore(elementTo, elementFrom);
+    }
+
+    decorateProposition(elementTo);
+    decorateProposition(elementFrom);
+
+    resolve();
```
ninaceban, line 37: here as well

src/components/Personalization/dom-actions/setText.js
```diff
@@ -10,6 +10,8 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-export default (container, text) => {
+export default (container, text, decorateProposition) => {
+  decorateProposition(container);
   container.textContent = text;
+  return Promise.resolve();
```
ninaceban, line 16: here too

src/components/Personalization/dom-actions/swapImage.js
```diff
@@ -11,20 +11,23 @@ governing permissions and limitations under the License.
...
   loadImage(url);
 
+  decorateProposition(container);
+
   // Remove "src" so there is no flicker
   removeAttribute(container, SRC);
 
   // Replace the image "src"
   setAttribute(container, SRC, url);
+  return Promise.resolve();
```
ninaceban, line 32: here

jasonwaters:
> > BTW, you're right about target services not fully benefiting from automatic interaction tracking yet. I'm not sure the product roadmap priorities for target, but I believe it is on the backlog.
> 
> Given the uncertainty surrounding Target's plans to adopt the automatic click tracking feature, it would be advisable to avoid adding it now. The same applies to json-content-item items.
> 
> It's always easier to add features later than to remove them.

@ninaceban I'm not sure if there is an action you want me to take here.  can you please be specific?  It would be advisable to avoid adding what? If you are asking to not add automatic proposition interaction tracking for target then no worries -- that's already done.  It is disabled for target by default.

As for the `json-content-item` tracking via `applyPropositions`, could you please raise your concern with Erik?  He explicitly asked for this gap to be covered.

PR #1114, opened 2024-05-16
Update dependencies.

dompuiu:
## Description

The dependencies on this project were not updated in a long time. 

## Motivation and Context

The dependencies were very old. A lot of dependencies moved to ES modules. TestCafe changed a lot. This project was using a 2 major versions old Prettier. ESLint released a new config format file. So a lot of reasons.

Big changes moving forward:
- Cannot longer use require;
- Imports must provide paths with extension.

~~One know issue is that I cannot make the script `test:functional:watch` work. Needs more investigation on how to make that work.~~ `test:functional:watch` seems to have issue when the `--esm` flag is used. Since we no longer use it, now the watch command works.

Update: It seems that Chrome and Edge are controlled using native automation protocol, while Firefox or Safari are controlled using a reverse proxy (the method that was used to control all the browsers in TestCafe 2.6). Because of that, I was forced to undo or tweak some of the fixes I made initially for Chrome. Examples of differences:

- The response bodies are automatically unzipped when test are run in Chrome. But that is not true for Safari or Firefox. So in the helper method that handles the responses, I try to unzip the body always. In case the unzip fails, I consider that the body is already a JSON. Yuck, but it works.
- There is a test where we check that the same ECID is returned when a first party cookie is set. For the test to pass in Chroe, the cookie needs to be set on domain `.alloyio.com`. In Safari and Firefox it needs to be set on `alloyio.com`, otherwise the TestCafe reverse proxy ignores it. So the solution I choose is to set one cookie for each domain.

I was forced to add a few tests to the ignore list: C8631576 and C8631577. I talked with Don and he said he will take a look. If he find a way to fix those tests, I will remove them from the ignore list.

## Types of changes

## Checklist:


jonsnyder (approved):

jfkhoury:
Thank you so much, @dompuiu, for taking on this effort!

PR #1130, opened 2024-06-11
Move the Terminal UI (TUI) script in alloy.

dompuiu:
## Description

Moved the terminal ui build script in Alloy.

## Types of changes

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
     "libEs6"
   ],
   "bin": {
+    "alloy": "scripts/tui-builder.js",
     "build-alloy": "scripts/build-alloy.js"
   },
   "scripts": {
@@ -72,6 +73,7 @@
     "@adobe/reactor-object-assign": "^1.0.0",
     "@adobe/reactor-query-string": "^1.0.0",
     "css.escape": "^1.5.1",
+    "inquirer": "^9.2.23",
     "js-cookie": "3.0.5",
     "parse-uri": "^1.0.9",
     "uuid": "^9.0.1"
--- a/scripts/build-alloy.js
+++ b/scripts/build-alloy.js
@@ -64,7 +64,7 @@ const argv = yargs(hideBin(process.argv))
   })
   .option("minify", {
     type: "boolean",
-    default: false,
+    default: true,
     alias: "m",
   })
   .option("outputDir", {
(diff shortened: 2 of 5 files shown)


scripts/build-alloy.js
```diff
@@ -64,7 +64,7 @@ const argv = yargs(hideBin(process.argv))
   })
   .option("minify", {
     type: "boolean",
-    default: false,
+    default: true,
```
dompuiu, line 67: Changed the CLI minify flag to true by default.

scripts/helpers/alloyComponents.js
```diff
@@ -0,0 +1,42 @@
...
+    checked: true,
+  },
+  {
+    name: "Decision Engine",
+    value: "decisioningengine",
+    checked: true,
+  },
+  {
+    name: "Machine Learning",
+    value: "machinelearning",
```
ninaceban, line 40: We'll need to add Streaming Media and Legacy Media Analytics components here too.

carterworks (approved):

shammowla (approved):

package.json
```diff
@@ -72,6 +73,7 @@
     "@adobe/reactor-object-assign": "^1.0.0",
     "@adobe/reactor-query-string": "^1.0.0",
     "css.escape": "^1.5.1",
+    "inquirer": "^9.2.23",
```
jonsnyder, line 76: Can this be a devDependency?

package.json
```diff
@@ -10,6 +10,7 @@
     "libEs6"
   ],
   "bin": {
+    "alloy": "scripts/tui-builder.js",
     "build-alloy": "scripts/build-alloy.js"
```
jonsnyder, line 14: If we just provide one executable, then it is much more simple to run because npx can determine which to run. What if both used the same script, but you could pass an --interactive flag?

scripts/build-alloy.js
jonsnyder: I was thinking that we'd rename the flag to "dev" which would default to false. So to get the un-minified version you would do "npx @adobe/alloy --dev"

package.json
dompuiu: From what I saw on internet, npx doesn't install dev dependencies. I don't know how I can test that in order to be sure.

PR #1139, opened 2024-06-26
Remove support for Internet Explorer

carterworks:
## Description

- Remove IE 11 and 10 from the Babel transpiration targets
- Remove and re-write the utilities that worked around IE limitations
- Remove unused dependencies

Size reduction:

`npm run clean && npm run build && gzip --keep dist/alloy.min.js && du -ah dist`

File name | Size with IE | Size without IE | % change
--- | --- | --- | ---
alloy.js | 620 KB | 540 KB | 15.6% smaller 
alloy.min.js | 144 KB | 112 KB | 15.2% smaller
alloy.min.js.gz | 44 KB | 40 KB | 9% smaller

## Related Issue

[PDCL-11084](https://jira.corp.adobe.com/browse/PDCL-11084)

## Types of changes

## Checklist:


src/utils/assignIf.js
```diff
@@ -10,11 +10,9 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
 
-import assign from "./assign.js";
-
 export default (target, source, predicate) => {
   if (predicate()) {
-    assign(target, source);
+    target = { ...target, ...source };
   }
 
   return target;
```
jonsnyder, line 18: I don't see any usage of assignIf. Could be a good time to remove it.

src/utils/padStart.js
```diff
@@ -12,14 +12,5 @@ governing permissions and limitations under the License.
...
-  let repeatedPadString = String(padString);
-  if (originalString.length >= targetLength || repeatedPadString.length === 0) {
-    return originalString;
-  }
-  const lengthToAdd = targetLength - originalString.length;
-  while (lengthToAdd > repeatedPadString.length) {
-    repeatedPadString += repeatedPadString;
-  }
-  return repeatedPadString.slice(0, lengthToAdd) + originalString;
+  return `${string}`.padStart(targetLength, padString);
```
jonsnyder, line 15: you can just move this to toISOStringLocal

test/unit/specs/core/network/requestMethods/injectSendXhrRequest.spec.js
```diff
@@ -27,9 +27,11 @@ describe("sendXhrRequest", () => {
...
     ]);
-    XMLHttpRequest = () => {
+    // use non-arrow function because we need to use `this`
+    XMLHttpRequest = function constructXMLHttpRequest() {
       return request;
     };
+
     sendXhrRequest = injectSendXhrRequest({ XMLHttpRequest });
     body = { a: "b" };
   });
```
jonsnyder, line 37: We can remove sendXhrRequest entirely

jonsnyder (changes requested):
Awesome stuff. I love the stats. We should definitely include those in the release notes.

@ninaceban is there more we can do with the personalization selectors or Dom utils now that we not supporting IE?

carterworks:
Removing XHR fallback broke the "C1472436: Set-consent is called when consent cookie is missing even though consent is the same" test in Safari only.

carterworks:
> I love the stats. We should definitely include those in the release notes.

I have created a proof-of-concept rollup plugin that will print out the sizes of each build output to the console. I also found [an old plugin that writes a file and keeps track of the size change on each build](https://github.com/luwes/rollup-plugin-size/). We can talk about the best path forward later.

```js
import { gzip, brotliCompress as br } from "node:zlib";

/**
 * @returns {Partial<import("rollup").PluginHooks>}
 */
const bundleSizePlugin = () => {
  const gzipCompress = promisify(gzip);
  const brotliCompress = promisify(br);
  /**
   * @param {import("node:zlib").InputType} source the source code to compress
   * @returns {number} size in bytes
   */
  const getGzippedSize = async (source, options = {}) => {
    const compressed = await gzipCompress(source, options);
    const byteSize = Number.parseInt(compressed.byteLength, 10);
    return byteSize;
  };
  const getBrotiliSize = async (source, options = {}) => {
    const compressed = await brotliCompress(source, options);
    const byteSize = Number.parseInt(compressed.byteLength, 10);
    return byteSize;
  };
  return {
    name: "bundle-size",
    async generateBundle(_rollupOptions, bundle) {
      // keep sizes in bytes until displaying them
      const sizes = await Promise.all(
        Object.values(bundle)
          .filter((outputFile) => outputFile.type === "chunk")
          .map(async (chunk) => ({
            fileName: chunk.fileName,
            uncompressedSize: Buffer.from(chunk.code).byteLength,
            gzippedSize: await getGzippedSize(chunk.code),
            brotiliSize: await getBrotiliSize(chunk.code),
          })),
      );
      console.table({
        ...sizes.map((size) => ({
          ...size,
          gzippedSize: size.gzippedSize / 1024,
          brotiliSize: size.brotiliSize / 1024,
          uncompressedSize: size.uncompressedSize / 1024,
        })),
      });
    },
  };
};
```

jonsnyder:
I don't understand how removing XHR would cause that test failure. Do you think it was unrelated?

It would be great to include that data within the release notes on Github. I would propose writing a file with the info as part of the build process, exporting that as a build artifact, and then when building the release notes, we could look at that artifact.

jonsnyder (approved):

carterworks:
Different test failed in Safari:

> C8085776: At.js 2.x to Web SDK - Assert same session ID, edge cluster are used for both of the requests interact and delivery API

The Firefox one timed out after 30 minutes.

dompuiu:
@carterworks It seems your branch is still using v1 of `@adobe/reactor-query-string`. Shouldn't you upgrade to v2? Or do you want to merge this first and then we will merge my PR [here](https://github.com/adobe/alloy/pull/1138).

v2 of `@adobe/reactor-query-string` would solve this issue: [CORE-109630](https://jira.corp.adobe.com/browse/CORE-109630).

carterworks:
> @carterworks It seems your branch is still using v1 of `@adobe/reactor-query-string`. Shouldn't you upgrade to v2? Or do you want to merge this first and then we will merge my PR [here](https://github.com/adobe/alloy/pull/1138).
> 
> v2 of `@adobe/reactor-query-string` would solve this issue: [CORE-109630](https://jira.corp.adobe.com/browse/CORE-109630).

Your PR has 3 approvals. Go ahead and merge yours and then I will update this PR

dompuiu (approved):

carterworks:
> Go ahead and merge yours and then I will update this PR

Done.

PR #1147, opened 2024-07-22
Report bundle size when building alloy

carterworks:
## Description

[@jonsnyder once said](https://github.com/adobe/alloy/pull/1139#issuecomment-2218776557)

> It would be great to include that data within the release notes on Github. I would propose writing a file with the info as part of the build process, exporting that as a build artifact, and then when building the release notes, we could look at that artifact.

This PR includes a small Rollup plugin that generates the base, gzip, and brotli compressed size of each build artifact. The idea is that each PR will include the record of artifact sizes and we can track the history of build sizes with git.

Size is reported in bytes in `./bundlesize.json`. 

Files are compressed with default Node compression level (which tries to find a balance between speed and file size), but it is negotiable. The compression level of a webserver is configurable—for gzip, Apache defaults to Zlib's 6 of 9(same as Node), but nginx defaults to the lowest level of compresison (1). For Brotli, Apache defaults to level 5 of 11, Nginx defaults to 6, and Node doesn't specify which level (but I bet it is somewhere in the middle).

## Types of changes

## Checklist:


rollup.config.js
```diff
@@ -10,14 +10,116 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import path from "path";
+import path from "node:path";
```
dompuiu, line 13: Why do you need to put `node:path`?

rollup.config.js
```diff
@@ -48,6 +150,9 @@ const buildPlugins = ({ variant, minify, babelPlugins }) => {
       configFile: path.resolve(dirname, "babel.config.cjs"),
       plugins: babelPlugins,
     }),
+    bundleSizePlugin({
+      output: "bundlesize.json",
```
dompuiu, line 154: If you run the custom build command from a folder outside the alloy directory, you will get the `bundlesize.json` file saved along with the alloy library file.

I think we should either not run this plugin when the custom build command is executed or change the path and use the dirname variable that is available in the file.

rollup.config.js
carterworks: It's personal preference. I like to know the difference between libraries that are listed in `package.json` and libraries that are built in.

dompuiu: I think we should choose a way and stick to it everywhere. Other scripts have it without `node:`.

dompuiu (approved):

jonsnyder (approved):
We'll need to change the deploy script so that it commits this file. Also, it is annoying if this file is generated when building locally for tests or for development. You'd have to ignore it every time when committing

carterworks:
> We'll need to change the deploy script so that it commits this file.

@jonsnyder What is better: updating this file when we deploy a new version (`deployRelease.yml`), or updating it every time we merge to main (`dev.yml`)?

jonsnyder:
Yes (scripts/deploy.js) is where we should add this file to the commit.

carterworks:
Before updating the version number, the GitHub action will run `npm run build` and then commit the resulting `bundlesize.json`

PR #1152, opened 2024-08-02
Changes in custom build script so that people that use the NPM package version can use rollup treeshake feature.

dompuiu:
## Description

Changed the way custom builder works.
- Got rid of the `skip_when` comments.
- I split the modules in 2 files:`src/core/componentCreators.js` for optional modules and `src/core/requiredComponentCreators.js` for required modules (data collector, library info and identity.
- I expose the optional modules to users that use the NPM package version of alloy.
- I added a new method `createCustomInstance` that will accept an object of optional modules to be bundled when rollup is executed. `createInstance` works as before for backward compatibility reasons (all optional modules will be used). 

## Types of changes

## Checklist:


src/standalone.js
```diff
@@ -13,5 +13,8 @@ governing permissions and limitations under the License.
 // This file is used by rollup to create the browser version that is uploaded to cdn
 
 import core from "./core/index.js";
+import * as components from "./core/componentCreators.js";
 
-core();
+// If you change this line, check if the custom build script is still working.
+// You might need to change the babel plugin in scripts/helpers/entryPointGeneratorBabelPlugin.js.
+core({ components });
```
dompuiu, line 20: `generateInputEntryFile` method from `alloyBuild.js` script, will read this file and transform it using Babel to use just the selected optional components that the user chooses. For example, if a user wants custom build with Privacy and Personalization components, this file will be transformed to a custom entry file with the following content:

```
import core from "./core/index.js";
import * as components from "./core/componentCreators.js";

core({
  components: {
    privacy: components.privacy,
    personalization: components.personalization,
  },
});
```

This will become the entry point file used by rollup. Tree shake will work.

src/index.js
```diff
@@ -41,7 +45,30 @@ export const createInstance = (options = {}) => {
...
+  const { name, monitors } = eventOptionsValidator(options);
+
+  return createCustomInstance({
+    name,
+    monitors,
+    components: optionalComponents,
+  });
+};
+
+export { optionalComponents as components };
```
dompuiu, line 74: If somebody wants to use alloy in their NPM project they can add something like the following code in their main.js file (tree shake will work in this case):

```
import { components, createCustomInstance } from "@adobe/alloy";

const alloy = createCustomInstance({
  name: "alloy",
  components: {
    privacy: components.privacy,
    personalization: components.personalization,
  },
});

alloy("configure", {
  datastreamId: "bc1a10...",
  orgId: "5B....@AdobeOrg",
  debugEnabled: true,
});
```
Tree shake doesn't function in all the case. For example, tree shake will not function in the following code:

```
import { components, createCustomInstance } from "@adobe/alloy";

const { privacy, personalization } = components; // This is the line that would break the tree shake.

const alloy = createCustomInstance({
  name: "alloy",
  components: {
    privacy,
    personalization
  },
});

alloy("configure", {
  datastreamId: "bc1a10...",
  orgId: "5B....@AdobeOrg",
  debugEnabled: true,
});
```

test/functional/specs/LibraryInfo/C2589.js
```diff
@@ -50,10 +50,15 @@ test("C2589: getLibraryInfo command returns library information.", async () => {
     "applyResponse",
     "configure",
     "createEventMergeId",
+    "createMediaSession",
```
dompuiu, line 53: I am not sure how this test was not failing before.

carterworks (approved):
Nicely done.

src/index.js
jonsnyder: It seems a more natural API is passing in an array. I see why you accept an object though because you can just "components: optionalComponents", but you could just do "components: Object.values(optionalComponents)".

test/unit/specs/core/componentCreators.spec.js
```diff
@@ -10,13 +10,18 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import componentCreators from "../../../../src/core/componentCreators.js";
+import * as componentCreators from "../../../../src/core/componentCreators.js";
 
 describe("componentCreators", () => {
   it("is an array of component creators", () => {
```
jonsnyder, line 16: It's an object right?

scripts/alloyBuilder.js
```diff
@@ -208,7 +234,7 @@ program
   .description(
     "Tool for generating custom Adobe Experience Platform Web SDK builds.",
   )
-  .version("1.0.0");
+  .version("1.1.0");
```
jonsnyder, line 237: How is this version used? Is it included in the build? Since this is included as part of the same repo and versioned with Alloy, do we need a version?

dompuiu: It is just the script version. You can see it when your run the script with the `-v` flag. I can remove it, but it seems more like a standard script that you see everywhere these days.

dompuiu: Alternatively, we could add the alloy version instead.

PR #1160, opened 2024-08-13
PLATIR-42315 Ensure audience cookies are set on the default path.

dompuiu:
## Description

1. Let's say you start with a js-cookie instance:

```
import export default {
  get: cookie.get,
  set: cookie.set,
  remove: cookie.remove,
  withConverter: cookie.withConverter
}; from 'js-cookie'
```

You will get back an object that has the regular methods (`get`, `set`, `remove`). The object contains also an `attributes` property that by default has the value `{path: '/'}`. If you then call 

```
Cookies.set('name', 'value');
```
the cookie's path will be `/`.

2.  If then we would do: 

```
Cookies = Cookies.withConverter({ write: function () { .... } });
Cookies.set('name', 'value');
```
The cookie's path will still be `/` in this case. js-cookie is supporting this scenario [here](https://github.com/js-cookie/js-cookie/blob/8eb77f5f5f17cf3d08ac8b6f74438d9425fdad6c/src/api.mjs#L89-L94). Pay attention that the code uses `this.attributes`.

3. We are using js-cookies in a different way [here](https://github.com/adobe/alloy/blob/08fd9856c92032edeacd9b6c80e0ad6512d78291/src/utils/cookieJar.js):

```
export default {
  get: cookie.get,
  set: cookie.set,
  remove: cookie.remove,
  withConverter: cookie.withConverter
};
```

We then call in the Audience module  `cookieJar.withConverter`. At this point, we will get back as instance without the default path being `/`. That happens because the `this.attributes` in js-cookie code will refer to the object that we export in the `cookieJar.js` file. And that object doesn't contain any attributes property.

4. Initially I was thinking to just export the js-cookie module inside `cookieJar.js`. The cookieJar code would have become something like this:

```
import cookie from "js-cookie";
export default cookie;
```

There is a problem though. The `Cookies` variable that is imported is an object that contains [2 properties](https://github.com/js-cookie/js-cookie/blob/main/src/api.mjs#L96-L99): `attributes` and `converter`. 

The methods `get`, `set` and `remove` are set on the prototype chain of that object. 

And when we do [here](https://github.com/adobe/alloy/blob/08fd9856c92032edeacd9b6c80e0ad6512d78291/src/utils/createLoggingCookieJar.js#L14-L21)

```
 return {
    ...cookieJar,
    set(key, value, options) {
        ...
    },
  };
```
the spread operation is working only on the properties. Not on the whole prototype chain. The result was that you would have got back an object that didn't contain `get` or `remove` properties anymore. This is why I got to the solution that Jon suggested to bind the js-cookie initial context.

 

## Related Issue
https://jira.corp.adobe.com/browse/PLATIR-42315

## Types of changes

## Checklist:

### DIFF
--- a/test/functional/specs/Audiences/C12412.js
+++ b/test/functional/specs/Audiences/C12412.js
@@ -66,6 +66,8 @@ test(`Verify cookie is set on the / path `, async () => {
 
   const cookies = await t.getCookies("C12412");
 
+  // In Firefox, the t.getCookies method returns an empty array even if the cookie is present.
+  // The if condition below is to handle this issue.
   if (cookies.length > 0) {
     await t.expect(cookies[0].path).eql("/");
   }
--- a/test/functional/specs/Audiences/C12412.js
+++ b/test/functional/specs/Audiences/C12412.js
@@ -65,5 +65,8 @@ test(`Verify cookie is set on the / path `, async () => {
   await alloy.sendEvent();
 
   const cookies = await t.getCookies("C12412");
-  await t.expect(cookies[0].path).eql("/");
+
+  if (cookies.length > 0) {
+    await t.expect(cookies[0].path).eql("/");
+  }
 });
--- a/test/functional/specs/Audiences/C12412.js
+++ b/test/functional/specs/Audiences/C12412.js
@@ -57,3 +57,13 @@ test(`Verify cookie destinations are returned in the response when turned on in 
   await t.expect(setCookieAttributes[0].sameSite).eql("none");
   await t.expect(setCookieAttributes[0].secure).eql(true);
 });
+
+test(`Verify cookie is set on the / path `, async () => {
+  const alloy = createAlloyProxy();
+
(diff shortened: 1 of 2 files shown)


carterworks (approved):

jonsnyder (approved):
Thanks for looking into this!

test/functional/specs/Audiences/C12412.js
```diff
@@ -65,5 +65,8 @@ test(`Verify cookie is set on the / path `, async () => {
   await alloy.sendEvent();
 
   const cookies = await t.getCookies("C12412");
-  await t.expect(cookies[0].path).eql("/");
+
+  if (cookies.length > 0) {
```
jonsnyder, line 69: Can you add a comment here to explain that Firefox cannot read the cookies path? Also, does Firefox just return an empty object?

dompuiu: Done. In Firefox an empty array is returned even if the cookie exists.

PR #1161, opened 2024-08-14
Remove references to Map, refactor getBrowser to lazy call

jonsnyder:
## Description

One customer has had an issue with deploying because they are redefining "Map" without a "has" function. Since we are only using Map in one insignificant place. I've removed references to it. Instead of using memoize, I've changed to use lazy since there was only ever once call that needed to be cached.

## Related Issue

https://jira.corp.adobe.com/browse/PDCL-12356

## Types of changes

## Checklist:


carterworks (approved):

ninaceban (commented):
As far as I can see the issue is only in using `Map`, because the customer has a pollyfill which is incomplete, it doesn't contain `has` method.  
I was thinking if changing `memoize` to use an ordinary JavaScript object `{}` is a better approach, since all the changes will affect just the `memoize` function instead of adjusting other files.

jonsnyder:
I considered that as well. In fact I went down that route at first. What I found was that Map is different from {} in that it accepts non-strings as keys. In fact anything can be a key even objects. This makes it good for memoization because you can use anything as the key. After looking deeper though I realized that we only use memoize in one place and we are using the userAgent as the key which shouldn't change on a page load anyways.

dompuiu (approved):

PR #1203, opened 2024-10-24
Web SDK Error in applyPropositions due to Default Content or Tracking Event

ninaceban:
## Description

The issue appeared after this [PR](https://github.com/adobe/alloy/pull/1101/files#diff-f045242682e2082ad8db0b9e09d37458bbe579ae746e9a388f65ec75471c87c2R23) has been released.
`default-content-item` schemas items do not have `item.data`, thus this was leading into validation error and propositions were never applied. Customers do not know when `default-content-item` is returned, which was breaking their implementation and their could not see the impressions. This is a critical issue because it impacts all the customers and can affect any other activities not only those that have `default-content-item`. 
## Types of changes

## Checklist:

### DIFF
--- a/src/components/Personalization/validateApplyPropositionsOptions.js
+++ b/src/components/Personalization/validateApplyPropositionsOptions.js
@@ -31,7 +31,7 @@ export default ({ logger, options }) => {
           objectOf({
             id: string().required(),
             schema: string().required(),
-            data: anything().required(),
+            data: objectOf(anything()),
           }),
         )
           .nonEmpty()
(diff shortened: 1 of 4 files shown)


carterworks (approved):

src/components/Personalization/validateApplyPropositionsOptions.js
```diff
@@ -31,7 +31,7 @@ export default ({ logger, options }) => {
           objectOf({
             id: string().required(),
             schema: string().required(),
-            data: anything().required(),
+            data: objectOf(anything()),
```
jonsnyder, line 34: I think you need to do objectOf({}). But why change data to have to be an object?. It seems you could just remove this line entirely and allow data to be set or not.

jonsnyder (approved):
Awesome work finding this!

PR #1199, opened 2024-10-17
PLATIR-44580 -  Do not send empty meta.configOverrides as it invalidates server side config overrides

ninaceban:
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Improvement (non-breaking change which does not add functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] I have signed the [Adobe Open Source CLA](https://opensource.adobe.com/cla.html) or I'm an Adobe employee.
- [ ] I have made any necessary test changes and all tests pass.
- [ ] I have run the Sandbox successfully.

PR #1205, opened 2024-10-29
Fix tests for edgeConfigOverrides and Streaming Media

ninaceban:
## Types of changes

## Checklist:

### DIFF
--- a/test/unit/specs/components/StreamingMedia/createMediaEventManager.spec.js
+++ b/test/unit/specs/components/StreamingMedia/createMediaEventManager.spec.js
@@ -130,6 +130,7 @@ describe("StreamingMedia::createMediaEventManager", () => {
     });
     expect(eventManager.sendEvent).toHaveBeenCalledWith(event, {
       mediaOptions,
+      edgeConfigOverrides: undefined,
     });
   });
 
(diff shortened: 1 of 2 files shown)


test/unit/specs/components/StreamingMedia/createMediaEventManager.spec.js
```diff
@@ -130,6 +130,7 @@ describe("StreamingMedia::createMediaEventManager", () => {
     });
     expect(eventManager.sendEvent).toHaveBeenCalledWith(event, {
       mediaOptions,
+      edgeConfigOverrides: undefined,
```
carterworks, line 133: Instead of changing the test to have `edgeConfigOverrides: undefined`, we should continue the work of #1199 and modify [`src/core/createEventManager.js`](https://github.com/adobe/alloy/blob/main/src/components/StreamingMedia/createMediaEventManager.js#L80) to not include empty `edgeConfigOverrides` in the call to `sendEvent`

carterworks (changes requested):

test/unit/specs/components/StreamingMedia/createMediaEventManager.spec.js
ninaceban: It is using src/core/createEventManager.js down the line, which checks the `edgeConfigOverrides` and construct the request payload based on that.

PR #1219, opened 2024-11-18
Fixing flickering in Activities of type redirect

ninaceban:
## Types of changes

## Checklist:

### DIFF
--- a/src/components/Personalization/handlers/createProcessRedirect.js
+++ b/src/components/Personalization/handlers/createProcessRedirect.js
@@ -9,8 +9,9 @@ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTA
 OF ANY KIND, either express or implied. See the License for the specific language
 governing permissions and limitations under the License.
 */
-import { hideElements } from "../flicker/index.js";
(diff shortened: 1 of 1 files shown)


carterworks (approved):

jonsnyder (approved):
Can you explain why the hide body was there in the first place and why it is no longer needed?

ninaceban:
> > Can you explain why the hide body was there in the first place and why it is no longer needed?
> 
> @jonsnyder It's the opposite—`hideElements("BODY")` was not there in the first place.

Earlier in Personalization component when a redirect type of item was returned from the Edge, it was turning Web SDK into a different mode than what it was running for rest of propositions type. 

After the refactoring was done, the processing and rendering is happening in an async mode while showContainers is called in a sync mode.
What happens now is that processing/rendering of the propositions happens async and till redirect executes, `showContainers` which is called sync already kicks in and unhides the body. By hiding `Body` in case redirect item is executed we can avoid flickering. This is the only quick fix that would fix flickering and not touch current refactoring of the Personalization component.

carterworks (approved):

jonsnyder (approved):
I've reviewed this again and see that you are only hiding the body when executing a redirect. I was thinking it was more than this.

PR #1239, opened 2025-02-25
Fixed an issue where identities were forgotten after a /collect call was sent

carterworks:
## Description

No longer preserves empty `namespace` objects after a response.

<details>
<summary>For the curious, this PR was 95% written by <a href="https://cursor.com">Cursor</a>, including the test.</summary>  

I prompted Cursor, using the "Composer" in "Agent Mode" (which can edit all files and even run cli commands) and the brand-new Sonnet 3.7 Thinking with the following:

> Here's a bug report.
> 
> > It looks like there is an issue with AEP Web SDK with the "getIdentity" command with version 2.24.0.  The highlighted line in the screenshot causes "namespaces" to be overwritten by any alloy('sendEvent') call that does not return the ECID in the response...which seems to happen for all our collect calls.  
> 
> Start by reproducing it with a functional test.

and telling it to put the tests in a specific file, using `decodeKndctrCookie.js` as a reference for style. It wrote a basic version of the test, but I noticed a few errors. I  told it to

> Use the identity component (which defines the getIdentity command) as a reference in writing the test. 

It searched the codebase and edited the test to make it fit better. It was still making the error of including the `CORE` namespace without third-party cookies, so I told it 

> CORE is only included when third party cookies are enabled, so we don't need to include it here in our test.

and it removed it from the test. I couldn't remember exactly how to force a `sendBeacon`, so I asked it 

> Find out how to send a `/collect` beacon with `sendEvent`. Something like "documentUnloading: true" is the easiest, but verify against the code.

and it searched through to code and modified the test accordingly. I then manually verified that the test it wrote was failing.

Now it was time to start fixing it.

> I've run the functional tests with `npm run test:functional:build:int && npm run test:functional` and it looks like they are failing successfully.
> 
> Where do we start with fixing this? Probably in identity/createcomponent

and it wrote the code into `Identity/createComponent.js`. The end result is the code in the diff as it currently stands. I ran the tests to verify that the test was now passing. 

It is a very simple ticket and fix, but it was still interesting to see that it worked decently and automatically. I think it's worth checking out.
</details>

## Related Issue

[PDCL-13107: Potential Bug in "getIdentity" command](https://jira.corp.adobe.com/browse/PDCL-13107)

## Types of changes

## Checklist:

### DIFF
--- a/test/functional/specs/Identity/getIdentityAfterCollectCall.js
+++ b/test/functional/specs/Identity/getIdentityAfterCollectCall.js
@@ -1,0 +1,67 @@
+/**
+ * Copyright 2025 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import { t } from "testcafe";
+import createNetworkLogger from "../../helpers/networkLogger/index.js";
+import createFixture from "../../helpers/createFixture/index.js";
+import {
+  compose,
+  orgMainConfigMain,
+  debugEnabled,
+} from "../../helpers/constants/configParts/index.js";
+import createAlloyProxy from "../../helpers/createAlloyProxy.js";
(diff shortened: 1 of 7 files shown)


test/functional/specs/Identity/getIdentityAfterCollectCall.js
```diff
@@ -0,0 +1,67 @@
+/**
```
jonsnyder, line 1: Can you create a test rail number for this test?

jonsnyder (approved):

test/functional/specs/Identity/getIdentityAfterCollectCall.js
dompuiu: Didn’t we say we won’t use that anymore?

carterworks: I don't remember any formal decision, so I've created a testrail case for each of the functional tests that were missing one.

PR #1270, opened 2025-03-20
Add generate event hash method.

dompuiu:
## Description

Add the new method that will generate the event history key hash from an event object. It will remove the empty keys, sort the keys, make a string from it and generate the hash using FNV1A32 algorithm.

## Types of changes

## Checklist:

### DIFF
--- a/src/components/RulesEngine/utils/generateEventHash.js
+++ b/src/components/RulesEngine/utils/generateEventHash.js
@@ -9,7 +9,6 @@ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTA
 OF ANY KIND, either express or implied. See the License for the specific language
 governing permissions and limitations under the License.
 */
-import sortObjectKeysRecursively from "../../../utils/sortObjectKeysRecursively.js";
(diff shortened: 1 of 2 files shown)


src/components/RulesEngine/utils/generateEventHash.js
```diff
@@ -0,0 +1,39 @@
...
+export default (o) => {
+  const obj = structuredClone(o);
+
+  const objectString = Object.keys(obj)
+    .sort()
+    .reduce((result, key) => {
+      const value = obj[key];
+
+      // eslint-disable-next-line eqeqeq
+      if (value == undefined || value === "") {
```
carterworks, line 29: This seems to be just a top-level hashing i.e. it does not descend into nested objects. Is that any concern?

dompuiu: Not at this moment. The objects that this method will receive are 1 level depth with string keys and values.

I asked the AJO mobile team how do they serialize objects or arrays with higher depth and they couldn’t provide an answer.

carterworks (approved):

PR #1213, opened 2024-11-07
Mark dom-action items that were executed on the page to not re-apply them

ninaceban:
## Description

If the customer is having a React SPA application, we recommend to subscribe to React lifecycle hook to re-apply the view when the component was re-rendered. In this situation when the experience has some destructive actions like `appendHTML`, `insertBefore`, `insertAfter`. The same experience get's re-applied which can lead into doubled activity showed on the page.

In the SPA context, when a view get's rendered we will mark the container that was augmented by Web SDK with the item identifier - `alloy-{item.id}`. When re-applying the view we will check to see if that container has already the experience(the item id), we will not re-apply it. 

## Related Issue

https://jira.corp.adobe.com/browse/PDCL-12754
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Improvement (non-breaking change which does not add functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] I have signed the [Adobe Open Source CLA](https://opensource.adobe.com/cla.html) or I'm an Adobe employee.
- [ ] I have made any necessary test changes and all tests pass.
- [ ] I have run the Sandbox successfully.

PR #1271, opened 2025-03-21
Fixed an issue where propositions would be reapplied in SPA use cases

carterworks:
## Description

Closes #1213—is a reimplementation.

When a proposition is rendered, it will put the proposition ID in a list contained in the `data-adobe-proposition-ids` property of the target element.

Before a proposition is rendered, it will look at the target element and read the `data-adobe-proposition-ids` property. If the ID of the proposition to be rendered is in the list, it will skip the rendering.

This resolves an issue where, in SPAs, non-idempotent proposition actions like `appendHtml`, `prependHtml`, `insertAfter`, and `insertBefore` would be rendered twice by `alloy("applyPropositions", { /* ... */ }`.

## Related Issue

https://jira.corp.adobe.com/browse/PDCL-12754

## Types of changes

## Checklist:


src/components/Personalization/handlers/createRenderStatusHandler.js
```diff
@@ -0,0 +1,57 @@
...
+    /**
+     * Determines if an element should be rendered based on scopeType and previous render status
+     *
+     * @param {Element|null} propositionContainer - The DOM element to check
+     * @returns {boolean} True if the element should be rendered, false otherwise
+     */
+    shouldRender: (propositionContainer) => {
+      // TODO: Figure out why Nina was filtering for only VIEW_SCOPE_TYPE.
+      // What about page scope?
+      if (propositionContainer) {
```
carterworks, line 31: One thing that I am unsure about: [in Nina's original PR, she only prevents rendering on propositions in the `view` scope](https://github.com/adobe/alloy/pull/1213/files#diff-3a0d487fd8c8e4f999642805edc6415589b1fd8c6e955db7a21455391bb95152R24). I don't know why, but I also am not very familiar with Target scopes. Should I be doing the same thing here?

jonsnyder: Nina was fixing a specific problem where view scoped activities were being applied multiple times. This would happen in a SPA where you visited a view multiple times. Let's have a conversation about when else this will apply.

dompuiu (approved):

test/functional/specs/Personalization/C22098199.js
```diff
@@ -0,0 +1,143 @@
...
+
+  // Second application - should not append again
+
+  await alloy.applyPropositions({
+    propositions,
+  });
+
+  // Verify the content was not appended again
+  await t.expect(renderedContent.count).eql(1);
+};
```
jonsnyder, line 134: Can you add sendEvent calls with "includeRenderedPropositions: true" after the applyProposition calls to check that the display notification are still sent even when not rendered? Also can you add a validation on the return value of applyPropositions that the proposition was marked rendered?

test/unit/specs/components/Personalization/dom-actions/collectInteractions.spec.js
```diff
@@ -19,6 +19,7 @@ import cleanUpDomChanges from "../../../../helpers/cleanUpDomChanges.js";
 import { INTERACT_ID_DATA_ATTRIBUTE } from "../../../../../../src/components/Personalization/handlers/createDecorateProposition.js";
 import { getAttribute } from "../../../../../../src/components/Personalization/dom-actions/dom/index.js";
 import createDecoratePropositionForTest from "../../../../helpers/createDecoratePropositionForTest.js";
+import createRenderStatusHandler from "../../../../../../src/components/Personalization/handlers/createRenderStatusHandler.js";
```
jonsnyder, line 22: Generally, you don't want to import production code other than the code under test. You could mock the createRenderStatusHandler pretty easily with this:
```
export default {
      shouldRender: () => true,
      markAsRendered: () => {},
    };
```

jonsnyder (commented):
I love the abstraction for RenderStatusHandlers. Here are a couple comments to look at when you have a minute

Issue #1276, opened 2025-04-01
Missing typescript types: Could not find a declaration file for module '@adobe/alloy'


cheesestringer:
**Issue**
Alloy should include some basic typescript types so this library works out of the box for typescript applications.
Currently when trying to consume the library you get an import error of `Could not find a declaration file for module '@adobe/alloy'`

**Related Issues**
https://github.com/adobe/alloy/issues/799

**Reproduction steps:**

1. Create a new typescript project, like rspack for example `npm create rsbuild@latest`
See https://rspack.dev/guide/start/quick-start#create-a-new-project
2. In the terminal select `React` (version doesn't matter) and then select `Typescript`
3. Install the alloy library `npm i @adobe/alloy`
4. Add `import { createInstance } from "@adobe/alloy";` in the `App.tsx`
5. Hover over the import to see the error `Could not find a declaration file for module '@adobe/alloy'.`

There are hacks and workarounds to get past this error, but some basic types should be included in the package to avoid them, as well as give our applications some type safety.

Issue #799, opened 2022-01-28
typescript support for clients using web sdk


aniham:
it would be useful/unbreak some linters if alloy exports a set of types that can be referenced by a typescript client using it.

alex-arriaga:
Hi Alloy team

Are there any plans to support TypeScript types?

PR #1277, opened 2025-04-01
Add auto generation of TypeScript definitions from JavaScript using JSDoc

cheesestringer:
## Description

* Add auto generated TypeScript definitions for Alloy so when the library is installed it doesn't immediately throw warnings.
* Add JSDoc comments so IDEs can show inline documentation.

See https://www.typescriptlang.org/docs/handbook/declaration-files/dts-from-js.html

Fixes https://github.com/adobe/alloy/issues/1276
Fixes https://github.com/adobe/alloy/issues/799

## Related Issue
https://github.com/adobe/alloy/issues/1276
https://github.com/adobe/alloy/issues/799

## Motivation and Context

Adding TypeScript definitions to the Alloy library aligns with modern best practices, as TypeScript is widely adopted in the JavaScript ecosystem. Providing type definitions out of the box makes it easier for developers to get started with the library, as they benefit from features like type safety, autocompletion, and inline documentation in their IDEs.

The current developer experience when installing Alloy into a brand new TypeScript project is:

```
Could not find a declaration file for module '@adobe/alloy'. 'c:/git/alloy-demo/alloy-demo/node_modules/@adobe/alloy/libEs5/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/adobe__alloy` if it exists or add a new declaration (.d.ts) file containing `declare module '@adobe/alloy';`
```

There are ways around this using hacks like `// @ts-ignore` which sometimes needs to be done in a few different places depending on the framework. We can make the developer experience nicer by not needing workarounds to get the library working straight away.

## Screenshots
**Before**
image

**After**
image

## Testing
I ran `npm pack` to generate the npm package and directly installed it in a brand-new project. The `Could not find a declaration file for module '@adobe/alloy'` error disappeared (see after screenshot) and I was able to create a new alloy instance and send `configure` without issues.

## Types of changes

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,9 +5,11 @@
   "type": "module",
   "main": "libEs5/index.js",
   "module": "libEs6/index.js",
+  "types": "types/index.d.ts",
   "files": [
     "libEs5",
     "libEs6",
+    "types",
     "LICENSE_BANNER",
     "babel.config.js",
     "rollup.config.js",
@@ -20,9 +22,10 @@
     "alloyBuilder": "scripts/alloyBuilder.js"
   },
   "scripts": {
-    "clean": "rimraf dist distTest libEs5 libEs6",
+    "clean": "rimraf dist distTest libEs5 libEs6 types",
     "lint": "eslint --cache --fix \"*.{js,cjs,mjs,jsx}\" \"{src,test,scripts,sandbox}/**/*.{js,cjs,mjs,jsx}\"",
     "format": "prettier --write \"*.{html,js,cjs,mjs,jsx}\" \"{sandbox,src,test,scripts}/**/*.{html,js,cjs,mjs,jsx}\"",
+    "types": "tsc",
     "test": "npm run test:unit && npm run test:scripts",
     "test:unit": "npx playwright install chromium && vitest run",
     "test:unit:debug": "npx playwright install chromium && vitest --no-file-parallelism --browser.headless=false",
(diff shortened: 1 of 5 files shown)


cheesestringer:
@carterworks Sorry to tag directly but doesn't look like I can request a review 😅 Could you let me know if you see any issues with this one?

cheesestringer:
Nice, I'll check this out over the weekend!

carterworks:
Actually, don't worry about it. @dompuiu from our team is working on it right now. Thank you for bringing this up and getting the ball rolling.

cheesestringer:
@carterworks Had already made the changes and just needed to tidy up / push but feel free to use this as a reference 😅

The only gotcha with JSDoc comments seems to be for arrow functions, the comments won't appear on the compiled types unless you add an `@type` property (see https://github.com/microsoft/TypeScript/issues/34860). An issue with using `@type` is you need to manually set the type yourself, which might not match the actual contract. Using regular functions, the type will be automatically generated.

src/index.js
```diff
@@ -61,7 +72,21 @@ export const createCustomInstance = (options = {}) => {
...
+ * @param {Array<Object>} [options.monitors] - (Optional) Monitors for the instance.
+ * @returns {(commandName: string, options?: Object) => Promise<any>} A callable Alloy instance.
+ * 
+ * @example
+ * const alloy = createInstance({ name: "myInstance" });
+ * alloy("configure", { datastreamId: "myDatastreamId", orgId: "myOrgId" });
+ * 
+ * @see {@link https://experienceleague.adobe.com/en/docs/experience-platform/web-sdk/install/npm} for more details.
+ */
+export function createInstance(options = {}) {
```
cheesestringer, line 89: An issue with creating types from arrow functions is the JSDoc comments won't appear unless you use `@type` to manually set the type. I changed these to regular functions so the type can be automatically generated which will avoid any user error defining the contract.

If arrow functions are preferred, we can add `@type {(options?: Object) => Function}` to the JSDoc comment but need to remember if the contract changes the `@type` fields need to be manually updated.

tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "include": ["src/**/*"],
```
dompuiu, line 2: Can you add `  "exclude": ["node_modules"]`?

package.json
```diff
@@ -38,7 +41,7 @@
     "sandbox:build": "rollup -c --environment SANDBOX && cd sandbox && npm run build",
     "dev": "concurrently --names build,sandbox \"rollup -c -w --environment SANDBOX\" \"cd sandbox && export REACT_APP_NONCE=321 && npm start\"",
     "dev:standalone": "npm run clean && rollup -c -w --environment STANDALONE",
-    "build": "npm run clean && rollup -c --environment BASE_CODE_MIN,STANDALONE,STANDALONE_MIN,BUNDLESIZE && echo \"Base Code:\" && cat distTest/baseCode.min.js",
+    "build": "npm run clean && npm run types && rollup -c --environment BASE_CODE_MIN,STANDALONE,STANDALONE_MIN,BUNDLESIZE && echo \"Base Code:\" && cat distTest/baseCode.min.js",
```
dompuiu, line 44: I think we need to move `npm run types` to the `prepack` script.

src/index.js
cheesestringer: Easy, will update

package.json
cheesestringer: Good idea, fixed!

dompuiu (approved):

PR #1289, opened 2025-04-24
Rework sandbox

dompuiu:
## Description

I tried to change how Sandbox was working. Now on each page, a new instance is initialized. 

## Types of changes

## Checklist:


sandbox/src/helpers/getUrlParameter.js
```diff
@@ -0,0 +1,8 @@
+export default (name) => {
+  name = name.replace(/[[]/, "\\[").replace(/[\]]/, "\\]");
```
carterworks, line 2: This could be greatly simplified with `URLSearchParam`.

Example:

```

/**
 * @param {string} name
 * @returns {string | null}
 */
export default (name) => {
  const params = new URLSearchParams(window.location.search)
  return params.get(name);
}
```

sandbox/src/helpers/setupAlloy.js
```diff
@@ -0,0 +1,44 @@
...
+      Visitor.getInstance("53A16ACB5CC1D3760A495C99@AdobeOrg", {
+        doesOptInApply: getUrlParameter("legacyOptIn") === "true",
+      });
+      // Alloy only looks for window.Visitor when it initially loads, so only load Alloy after Visitor loaded.
+      includeScript("/alloy.js");
+    });
+  } else {
+    includeScript("/alloy.js");
+  }
+};
```
carterworks, line 44: Since `setupAlloy()` and `configureAlloy()` are 

1. always called together
2. only used in a react context

I'd prefer to see them combined into a React hook. Something that can be used like this:

```js
const alloy = useAlloy({
  instanceName: "alloy",
  config: {
    /* ... */
  }
});
```

Open for debate, since this is only a developer sandbox.

carterworks (approved):
Neither of my two comments are blockers, just preferences.

PR #1290, opened 2025-04-29
Expand list of supported browsers and introduce compatibility ESLint plugin

carterworks:
### Adjust our Babel transpilation targets

Previously, our list of supported browsers was ["the last two versions of Chrome, Firefox, Edge, and Safari"](https://browsersl.ist/#q=last+2+Chrome+versions%0Alast+2+Firefox+versions%0Alast+2+Safari+versions%0Alast+2+Edge+versions). This covers 15.5% of global browser usage
image

This PR will adjust the target to ["defaults and not op_mini all"](https://browsersl.ist/#q=defaults%2C+not+op_mini+all) which is a shorthand for "> 0.5%, last 2 versions, Firefox ESR, not dead, not op_mini"
image

Does this change any code that gets generated? **No**. But, it sets us up for the other change in this PR…

## Introduce `eslint-plugin-compat`

[`eslint-plugin-compat`](https://github.com/amilajack/eslint-plugin-compat) will throw a linting error when a JS feature outside of that supported browsers list is used. 

This will help in two ways:

1. We can give future AEP Universal JS SDK packages different browser compatibility rules. For example the `@adobe/alloy-node` can target [`maintained node versions`](https://browsersl.ist/#q=maintained+node+versions)
2. We can confidently get rid of babel, which will simplify our build process (my current task). We don't need it, it doesn't do anything to change the level of targets we support, and I'd rather we ship the raw source as the NPM library instead.

## Types of changes

## Checklist:

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,4 +1,2 @@
-last 2 Chrome versions
-last 2 Firefox versions
-last 2 Safari versions
-last 2 Edge versions
+defaults
+not op_mini all
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -11,16 +11,17 @@ governing permissions and limitations under the License.
 */
 
 import { FlatCompat } from "@eslint/eslintrc";
+import pluginJs from "@eslint/js";
+import vitest from "@vitest/eslint-plugin";
+import eslintPluginPrettierRecommended from "eslint-plugin-prettier/recommended";
+import react from "eslint-plugin-react";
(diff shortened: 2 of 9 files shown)


dompuiu (approved):
Related to your comment that you want to remove babel. I am not sure how easy that will be. If you remove it, you need to think on how will you still keep the custom build working (in alloy and in the extension).

jonsnyder (commented):
Babel also does the version number replacement, but it's always great when we can get away from a dependency.

PR #1295, opened 2025-05-09
Simplify and modernize build.

carterworks:
## Description

The goals of this PR are:

- Use only rollup to build the library
- Remove unnecessary steps from the rollup build process  
- Move all build outputs to the `dist/` directory
- Use the `exports` field of the `package.json` file to control what is exported from the library

### Use only rollup to build the library

The `main` rollup config would build specific outputs based on environment variables. This made the control flow very complex, and it was hard to make a change and know exactly which output(s) would be affected. This PR removes that ability and just builds everything, all the time. This sounds like more work, but on my laptop it takes 6 seconds instead of the 12 it used to take.

### Remove unnecessary steps from the rollup build process

The `main` would build the `npm` library with the `babel` tool, outputting transpiled versions of the source code into the `libEs6/` (for ES modules) and `libEs5/` (for CommonJS) directories. This PR removes the `babel` tool (it wasn't doing anything, and added a significant amount of time to the build process). Other options have been removed if they were similarly ineffective.

### Move all build outputs to the `dist/` directory

The `main` build would output to the following directories:

* `distTest/` - Builds for functional tests
* `dist/` - Standalone builds for the CDN
* `sandbox/public/` - Standalone builds for the sandbox
* `libEs6/` - ES modules for NPM
* `libEs5/` - CommonJS for NPM

This PR moves all of these outputs into the `dist/` directory. The final build outputs are:

* `dist/alloy.js` - The ES module version of the library, for npm
* `dist/alloy.cjs` - The CommonJS version of the library, for npm
* `dist/alloy.standalone.js` - The standalone version of the library, for the CDN
* `dist/alloy.standalone.min.js` - The minified standalone version of the library, for the CDN
* `dist/baseCode.js` - The base code version of the library, for use with the CDN
* `dist/baseCode.min.js` - The minified base code version of the library, for use with the CDN

Additionally, there is an additional `utils` bundle, in CJS and ESM formats, that is intended to be used in the `reactor-extension-alloy`, which imports the `deepAssign` and `createEventMergeId` functions from `alloy`. The goal is that all environments-tests, sandbox, CDN, NPM-should use the same outputs.

### Use the `exports` field of the `package.json` file to control what is exported from the library

The `main` branch of alloy defines the exports using 3 different fields in the `package.json` file:

* `main` - The main entry point for the library
* `module` - The ES module version of the library
* `types` - The type definitions for the library

This way of defining exports is deprecated in Node 16, and the `exports` field is the preferred way to define exports from a module. The benefits is that there is a single key to define the exports. It is also a form of access control: only the keys defined in the `exports` field are available when the library is imported. Nothing else is available.

## To-do

- [x] Re-add tests for node_modules/alloy.cjs
- [x] Re-add tests for dist/alloy.cjs
- [x] Fix `test:*:watch` scripts
- [x] Reconcile the `package-lock.json`

## Types of changes

## Checklist:


dompuiu:
After the last commit, there is still a file called alloy.cjs beign created. There is also a filed called alloy.js. Is this intentional?

carterworks:
> After the last commit, there is still a file called alloy.cjs beign created. There is also a filed called alloy.js. Is this intentional? Is the one ending in `.cjs` for es5 (require) and the one ending in `.js` for es6 (import/export)?

Yes this is correct. 'alloy.js' and '.cjs' are the EMS and CJS builds of the npm library. 'alloy.standalone.js' is an iife build for the CDN

dompuiu:
@carterworks Seems the package-lock.json is out of sync.

dompuiu:
`@rollup/plugin-json` needs to be a dependency not a devDependency. Otherwise the alloyBuild.js script is failing.

On a different note, can't we read the version in a way so that we don't need a plugin?

dompuiu:
The standalone version is still added to the NPM package.

carterworks:
> The standalone version and the basecode is still added to the NPM package.

@dompuiu 

Fixed. The `files` key in the `package.json` has been restricted to just specific files in the `dist/` dir.

The `test:*:watch` scripts work now. Vite uses rollup under the hood, so it can import the raw source code without any issues.

This PR is ready for review.

dompuiu (approved):

PR #1317, opened 2025-06-11
Reorganize ESLint config and remove Airbnb style plugin

carterworks:
## Description

[the Airbnb style plugin for ESLint does not support ESLint v9, which we use](https://github.com/airbnb/javascript/issues/2804#issuecomment-2118873341). ESLint v9 was released a while ago and yet this plugin has not been updated, and the maintainer is very busy.

Because it required v8 and we require v9 everywhere else, it was causing problems with our dependency resolution. My initial monorepo prototypes had issues when running `npm install` because of this version conflict.

So I have taken the most relevant rules from the plugin and copy-pasted them into our eslint configuration. I then moved this around so that which config applies to which section is more clear.

I have also removed the testcafe ESLint plugin, since it was just two lines that define globals, and reproduced it in the config section applied to the functional tests.

## Types of changes

## Checklist:


eslint.config.js
```diff
@@ -10,62 +10,120 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
-      "prefer-destructuring": "off",
       "import/prefer-default-export": "off",
+    },
+  },
+  {
+    name: "alloy/src",
+    files: ["src/**/*.{cjs,js}"],
+    languageOptions: {
+      globals: {
+        turbine: "readonly",
```
jonsnyder: Is turbine used in the Web SDK library?

carterworks: It is not, but the line was present previously and I opted not to remove it.

PR #1322, opened 2025-06-25
Fixed an issue where HTML content items failed to render due to missing renderStatusHandler

carterworks:
## Description

Fixed a bug where HTML content items (like those used in Adobe Target) would fail to render with the error "Cannot read properties of undefined (reading 'shouldRender')". The issue was caused by `createProcessHtmlContent.js` not passing the required `renderStatusHandler` parameter to DOM action modules, while `createProcessDomAction.js` was correctly passing it.

## Related Issue

* [TGT-52945 - Apply Propositions Function Broken (Version: 2.30.1)](https://jira.corp.adobe.com/browse/TGT-52945)
* [PLATIR-51065 - Activity not rendering after web sdk extension update](https://jira.corp.adobe.com/browse/PLATIR-51065)

## Motivation and Context

Recent changes in PR #1271 introduced render status handling to prevent duplicate rendering in SPAs. This required all DOM action calls to receive a `renderStatusHandler` parameter. While `createProcessDomAction.js` was updated to pass this parameter, `createProcessHtmlContent.js` was missed, causing HTML content items to fail with `"shouldRender"` errors.
This fix ensures both DOM action and HTML content processing paths consistently pass the `renderStatusHandler`, resolving the rendering failures reported by customers after upgrading to v2.30.1.

## Checklist:

### DIFF
--- a/src/components/Personalization/handlers/createProcessHtmlContent.js
+++ b/src/components/Personalization/handlers/createProcessHtmlContent.js
@@ -10,6 +10,7 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 import createDecorateProposition from "./createDecorateProposition.js";
+import createRenderStatusHandler from "./createRenderStatusHandler.js";
 
 export default ({
     modules,
@@ -40,9 +41,18 @@ export default ({
       storeInteractionMeta,
     );
 
+    const renderStatusHandler = createRenderStatusHandler(
+      item.getProposition().getScopeType(),
+      item.getId(),
+    );
+
     return {
       render: () => {
-        return modules[type](item.getData(), decorateProposition);
+        return modules[type](
+          item.getData(),
+          decorateProposition,
+          renderStatusHandler,
+        );
       },
       setRenderAttempted: true,
       includeInNotification: true,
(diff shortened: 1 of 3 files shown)


src/components/Personalization/handlers/createProcessHtmlContent.js
```diff
@@ -40,9 +41,18 @@ export default ({
...
+      item.getId(),
+    );
+
     return {
       render: () => {
-        return modules[type](item.getData(), decorateProposition);
+        return modules[type](
+          item.getData(),
+          decorateProposition,
+          renderStatusHandler,
```
jonsnyder, line 54: Do you need to add the renderStatusHandler here too?

https://github.com/adobe/alloy/blob/f7b6e1024ccb69bd87c0cd881ac7f20c00ac4fa4/src/components/Personalization/handlers/createProcessInAppMessage.js#L81

carterworks: I don't think so, because renderStatusHandler is only used for DOM action modules and I don't think in app messages are processed with those?

See: https://github.com/adobe/alloy/blob/c540095d5520d87921790f4f28a53dace719b987/src/components/Personalization/index.js#L114-L115

carterworks:
> Hi @carterworks,
> 
> first - thanks for the fix.
> 
> This bug indeed broke Target integration for many of our clients.
> 
> Can you give a hint when the new version is updated in the Adobe Tags Extension Catalogue and can be installed?
> 
> This is highly awaited as we cannot roll back to old Extension Version in Adobe Launch.
> 
> Or do we know a workaround until the publication?
> 
> 
> 
> Cheers and kind regards,
> 
> Martin
> 
> 

We are targeting (pun not intended) early next week. I can put a note here and on the Jira ticket.

Issue #1324, opened 2025-07-15
ActivityCollector collects wrong URLs with non-HTTP/FTP/File protocols


yuhui:
**This issue should be considered to be a MAJOR BUG!**

### Expected Behaviour
ActivityCollector collects link URLs that conform to the expected URI schema. For example:
- `https://www.domain.com/path1/path2/page`
- `file:///computer/folder1/folder2/file`
- `mailto:email@domain.com`
- `tel:8675309`
- `javascript:`

### Actual Behaviour
For non-HTTP/FTP/File protocols, ActivityCollector inserts the current page's URL path into the link URL. For example:
- `mailto:email@domain.com` becomes `mailto://www.domain.com/path1/path2/mailto:email@domain.com`
- `tel:8675309` becomes `tel://www.domain.com/path1/path2/tel:8675309`
- `javascript:` becomes `javascript://www.domain.com/path1/path2/javascript:`

### Reason
When determining the URL to use with the link, ActivityCollector relies on https://github.com/adobe/alloy/blob/c14795bf4ce8aad3ab3f5704b2c253ecafc1f5a8/src/components/ActivityCollector/utils/urlStartsWithScheme.js#L14

As you can see, this code not only checks for an alphanumeric string, but also for `://`.

But non-HTTP/FTP/File URLs **do not** have `://` in their URLs. For example, `mailto:email@domain.com` is a valid URL.

As a result, in this code block: https://github.com/adobe/alloy/blob/c14795bf4ce8aad3ab3f5704b2c253ecafc1f5a8/src/components/ActivityCollector/utils/dom/getAbsoluteUrlFromAnchorElement.js#L23-L38

Since `urlStartsWithScheme()` returns `false`, so the link URL is assumed to be invalid or assumed to be a path only. So the current page's domain and path get inserted into the link URL. As a result, `mailto:email@domain.com` becomes `mailto://www.domain.com/path1/path2/mailto:email@domain.com`.

#### Steps to Reproduce
1. In a test web page, find a non-HTTP/FTP/File link, for example, a mailto:, tel: or javascript: link.
2. Click the link.
3. Observe the link URL that ActivityCollector collects.

#### Platform and Version
All

### Suggestions
[...]

PR #1325, opened 2025-07-15
Fixed activity map url collection for non-standard URLs

jonsnyder:
## Description

Use URL constructor to figure out what the absolute URL should be.

## Related Issue

https://github.com/adobe/alloy/issues/1324

## Motivation and Context

Fixes issues with non-standard urls.

## Checklist:


carterworks (approved):

dompuiu (approved):

dompuiu (approved):
The `URL` constructor can throw exeptions: https://developer.mozilla.org/en-US/docs/Web/API/URL/URL#exceptions. It might be a very rare occurrence, but it might be worth to test what happens if you throw an error in that place. What will appear in the browser console?

PR #1318, opened 2025-06-11
Fixed identity management code to handle promise rejections for identity not found errors.

jonsnyder:
## Description

We handled this already for the first request, but there was a different code path for the second request. I fixed the issue for the first request in https://github.com/adobe/alloy/pull/1086, which was released in v2.19.2.

## Related Issue

https://jira.corp.adobe.com/browse/PLATIR-51159

## Motivation and Context

Customer was seeing an unhandled promise rejection error in the console. They were using a third-party tool to log any errors on their webpage and were seeing a bunch of these.

## Checklist:

- [x] I have signed the [Adobe Open Source CLA](https://opensource.adobe.com/cla.html) or I'm an Adobe employee.
- [x] I have made any necessary test changes and all tests pass.
- [x] I have run the Sandbox successfully.

PR #1331, opened 2025-07-22
Suppress unhandled promise rejection error

jonsnyder:
## Description

This is the real fix for #1318 

This console log was instructive to understand when the errors are shown:

```
> const makePromise2 = () => { const promise = new Promise((resolve,reject) => setTimeout(() => reject("Foo"), 1000)); promise.catch(() => {}); return promise; }
< undefined
> const p5 = makePromise2();
< undefined
> p5.then(() => "Hello")
< Promise {status: "pending"}
[Error] Unhandled Promise Rejection: Foo
> const p6 = p5.catch(makePromise2);
< undefined
[Error] Unhandled Promise Rejection: Foo (x2)
> const p7 = p5.catch(makePromise2); p7.catch(() => {});
< Promise {status: "pending"}
```

## Checklist:

### DIFF
--- a/src/components/Identity/injectEnsureSingleIdentity.js
+++ b/src/components/Identity/injectEnsureSingleIdentity.js
@@ -64,6 +64,8 @@ export default ({
       obtainedIdentityPromise = previousObtainedIdentityPromise.catch(() => {
         return awaitIdentityCookie({ onResponse, onRequestFailure });
       });
+      // This prevents an un-caught promise in the console when the identity isn't set.
+      obtainedIdentityPromise.catch(() => {});
 
       // When this returned promise resolves, the request will go out.
       return (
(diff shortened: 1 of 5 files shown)


src/components/Identity/injectEnsureSingleIdentity.js
```diff
@@ -89,6 +91,9 @@ export default ({
       onResponse,
       onRequestFailure,
     });
+    // This prevents an un-caught promise in the console when the identity isn't set.
+    obtainedIdentityPromise.catch(() => {});
```
carterworks, line 95: Would it be appropriate to use the logger to log some kind of debug log only when `{ debugEnabled: true }`?

jonsnyder: No need to log because we log before we throw the exception

PR #1341, opened 2025-08-15
Add a way for waiting getting identity information inside a component.

dompuiu:
## Description

For push notifications we need to send the ECID inside the profile update payload. For the case when ECID is not present, we need a way to wait for the ECID to be fetch by another send event. I followed the pattern we used for waiting for obtaining user consent.

## Checklist:


src/core/identity/createIdentity.js
```diff
@@ -0,0 +1,46 @@
...
+      }
+    },
+
+    setIdentityAcquired() {
+      awaitIdentityResolve();
+    },
+
+    async awaitIdentity() {
+      return await awaitIdentityPromise;
+    },
```
carterworks: 1. I don't see this used anywhere except tests
2. This can be just

```js
    awaitIdentity() {
      return awaitIdentityPromise;
    }
```
no need to construct a new promise.

dompuiu: Related to 1, you are right. Is not used yet. I will use it inside the PushNotifications component.

dompuiu: For 2, good catch.

PR #1344, opened 2025-08-22
Add support for sending push notification subscription details

dompuiu:
## Description

Adding a new Push notifications module. This module adds a command that will send the push subscription details to the profile.

## Checklist:


eslint.config.js
```diff
@@ -230,6 +230,7 @@ export default defineConfig([
     },
     rules: {
       ...react.configs.recommended.rules,
+      "react/prop-types": "off",
```
dompuiu, line 233: This one was showing some lint errors in sandbox.

carterworks (approved):
Looks good. This only handles sending the push subscription details to AEP. The other stuff, like interact notifications and the service worker, are coming later because the requirements are still muddy, correct?

src/components/PushNotifications/index.js
```diff
@@ -0,0 +1,101 @@
...
+  consent,
+  identity,
+  sendEdgeNetworkRequest,
+}) => {
+  return {
+    commands: {
+      sendPushSubscription: {
+        run: async () => {
+          if (isComponentDisabled(config)) {
+            return Promise.reject(
```
carterworks: Because this is an `async` function, you can just throw the error. No need to reject the promise manually.

dompuiu: Good catch.

src/components/PushNotifications/helpers/getPushSubscriptionDetails.js
```diff
@@ -0,0 +1,118 @@
...
+
+  if (!("PushManager" in window) || !("Notification" in window)) {
+    throw new Error("Push notifications are not supported in this browser.");
+  }
+
+  /** @type {object} */
+  const notification = window.Notification;
+  if (notification.permission !== "granted") {
+    throw new Error(
+      "The user didn't grant permissions for push notifications.",
```
jonsnyder: "The user has not given permission to send push notifications."

src/components/PushNotifications/helpers/getPushSubscriptionDetails.js
```diff
@@ -0,0 +1,118 @@
...
+    if (!subscription) {
+      throw e;
+    }
+
+    const unsubscribeResult = await subscription.unsubscribe();
+    if (!unsubscribeResult) {
+      throw e;
+    }
+
+    return getPushSubscriptionDetails({ vapidPublicKey, window });
```
jonsnyder, line 114: Could this lead to an infinite loop? What do you think about adding a retries count param?

src/components/PushNotifications/index.js
```diff
@@ -0,0 +1,99 @@
...
+/** @import { EventManager, Logger } from '../../core/types.js' */
+/** @import { IdentityManager } from '../../core/identity/types.js' */
+/** @import { ConsentManager } from '../../core/consent/types.js' */
+/** @import { EdgeRequestExecutor } from '../../core/edgeNetwork/types.js' */
+
+import { objectOf, string } from "../../utils/validation/index.js";
+import { sanitizeOrgIdForCookieName } from "../../utils/index.js";
+import makeSendPushSubscriptionRequest from "./request/makeSendPushSubscriptionRequest.js";
+
+const isComponentDisabled = ({
```
jonsnyder: "isComponentConfigured" may be a better name.

src/utils/createLoggingCookieJar.js
```diff
@@ -10,12 +10,26 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+ *
+ * @returns {CookieJar}
+ */
 export default ({ logger, cookieJar }) => {
   return {
     ...cookieJar,
     set(key, value, options) {
       logger.info("Setting cookie", { name: key, value, ...options });
-      cookieJar.set(key, value, options);
+      return cookieJar.set(key, value, options);
```
jonsnyder, line 32: Does cookieJar.set return a value? is this change intented?

src/components/PushNotifications/helpers/getPushSubscriptionDetails.js
dompuiu: done

src/utils/createLoggingCookieJar.js
dompuiu: It was intentional and it seems more correct. If cookieJar.set doesn't return a value, meaning it returns undefined then that is what we will also return. If it returns a value, then we also return a value. Do you see any problems with this approach?

src/components/PushNotifications/index.js
dompuiu: done

src/components/PushNotifications/helpers/getPushSubscriptionDetails.js
dompuiu: The idea is that it just retries once in case of a failure. I don't see how it would get into an infinite loop. How do you think it could get into one?

jonsnyder: Consider if subscribe threw an exception every time, and unsubscribe always returned a value. I know this doesn't happen in the browsers you've tested, but who knows what could happen with these APIs?

dompuiu: subscription.unsubscribe is returning a promise that is resolved to boolean true once the subscription was removed. 

It then seems that there are the following 4 paths:

- subscription.unsubscribe resolves to true - then we try one more time. Then subscribe throws again. Then getSubscription is no longer returning anything (otherwise subscription.unsubscribe would not have resolved to true) and we throw an error;
- subscription.unsubscribe resolves to false - we throw the error;
- subscription.unsubscribe never resolves, we never return a result but we don't get stuck into an infinite loop;
- subscription.unsubscribe throws an errors - that error is not caught, so that goes up the stack.

If what you describe in your example would ever happen, I think that the browser should fix this way to behave.

I can add that param if you have a strong opinion about this. But it seems unnecessary imho.

jonsnyder: No, I don't have a strong opinion. Just wanted to make sure you've thought it through.

PR #1355, opened 2025-08-29
Add push notifications service worker code.

dompuiu:
## Description

This is how the service worker for push notification could potentially look. At this moment I am not able to test what Herald will send to us. It will suffer some changes.

## Checklist:

### DIFF
--- a/src/components/PushNotifications/helpers/serviceWorkerNotificationClickListener.js
+++ b/src/components/PushNotifications/helpers/serviceWorkerNotificationClickListener.js
@@ -29,7 +29,7 @@ const canHandleUrl = (type) => ["DEEPLINK", "WEBURL"].includes(type);
  * @param {Object} options
  * @param {ServiceWorkerGlobalScope} options.sw
  * @param {NotificationEvent} options.event
- * @param {Function} options.fetch
+ * @param {(url: string, options: object) => Promise<Response>} options.fetch
  * @param {ServiceWorkerLogger} options.logger
  */
 export default ({ event, sw, logger, fetch }) => {
--- a/src/components/PushNotifications/request/makeSendServiceWorkerTrackingData.js
+++ b/src/components/PushNotifications/request/makeSendServiceWorkerTrackingData.js
@@ -27,7 +27,7 @@ import uuidv4 from "../../../utils/uuid.js";
  * @param {number} [options.applicationLaunches=0]
  * @param {Object} utils
  * @param {ServiceWorkerLogger} utils.logger
- * @param {Function} utils.fetch
+ * @param {(url: string, options: object) => Promise<Response>} utils.fetch
  *
  * @returns {Promise<boolean>}
  * @throws {Error}
--- a/test/unit/specs/components/PushNotifications/helpers/serviceWorkerNotificationClickListener.spec.js
+++ b/test/unit/specs/components/PushNotifications/helpers/serviceWorkerNotificationClickListener.spec.js
@@ -1,0 +1,327 @@
+/*
+Copyright 2025 Adobe. All rights reserved.
+This file is licensed to you under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License. You may obtain a copy
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under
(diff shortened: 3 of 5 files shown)


scripts/alloyBuilder.js
```diff
@@ -190,6 +248,76 @@ const getMakeBuildCommand = () => {
...
+        .default(getProjectRoot())
+        .argParser((value) => {
+          if (!path.isAbsolute(value)) {
+            value = path.join(process.cwd(), value);
+          }
+
+          try {
+            const stats = fs.statSync(value);
+            if (!stats.isDirectory()) {
+              throw new InvalidOptionArgumentError(
```
carterworks, line 269: This exception will get caught by the `catch` on line 276, thus duplicating the exception message.

carterworks (approved):

sandbox/src/PushNotifications.jsx
```diff
@@ -19,18 +19,22 @@ export default function Home() {
   useAlloy({
     configurations: {
       alloy: {
-        orgId: "97D1F3F459CE0AD80A495CBE@AdobeOrg",
-        datastreamId: "9db94cb5-a892-4ae9-b117-a667a4ed67a8",
+        orgId: "745F37C35E4B776E0A49421B@AdobeOrg",
+        datastreamId: "5aba15dc-b57b-41eb-99a5-5b7b083f2120",
```
Spencer-Smith, line 23: Just curious here, why are changing the orgId and datastreamId for the sandbox? I'm assuming these new ones were set up to test the push notifications, but were we not able to do the same setup for the existing org/datastream?

scripts/alloyBuilder.js
dompuiu: Good catch. Fixed. Thanks!

sandbox/src/PushNotifications.jsx
dompuiu: For the moment, the changes related backend services are deployed only on stage. We change also the `edgeDomain`. At some point when everything will be deployed on production, we will change these back.

src/components/PushNotifications/serviceWorker.js
```diff
@@ -0,0 +1,412 @@
...
+  try {
+    for (const field of requiredFields) {
+      if (!configData[field.name]) {
+        throw new Error(
+          `Cannot send tracking call. ${field.errorField} is missing.`,
+        );
+      }
+    }
+
+    const url = `https://${edgeDomain}/${edgeBasePath}/v1/interact?configId=${datastreamId}`;
```
jonsnyder: No request id?

src/components/PushNotifications/serviceWorker.js
```diff
@@ -0,0 +1,412 @@
...
+            },
+          },
+          meta: {
+            collect: {
+              datasetId,
+            },
+          },
+        },
+      ],
+    };
```
jonsnyder: Probably should include implementationDetails

src/components/PushNotifications/serviceWorker.js
```diff
@@ -0,0 +1,412 @@
...
+  if (!event.data) {
+    return;
+  }
+
+  /** @type {PushNotificationData} */
+  let notificationData;
+  try {
+    notificationData = event.data.json();
+  } catch {
+    return;
```
jonsnyder: Should we log these failure conditions?

src/components/PushNotifications/serviceWorker.js
```diff
@@ -0,0 +1,412 @@
...
+sw.addEventListener("notificationclose", (event) => {
+  const data = event.notification.data;
+
+  sendTrackingCall({
+    xdm: data._xdm.mixins,
+    actionLabel: "Dismiss",
+  }).catch((error) => {
+    logger.error("Failed to send tracking call:", error);
+  });
+});
```
jonsnyder, line 89: You could split this into multiple files to facilitate unit testing.

jonsnyder (approved):

src/components/PushNotifications/serviceWorker.js
dompuiu: Seems the server generates one if I don't provide one.

src/components/PushNotifications/serviceWorker.js
dompuiu: Why do you think we should include the implementationDetails? I don't think they are used by the tracking.

PR #1359, opened 2025-09-08
Allow appID field to be provided by the user.

dompuiu:
## Description

The application ID needs to be provided by the customer.

## Checklist:

### DIFF
--- a/test/unit/specs/components/PushNotifications/configValidators.spec.js
+++ b/test/unit/specs/components/PushNotifications/configValidators.spec.js
@@ -22,7 +22,7 @@ describe("PushNotifications configValidators", () => {
     }).not.toThrow();
   });
 
-  it("should require both vapidPublicKey and appId when pushNotifications is provided", () => {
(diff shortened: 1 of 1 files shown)


src/components/PushNotifications/index.js
```diff
@@ -54,15 +55,14 @@ const createPushNotifications = ({
...
+            throw new Error("Push notifications module is not configured.");
           }
 
           const {
             orgId,
-            pushNotifications: { vapidPublicKey } = {
+            pushNotifications: { vapidPublicKey, appId } = {
               vapidPublicKey: undefined,
+              appId: undefined,
             },
```
jonsnyder, line 66: You can just do `pushNotifications: { vapidPublicKey, appId } = {},`, but really you never use vapidPublicKey and appId except to pass along to makeSendPushSubscriptionRequest, so you don't need to spell out all the fields. You could just do `pushNotifications,`

jonsnyder (approved):

PR #1368, opened 2025-09-23
Rename service worker commands to be more generic.

dompuiu:
## Description

Initially I wanted to generate a service worker specifically for push notifications. After talking with Jon, it seems we should just have one service worker for other potential features.

## Checklist:

### DIFF
--- a/scripts/alloyBuilder.js
+++ b/scripts/alloyBuilder.js
@@ -247,8 +247,8 @@ const getMakeBuildCommand = () => {
 };
 
 const getPushNotificationsServiceWorkerBuildCommand = () =>
-  new Command("build-pushnotifications-sw")
-    .description("Build the push notification service worker.")
+  new Command("build-sw")
+    .description("Build the s ervice worker.")
     .addOption(
       new Option("-m, --minify", "enable code minification").default(false),
     )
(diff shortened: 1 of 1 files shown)


scripts/alloyBuilder.js
```diff
@@ -247,8 +247,8 @@ const getMakeBuildCommand = () => {
 };
 
 const getPushNotificationsServiceWorkerBuildCommand = () =>
-  new Command("build-pushnotifications-sw")
-    .description("Build the push notification service worker.")
+  new Command("build-sw")
+    .description("Build the s ervice worker.")
```
carterworks, line 251:
```suggestion
    .description("Build the service worker.")
```

carterworks (changes requested):
If the service worker is to be more generic, I think we should put the entry point at the top level and import push notification functionality into it.

scripts/alloyBuilder.js
dompuiu: I know, I know. I was thinking similarly. But Jon pointed out that we can change the entrypoint later without affecting the commands. It is something we can do behind the scenes at a later point without breaking anything.

jonsnyder (approved):

PR #1449, opened 2026-02-24
Pass inbox items during subscribe ruleset callback

Spencer-Smith:
## Description

Allows for items with `inbox-item` schemas to be passed back through during subscribeRuleset callback

## Related Issue

[Jira](https://jira.corp.adobe.com/browse/PDCL-15026)

## Motivation and Context

Enable message inbox through AJO

Sandbox demo:
Screenshot 2026-02-25 at 3 22 35 PM
_New sandbox page provides buttons to send "messages"_

Screenshot 2026-02-25 at 3 22 48 PM

_Messages appear as toasts when first received/applied_

Screenshot 2026-02-25 at 3 23 00 PM

_After toast disappears (unless dismissed), inbox shows new messages_

Screenshot 2026-02-25 at 3 23 19 PM

_Opening inbox shows all messages received_

## Checklist:

- [x] I have added a Changeset (`pnpm changeset`) or it is not necessary because this PR is not consumer-facing.


packages/core/src/components/RulesEngine/createDecisionProvider.js
```diff
@@ -28,6 +29,8 @@ export default ({ eventRegistry }) => {
 
     if (evaluableRulesetPayload.isEvaluable) {
       payloadsBasedOnActivityId[activityId] = evaluableRulesetPayload;
+    } else if (Array.isArray(payload.items) && payload.items.length > 0) {
```
dompuiu: Can you explain what are you trying to check in this IF?

Spencer-Smith: Updated this to only add payloads with a `inbox-item`

packages/core/src/components/RulesEngine/createDecisionProvider.js
```diff
@@ -9,11 +9,17 @@ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTA
...
 import createEvaluableRulesetPayload from "./createEvaluableRulesetPayload.js";
 import { getActivityId } from "./utils/index.js";
 
+const hasInboxItem = (payload) =>
+  Array.isArray(payload.items) &&
+  payload.items.some((item) => item.schema === INBOX_ITEM);
+
 export default ({ eventRegistry }) => {
   const payloadsBasedOnActivityId = {};
+  const passthroughPayloads = [];
```
dompuiu: DecisionProvider is a singleton. Every time a sendEvent call will return an inbox item, that will be added in the passthroughPayloads. Every subsequent evaluate() call returns all previously accumulated inbox payloads from every prior response, not just the latest ones. This is probably not something we want.

But on a different thread, this is bypassing the evaluate step entirely. I am not sure if we want this. Can you explain how the data flows? I would think there is some place where we check the schemas of payloads and just let the content cards to be returned to the subscribeRulesetItems callback.

dompuiu (changes requested):

packages/core/src/components/RulesEngine/createDecisionProvider.js
Spencer-Smith: Updated now to remove changes in the decision provider, instead `createEvaluableRulesetPayload` allows payloads with `inbox-item`(s) to be evaluated (simply returns the inbox item(s)).

packages/core/src/components/RulesEngine/createOnResponseHandler.js
```diff
@@ -27,16 +36,20 @@ export default ({
   };
 
   return ({ response }) => {
-    decisionProvider.addPayloads(
-      response.getPayloadsByType(PERSONALIZATION_DECISIONS_HANDLE),
+    const personalizationPayloads = response.getPayloadsByType(
+      PERSONALIZATION_DECISIONS_HANDLE,
     );
+    decisionProvider.addPayloads(personalizationPayloads);
```
jonsnyder, line 42: We don't need to extract the inbox propositions here. This code is run inside the onResponse lifecycle event inside the RulesEngine component so it only needs to return the matching propositions. The regular personalization process with include the inbox proposition in the response already.

So there should be no changes to this file.

sandboxes/browser/src/components/MessageInboxDemo/MessageInbox.jsx
```diff
@@ -0,0 +1,448 @@
...
+  useEffect(() => {
+    if (!window.alloy) return;
+
+    const sub = window.alloy("subscribeRulesetItems", {
+      surfaces: [SURFACE],
+      schemas: [CONTENT_CARD_SCHEMA, INBOX_ITEM_SCHEMA],
+      callback: (result, collectEvent) => {
+        const { propositions = [] } = result;
+        const newCards = extractContentCards(propositions);
+        const configFromPayload = extractInboxConfig(propositions);
```
jonsnyder, line 248: So the inbox shouldn't be coming from here. it should come from the response on the sendEvent call.

sandboxes/browser/src/components/MessageInboxDemo/MessageInbox.jsx
```diff
@@ -0,0 +1,448 @@
...
+
+    unsubscribeRef.current = sub;
+
+    window.alloy("applyResponse", {
+      renderDecisions: true,
+      responseBody: INITIAL_INBOX_RESPONSE,
+      personalization: {
+        decisionContext: { "~type": "always" },
+      },
+    });
```
jonsnyder, line 279: add a ".then(" at the end of the applyResponse call to get the inbox proposition.

jonsnyder (changes requested):

packages/core/src/components/RulesEngine/createOnResponseHandler.js
Spencer-Smith: They are not ruleset payloads, though, so they don't get evaluated/extracted in the RulesEngine. That was one approach we considered (and in fact originally implemented), but there was a desire to move away from it.

Spencer-Smith: Oh wait I forgot that Serban had mentioned this strategy the other day. In that case, I'm not sure any code changes are necessary to the SDK at all? We can just close this MR, right?

dompuiu: I think we can close it. Sorry for making you work for nothing.

Spencer-Smith: I actually still got paid for working on this, so it wasn't for nothing

PR #1477, opened 2026-04-02
Create package-level components.json for custom build

carterworks:
## Description

- Added `components.json` manifests to `packages/core` and `packages/browser`
- Replaced fragile regex-based component discovery in `componentMetadata.js` (which parsed barrel file exports at build time using `/export\s+\{\s*default\s+as\s+([\w$]+)\s*\}/g`) with simple JSON reads
- Replaced `entryPointGeneratorBabelPlugin.js` (Babel AST transform for custom builds) with `generateEntryPoint.js` (plain string template)
- Replaced `versionBabelPlugin.js` with `@rollup/plugin-replace` for `__VERSION__` substitution
- Removed all `@babel/*` packages and `@rollup/plugin-babel` — including `@babel/core` which was shipping as a production dependency to npm consumers via `alloyBuilder.js`
- Added drift-detection tests that fail CI if a barrel file is edited without updating the corresponding manifest

## Related Issue

Motivated by the suggestion in adobe/alloy#1474 (comment): https://github.com/adobe/alloy/pull/1474#discussion_r3030116255

## Motivation and Context

The build system discovered components by regex-parsing JS barrel files, which is fragile and tightly coupled to export syntax. Babel was used for four things: (1) AST-rewriting `standalone.js` imports for custom builds, (2) `__VERSION__` string replacement, (3) `preset-env` transpilation, (4) template literal downleveling — all of which have lighter-weight replacements. The `npmEs5`/`npmEs6` Babel envs were verified dead code: no build script or CI step invokes them, and `libEs5`/`libEs6` directories are absent from the published `@adobe/alloy@2.32.0` tarball. The 2026 `defaults` browserslist targets (Chrome 109+, Firefox 128+, Safari 26.1+) all support ES6+ natively, so `preset-env` is no longer needed.

N/A

## Checklist:

- [x] I have added a Changeset (`pnpm changeset`) or it is not necessary because this PR is not consumer-facing.


packages/browser/components.json
```diff
@@ -0,0 +1,9 @@
+{
+  "required": [],
+  "optional": [
+    {
+      "name": "activityCollector",
+      "path": "./src/components/ActivityCollector/index.js"
```
dompuiu: Is the path used anywhere?

packages/browser/rollup.config.js
```diff
@@ -148,8 +150,12 @@ export const utilityExportBuilds = [
   },
 ];
 
-const buildPlugins = ({ variant, minify, babelPlugins }) => {
+const buildPlugins = ({ variant, minify }) => {
   const plugins = [
+    replace({
+      __VERSION__: version,
```
dompuiu, line 156: Did you confirm that __VERSION__ is replaced in this new paradigm?

packages/browser/src/standalone.js
```diff
@@ -35,6 +35,5 @@ const initializeStandalone = ({ components }) => {
   }
 };
 
-// If you change this line, check if the custom build script is still working.
-// You might need to change the babel plugin in scripts/helpers/entryPointGeneratorBabelPlugin.js.
+// Custom builds use scripts/helpers/generateEntryPoint.js to generate a modified version of this file.
```
dompuiu, line 16: This file might get changed and drift away from the template used in generateEntryPoint. Can we unify them somehow? To have one source of truth?

dompuiu:
Seems the bundle size increased since we removed babel. Probably preset-env was doing some optimizations that were making the final library smaller. Are we ok with increased size in the final library?

packages/browser/scripts/alloyBuilder.js
dompuiu: libES6 i think is no longer a thing since you removed babel.

packages/browser/scripts/alloyBuilder.js
```diff
@@ -116,16 +113,20 @@ const build = async (argv) => {
...
-  );
+    console.log(
+      `🎉 Wrote ${
+        path.isAbsolute(argv.outputDir)
+          ? rollupConfig.output[0].file
+          : path.relative(process.cwd(), rollupConfig.output[0].file)
+      } (${getFileSizeInKB(rollupConfig.output[0].file)}).`,
+    );
+  } finally {
+    fs.rmSync(inputFile, { force: true });
```
dompuiu, line 115: This was a problem before too but might be worth fixing it here. We should probably add `await bundle.close();` in the finally clause to release the internal file watchers.

carterworks:
> Seems the bundle size increased since we removed babel. Probably preset-env was doing some optimizations that were making the final library smaller. Are we ok with increased size in the final library?

it's a difference of 211 bytes minzipped. 

A Claude investigation says the increase came from removing the babel template literal plugin. `@babel/plugin-transform-template-literals` changed all template literals (e.g. `1 + 1 = ${answer}`) into string concatenations (`"1 + 1 = " + answer`). Template literals are slightly more verbose and less gzippable

```
The answer is precise now. The 154-byte increase in the minified bundle comes entirely from template literals that Babel   compiled away but Terser doesn't.

  Babel's @babel/plugin-transform-template-literals (loose mode) converted template literals to string concatenation:

  ┌────────────────────────────┬────────────────────────────┬─────┐
  │     Branch (no Babel)      │     Main (with Babel)      │  Δ  │
  ├────────────────────────────┼────────────────────────────┼─────┤
  │ `:scope ${t}` (13)         │ ":scope "+t (11)           │ +2  │
  ├────────────────────────────┼────────────────────────────┼─────┤
  │ `${e}` (6)                 │ ""+e (4)                   │ +2  │
  ├────────────────────────────┼────────────────────────────┼─────┤
  │ `${r}.${t}` (11)           │ r+"."+t (7)                │ +4  │
  ├────────────────────────────┼────────────────────────────┼─────┤
  │ `${n}[${i}]` (12)          │ n+"["+i+"]" (11)           │ +1  │
  ├────────────────────────────┼────────────────────────────┼─────┤
  │ `UUID byte range ...` (54) │ "UUID byte range "... (58) │ −4  │
  └────────────────────────────┴────────────────────────────┴─────┘

  - The branch minified has 139 template literals vs 2 in main (Babel compiled 137 away)
  - Avg overhead: ~1.1 bytes per template literal × 137 ≈ ~150 bytes, accounting for almost all 154 bytes
  - Some complex multi-expression templates actually get longer in concatenated form (like the UUID bounds check), partially   offsetting the total

  So the minified difference is: the ~10 characters of overhead per simple template literal (backticks + ${}) that Babel   eliminates by compiling to "literal "+var, but Terser leaves as-is. It's a minor cost of removing Babel, not a functional   concern.

✻ Brewed for 5m 23s · 1 shell still running

❯ that's 154 bytes. the other 45 bytes?

The gzipped minified went from 50,575 → 50,786, a +211 byte increase, but the uncompressed only increased by 154 bytes. So   gzip made things worse by 57 bytes on top of the raw content change.

  This happens because template literal syntax is less compressible than string concatenation. Gzip works by finding repeated   patterns in a sliding window. String concatenation like "text "+e+"more" reuses patterns that already appear dozens of   times throughout the bundle — gzip has good dictionary entries for " + + + identifier. Template literal syntax like `text   ${e}more` introduces the `${ and }`  delimiters, which are less common, resulting in longer Huffman codes and fewer   back-references.

  In other words: the 154 bytes of extra characters compress to 211 bytes because the `${}` pattern has higher entropy in the   context of this file than ""+ does.
```

Spencer-Smith (approved):

packages/browser/rollup.config.js
carterworks: Yes

packages/browser/src/standalone.js
carterworks: simplified as much as I can, and there is a test to make sure that the other files stay connected.

dompuiu (approved):

packages/core/components.json
```diff
@@ -0,0 +1,20 @@
...
+    { "name": "eventMerge" },
+    { "name": "mediaAnalyticsBridge" },
+    { "name": "personalization" },
+    { "name": "rulesEngine" },
+    { "name": "streamingMedia" },
+    { "name": "advertising" },
+    { "name": "pushNotifications" },
+    { "name": "brandConcierge" }
+  ]
+}
```
jonsnyder, line 20: Somewhere we keep track of default and deprecated flags on the components. (Maybe just in the extension?). It may make more sense to change this to:
```
{
  "components": [
    {
      "name": "context",
      "required": true,
    },
    {
      "name": "mediaAnalyticsBridge",
      "required": false,
      "default": false,
    },
  ]
}
```

Or even just an array at the top level, but that precludes including properties outside of components. Maybe we don't need any as we can get the name, description, etc from package.json.

PR #1485, opened 2026-04-23
Added @adobe/alloy-node package scaffold

carterworks:
## Description

Scaffolds a new `@adobe/alloy-node` workspace package as the eventual Node.js entrypoint for the Web SDK, plus a matching bare-bones sandbox at `sandboxes/node`. The package is marked **pre-alpha / non-functional**: importing it today throws because `@adobe/alloy-core` still references browser globals (`window`, `document`, `navigator`, etc.) at module scope. This PR is purely the starting point — no component moves, no service abstractions, no core refactors.

What's here:

- `packages/node/` — ESM package, `engines.node: ">=18"`, `createInstance(options)` forwarding to `createCustomInstance` with an empty components array.
- README + JSDoc warn it's non-functional and document the anticipated `(options, platformServices)` signature drift per `UNIVERSAL_JS_MIGRATION.md`.
- Ambient `alloy-core.d.ts` type shim + `jsconfig.json` with `paths` mapping so `tsc` doesn't walk into the real core sources (avoids hundreds of unrelated TS errors).
- `sandboxes/node/` — minimal consumer example mirroring the skipped integration suite (`configure` + `sendEvent` + log). Throws today; unblocks the moment core becomes import-safe.
- Node integration vitest project (`environment: "node"`, `pool: "threads"`) with a `describe.skip` consumer suite.
- `typecheck` added to the Alloy Dev GitHub workflow so regressions in the new `jsconfig.json` don't sneak through.

## Related Issue

N/A — scaffolding work; follows the "Node SDK package" future-phase item in `packages/browser/UNIVERSAL_JS_MIGRATION.md`.

## Motivation and Context

The Universal JS migration plan calls out a Node SDK package as a future phase blocked on core portability + async storage. Landing the empty package + sandbox now gives the migration a concrete consumer target to validate against (and a first working smoke test the moment core stops touching `window` at module scope). It also forces us to stake out the public shape (`createInstance(options)`) and the test harness before the substantive work starts.

N/A — no runtime behavior yet.

## Checklist:

- [x] I have run the Sandbox successfully. (The new Node sandbox throws at import by design — same blocker as the skipped integration tests, documented in its README.)
- [x] I have added a Changeset (`pnpm changeset`) or it is not necessary because this PR is not consumer-facing. (Package is `"private": true` and pre-alpha; no consumer-facing change.)

### DIFF
--- a/.github/workflows/quality-checks.yml
+++ b/.github/workflows/quality-checks.yml
@@ -16,6 +16,7 @@ jobs:
           - build
           - lint
           - "format:check"
+          - typecheck
     steps:
       - uses: actions/checkout@v4
       - uses: pnpm/action-setup@v4
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -28,7 +28,7 @@ import license from "./scripts/eslint/licenseRule.js";
 const allComponentPaths = glob.sync("packages/core/src/components/*/");
 
 const sharedIgnores = [
-  "sandboxes/**",
+  "sandboxes/browser/**",
   "dist/**",
   "distTest/**",
   "packages/**/dist/**",
@@ -44,7 +44,7 @@ export default defineConfig([
   pluginJs.configs.recommended,
   eslintPluginPrettierRecommended,
   globalIgnores([
-    "sandboxes/**",
+    "sandboxes/browser/**",
     "node_modules/",
     "launch*.js",
     "packages/reactor-extension/dist/**",
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -349,6 +349,21 @@ export default defineConfig([
     },
   },
   {
+    name: "alloy/node-sandbox",
+    files: ["sandboxes/node/src/**/*.{cjs,js,mjs}"],
+    settings: {
+      "import/core-modules": ["@adobe/alloy-node"],
+    },
+    languageOptions: {
+      globals: {
+        ...globals.node,
+      },
+    },
+    rules: {
+      "no-console": "off",
+    },
+  },
+  {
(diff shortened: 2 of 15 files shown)


Spencer-Smith (approved):
I don't love the package directory being called `node` as that seems a little confusing, but I don't have a better idea right now

carterworks:
> I don't love the package directory being called `node` as that seems a little confusing, but I don't have a better idea right now

some options

* `server`
* `node-server`
* `backend`

Spencer-Smith:
`server` is probably most accurate and does feel like it has some symmetry with the `browser` package.

Edit: that said, I don't usually like naming things for their use case, it still sounds like this is the alloy package's OWN server, which is obviously incorrect. Maybe `server-sdk` would be more accurate. Then again, I'm trying to read these packages as though I have NO knowledge whatsoever of alloy, and that might be a little too naive, it might be reasonable to assume a person is reading them in the context of being in this repo

carterworks:
I like `server-sdk`. Would you name the package `@adobe/alloy-server`?

Spencer-Smith:
Yeah I think that makes sense. I also realize I'm thinking backwards, probably naming something for its use case is better than naming it for its implementation details. I'm not actually sure how well that holds in this realm, because realistically were we to support a future implementation of alloy that is usable for servers, the distinction would probably be the language/framework it was implemented in.

Naming is hard 🤷

carterworks:
I mean, Digant does want to add a python server sdk, so it is not out of the question...

carterworks:
i'm gonna keep `@adobe/alloy-node`

PR #1490, opened 2026-04-30
Run script tests on CI.

dompuiu:
## Changed Packages

- [ ] core
- [ ] reactor-extension 

## Description

Running script tests on every commit is not optimal IMHO. I moved them to run only before pushing to github.

## Types of changes

## Checklist:


.changeset/short-radios-throw.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
carterworks: a changeset is unnecessary, since the change does not need to be published as a beta.

.husky/pre-push
```diff
@@ -0,0 +1 @@
+pnpm vitest run --project=unit --project=scripts
```
carterworks: what do you think about also running the integration tests (aka all the tests)?

carterworks: in the future, we can do more complex handling around running only the tests for things that have changed, but I think that means adding a dependency like turbopack or nx and we are not there yet.

carterworks (changes requested):

.changeset/short-radios-throw.md
dompuiu: In the PR template we have this:
```<!--- If your changes are not consumer facing, use `pnpm changeset --empty` to add an empty changeset file. -->```. 

Should we remove that from the template?

.husky/pre-push
dompuiu: Integration tests are too slow for my taste. And they are already run on the CI. I would prefer to move even the scripts tests to be run on CI if that would be possible. If you really want to add it there, I am ok with that. I will start doing pushes with `--no-verify`.

carterworks: 👿

my goal is to get everyone to run as many (good) tests as possible. if you are going to ignore then AND they pass on CI, then let's just run them on CI.

Will you modify `scripts/extensionBuild.spec.js` and `scripts/customBuild.spec.js` and enable them to run on CI?

.changeset/short-radios-throw.md
carterworks: Yes

* Do you need to trigger a release?
  * YES—Do you need a note in the customer-facing changelog?
    * YES—`pnpm changeset`
    * NO—`pnpm changeset --empty`
  * NO—Push without a changeset

.husky/pre-push
dompuiu: I changed the tests.

carterworks (approved):

PR #1491, opened 2026-04-30
Do not send advertising data when advertising is unconfigured

adhir111:
## Changed Packages
- [x] core
- [ ] reactor-extension

## Description

Adds an early-exit guard to `createOnBeforeSendEventHandler` so that advertising
identity resolution (Surfer ID, ID5, RampID) is **skipped** when no enabled
`advertiserSettings` are configured. Without this guard, every customer
`sendEvent` was loading third-party scripts and injecting iframes even though
there was nothing meaningful to attach to the outgoing event downstream.

The new gate uses the existing `normalizeAdvertiser` helper from
`utils/helpers.js` and is wired through the factory's dependency injection, so:

- `componentConfig.advertiserSettings` is `undefined`, `[]`, or has all entries
  with `enabled: false` → handler returns immediately.
- At least one `{ advertiserId, enabled: true }` is present → handler proceeds
  exactly as before.

This mirrors the equivalent gate already present in `sendAdConversion.js` for
the view-through code path, so both surfaces now have consistent behavior.

### Files changed
- `packages/core/src/components/Advertising/handlers/createOnBeforeSendEventHandler.js`
  — added `normalizeAdvertiser` to the destructured deps + new `!activeAdvertiserIds`
  short-circuit in the early-return condition.
- `packages/core/src/components/Advertising/index.js` — imports
  `normalizeAdvertiser` from `./utils/helpers.js` and passes it into the factory.
- `packages/core/test/unit/specs/components/Advertising/handlers/createOnBeforeSendEventHandler.spec.js`
  — passes a `normalizeAdvertiser` mock through DI; adds 4 new test cases for
  the gate (missing settings / empty array / all-disabled / at least one
  enabled).

## Related Issue

[
](https://jira.corp.adobe.com/browse/AMO-181188)

## Motivation and Context

When Alloy is configured with the Advertising component but no enabled
`advertiserSettings` (for example, the customer has enabled the component but
not yet configured advertisers, or has disabled all of them), every
`sendEvent` was triggering:

- A hidden iframe to `pixel.everestjs.net` for Surfer ID resolution.
- Loading the `id5-api.js` script for ID5.
- Loading the configured ATS script for RampID.

These calls produced no useful data because `appendAdvertisingIdQueryToEvent`
relies on `advertiserSettings` to populate the `advIds` field, and there are no
advertisers to stitch identities to. The result was wasted network requests,
unnecessary third-party script execution, additional latency on the
`sendEvent` lifecycle, and noise in the consent / cookie surface for users
who effectively aren't using the advertising features.

This PR aligns the `onBeforeEvent` lifecycle with the already-shipped behavior
of the view-through conversion path in `sendAdConversion.js`, where the same
gate (`activeAdvertiserIds`) is already used to skip work when no advertiser
is configured.

N/A — code-only change. Verified in the sandbox at `/advertising` with both
`?advertiserSettings=on` and `?advertiserSettings=off`:
- **off** → no `[Advertising]` log lines, no third-party network requests.
- **on**  → identity collectors run as before.

## Documentation

No consumer-facing documentation update required. The new gate is a no-op for
any customer that has properly configured `advertiserSettings` (which is the
documented usage); only customers in the "Advertising component enabled but
no advertisers configured" state are affected, and the new behavior is the
correct one for that state.

## Types of changes

## Checklist:

- [x ] I have signed the [Adobe Open Source CLA](https://opensource.adobe.com/cla.html) or I'm an Adobe employee.
      - Unit: `core/unit` → `createOnBeforeSendEventHandler.spec.js` 12/12 passing.
      - Integration: `browser/integration` → all 7 `Advertising/*.spec.js` files, 12/12 passing.
      - Verified at `http://localhost:3000/advertising` with both
        `?advertiserSettings=on` and `?advertiserSettings=off`; no third-party
        network calls in the "off" case.

### DIFF
--- a/packages/core/src/components/Advertising/handlers/createOnBeforeSendEventHandler.js
+++ b/packages/core/src/components/Advertising/handlers/createOnBeforeSendEventHandler.js
@@ -39,6 +39,7 @@ const waitForAdvertisingId = (advertising) => {
  * @param {Function} dependencies.appendAdvertisingIdQueryToEvent
  * @param {Function} dependencies.getUrlParams
  * @param {Function} dependencies.isThrottled
+ * @param {Function} dependencies.normalizeAdvertiser
  */
 export default ({
   cookieManager,
@@ -52,6 +53,7 @@ export default ({
   appendAdvertisingIdQueryToEvent,
   getUrlParams,
   isThrottled,
+  normalizeAdvertiser,
 }) => {
   /**
    * Appends advertising identity IDs to AEP event query if not already added.
@@ -68,9 +70,15 @@ export default ({
     }
     const { skwcid, efid } = getUrlParams();
     const isClickThru = !!(skwcid && efid);
+    // Skip identity resolution when no enabled advertiser IDs are configured —
+    // there is nothing meaningful to enrich the event with downstream.
+    const activeAdvertiserIds = normalizeAdvertiser(
+      componentConfig?.advertiserSettings,
+    );
     if (
       isAdvertisingDisabled(advertising) ||
       isClickThru ||
+      !activeAdvertiserIds ||
       (isThrottled(SURFER_ID, cookieManager) &&
         isThrottled(ID5_ID, cookieManager) &&
         isThrottled(RAMP_ID, cookieManager))
--- a/packages/core/src/components/Advertising/handlers/createOnBeforeSendEventHandler.js
+++ b/packages/core/src/components/Advertising/handlers/createOnBeforeSendEventHandler.js
@@ -75,10 +75,11 @@ export default ({
     const activeAdvertiserIds = normalizeAdvertiser(
       componentConfig?.advertiserSettings,
     );
+    const hasAdvertiserIds = activeAdvertiserIds.length > 0;
     if (
       isAdvertisingDisabled(advertising) ||
       isClickThru ||
-      !activeAdvertiserIds ||
+      !hasAdvertiserIds ||
       (isThrottled(SURFER_ID, cookieManager) &&
         isThrottled(ID5_ID, cookieManager) &&
         isThrottled(RAMP_ID, cookieManager))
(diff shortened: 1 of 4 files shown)


packages/core/src/components/Advertising/handlers/createOnBeforeSendEventHandler.js
```diff
@@ -68,9 +70,15 @@ export default ({
...
     const isClickThru = !!(skwcid && efid);
+    // Skip identity resolution when no enabled advertiser IDs are configured —
+    // there is nothing meaningful to enrich the event with downstream.
+    const activeAdvertiserIds = normalizeAdvertiser(
+      componentConfig?.advertiserSettings,
+    );
     if (
       isAdvertisingDisabled(advertising) ||
       isClickThru ||
+      !activeAdvertiserIds ||
```
carterworks: don't assume that UNKNOWN_ADVERTISER will always be `""`.

add an intermediate variable, something like 

```js
const hasAdvertisers = activeAdvertisersIds.length > 0;
```

or

```js
const hasAdvertisers = !activeAdvertisersIds?.includes(UNKNOWN_ADVERTISer);
```

```suggestion
      hasAdvertiserIds
```

packages/core/test/unit/specs/components/Advertising/handlers/createOnBeforeSendEventHandler.spec.js
```diff
@@ -209,6 +215,67 @@ describe("Advertising::createOnBeforeSendEventHandler", () => {
...
+
+      await onBeforeEvent({
+        event,
+        advertising: { handleAdvertisingData: "auto" },
+      });
+
+      expect(collectSurferIdFn).not.toHaveBeenCalled();
+      expect(getID5IdFn).not.toHaveBeenCalled();
+      expect(getRampIdFn).not.toHaveBeenCalled();
+      expect(event.mergeQuery).not.toHaveBeenCalled();
```
carterworks, line 291: these three tests are identical, but based on the test name I believe they should be different.

carterworks (changes requested):
two very minor changes. fix those and we are good to go.

carterworks (approved):

PR #1486, opened 2026-04-28
Migrated browser components

Spencer-Smith:
## Changed Packages

- [x] core
- [x] browser
- [ ] reactor-extension 

## Description

Moves browser-specific components (Personalization, Context, Advertising, PushNotifications, DOM utils, etc.) from the core package to browser

## Related Issue

N/A

## Motivation and Context

For universal JS, we need to trim the core package down to only functionality shared between the various implementations (browser, node).

## Screenshots (if appropriate):

N/A

## Documentation

N/A (No user-facing changes)

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Improvement (non-breaking change which does not add functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have added a Changeset file with a consumer-facing description of my changes.
- [x] I have signed the [Adobe Open Source CLA](https://opensource.adobe.com/cla.html) or I'm an Adobe employee.
- [x] I have made any necessary test changes and all tests pass.
- [x] I have run the Sandbox successfully.

PR #1500, opened 2026-05-07
Fixed Context component being excluded from createCustomInstance in browser package

carterworks:
## Changed Packages
- [x] browser

## Description

After moving the Context component to the browser package (PRs [#1494](https://github.com/adobe/alloy/pull/1494) and [#1486](https://github.com/adobe/alloy/pull/1486)), the tags extension builds without it because the Web SDK doesn’t force required components-from-browser-package to be built.

Tags [always](https://github.com/adobe/alloy/blob/main/packages/reactor-extension/src/lib/instanceManager/injectWrapOnBeforeEventSend.js) sets content.xdm.implementationDetails.name = `${content.xdm.implementationDetails.name}/reactor` but because the Context-created `context.implementationDetails` object isn’t there, *every* `sendEvent` fails.

## Related Issue

[PDCL-15715](https://jira.corp.adobe.com/browse/PDCL-15715)

## Motivation and Context

Without this change, the `Context` component (and other required components) could be excluded when consumers used `createCustomInstance` directly, leading to events being dropped or missing required data in some environments (notably AdCloud).

## Types of changes

## Checklist:


dompuiu (approved):

jonsnyder:
What can we change to our automated tests to catch failures like this in the future?

carterworks:
> @jonsnyder:
> What can we change to our automated tests to catch failures like this in the future?

The on-page, alloy wrapper part of the extension is under-covered by our integration test suite. Something there that sends an event would have caught it.

PR #1503, opened 2026-05-11
FIX: Updated data element lookup for stale DE references after copy

Spencer-Smith:
## Changed Packages

- [ ] core
- [x] reactor-extension 

## Description

Fixes two bugs:
1. During data element lookup by name, uses same utility function as the lookup that checks whether to automatically fill the data element, which stops searching after finding two data elements, which limits our replacement in scenarios where the property has more than two data elements.
2. Due to logic for resolving the stale reference, it previously required two open/saves to resolve the stale DE reference to point to the same-named DE within the new property; the first lookup saved the data element name to the settings and the second looks up the correct DE ID by that name. This has been updated to check for the name and update the ID on the first load/save.

Unclear to me if the second issue is a regression (did not notice it in my initial testing when introducing the feature), but the first bug seems to have been there the whole time and never detected because my tests were on properties with exactly two data elements.

## Related Issue

N/A

## Motivation and Context

Reported issue by client that the previous solution was not (no longer?) working.

N/A

## Documentation

N/A

## Types of changes

## Checklist:


.changeset/social-coins-clean.md
```diff
@@ -0,0 +1,6 @@
+---
+"reactor-extension-alloy": patch
+"adobe-alloy-monorepo": patch
```
carterworks: why are you bumping the version of the monorepo

packages/reactor-extension/src/view/utils/fetchDataElements.js
```diff
@@ -15,17 +15,28 @@ import UserReportableError from "../errors/userReportableError";
...
+/**
+ * Fetches variable data elements for a property, one page per call.
+ *
+ * Pass `minResults` to keep fetching pages until that many variable data
+ * elements have accumulated — useful when you need to know whether 0, 1,
+ * or 2+ variable data elements exist on the property without loading all of
+ * them.
+ *
+ * To search by exact name across pages, use `fetchDataElementByName` instead.
+ */
```
carterworks, line 36: (nit) some jsdoc types would be nice https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html

carterworks (approved):

.changeset/social-coins-clean.md
Spencer-Smith: Ah I thought that some of the updated utils were outside the extension package but not so. NVM then

PR #1504, opened 2026-05-12
Adding region configuration to Brand Concierge component

ninaceban:
## Changed Packages

- [ ] core
- [ ] reactor-extension 

## Description

  Added a region configuration option to the Brand Concierge component. When set, the region routes conversation requests to a specific Adobe edge node.

  Core (@adobe/alloy-core):
  - Added region to the conversation config schema, validated with string().matches(/^[a-z]{2,4}[0-9]{1,2}$/i) (e.g. va7, or2, irl1). The field is optional.
  - Removed the single-purpose regionValidator.js in favour of the existing string().matches() chain.
  - Fixed an unused edgeBasePath variable in createSendConversationEvent.js.

  Tags extension (reactor-extension-alloy):
  - Added a Region text field to the Brand Concierge configuration section in the Tags UI.
  - Validation in the Tags extension mirrors the SDK: accepts va7-style strings, rejects non-matching input with a user-friendly message.

  Motivation and Context

  Brand Concierge needs to support region-specific deployments. Without a configurable region, all conversation traffic goes to the default edge node, which
   does not work for region-restricted rollouts.

Testing Steps

  1. SDK — valid region accepted:
  Configure Alloy with brandConcierge: { conversation: { region: "va7" } } and confirm no validation error is thrown on startup.
  2. SDK — invalid region rejected:
  Configure with region: "invalid-region" and confirm the SDK throws a validation error describing the expected format.
  3. SDK — region omitted (default behaviour unchanged):
  Configure without a region field and confirm conversation events fire normally with no regression.
  4. Tags UI — field renders:
  Open the Brand Concierge configuration section with the component enabled. Confirm the Region field appears above the other settings.
  5. Tags UI — valid region saves:
  Enter va7 in the Region field, save the configuration, and reload — confirm va7 is preserved.
  6. Tags UI — invalid region shows error:
  Enter invalid-region in the Region field and save. Confirm the inline error "Please enter a valid region (e.g. va7, or2, irl1)." appears and the form does
   not save.
  7. Tags UI — empty region not written to settings:
  Leave the Region field blank and save. Confirm conversation.region is absent from the saved settings object.

## Types of changes

## Checklist:


packages/core/src/components/BrandConcierge/configValidators.js
```diff
@@ -20,6 +20,7 @@ export default objectOf({
       .minimum(STREAM_START_TIMEOUT_MS)
       .default(STREAM_START_TIMEOUT_MS),
     collectSources: boolean().default(false),
+    region: string().region(),
```
carterworks: We already have a regex-based `string.matches()` validator that can be used.

```suggestion
    region: string().matches(/^[a-z]{2,4}[0-9]{1,2}$/i),),
```

Then deleted `packages/core/src/utils/validation/regionValidator.js`.

carterworks (changes requested):
* Use the `string().matches(/regex/)` validator instead of creating a new single-purpose region validator.
* Address linting and formatting errors
* Run `pnpm changeset` and write the changelog entry
* Fill out the PR description with some testing steps
* Will this require a new configuration in the Tags extension? If so, you can add it in this branch to `packages/reactor-extension/src/view/configuration/brandConciergeSection.jsx` (we recently merged the Tag extension repo and this repo).
* Update this branch with `main` before merging.

packages/core/src/components/BrandConcierge/configValidators.js
ninaceban: thank you Carter, I'll add your suggestions and the Tags changes.

packages/core/src/components/BrandConcierge/createBuildEndpointUrl.js
```diff
@@ -11,13 +11,14 @@ governing permissions and limitations under the License.
...
 
 export default ({ queryString }) => {
-  return ({ edgeDomain, request, datastreamId }) => {
+  return ({ edgeDomain, region, request, datastreamId }) => {
     const params = request.getRequestParams();
     const configId = request.getDatastreamIdOverride() || datastreamId;
     params.requestId = request.getId();
     params.configId = configId;
     const stringifiedRequestParams = queryString.stringify({ ...params });
+    const regionSegment = region ? `/${region}` : "";
```
jonsnyder, line 20: I'm assuming conversations should be sticky to a region. Consider the case where you are not specifying a region in the config. If a user moves internet connections, the session should still be sticky to the same region. I'm thinking we should use the configured region first, and if that is not set, use the cluster cookie.

carterworks:
You can safely ignore the failing "Generate Coverage Report and SonarQube Analysis" jobs - that test has been flakey and @Spencer-Smith is looking at it right now

packages/core/src/components/BrandConcierge/createBuildEndpointUrl.js
ninaceban: We will have customers deployed to regions so that the traffic is routed only to that region irrespective of the end user location.

carterworks (approved):

dompuiu (approved):

PR #1510, opened 2026-05-14
Fix the region validation error for extension

ninaceban:
## Changed Packages

- [ ] core
- [ ] reactor-extension 

## Types of changes

## Checklist:


carterworks (changes requested):
What exactly was the issue? "" was failing the regex validation, or conversation: {} was failing the schema validation?

Also, run `pnpm changeset --empty` and add the resulting file.

ninaceban:
> What exactly was the issue? "" was failing the regex validation, or conversation: {} was failing the schema validation?
> 
> Also, run `pnpm changeset --empty` and add the resulting file.

`string().matches(regex)` validates empty strings — "" doesn't match the pattern so it throws. Using `test` would only apply the regex when something has actually been entered.

carterworks (approved):

PR #1522, opened 2026-06-03
Adding munchkin cookie to state payload

ninaceban:
The munchkin cookie is created by the Marketo SDK and used for integrations with Marketo like services: 
- meeting booking
- live chat feature
- AEM forms integration
We have defined a way in BC runtime to standardize the cookie value propagation down to the agents so that it aligns with Konductor strategy to share relevant cookies with the subscribed Upstreams. 
Initialy the munchkin cookie was read and added to payload in Web Client, while Web SDK has the right knowledge of how to read and share cookies with the Edge Network and BC runtime.
My proposal around **Uniform Cookie-Sharing Protocol for Brand Concierge Agents**: https://wiki.corp.adobe.com/spaces/brandconcierge/pages/3973993839/Decision+Record+Uniform+Cookie-Sharing+Protocol+for+Brand+Concierge+Agents

## Changed Packages

- [ ] core
- [ ] reactor-extension 

## Types of changes

## Checklist:

### DIFF
--- a/packages/core/src/components/BrandConcierge/constants.js
+++ b/packages/core/src/components/BrandConcierge/constants.js
@@ -13,4 +13,4 @@ export const BC_SESSION_COOKIE_NAME = "bc_session_id";
 export const STREAM_START_TIMEOUT_MS = 10000;
 export const SOURCES_QUERY_PARAM = "adobe_brand_concierge_source";
 // This is the cookie that marketo js sets and we need it for Marketo integrations
-export const MUNCHKIN_COOKIE_NAME = "munchkin"; 
+export const MUNCHKIN_COOKIE_NAME = "munchkin";
--- a/packages/core/src/components/BrandConcierge/constants.js
+++ b/packages/core/src/components/BrandConcierge/constants.js
@@ -12,4 +12,5 @@ governing permissions and limitations under the License.
 export const BC_SESSION_COOKIE_NAME = "bc_session_id";
 export const STREAM_START_TIMEOUT_MS = 10000;
 export const SOURCES_QUERY_PARAM = "adobe_brand_concierge_source";
-export const MUNCHKIN_COOKIE_NAME = "munchkin";
+// This is the cookie that marketo js sets and we need it for Marketo integrations
+export const MUNCHKIN_COOKIE_NAME = "munchkin"; 
--- a/packages/core/src/components/BrandConcierge/constants.js
+++ b/packages/core/src/components/BrandConcierge/constants.js
@@ -12,3 +12,4 @@ governing permissions and limitations under the License.
 export const BC_SESSION_COOKIE_NAME = "bc_session_id";
 export const STREAM_START_TIMEOUT_MS = 10000;
 export const SOURCES_QUERY_PARAM = "adobe_brand_concierge_source";
+export const MUNCHKIN_COOKIE_NAME = "munchkin";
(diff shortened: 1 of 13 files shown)


packages/core/src/core/createCookieTransfer.js
```diff
@@ -33,24 +33,31 @@ export default ({
         cookiesEnabled: true,
       };
 
+      const cookies = cookieJar.get();
+      const entries = [];
```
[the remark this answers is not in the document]
jonsnyder: Yes, the munchkin cookie should only be included if endpoint is third party.

packages/core/src/components/BrandConcierge/constants.js
```diff
@@ -12,3 +12,4 @@ governing permissions and limitations under the License.
 export const BC_SESSION_COOKIE_NAME = "bc_session_id";
 export const STREAM_START_TIMEOUT_MS = 10000;
 export const SOURCES_QUERY_PARAM = "adobe_brand_concierge_source";
+export const MUNCHKIN_COOKIE_NAME = "munchkin";
```
jonsnyder: Please include a comment here to explain what product uses the munchkin cookie.

packages/core/src/core/createCookieTransfer.js
```diff
@@ -36,18 +39,26 @@ export default ({
       // If the endpoint is first-party, there's no need to transfer cookies
       // to the payload since they'll be automatically passed through cookie
       // headers.
-      if (!isEndpointFirstParty) {
+      if (!isEndpointFirstParty || isNonEmptyArray(extraCookieNames)) {
```
jonsnyder, line 42: Why include the munchkin cookie in the payload for first party collection endpoints? The munchkin cookie will already be included in the request headers.

jonsnyder: Nina said backend only reads from payload. It's okay to do it this way, but the backend should change at some point.

packages/core/src/core/createCookieTransfer.js
```diff
@@ -33,24 +33,31 @@ export default ({
...
           });
+      }
 
-        if (entries.length) {
-          state.entries = entries;
+      extraCookieNames.forEach((name) => {
+        if (cookies[name] !== undefined) {
+          entries.push({ key: name, value: cookies[name] });
         }
+      });
```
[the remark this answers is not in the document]
ninaceban: will include the deduplication logic if we make this configurable and will make sense, for now we only add the extra cookies from BC component, we are not allowing the customer to provide that

carterworks (approved):

packages/core/src/components/BrandConcierge/createSendConversationEvent.js
```diff
@@ -103,9 +104,11 @@ export default ({
         }
 
         payload.addEvent(event);
-        if (config.conversation.stickyConversationSession === true) {
-          cookieTransfer.cookiesToPayload(payload, edgeDomain);
-        }
+        cookieTransfer.cookiesToPayload(
```
carterworks, line 107: The old code called `cookiesToPayload` only when `stickyConversationSession === true`; it now calls it unconditionally. For a third-party edge domain, this means the standard qualifying identity cookies (`kndctr_*`, `mbox`, `at_qa`) are now transferred into the conversation payload for non-sticky sessions too — previously those sessions transferred zero cookies. This is broader than the stated intent (forward munchkin regardless of sticky; the decision-record payload example shows only munchkin). It's harmless when BC's edge domain is first-party (standard-cookie loop is skipped, only munchkin is added), but Adobe default edge domains are third-party.

I just want to make sure this is intentional

carterworks (changes requested):

carterworks (approved):

jonsnyder (approved):
Thanks for adding the munchkin cookie globally too.

PR #1524, opened 2026-06-04
UJS: Refactored core package to be platform agnostic

Spencer-Smith:
## Changed Packages

- [x] core
- [ ] reactor-extension 

## Description

Moved platform-specific implementation details (ie 'window' and 'document' that are specific to browser implementation, as well as upstream dependencies that use these artifacts) to the browser's platform services. C

## Related Issue

(Related to Universal JavaScript migration)

## Motivation and Context

Node package for WebSDK will have failures when encountering references to browser resources.

N/A

## Documentation

(All UJS should be documented together)

## Types of changes

## Checklist:


.changeset/vast-jeans-walk.md
```diff
@@ -0,0 +1,8 @@
+---
+"@adobe/alloy": patch
+"@adobe/alloy-core": patch
```
carterworks, line 3: this is pretty big. should be at least a minor version patch.

packages/browser/src/services/createBrowserGlobalsService.js
```diff
@@ -23,8 +23,20 @@ const createBrowserGlobalsService = () => ({
...
+  isPageSsl: () => window.location.protocol === "https:",
+  getWindowContext: () => ({
+    title: document.title,
+    url: window.location.href,
+    referrer: document.referrer,
+    height: window.innerHeight,
+    width: window.innerWidth,
+    scrollY: window.scrollY,
+    scrollX: window.scrollX,
+  }),
```
carterworks, line 41: This abstraction is fine for now, but I suspect that it may need to change when we start implementing nodejs clients - as is, it is quite leaky. What would be the implementation of these in the node sdk? ya can't do it.

packages/browser/src/services/createBrowserNetworkService.js
```diff
@@ -13,28 +13,22 @@ governing permissions and limitations under the License.
...
 /**
- * @param {{ logger: import('@adobe/alloy-core/core/types.js').Logger }} dependencies
  * @returns {NetworkService}
  */
-const createBrowserNetworkService = ({ logger }) => {
+const createBrowserNetworkService = () => {
   const { fetch, navigator } = window;
   const sendFetchRequest = injectSendFetchRequest({ fetch });
-  const sendBeaconRequest =
+  // bind() avoids "illegal invocation" when sendBeacon loses its receiver.
```
carterworks: nit: unnecessary comment

packages/browser/src/services/createBrowserNetworkService.js
```diff
@@ -13,28 +13,22 @@ governing permissions and limitations under the License.
...
+  const sendBeacon =
     typeof navigator.sendBeacon === "function"
-      ? injectSendBeaconRequest({
-          // bind() avoids "illegal invocation" when sendBeacon loses its receiver.
-          sendBeacon: navigator.sendBeacon.bind(navigator),
-          sendFetchRequest,
-          logger,
-        })
-      : sendFetchRequest;
+      ? navigator.sendBeacon.bind(navigator)
```
carterworks: this needs to be `injectSendBeacon` - they are not drop in replaceable.

carterworks: `sendBeacon` should fall back to fetch, it should accept just a url and a json body, it should return a promise - native `navigator.sendBeacon` does none of these.

packages/browser/test/unit/specs/services/createBrowserNetworkService.spec.js
```diff
@@ -24,21 +24,21 @@ afterEach(() => {
...
-    const network = createBrowserNetworkService({ logger: createLogger() });
+  it("exposes sendFetchRequest as a function and sendBeacon when available", () => {
+    const network = createBrowserNetworkService();
     expect(typeof network.sendFetchRequest).toBe("function");
-    expect(typeof network.sendBeaconRequest).toBe("function");
+    expect(typeof network.sendBeacon).toBe("function");
   });
 
-  it("falls back to sendFetchRequest when navigator.sendBeacon is unavailable", () => {
+  it("sets sendBeacon to null when navigator.sendBeacon is unavailable", () => {
```
carterworks: 👎

packages/browser/src/services/createBrowserGlobalsService.js
carterworks: But, I'm fine keeping it and have the eventual shape be moulded by our real needs. I thought about it a lot previously and I was never able to guess a good abstraction without any data. We just need to be flexible in the future. Otherwise, this interface will not work.

packages/core/src/components/RulesEngine/createEventRegistry.js
```diff
@@ -61,7 +64,7 @@ export default ({ storage, logger }) => {
 
     save = createSaveStorage(currentStorage, EVENT_HISTORY_STORAGE_KEY);
 
-    [eventRegistry, eventRegistrySize] = restore({});
+    [eventRegistry, eventRegistrySize] = await restore({});
```
carterworks, line 65: **Race window between `save` reassignment and `await restore`.** `save` is reassigned to write to `currentStorage` on line 65, but `eventRegistry` isn't replaced until this `await restore({})` resolves. Any `addEvent` firing in that microtask gap appends to the stale in-memory `eventRegistry`, persists it via the already-updated `save`, and is then clobbered when `restore` completes and overwrites `eventRegistry`. Move `save = createSaveStorage(...)` to *after* the `await`, or snapshot/merge in-flight events. Benign with today's synchronous browser storage, but a latent bug for the async backends this refactor enables.

---
🤖 _Drafted by Claude Code during review assistance._

packages/core/src/components/RulesEngine/utils/storage.js
```diff
@@ -11,9 +11,9 @@ governing permissions and limitations under the License.
 */
 
 export const createRestoreStorage = (storage, storageKey) => {
-  return (defaultValue) => {
+  return async (defaultValue) => {
     try {
-      const stored = storage.getItem(storageKey);
+      const stored = await storage.getItem(storageKey);
```
carterworks, line 16: **Fragile null handling.** When `getItem` returns `null`, `JSON.parse(null)` yields `null` (no throw) and the code only falls through to defaults because `stored.length` then throws on `null`. It works by accident. Prefer an explicit guard: `if (stored == null) return [defaultValue, 0];`. Also note the surrounding `catch {}` now masks genuine parse/storage errors, not just the no-data-yet case.

---
🤖 _Drafted by Claude Code during review assistance._

packages/core/src/core/network/injectSendNetworkRequest.js
```diff
@@ -11,14 +11,18 @@ governing permissions and limitations under the License.
 */
 
 import { stackError } from "../../utils/index.js";
+import injectSendBeaconRequest from "./requestMethods/injectSendBeaconRequest.js";
```
carterworks: IMO, `injectSendBeaconRequest` and its ilk should live in browser. As-is, it's a wrapper of a wrapper and is pretty messy.

packages/core/src/core/index.js
```diff
@@ -71,23 +68,17 @@ export const createExecuteCommand = ({
...
-  const createNamespacedStorage = injectStorage(window);
+  const platformServices = createPlatformServices();
 
   const logController = createLogController({
     console: globalThis.console,
     createLogger,
     instanceName,
-    createNamespacedStorage,
-    getMonitors: getMonitors.bind(null, monitors),
+    getMonitors: () => [...platformServices.globals.getMonitors(), ...monitors],
```
carterworks: why keep two levels of monitors? ideally, it would be cleaner just o have one.

packages/core/src/utils/request/createGetAssuranceValidationTokenParams.js
```diff
@@ -13,28 +13,19 @@ governing permissions and limitations under the License.
...
-  let clientId = storage.persistent.getItem(CLIENT_ID);
-  if (!clientId) {
-    clientId = uuid();
-    storage.persistent.setItem(CLIENT_ID, clientId);
-  }
-  return clientId;
-};
+export default ({ getLocationSearch }) => {
+  // Generated once per instance — Assurance is a browser debugging tool,
+  // so a fresh ID per instance startup is sufficient.
```
carterworks: Disagree with this comment - it would be useful for an assurance session to persist across link clicks/between pages.

eslint.config.js
```diff
@@ -185,6 +185,58 @@ export default defineConfig([
...
+      // things like setTimeout, TextEncoder, fetch, etc. The no-restricted-globals
+      // rule below then blocks the ones that aren't actually portable.
+      globals: {
+        ...globals.browser,
+        ...globals.node,
+      },
+    },
+  },
+  {
+    name: "alloy/core-src-no-browser-globals",
```
carterworks, line 202: 🔥 I like this

package.json
```diff
@@ -8,7 +8,7 @@
     "clean": "pnpm -r --if-present run clean && rimraf coverage",
     "dev": "concurrently --names build,sandbox \"pnpm --filter @adobe/alloy run build:watch\" \"pnpm --filter @adobe/alloy-sandbox-browser run start\"",
     "lint": "eslint --cache --fix \"*.{js,cjs,mjs,jsx}\" \"{sandboxes/*/src,packages/*/{src,test,scripts},scripts}/**/*.{js,cjs,mjs,jsx}\"",
-    "format": "prettier --write \"*.{html,js,cjs,mjs,jsx}\" \"{sandboxes/*/src,packages/*/{src,test,scripts},scripts}/**/*.{html,js,cjs,mjs,jsx}\"",
+    "format": "prettier --log-level warn --write \"*.{html,js,cjs,mjs,jsx}\" \"{sandboxes/*/src,packages/*/{src,test,scripts},scripts}/**/*.{html,js,cjs,mjs,jsx}\"",
```
carterworks, line 11: why

carterworks (changes requested):

.changeset/vast-jeans-walk.md
jonsnyder: If it doesn't change any functionality it is a patch no matter how big.

packages/core/src/components/Audiences/index.js
```diff
@@ -10,13 +10,17 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
```
jonsnyder, line 12: I think running url destination syncs only makes sense in the browser. As I understand it the url destination sync is just a way to tie this user to users on other domains by making a call to other domains with their 3rd party cookie. When on a server you do not have any access to their 3rd party cookies. We make need to make something like platformServices.runUrlDestination, where on server it just logs a warning and doesn't do anything.

jonsnyder (commented):
I'm a bit concerned that we are going to introduce a ton of regressions with this migration. One way to mitigate that risk is with automated testing at the SDK level. If we can develop a set of tests, and then run it against an old SDK version without failures, we can be more certain that this migration isn't creating problems. We could have AI create a bunch of tests, or we could do some random testing to generate a test set.

package.json
Spencer-Smith: This makes it so that prettier only outputs changed files instead of all the files it reads, which just makes the output cleaner (and also the command runs faster since it's doing less output)

packages/core/src/core/network/requestMethods/injectSendBeaconRequest.js
```diff
@@ -10,11 +10,12 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
 */
 
-export default ({ sendBeacon, sendFetchRequest, logger }) => {
+export default ({ sendBeacon, sendFetchRequest }) => {
   return (url, body) => {
     const blob = new Blob([body], { type: "text/plain; charset=UTF-8" });
     if (!sendBeacon(url, blob)) {
-      logger.info("Unable to use `sendBeacon`; falling back to `fetch`.");
+      // eslint-disable-next-line no-console
+      console.info("Unable to use `sendBeacon`; falling back to `fetch`.");
```
jonsnyder: What's the rational behind this change? All logging should be done on the logger.

packages/core/src/core/network/requestMethods/injectSendBeaconRequest.js
```diff
@@ -10,11 +10,12 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-export default ({ sendBeacon, sendFetchRequest, logger }) => {
+export default ({ sendBeacon, sendFetchRequest }) => {
```
carterworks: revert this change - use the logger

packages/core/package.json
```diff
@@ -24,7 +24,9 @@
   "dependencies": {
     "@adobe/aep-rules-engine": "^3.1.1",
     "@adobe/reactor-query-string": "^2.0.0",
-    "js-cookie": "3.0.7",
     "uuid": "^14.0.0"
+  },
+  "devDependencies": {
+    "js-cookie": "3.0.7"
```
jonsnyder: Are we using js-cookie in tests? Is that why this needs to be a dev dependency?

packages/core/src/core/index.js
```diff
@@ -92,10 +95,13 @@ export const createExecuteCommand = ({
     platformServices.globals.getHostname(),
     platformServices.cookie,
   );
-  const fireReferrerHideableImage = injectFireReferrerHideableImage();
+  const { fireReferrerHideableImage } = platformServices.globals;
   const getAssuranceValidationTokenParams =
     createGetAssuranceValidationTokenParams({
       getLocationSearch: () => platformServices.globals.getLocationSearch(),
+      storage:
```
carterworks: will this collide with the storage of other instances? above, on line 78, we namespace it by instance name, but we don't do that here or in other spots

eslint.config.js
```diff
@@ -265,7 +317,9 @@ export default defineConfig([
     files: [
       "packages/browser/**/*.{cjs,js,mjs,jsx}",
       "sandboxes/browser/**/*.{cjs,js,mjs,jsx}",
-      "packages/core/**/*.{cjs,js,mjs,jsx}", // TODO: Remove this once browser APIs are removed from core.
+      "packages/core/src/utils/dom/**/*.{cjs,js,mjs,jsx}",
+      "packages/core/src/utils/fireImage.js",
+      "packages/core/src/utils/injectFireReferrerHideableImage.js",
```
jonsnyder: Why specify these individually?

carterworks (changes requested):
two small things. good job on the tests

packages/core/src/core/network/requestMethods/injectSendBeaconRequest.js
Spencer-Smith: There was an awkward circular dependency because the logger needs the storage component from platform services but the network platform service wants the logger during setup. I tried simplifying that by removing the logger dependency, but yeah obviously this is a bit of a naive approach.

Current fix was to change the way the dependency is injected to not build a custom sendBeaconRequest function since the browser just uses its own implementation and with a node server we just want to fallback to fetch anyway. I'm not sure I love that compromise but it's documented and we can change the contract later without too much disruption if we want to do a platform-specific implementation of sendBeacon.

eslint.config.js
Spencer-Smith: Seems like those utils should be under the `utils/dom/` directory anyway so I moved them there are removed their individual mentions here

packages/browser/src/services/createBrowserNetworkService.js
```diff
@@ -13,25 +13,21 @@ governing permissions and limitations under the License.
...
   const sendFetchRequest = injectSendFetchRequest({ fetch });
-  const sendBeaconRequest =
+  const sendBeacon =
     typeof navigator.sendBeacon === "function"
-      ? injectSendBeaconRequest({
-          sendBeacon: navigator.sendBeacon.bind(navigator),
-          sendFetchRequest,
-        })
-      : sendFetchRequest;
+      ? navigator.sendBeacon.bind(navigator)
```
carterworks: consider moving `injectSendFetchRequest` and `injectSendBeaconRequest` into the browser module. Ideally, all usages of those files are replaced by this injectable network service.

carterworks (changes requested):

packages/browser/src/services/createBrowserNetworkService.js
Spencer-Smith: I think there's just a bit of a weird circular dependency in the platform services because the logger needs the storage service but the network needs the logger. I've been doing a bunch of weird adjustments to sort that out so the instantiation isn't so weird or prone to failure, but I think it might be easier to just split up the platform services into two groups ("base" and something else(?)), create the first group in one go, use that to create stuff like logger, then create the remaining services.

carterworks (approved):

packages/browser/src/services/createBrowserNetworkService.js
carterworks: yeah that is tricky. that darn `debug` guy.

carterworks (approved):

PR #1532, opened 2026-06-12
Added MSW handlers, mock fixtures, and Vitest+Playwright test harness for functional test migration

carterworks:
## Changed Packages
- [x] core
- [x] reactor-extension
- [x] browser

## Description

Part of the [functional-to-integration test migration stack](https://github.com/adobe/alloy/pull/1532).

**Scope**
- Adds the shared Vitest + Playwright + MSW migration infrastructure, migration plan, and supporting test-tree quality updates.

**Coverage**
- Adds acquire and consent MSW handling, seven deterministic response fixtures, consent constants, CookieStore legacy-cookie helpers, and cleanup/reload support.

**Functional test disposition**
- Deletes no functional tests and corrects four existing functional specs: C81183, C81184, C17409728, and C5298194.

**Verification**
- Focused integration tests and current quality and coverage checks pass.

## Related Issue

- Stack base: https://github.com/adobe/alloy/pull/1532

## Motivation and Context

Migrates atrophied TestCafe functional coverage to deterministic Vitest + Playwright + MSW integration coverage while preserving behavioral intent.

## Screenshots (if appropriate):

Not applicable.

## Documentation

No consumer-facing documentation changes.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Improvement (non-breaking change which does not add functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] I have added a Changeset file with a consumer-facing description of my changes.
- [x] I have signed the [Adobe Open Source CLA](https://opensource.adobe.com/cla.html) or I'm an Adobe employee.
- [x] I have made any necessary test changes and all tests pass.
- [ ] I have run the Sandbox successfully.

### [Stack](https://github.com/kitlangton/stack)

1. **#1532** 👈 current
2. #1533

PR #1539, opened 2026-06-12
Migrated Data Collector functional tests to Vitest+Playwright+MSW

carterworks:
## Changed Packages
- [ ] core
- [ ] reactor-extension
- [x] browser

## Description

**Scope**
- Migrates 13 Data Collector cases into 45 active Vitest + Playwright + MSW integration tests with no skips.
- Adds succinct harness support for DOM setup/cleanup, observable `sendBeacon` calls, complete network-call waiting, response handling, and referrer capture.

**Coverage**
- Covers interact/collect routing and payloads, link-click collection, personalization interaction metrics, monitor initialization, and referrer propagation.

- Deletes six replaced originals: C2592, C225010, C455258, C81184, C11693274, and C9369211.
- Retains seven covered originals: C75372, C1715149, C8118, C8119, C81181, C81182, and C81183.

**Verification**
- Focused Data Collector integration tests pass in CI; quality checks (build, lint, format:check, and typecheck), Tests, and Generate Coverage Report also pass on the current head.

## Related Issue

## Motivation and Context

Not applicable.

## Documentation

## Types of changes

## Checklist:

1. #1532
2. #1533
3. **#1539** 👈 current


carterworks:
## Changes since the initial migration

Follow-up work hardening the migrated `dataCollector.spec.js`. Grouped by theme; the suite is green throughout (**65 passed, 0 failed, 8 skipped** in `browser/integration`).

### Idiomatic refactor
- Replaced the hand-rolled `waitUntil` poller with `expect.poll`; dropped the redundant re-assert.
- `callbackInvoked` booleans → `vi.fn()` spies; manual try/catch → `await expect(...).rejects.toThrow(...)`.
- Added `restoreMocks: true` to the integration project and removed manual `mockRestore()` boilerplate.
- Extracted helpers: `appendLink`/`clickLink` (`domHelpers.js`) and a synchronous `interactCalls()` (correct for "no request fired" assertions, unlike the retrying `findCalls`).
- Centralized the negative-assertion wait as `NO_REQUEST_WAIT_MS` and reused the existing `waitFor` util.

### Coverage parity (vs the functional originals)
- **C2592** — restored the `meta.configOverrides` assertion (the functional `.event` check was a no-op against `undefined`; this asserts the `datasetId` actually propagates).
- **C11693274** & **C81181 augment** — assert the full `webInteraction` object (`name`/`region`/`type`/`linkClicks`), not just a subset.
- **C81183** — restored the dropped "non-clickable element ⇒ undefined fields" assertion (was only `not.toThrow`).
- **C81181** — 5 functional sub-tests that weren't migrated are now explicit `test.skip` stubs (interact link-XDM variations, deferred scope, not blocked).

### New: collect-endpoint routing (previously skipped)
- **C455258** is now migrated. Since `navigator.sendBeacon` is invisible to MSW/JS, added a hermetic **sendBeacon recorder** (installed before the library loads, returns `true`, records calls) plus a `sendEventWithIdentityHandler` so the interact→collect routing transition can be exercised without hitting the network.
- **C8118 / C9369211 / C81182** stay skipped with accurate, verified rationale (C9369211's `Referer` is a browser-set forbidden header `networkRecorder` can't capture — confirmed empirically; C81182's source tests are themselves `test.skip`).

### Organization & hygiene
- Centralized per-test lifecycle in the `alloy` fixture (DOM cleanup, sendBeacon reset; `___getLinkDetails` teardown moved into `cleanAlloy`); removed the top-level `afterEach`.
- Hoisted the repeated `__alloyMonitors` setup into a `beforeEach`; grouped the loose top-level skip stubs into a "Not migrated" describe.
- Audited all comments for terseness/accuracy (corrected the stale C8118 rationale).

carterworks:
Thanks for the thorough review — all nine findings are addressed. Summary below; each is its own commit.

**High — C81181 five skipped tests** (`a3ea7b3`→implemented): All five were ported, not re-stubbed. No-callback collection, conditional cancel via `onBeforeLinkClickSend`, conditional cancel via `filterClickDetails`, `filterClickDetails` augmentation, and `sessionStorageEnabled:false`. Each pins the full `webInteraction` (and activity map where the original did), adapted to the localhost host. (commit "implement the 5 dropped C81181 link-click tests")

**High — C8118 wholesale skip**: Migrated the whole suite (15 tests). The interact→collect routing case uses the sendBeacon recorder once the interact response establishes identity; the other 14 are interact-only — enabled/disabled download/internal/external, event grouping, cached-on-pageview, custom region/type, custom activity map, multiple grouped clicks, and custom XDM. (commit "migrate the C8118 link-click suite")

**Medium — C81181 augment coverage**: Now asserts `eventType`, the full `webInteraction`, and the activity-map data (link stays `"Test Link"` since the callback augmented only the xdm). (commit "assert eventType and activity map in C81181 augment")

**Medium — C225010 navigation**: The click now lets default navigation proceed and asserts `window.location.href` contains `#foo`, so a regression that blocks navigation would fail. (commit "assert C225010 click still navigates to #foo")

**Medium — C455258 beacon**: Now asserts the beacon URL matches `/v1/collect?configId=` and that the body parses to a valid `events` payload. (commit "verify C455258 beacon URL and body")

**Medium — C81183 loosening**: Restored exact internal-link pinning — `linkName`, `linkRegion`, `linkType: "other"`, `pageIDType`, resolved `linkUrl` and `pageName` (adapted to the localhost page) — and all five fields `undefined` for a null element. (commit "restore exact C81183 getLinkDetails assertions")

**Medium — global sendBeacon override**: I attempted opt-in (install in a `beforeEach`) and verified it fails — zero beacons recorded. Alloy creates its network service at bundle load, and `createBrowserNetworkService` binds `navigator.sendBeacon` *by value* (`.bind`) at that instant, so any per-test swap is too late. The global install (before the script is injected) is therefore required, and returning `true` is the deliberate hermetic default mirroring the functional suite (`sendBeaconMock.js`). Documented in `sendBeacon.js`. (commit "explain why the sendBeacon recorder must be global")

**Medium — broad endpoint regexes / `v1/interact` counting**:
- Fixed the spec's counter: `interactCalls` now pins the full host + path `^https://edge\.adobedc\.net/ee/(?:…/)?v1/interact\?configId=` instead of any URL containing `v1/interact`. (commit "pin interactCalls to full edge host and path")
- `handlers.js`: the one handler this PR added (`sendEventWithIdentityHandler`) follows the file-wide convention, and like every handler it gates on `configId.startsWith("bc1a10e0-…")` and **throws `"Handler not configured properly"` otherwise** — a wrong host/malformed path fails loudly rather than being silently mocked. Tightening only my handler would diverge from the rest of the file; rewriting all handlers is out of scope for a test migration.

**Medium — bundlesize.json**: Reverted to the base values; no production build input changed in this PR. (commit "restore bundlesize.json to base")

Full `Data Collector` spec: **39 pass, 2 skip** (only the genuinely-blocked C9369211 Referer test, and C81182 which is `test.skip` in the functional source too). Unrelated integration specs still pass.

carterworks:
> **Verdict: Approve with comments — 🟡** Parity on the two rich suites (C8118, C81181) is excellent; the two active skips both have rationale; no test that cleanly ran is silently dropped. Three items warrant a second look: the C2592 `domain` assertion is environment-forced (🟡 acceptable), C9369211's interact-side test ran and passed but can't be replicated in MSW (🟡 not a silent drop — old file is retained), and a minor `getLinkDetails`-for-external-link path from C81184 is not re-tested.

---

## Summary table

| Criterion | Status | Note |
|---|---|---|
| 1a C-file coverage | 🟢 | 13 old → 11 describe + 2 labeled skips |
| 1b Subtest coverage | 🟢 | All active sub-tests accounted for (see matrix) |
| 1c Assertion fidelity | 🟡 | C2592 `domain` weakened; env-forced; documented |
| 2 Skip justification | 🟡 | C9369211 interact test ran and passed; not a silent drop (old file kept), but rationale should note SW can't read forbidden headers |
| 3 Mock fidelity | 🟢 | `sendEventHandler` → no identity cookie (interact stays); `sendEventWithIdentityCookieHandler` → `_identity` state cookie (routing flips to collect) — correct |
| 4 Timing / flake | 🟢 | Negative assertions use `waitFor(200)`; fire-and-forget clicks use event-driven `waitForCall` |
| 5 Isolation | 🟢 | `networkRecorder` and `worker` reset each test; cookies cleared; DOM cleaned; `sendBeaconCalls` reset |
| 6 Hygiene | 🟡 | License header present; `test.meta` SEVERITY/TEST_RUN lost (accepted Vitest tradeoff); deletions align with migrated set; spec auto-picked by `integration/**/*.spec.js` glob |

---

## Parity matrix

| Old C-file | Describe in new spec | Old subtests | New active | Notes |
|---|---|---|---|---|
| C2592.js | C2592 | 1 | 1 | `domain` assertion weakened (see §1c) |
| C75372.js | C75372 | 1 | 1 | Exact parity |
| C1715149.js | C1715149 | 3 | 3 | Exact parity |
| C8119.js | C8119 | 1 | 1 | Exact parity |
| C81184.js | C81184 | 4 | 4 | Old test 4 (getLinkDetails external-link) is replaced by warning variant; see §1b note |
| C81181.js | C81181 | 8 | 8 | Exact parity |
| C11693274.js | C11693274 | 1 | 1 | Exact parity |
| C225010.js | C225010 | 1 | 1 | Exact parity |
| C455258.js | C455258 | 1 | 1 | Exact parity |
| C8118.js | C8118 | 14 | 14 | Exact parity |
| C81183.js | C81183 | 3 | 3 | Test 3 changed from `cancel-alloy-link-test` element to `null`; spirit preserved |
| C9369211.js | "Not migrated" | 2 active | 0 | 🟡 interact test ran and passed (see §2) |
| C81182.js | "Not migrated" | 3 (all test.skip) | 0 | 🟢 all originals were skipped |

**Deletion alignment:** The `--diff-filter=D` output lists exactly 11 deleted functional files, matching the 11 active describes. `C9369211.js` and `C81182.js` are **not deleted** — their functional coverage persists if the functional suite runs. Correct.

---

## §1c — Assertion fidelity

**C2592 `domain`** (`dataCollector.spec.js:110`):
```js
// Old (testcafe):  await t.expect(request.meta.state.domain).ok();
// New (vitest):    expect(typeof body.meta.state.domain).toBe("string");
```
Old `.ok()` requires the value to be truthy, which `""` (empty string on localhost) would fail. The weakening to `typeof === "string"` is environment-forced: localhost returns `domain: ""` and any stronger check would always fail in the MSW/localhost environment. The comment at line 109 documents the reason. **🟡 Acceptable; documented.**

**configOverrides shape** (`dataCollector.spec.js:101-107`): The old test asserted the entire `com_adobe_experience_platform.event` object; the new test only asserts `datasets.event.datasetId`. This is a minor scope reduction — not a hardening regression, but the old test verified the full overrides round-trip. Worth noting but not a blocker.

---

## §2 — Skip justification

**C9369211** (`dataCollector.spec.js:1263`): The skip comment says "forbidden header ... MSW can't capture it (verified)." The interact-side original test **ran and passed** against real edge (`referer === TEST_PAGE`). The collect-side test was already a TODO — the header check was commented out in the functional source itself.

The rationale is plausible: `Referer` is a [forbidden request header](https://fetch.spec.whatwg.org/#forbidden-request-header) that browsers set after the service worker intercepts the request, so the SW never sees it in `request.headers`. However: the Request *property* `request.referrer` (distinct from the `Referer` header) might be capturable — this wasn't explored in the PR. Since the old file is not deleted, functional coverage survives. **🟡 should-note — the "interact-side test ran and passed" provenance should be explicit in the skip rationale, and `request.referrer` (the property) could be evaluated as a partial alternative.**

**C81182** (`dataCollector.spec.js:1267`): All three originals were `test.skip` in the functional source. **🟢 Justified.**

---

## §3 — Mock fidelity

`sendEventHandler` → returns `sendEventResponse.json` which includes `identity:result` (ECID) and `state:store` containing only `kndctr_..._cluster` (no `_identity` cookie). This means clicks on the first interaction correctly stay on interact, not collect. Good.

`sendEventWithIdentityCookieHandler` → returns `sendEventWithIdentityCookieResponse.json` which includes `state:store` with `kndctr_..._identity` (the 34-day identity cookie). This correctly causes the SDK to flip subsequent `documentUnloading` calls to `sendBeacon/collect`. The C455258 and C8118 routing assertions depend on this behavior and the mock faithfully replicates it.

`setConsentHandler` → full consent-out logic implemented, sets `general=out` cookie. Adequate for C225010.

---

## §4 — Timing / flake

Negative assertions ("no request fired") all use `waitFor(NO_REQUEST_WAIT_MS)` where `NO_REQUEST_WAIT_MS = 200` (`dataCollector.spec.js:39`). The constant is named and documented at line 38. Good.

Fire-and-forget link-click positives use `findInteractCall` → `networkRecorder.waitForCall(/v1\/interact/)` which is event-driven (resolves on `notifyWaiters`, not retry-polling). The `waitForCall` has a 5 s timeout (line 214 of `networkRecorder.js`), which is generous. This is the right approach and avoids the flake that plagued prior iterations.

One pattern in C8118's second click test uses `expect.poll(() => sendBeaconCalls().length).toBe(1)` (line 790) — this is correct for sendBeacon which doesn't produce a network event, so polling is the only available signal.

---

## §5 — Isolation

- `networkRecorder.reset()` called before and after each test via the `networkRecorder` fixture (`extend.js:49-55`).
- `worker.resetHandlers()` called after each test (`extend.js:43`), preventing handler bleed.
- Cookies cleared via `cookieStore.getAll()` + `cookieStore.delete()` before each alloy setup (`extend.js:61-64`).
- `cleanupDom()` called after each test (`extend.js:74`).
- `resetSendBeaconCalls()` called in alloy fixture setup (`extend.js:68`).
- Tests that do inline resets (C455258, C8118 line 787-788) are correct — they reset mid-test after establishing identity.

Clean.

---

## §6 — Hygiene

- License header: present, year `2026`. ✅
- `test.meta` (SEVERITY/TEST_RUN): dropped across all new specs — the Vitest framework has no equivalent mechanism. Accepted tradeoff for the whole migration; flag once. 🟡
- Spec registration: `vitest.projects.js` uses a glob `packages/browser/test/integration/**/*.{test,spec}.?(c|m)[jt]s?(x)` — the new file is auto-included. ✅
- Comments: explain *why* (environment constraints, fire-and-forget semantics, sendBeacon binding timing). Follow the style guide. ✅
- `C81183` old test 3 used `cancel-alloy-link-test` (a valid element with `onBeforeLinkClickSend` returning `false`); new test 3 uses `null`. Both exercise the "no-link element → undefined fields" path. The old test 3 also contained a positive case (`alloy-link-test` → full details), but that's covered by tests 1 and 2. Equivalent coverage. ✅
- Old C81184 test 4 (`getLinkDetails works regardless of clickCollectionEnabled`) tested an **external link** (`linkType: "exit"`). The new spec covers the spirit via C81183 test 2 (internal link, `clickCollectionEnabled: false`) and C8118 external-link click tests, but the hook-API (`window.___getLinkDetails`) is only exercised for internal links. Minor gap — `exit` type in the hook path isn't explicitly tested. 🟡 Minor.

---

## Recommendation

**Approve with the following comments:**

1. **🟡 C9369211 skip comment** — add "the interact-side functional test (`C9369211.js`) ran and passed against real edge; the Referer header is browser-forbidden and added after MSW's SW intercepts the request. `request.referrer` (the property) was not evaluated as an alternative."
2. **🟡 C2592 configOverrides scope** — consider also asserting `body.meta.configOverrides.com_adobe_experience_platform.datasets.event.datasetId` matches `datasetId` (already done, line 104–107) — this is actually present; the reduced assertion vs. old (full `event` object) is cosmetic.
3. **🟡 Minor** — C81184 old test 4 (external link via `getLinkDetails` hook) has no direct counterpart. Either add a test or note it as out-of-scope.

No blockers. The migration is solid.

---

*This is a PR comment, not a formal approval or request-for-changes review.*

PR #1549, opened 2026-06-12
Migrated Brand Concierge functional tests to Vitest+Playwright+MSW

carterworks:
## Changed Packages
- [ ] core
- [ ] reactor-extension
- [x] browser

## Description

**Scope**
- Adds 5 active Brand Concierge integration tests covering BC1 through BC5.

**Coverage**
- Preserves response statuses, feedback ratings and data, network requests, ECID propagation, and the current permissive anyOf behavior for incomplete options.

- Deletes the Brand Concierge functional source after migrating BC1 through BC5.

**Verification**
- Current CI Tests, quality checks, and coverage generation pass for the remote branch.

## Related Issue

## Motivation and Context

Not applicable.

## Documentation

## Types of changes

## Checklist:

1. #1532
2. #1533
3. **#1549** 👈 current


carterworks:
## Integration Test Migration Review — BrandConcierge (`migrate-integration/17-brand-concierge`)

**Verdict: 🟡 CONDITIONAL PASS** — The migration is structurally sound and the new tests are better instrumented than the old ones. Two issues need resolution before merge: a deliberate-but-undocumented behavioral change in BC4/BC5, and a misused assertion idiom that makes those tests vacuous.

---

### Parity matrix

| Old case | New case | Subtests old→new | Status |
|---|---|---|---|
| C2590433 — message only | BC1/C2590433 | 5→5 | 🟢 |
| C2590434 — data object | BC2/C2590434 | 5→6 | 🟢 (extra surface-length assertion) |
| C2590435 — feedback in XDM | BC3/C2590435 | 5→5 | 🟡 payload field changed (see §3) |
| C2590436 — invalid options throws | BC4/C2590436 | 2→1 | 🔴 polarity inverted |
| C2590437 — missing data.type throws | BC5/C2590437 | 2→1 | 🔴 polarity inverted |

No old functional tests were deleted by this PR. Old spec: `packages/browser/test/functional/specs/BrandConcierge/sendConversationalEvent.js` (not modified).

---

### 1. Parity — 🟢 BC1–BC3, 🔴 BC4–BC5

BC1–BC3 map cleanly. Assertion counts are equal or higher. Assertion paths are corrected (see §3).

**BC4 / BC5 — polarity inversion.** The old tests assert `errorThrown === true`; the new tests assert `.resolves.not.toThrow()`.

The comment in the new spec (`brandConcierge.spec.js:177`) acknowledges this: *"Original test expected a throw; the anyOf validator is permissive so neither case throws."*

After tracing the validator:

```
// validateMessage.js:35 — branch 2 of anyOf
objectOf({ xdm: xdmValidator, voiceEnabled: boolean() }).required()
```

`createAnyOfValidator.js:17` iterates branches, returning the first that doesn't throw. Branch 2 accepts any non-null object because `xdm` and `voiceEnabled` are both optional, and `createObjectOfValidator.js:37` copies unknown keys through without error. So `{ onStreamResponse: fn }` and `{ data: { payload: … } }` both match branch 2, and **neither throws today**.

This means:
- If the old functional tests ever passed against a real edge, the validator has since been relaxed.
- If they never passed (aspirational spec), the new tests accurately document current behavior.

**This needs an explicit decision in the PR description / ticket.** Is the permissive anyOf intentional (product decision to not validate on the client)? Or is branch 2 missing a `.required()` on `xdm` to narrow it to xdm-only calls? Until that's confirmed, these two cases represent **reduced validation coverage**.

---

### 2. Skip justification — 🟢

No skipped or `.todo` tests in the new spec.

---

### 3. Mock fidelity — 🟢 (one data-schema note)

**Request path correction (🟢):** Old tests check `requestBody.query.conversation`; new tests correctly check `body.events[0].query.conversation`. The source (`createSendConversationEvent.js:54`) calls `event.mergeQuery(…)` which places the query inside the event object, not the payload root. The old path was wrong; the new path is right.

**BC3 feedback schema alignment (🟡):** Old test passes `{ feedback: { rating: 4, comment: "…" } }`, new test passes `{ feedback: { classification: "positive", comment: "…" } }` (`brandConcierge.spec.js:150–157`). The validator schema (`validateMessage.js:23–32`) defines `feedback.classification`, not `feedback.rating`, so the new payload is schema-valid where the old was not. This is a correction, but it's worth a comment in the PR that the payload was updated to match the actual schema.

**SSE mock fidelity (🟢):** `brandConciergeStreamingHandler` returns a proper `text/event-stream` response with `data: {…}\n\n` — adequate for exercising the stream parser.

**BC4/BC5 handler (🟡):** Both use `brandConciergeNoStreamHandler` (204). Since neither test asserts any network call (they assert only that the call doesn't throw), this is fine for now, but if the tests are fixed to use `rejects.toThrow()`, they won't reach the network at all — worth noting.

---

### 4. Timing / flake — 🟡

**BC1 stream callback** (`brandConcierge.spec.js:97–99`): Uses a local `waitUntil` polling helper — appropriate for async SSE, but the helper is defined inline in the spec file (`line 47–62`) rather than imported from `helpers/utils/`. Consider extracting it to the shared utils.

**BC4/BC5 assertion idiom — wrong pattern (🔴):**

```js
// brandConcierge.spec.js:185-189 and 199-208
await expect(
  alloy("sendConversationEvent", { … }),
).resolves.not.toThrow();
```

`.toThrow()` / `.not.toThrow()` applies to **functions**, not promise-resolved values. Chained after `.resolves`, it receives the settled value (here `undefined`) and `expect(undefined).not.toThrow()` passes trivially — it does not verify the promise resolved vs. rejected. Every other rejection test in the codebase uses `.rejects.toThrow()`. If the intent is "does not reject", the correct form is:

```js
await expect(promise).resolves.toBeUndefined();
// or simply:
await promise; // throws the test if the promise rejects
```

As written, BC4 and BC5 **cannot catch a regression** where the call starts throwing again. They are effectively no-ops.

---

### 5. Isolation — 🟢

- `networkRecorder.reset()` in before/after via `{ auto: true }` fixture (`extend.js:27–35`).
- `worker.resetHandlers()` after each test (`extend.js:21–24`).
- Cookie jar cleared before each test (`extend.js:52–56`).
- Per-test `worker.use(…)` for BrandConcierge handlers — no global handler registration that could bleed.

---

### 6. Hygiene — 🟡

| Check | Status | Note |
|---|---|---|
| License header | 🟡 | Year reads `2026` (`brandConcierge.spec.js:2`); other new specs use `2025`. Minor. |
| `test.meta` tags | 🟢 | IDs embedded in test names (`BC1/C2590433`, etc.) |
| `vitest.projects.js` | 🟢 | Glob `test/integration/**/*.spec.js` picks up the new file automatically — no change needed |
| Comments | 🟢 | Inline comment at line 177 explains the polarity change — good instinct, but the explanation should be in the PR description with a ticket ref, not just a code comment |
| `waitUntil` duplication | 🟡 | Same utility may exist in `helpers/utils/` — worth checking before merging |

---

### Summary of required actions

1. **🔴 BC4/BC5 assertion idiom** — Replace `resolves.not.toThrow()` with a form that actually validates behavior (either `resolves.toBeUndefined()` if non-throwing is the intent, or `rejects.toThrow()` if validation is tightened).

2. **🔴 BC4/BC5 behavioral decision** — Confirm in the PR (with a ticket or design doc reference) whether the permissive `anyOf` is intentional. If it's a known product decision, the test is correct documentation. If it's a validator bug, the tests should be written to `rejects.toThrow()` and the validator fixed first.

3. **🟡 BC3 payload change** — Call out `rating → classification` in the PR description as a schema alignment fix, not just a rename.

4. **🟡 `waitUntil`** — Extract to shared utils or confirm it doesn't already exist there.

PR #1556, opened 2026-07-20
Persist ECID reliably and self-heal subscription cache.

dompuiu:
## Changed Packages

- [ ] core
- [x] browser
- [ ] reactor-extension 

## Description

This PR fixes a few edge case that were not that important until now. More specifically, it was possible to store an ECID as undefined in the Index DB that was used then in the push notification service worker. Also the value is stored in the guard cache only if the index db operation succeeds.

## Motivation and Context

There will be some changes in the push notifications backend service. They will start sending an ECID in the notification payload an we need to check that value against the one we store in the IndexDB. If those values don't match, we won't show the notification. This is why it is more important now to have a correct ECID stored in the Index DB. 

## Types of changes

## Checklist:

### DIFF
--- a/packages/core/src/components/Identity/createComponent.js
+++ b/packages/core/src/components/Identity/createComponent.js
@@ -56,7 +56,9 @@ export default ({
         // so that sendBeacon requests don't override the edge info from before.
         edge = { ...edge, ...response.getEdge() };
 
-        identity.setIdentityAcquired();
+        if (identity.getEcidFromCookie()) {
+          identity.setIdentityAcquired();
+        }
 
         return handleResponseForIdSyncs(response);
       },
--- a/.changeset/quick-melons-guard.md
+++ b/.changeset/quick-melons-guard.md
@@ -1,0 +1,5 @@
+---
+"@adobe/alloy": patch
+---
+
+Fix Push Notifications module so the ECID used to target notifications is persisted reliably: skip sending when no ECID is available, and only cache the subscription details after the ECID is saved so a failed send or storage write self-heals on the next attempt.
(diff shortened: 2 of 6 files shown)


packages/core/src/components/Identity/createComponent.js
```diff
@@ -56,7 +56,9 @@ export default ({
         // so that sendBeacon requests don't override the edge info from before.
         edge = { ...edge, ...response.getEdge() };
 
-        identity.setIdentityAcquired();
+        if (identity.getEcidFromCookie()) {
```
dompuiu, line 59: This is another edge case I missed in the initial release. On a new user, without ecid, a request was made to Konductor. If the cookie from the response could not be read, the identity was still marked as acquired.

carterworks: Because of `await identity.awaitIdentity()` on [`sendPushSubscriptionRequest.js:50`](https://github.com/adobe/alloy/pull/1556/changes#diff-e3b56d86d03ddf3483edd962f14362ec436e9926bc15a2a7a6bd572969dfc951L60), if there is no ecid in the cookie, the `sendPushSubscriptionRequest` will never get to your new "no ECID" logic. Is that intended?

dompuiu: Yes. Before this change, if the ECID was not decoded from the cookie, the send push subscription command would have fired. Without an ECID, the profile cannot be updated with the subscription details.

dompuiu: When you say "NO ecid" logic are you referring to

```
  if (!ecid) {
    logger.info(
      "No ECID is available. Not sending push subscription details to the server.",
    );

    return;
  }
  ```
  ?

carterworks: Yes that is the logic I am talking about.

carterworks: It's basically dead code, is it not?

carterworks (approved):

Spencer-Smith (approved):

PR #1557, opened 2026-07-20
Preserved author attributes on remote scripts injected from Personalization offers

jonsnyder:
## Changed Packages

- [ ] core
- [ ] reactor-extension

## Description

When a Personalization offer (Target HTML offer / experience fragment) contains a remote `<script src="...">`, the Web SDK does not execute the script node inserted via the offer HTML. Instead it re-creates a fresh `<script>` element and appends it to `document.head` to force the browser to load/run it (this is what lets us know when the script finishes loading, so the pre-hiding snippet can be removed at the right time).

Previously that re-created element only had `src` set (plus `async`), so every other author-supplied attribute (`class`, `type`, `data-*`, `crossorigin`, `integrity`, …) was stripped.

This PR changes the remote-script pipeline in `packages/browser/src/components/Personalization/dom-actions/scripts.js` to carry the original `<script>` **elements** through to execution (instead of collapsing them to bare `src` URLs) and copy all author attributes onto the re-created head element. Invariants are still enforced last: `async` is forced on, and the CSP `nonce` is applied when available. Load handlers (`onload`/`onerror`) are assigned after attributes are copied so a copied inline handler can't break load-completion tracking.

- `getRemoteScriptsUrls` → `getRemoteScripts` (returns elements)
- `loadScript` copies attributes, then overrides `nonce`/`async`
- All four insertion actions updated: `appendHtml`, `prependHtml`, `insertHtmlBefore`, `insertHtmlAfter`
- `remapHeadOffers` needs no change (it delegates to `appendHtml`), so the head-remap path inherits the fix

Design/spec/tasks tracked under `openspec/changes/preserve-remote-script-attributes/`.

## Related Issue

PDCL-14174 — "Injected Target scripts from offers is getting duplicated into the head of the page incorrectly" (Adobe internal Jira). Reported by Vanguard: a required `class` attribute on the offer script was lost when the script was re-created in `<head>`.

## Motivation and Context

Customers rely on author-supplied script attributes (e.g. `class`) for downstream logic; stripping them on re-creation breaks their functionality. This preserves author intent faithfully while keeping the intentional async re-creation that drives pre-hiding removal.

## Documentation
N/A — behavioral fix; no consumer configuration changes. Covered by a changeset entry.

## Types of changes

## Checklist:

- [x] I have signed the Adobe Open Source CLA or I'm an Adobe employee.

### DIFF
--- a/packages/browser/src/components/Personalization/dom-actions/scripts.js
+++ b/packages/browser/src/components/Personalization/dom-actions/scripts.js
@@ -24,10 +24,24 @@ const getPromise = (url, script) => {
     };
   });
 };
-const loadScript = (url) => {
+const loadScript = (source) => {
+  const url = getAttribute(source, SRC);
   const script = document.createElement("script");
-  script.src = url;
+  // Preserve all author-supplied attributes (class, type, data-*, etc.) so
+  // consumers relying on them keep working.
+  const { attributes } = source;
+  for (let i = 0; i < attributes.length; i += 1) {
+    const { name, value } = attributes[i];
+    script.setAttribute(name, value);
+  }
+  // Override invariants last so they can't be clobbered by the source element.
+  const nonce = getNonce();
+  if (nonce) {
+    script.setAttribute("nonce", nonce);
+  }
   script.async = true;
+  // Assign load handlers after copying attributes so a copied inline
+  // onload/onerror attribute can't break load-completion tracking.
   const promise = getPromise(url, script);
   document.head.appendChild(script);
   return promise;
@@ -70,7 +84,7 @@ export const getInlineScripts = (fragment) => {
   return result;
 };
 
-export const getRemoteScriptsUrls = (fragment) => {
+export const getRemoteScripts = (fragment) => {
   const scripts = selectNodes(SCRIPT, fragment);
   const result = [];
   const { length } = scripts;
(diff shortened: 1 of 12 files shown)


Spencer-Smith (approved):

packages/browser/src/components/Personalization/dom-actions/scripts.js
```diff
@@ -24,10 +24,24 @@ const getPromise = (url, script) => {
...
 };
-const loadScript = (url) => {
+const loadScript = (source) => {
+  const url = getAttribute(source, SRC);
   const script = document.createElement("script");
-  script.src = url;
+  // Preserve all author-supplied attributes (class, type, data-*, etc.) so
+  // consumers relying on them keep working.
+  const { attributes } = source;
+  for (let i = 0; i < attributes.length; i += 1) {
```
carterworks, line 33: style nitpick: this could be a `for..of` loop

carterworks (approved):

packages/browser/src/components/Personalization/dom-actions/scripts.js
dompuiu: You are right but for..of is slower than the standard for due to iterator overhead.

PR #1561, opened 2026-08-03
FEAT: MVP node SDK

Spencer-Smith:
## Changed Packages

- [ ] core
- [ ] reactor-extension 
- [x] node

## Description

This adds `@adobe/alloy-node` plus a small backward-compatible core change to support it.

Instead of the Wen SDK's string-command dispatcher (which exists to support a pre-load stub queue that's meaningless in Node), `createInstance()` returns an object with real methods (`configure`, `sendEvent`, `getIdentity`, etc.). The key addition is `forRequest()`: call `configure()` once at startup, then `forRequest({ cookie })` per request to get a handle bound to that request's real cookies, so one long-lived instance resolves per-visitor identity across requests with no server-side session cache. This needed an optional `coreConfigValidators` param threaded through `createExecuteCommand`/`createCustomInstance`, since `forRequest()`'s repeated re-`configure()` would otherwise trip the cross-instance org/datastream uniqueness check (default behavior is unchanged).

Node also gets a fetch-only Personalization component (`sendEvent()` with `decisionScopes`/`surfaces` returns `{ propositions }` in the shape `applyPropositions` expects, no rendering machinery), and pluggable platform services (`network`/`storage`/`cookie`/`runtime`/`legacy`/`globals`), each independently overridable. Test coverage includes unit tests for every service and for `createNodeAlloy`'s wiring, plus real-Edge-Network integration tests proving both identity continuity (same cookie → same ECID) and isolation (different cookies → no cross-visitor leakage). `sandboxes/node` gained a few runnable scripts as manual smoke-test aids.

## Related Issue

Part of Universal Javascript

## Motivation and Context
Allows users to onboard into server-based uses for the SDK

N/A

## Documentation
N/A

## Types of changes

## Checklist:


packages/node/src/services/createNodeCookieService.js
```diff
@@ -0,0 +1,54 @@
...
+ * across process restarts or across requests in a server context.
+ *
+ * The `jar` and `converter` params are the recursion seam for `withConverter`,
+ * mirroring the browser cookie service built on js-cookie.
+ *
+ * @param {Map<string, string>} [jar]
+ * @param {{ read?: Function, write?: Function }} [converter]
+ * @returns {CookieService}
+ */
+const createNodeCookieService = (jar = new Map(), converter = {}) => ({
```
carterworks, line 27: This default cookie service is stateful, and since it is created in `createNodeInstance` instead of `forRequest`, cookies will (understandably) leak between requests

for example, if two visitors make two sequential requests come in, they will have the same ecid

```js
  it("forRequest() must not leak identity between visitors via a top-level cookie override", async () => {
    const sharedCookie = createNodeCookieService();
    const alloy = node.createInstance({
      platformServices: { cookie: sharedCookie },
    });
    await alloy.configure(config);

    const visitorA = alloy.forRequest();
    const identityA = await visitorA.getIdentity();

    const visitorB = alloy.forRequest();
    const identityB = await visitorB.getIdentity();

    expect(identityA.identity.ECID).not.toEqual(identityB.identity.ECID);
  });
```

packages/node/src/createNodeAlloy.js
```diff
@@ -0,0 +1,145 @@
...
+    );
+    requestExecuteCommand("configure", capturedConfig);
+    return bindCommandMethods(requestExecuteCommand, COMMAND_NAMES);
+  };
+
+  return {
+    ...bindCommandMethods(executeCommand, COMMAND_NAMES),
+    /** @param {Record<string, unknown>} options */
+    configure(options) {
+      capturedConfig = options;
```
carterworks: this should happen _after_ `executeCommand` validates the configuration. As-is, an invalid config can still be used in a `forRequest`

carterworks (changes requested):

.changeset/dirty-pugs-boil.md
```diff
@@ -0,0 +1,5 @@
+---
+"@adobe/alloy-node": major
+---
```
carterworks, line 4:
```suggestion
"@adobe/alloy-core": patch
---
```

need to bump the version to include the newly-expored `createCoreConfigs`

carterworks (changes requested):

carterworks (approved):

PR #1562, opened 2026-08-07
Run inline offer ES modules as modules; only extract executable scripts, with a functional attribute allowlist for re-created ones

jonsnyder:
## Why

Follow-up to #1557 (`preserve-remote-script-attributes`, PDCL-14174) and the ES Module request in **PDCL-14354**. Four issues:

1. **Inline ES modules don't run.** Inline offer scripts are re-created with only `textContent` + `nonce` and appended/removed **synchronously** — a model that assumes classic execution and drops `type`. So an inline `<script type="module">…</script>` ran as a **classic** script and its module-only `import`/`export` syntax threw a `SyntaxError`.
2. **Copying all attributes duplicated identity.** The original offer `<script>` stays in the page (inert — `innerHTML` scripts never run). #1557 copied its `id`/`class`/`data-*` onto the executed head element too, so `document.getElementById`/`querySelector` began matching the head copy instead of the original.
3. **Copying only `type`/`src`/`nonce` dropped attributes that change execution semantics.** `crossorigin`, `integrity`, `referrerpolicy`, `fetchpriority`, and `nomodule` all affect how the browser fetches/executes the script (CORS mode, SRI validation, referrer, load priority, module-support opt-out) — dropping them silently changes behavior rather than just identity.
4. **Extraction was applied to every `<script>` tag, regardless of type.** The whole reason for extraction/re-insertion is that `innerHTML` never executes scripts. But a `<script>` whose `type` isn't recognized as executable (e.g. `importmap`, `application/json`, a templating library's custom type) is a data block per the HTML spec — the browser never executes it, insertion method notwithstanding. Extracting these did nothing useful, and (before `type` was preserved) risked turning inert data into code that actually ran.

## What changed (all in `scripts.js`)

**1. Inline modules run as modules.** An inline `<script type="module">` is now classified as a **head script** (executed async in `document.head`) instead of a classic inline script:
- `getRemoteScripts` also returns inline module elements; `getInlineScripts` handles classic inline only (original behavior).
- `loadScript` re-creates a source with no `src` in `document.head`, carrying `type="module"` and the code, so it runs as a real module against the **document base URL** (relative `import`s resolve).
- Classic inline handling is unchanged.

**2. Head scripts copy a functional attribute allowlist, not all author attributes.** A `COPIED_ATTRIBUTES` constant (`type`, `nonce`, `crossorigin`, `integrity`, `referrerpolicy`, `fetchpriority`, `nomodule`) is copied from the source element when present. `id`/`class`/`data-*` are still never copied, so attribute lookups continue to resolve to the original (inert) offer script — this narrows the "copy all attributes" behavior shipped in #1557, while the addition of `crossorigin`/`integrity`/`referrerpolicy`/`fetchpriority`/`nomodule` keeps the re-created script's fetch/CORS/SRI/priority behavior matching the original.

**3. Nonce: source as fallback, page nonce takes priority.** `nonce` is copied from the source markup like the rest of the allowlist, but the page's current CSP nonce (from `getNonce()`) always overrides it when one is found, since that's the value the browser actually checks against the CSP header for this page load. This preserves prior behavior when a page nonce exists and adds a fallback when it doesn't.

**4. Non-executable script types are left untouched.** A new type gate (`isExecutableScript`) runs ahead of the inline/remote/head-script classification: a script is only eligible for extraction if its `type` is absent/empty, a recognized JavaScript MIME type, or `"module"`. Anything else (`importmap`, `application/json`, custom templating types, etc.) is skipped entirely — not extracted, not re-created, left exactly where it was. Re-created classic inline scripts also now preserve `nomodule`, so an author's module-browser fallback script is still correctly skipped after being forced through `appendChild`.

## Not tracked (by design)

Inline module scripts emit **no `load` event** (verified in Chromium; external module scripts do), so the inline-module head element resolves **immediately (fire-and-forget)**. A `blob:`/`data:` URL workaround would get a load event but change the base URL and **break relative imports**, so it's rejected. `executeInlineScripts` never returned a promise, so inline scripts were never awaited; remote-script load tracking (which gates pre-hiding removal) is unchanged.

## Behavior-change note for reviewers

This **reduces** what #1557 shipped: offers that relied on `id`/`class`/`data-*` appearing on the *executed* head script will no longer find them there (the attributes remain on the original offer element). This is deliberate — see issue 2 above. It also **restores** `crossorigin`/`integrity`/`referrerpolicy`/`fetchpriority`/`nomodule` copying (dropped by narrowing to `type`/`src`/`nonce`), since those affect execution rather than identity. It additionally **stops force-executing** scripts with a non-executable `type` (e.g. `importmap`) that the previous inline/remote logic didn't explicitly exclude — these were inert before and remain inert now, just without being needlessly cloned into the DOM.

## Tests

Added/updated in `scripts.spec.js` (real Chromium via Playwright):
- inline module routed to `getRemoteScripts`, excluded from `getInlineScripts`; classic inline stays inline;
- inline module with no code ignored by both paths;
- inline module injected into `document.head`, resolves without waiting, executes with ES module semantics (module-only `export` runs), and does **not** carry `class`/`data-*`;
- remote head script copies the full functional allowlist (`type`, `crossorigin`, `integrity`, `referrerpolicy`, `fetchpriority`, `nomodule`) and omits them when absent from the source (asserts `class`/`id`/`data-*` are still absent);
- nonce: page nonce applied when present, source nonce used as fallback when no page nonce is found, and page nonce overrides a source nonce when both are present;
- non-executable types (`text/x-template`, `application/json`, `importmap` inline and with `src`) are excluded from both inline and remote extraction;
- recognized JavaScript MIME type variants (`application/javascript`, mixed case with a `charset` parameter) and a padded/uppercase `"module"` are still classified correctly;
- `nomodule` preserved on a re-created classic inline script, and omitted when absent from the source;
- `async`, load/failure, and classic-inline removal still covered.

Full Personalization dom-actions suite passes (21 tests in `scripts.spec.js`; 286 tests across the whole Personalization suite); changed files lint clean. OpenSpec change: `openspec/changes/support-inline-module-scripts/` (validates `--strict`).


carterworks (approved):

packages/browser/src/components/Personalization/dom-actions/scripts.js
```diff
@@ -24,24 +27,40 @@ const getPromise = (url, script) => {
...
-  for (let i = 0; i < attributes.length; i += 1) {
-    const { name, value } = attributes[i];
-    script.setAttribute(name, value);
+  // Only carry the attributes required to execute the script correctly: `type`
+  // (e.g. "module"), `src` (below), and the CSP `nonce`. Author attributes such
+  // as `id`, `class`, and `data-*` are intentionally NOT copied: the original
+  // offer <script> is left in the page (inert), so copying its id/class onto
+  // this executed element would create duplicate matches for
+  // document.getElementById / querySelector.
+  if (type) {
```
carterworks: #1557 added all previously-authored elements, including `class`, which was specifically requred by Vanguard.

This PR removes it an only copies the `type`.

1. Can you explain to me how this isn't a regression of the case that #1557 was meant to fix?
2. We should also copy over `integrity` and `crossorigin`. Those can also be required to run the script (although I'd be surprised if anyone ever used them)

carterworks: > are intentionally NOT copied: the original
  // offer <script> is left in the page (inert), so copying its id/class onto
  // this executed element would create duplicate matches for
  // document.getElementById / querySelector.

I don't understand why we append inline scripts twice. Shouldn't this be the only proposition activation?

carterworks (changes requested):

carterworks (approved):

packages/browser/src/components/Personalization/dom-actions/scripts.js
jonsnyder: When you set innerHTML the browser doesn't execute scripts. So you include it in the regular HTML, but then also add it elsewhere as a full fledged script tag inserted into the DOM.

Yes, that is a good point there are other properties that we should be copying over. I identified these:
const COPIED_ATTRIBUTES = [
  "crossorigin",
  "integrity",
  "referrerpolicy",
  "fetchpriority",
  "nomodule",
];

packages/browser/src/components/Personalization/dom-actions/scripts.js
```diff
@@ -14,6 +14,55 @@ import { SCRIPT } from "@adobe/alloy-core/constants/tagName.js";
...
+  "text/javascript1.1",
+  "text/javascript1.2",
+  "text/javascript1.3",
+  "text/javascript1.4",
+  "text/javascript1.5",
+  "text/jscript",
+  "text/livescript",
+  "text/x-ecmascript",
+  "text/x-javascript",
+]);
```
jonsnyder: Can we just search the mime type value for the word "script"?

jonsnyder:
@carterworks can you re-review? I've added additional attributes and checks.

carterworks (approved):

PR #1564, opened 2026-08-24
FEAT: Implement consent and context components in Node SDK

Spencer-Smith:
## Changed Packages

- [x] core
- [x] node
- [ ] reactor-extension 

## Description

This branch implements two components to the Node SDK:
1. **Consent (`setConsent`)** — Node can now report a visitor's privacy consent state to Edge Network, matching the browser SDK. Previously there was no way to do this server-side at all.
2. **Context + real visitor signals** — server calls now tag events with basic metadata, capture what page the visitor was on, and forward the visitor's real browser/language info upstream instead of a generic Node signature — plus a fix ensuring that works on both classic (Express) and modern fetch-based (Next.js, edge runtime) backends.

## Related Issue

(Related to Universal JS Migration effort)

## Motivation and Context

Improves SDK functionality and enables new use cases for Node servers

N/A

## Documentation
[Doc ticket](https://jira.corp.adobe.com/browse/PDCL-16447)

## Types of changes

## Checklist:

### DIFF
--- a/packages/node/package.json
+++ b/packages/node/package.json
@@ -11,7 +11,7 @@
   "author": "Adobe Inc.",
   "license": "Apache-2.0",
   "engines": {
-    "node": ">=18"
+    "node": ">=22"
   },
   "files": [
     "src",
--- a/packages/node/package.json
+++ b/packages/node/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@adobe/alloy-node",
-  "version": "1.0.0-beta.0",
+  "version": "0.0.1-beta.0",
   "description": "Adobe Experience Platform Web SDK - Node.js entrypoint",
   "type": "module",
   "repository": {
--- a/packages/node/package.json
+++ b/packages/node/package.json
@@ -3,7 +3,6 @@
   "version": "1.0.0-beta.0",
   "description": "Adobe Experience Platform Web SDK - Node.js entrypoint",
   "type": "module",
-  "private": true,
   "repository": {
     "type": "git",
     "url": "git+https://github.com/adobe/alloy.git",
--- a/packages/node/src/components/context.js
+++ b/packages/node/src/components/context.js
@@ -12,31 +12,21 @@ governing permissions and limitations under the License.
 
 /** @import { NodeRequestLike } from "../services/createNodePlatformServices.js" */
 
+import { createRequire } from "node:module";
 import getHeader from "../services/getHeader.js";
 
-// Mirrors @adobe/alloy-core/constants/libraryName.js, hardcoded here rather
-// than imported to avoid pulling core's build-time __VERSION__ placeholder
-// (never replaced in Node's unbundled source) and to keep Node's own
-// implementationDetails.version tied to @adobe/alloy-node's own version,
-// which is independent of the browser bundle's.
+// Mirrors @adobe/alloy-core/constants/libraryName.js — hardcoded rather than
+// imported since it's independent of @adobe/alloy-node's own version.
 const LIBRARY_NAME = "https://ns.adobe.com/experience/alloy";
-const LIBRARY_VERSION = "1.0.0-beta.0";
+const { version: LIBRARY_VERSION } = createRequire(import.meta.url)(
+  "../../package.json",
+);
 
 /**
- * Required (always active) Node component analogous to the browser bundle's
- * Context component. There's no DOM to read device/viewport/timezone info
- * from, so unlike the browser version this only attaches:
- *
- * - `implementationDetails`, unconditionally.
- * - `web.webPageDetails.URL`, best-effort, from the `Referer` header of the
- *   request passed to `forRequest({ request })` — the closest Node
- *   equivalent of "the page the visitor is on," since it's usually the page
- *   that called this server's endpoint in a hybrid-personalization setup.
- *
- * Everything else the browser's Context collects (screen size, viewport,
- * local timezone) has no honest server-side source and is deliberately not
- * guessed at here — a caller who has real values for those can still merge
- * them directly via `sendEvent({ xdm })`.
+ * Node's analog to the browser bundle's Context component. No DOM to read
+ * device/viewport/timezone from, so this only attaches `implementationDetails`
+ * and, best-effort, `web.webPageDetails.URL` from the `Referer` header of
+ * `forRequest({ request })`.
  *
  * @param {Object} params
  * @param {{ request?: NodeRequestLike }} params.platformServices
--- a/packages/node/src/components/context.js
+++ b/packages/node/src/components/context.js
@@ -1,0 +1,70 @@
+/*
+Copyright 2026 Adobe. All rights reserved.
+This file is licensed to you under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License. You may obtain a copy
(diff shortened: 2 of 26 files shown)


packages/node/src/components/context.js
```diff
@@ -0,0 +1,70 @@
...
+
+import getHeader from "../services/getHeader.js";
+
+// Mirrors @adobe/alloy-core/constants/libraryName.js, hardcoded here rather
+// than imported to avoid pulling core's build-time __VERSION__ placeholder
+// (never replaced in Node's unbundled source) and to keep Node's own
+// implementationDetails.version tied to @adobe/alloy-node's own version,
+// which is independent of the browser bundle's.
+const LIBRARY_NAME = "https://ns.adobe.com/experience/alloy";
+const LIBRARY_VERSION = "1.0.0-beta.0";
```
carterworks: Pull this from the `package.json`, either by [import attributes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import/with) or with some bundle-time replacement like we do in the browser bundle

packages/node/src/services/pickForwardableHeaders.js
```diff
@@ -0,0 +1,40 @@
...
+/** @import { NodeRequestHeaders } from "./createNodePlatformServices.js" */
+
+import getHeader from "./getHeader.js";
+
+// Headers worth forwarding upstream to Edge Network when Node is proxying a
+// request on a real visitor's behalf, so Edge Network's own server-side
+// device/locale parsing (which normally reads these off the direct
+// browser -> Edge Network request) still has something real to parse instead
+// of whatever Node's own fetch() would send by default.
+const FORWARDABLE_HEADER_NAMES = ["user-agent", "accept-language"];
```
carterworks: We should consider forwarding [client hint headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Client_hints) by default, as well as some proxy/ip related headers, in order to send the client's ip address to the backend. 

As is, I think that  Platform will read the ip address from the server and start attributing all requests to the single geographic location of the server, instead of the clients.

packages/node/package.json
```diff
@@ -3,7 +3,6 @@
   "version": "1.0.0-beta.0",
   "description": "Adobe Experience Platform Web SDK - Node.js entrypoint",
   "type": "module",
-  "private": true,
```
carterworks, line 6: 😎

packages/node/test/unit/specs/createNodeAlloy.spec.js
```diff
@@ -155,4 +161,56 @@ describe("createNodeAlloy", () => {
...
+      const [, requestInit] = fetchMock.mock.calls.at(-1);
+      expect(requestInit.headers).toEqual(
+        expect.objectContaining({ "user-agent": "Mozilla/5.0" }),
+      );
+      const { xdm } = JSON.parse(requestInit.body).events[0];
+      expect(xdm.web).toEqual({
+        webPageDetails: { URL: "https://example.com/page" },
+      });
+    });
+  });
```
carterworks, line 315: I think this leaves a lot of ground untested. If you add a new test file in the integration tetss, you can set up some msw stuff to be able to inspect outgoing request bodies and incoming response bodies.

From Claude:

> ### Test coverage gaps on this branch
>
> The unit specs cover the **request side** of Context/header-forwarding well, and the two integration tests are useful smoke tests. But the added `Consent` component is only smoke-tested (a `_consent` cookie appears), and the whole **response side** is untested because the unit `fetch` stub returns empty handles. Could we add coverage for the following before merge?
>
> **Consent behavior** (component is wired, but none of the behavior is tested)
> - Consent gates events: with `defaultConsent: "pending"`, an event is held and sends only after opt-in
> - Opt-out blocks the event — no `/v1/interact` request fires and `sendEvent` resolves empty
> - Revoking consent after an opt-in blocks later events
> - `defaultConsent: "out"` blocks events, then opt-in lets them send
> - The consent **value** (in vs out) is transmitted to Edge and written to the cookie (the smoke test never checks the value)
> - `setConsent` forwards an `identityMap`
> - Consent persists across two requests sharing a cookie jar (both opt-in and opt-out)
> - Stored consent is cleared when the identity cookie is missing (the "consent is tied to identity" branch)
> - Adobe 2.0 consent standard (`value.collect.val` y/n) gates events
> - `setConsent` rejects invalid/malformed options (and an Edge 4xx) instead of resolving silently
> - Identity still resolves after a failed `setConsent`
> - Duplicate identical `setConsent` calls dedupe to a single `/privacy/set-consent` request
>
> **Response side** (untested — the unit `fetch` stub returns empty handles)
> - A `state:store` response actually writes the identity/consent cookies through the real cookie service
>
> **Context** (request side is covered; these are not)
> - `Accept-Language` is forwarded — the wired unit test asserts only `User-Agent`
> - Cross-request isolation: one request's referer-derived URL must not leak into a later request with no referer (shared, long-lived instance)
> - Multiple events from one `forRequest` handle each get `implementationDetails` and the URL
> - Precedence: a caller-supplied `web.webPageDetails.URL` vs the referer-derived URL — pin down which wins
>
> **Identity / consent interaction**
> - `getIdentity()` behavior while `defaultConsent: "pending"` (resolves, hangs, or rejects?)
> - `appendIdentityToUrl` end-to-end coverage
>
> **Lower priority**
> - Concurrent `setConsent` calls serialize and the last value wins
> - Consent carried inline on `sendEvent` instead of via `setConsent`

carterworks (changes requested):
please remove some of the redundant, claude-y comments.

packages/node/src/components/context.js
```diff
@@ -0,0 +1,60 @@
...
+
+import { createRequire } from "node:module";
+import getHeader from "../services/getHeader.js";
+
+// Mirrors @adobe/alloy-core/constants/libraryName.js — hardcoded rather than
+// imported since it's independent of @adobe/alloy-node's own version.
+const LIBRARY_NAME = "https://ns.adobe.com/experience/alloy";
+const { version: LIBRARY_VERSION } = createRequire(import.meta.url)(
+  "../../package.json",
+);
```
carterworks:
```suggestion
import packageJson from "../../package.json" with { type: "json" };
import getHeader from "../services/getHeader.js";

const LIBRARY_NAME = "https://ns.adobe.com/experience/alloy";
const { version: LIBRARY_VERSION } = packageJson;
```

packages/node/package.json
carterworks: 18 is end of life. Let's start with the oldest, supported lts - 22

carterworks (changes requested):
other than this small issue with `createRequire`, looks good.

Every time i try to get those claude models to use `with { type: "json" }`, they do that `createRequire` thing.

carterworks (approved):

packages/node/package.json
```diff
@@ -3,7 +3,6 @@
   "version": "1.0.0-beta.0",
```
carterworks:
```suggestion
  "version": "0.0.1-beta.0",
```

this will create a beta version on npm. the full version will be released the next time we release a full browser version. (the release process is synchronous - new version and stable version are all released at the same time)

carterworks (changes requested):

carterworks (approved):

PR #1566, opened 2026-09-01
feat(reactor-extension): add Brand Concierge transferCookies UI

jonsnyder:
claude run

## Summary

Adds UI for the Brand Concierge `conversation.transferCookies` configuration option (an array of first-party cookie names) to the Tags extension configuration view and manifest.

This is the extension-side counterpart to the SDK config field added on the `munchkin` branch (`conversation.transferCookies` in `packages/core/src/components/BrandConcierge/configValidators.js`). The UI emits `transferCookies` so the value is consumed by the SDK once that change lands.

> Note: this was requested as "extraCookies", but the actual SDK config field is `transferCookies`; the UI/manifest use `transferCookies` to match.

## Changes

- **`brandConciergeSection.jsx`** — a `FieldArray`-based "Transfer cookies" list (text fields with per-row delete buttons and an "Add cookie" button), plus default settings, load/save handling (trims entries, drops empty ones, and only writes when non-default), and validation (`array().of(string())`). Always shows at least one field; the delete button is disabled when only one field remains. Label and description are placed on the first and last fields respectively, avoiding redundant markup.
- **`createExtensionManifest.mjs`** — adds `transferCookies` as `{ type: "array", items: { type: "string" } }` under `conversation` in the generated `extension.json` schema.
- **`brandConciergeSection.spec.jsx`** — 5 new integration tests: load from settings, add/save, omit when empty, trim-and-drop-empty, and remove.
- Adds a changeset (`reactor-extension-alloy`: minor).

## Testing

- Integration suite: **18/18 pass** (13 existing + 5 new).
- ESLint clean on changed files.
- `extension.json` regenerates and passes the manifest validator.

### DIFF
--- a/packages/reactor-extension/src/view/configuration/brandConciergeSection.jsx
+++ b/packages/reactor-extension/src/view/configuration/brandConciergeSection.jsx
@@ -270,7 +270,7 @@ const BrandConciergeSection = ({ instanceFieldName }) => {
               data-test-id="transferCookiesDataElementOption"
               value={DATA_ELEMENT}
             >
-              Use a whole data element
+              Use a data element
             </Radio>
           </FormikRadioGroup>
           {transferCookiesInputMethod === DATA_ELEMENT ? (
--- a/packages/reactor-extension/src/view/configuration/brandConciergeSection.jsx
+++ b/packages/reactor-extension/src/view/configuration/brandConciergeSection.jsx
@@ -20,7 +20,6 @@ import {
   Flex,
   Button,
   ActionButton,
-  LabeledValue,
 } from "@adobe/react-spectrum";
 import Delete from "@spectrum-icons/workflow/Delete";
 import SectionHeader from "../components/sectionHeader";
@@ -70,6 +69,16 @@ export const bridge = {
     const streamTimeoutMs =
       instanceSettings.conversation?.streamTimeout ?? STREAM_TIMEOUT_MS;
     conversation.streamTimeout = streamTimeoutMs / 1000;
+
+    // Always show at least one transfer cookie field. An untouched empty
+    // field is trimmed and dropped in getInstanceSettings, so this does not
+    // emit a transferCookies setting on its own.
+    if (
+      !conversation.transferCookies ||
+      conversation.transferCookies.length === 0
+    ) {
+      conversation.transferCookies = [""];
+    }
 
     return { conversation };
   },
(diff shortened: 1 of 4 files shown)


packages/reactor-extension/src/view/configuration/brandConciergeSection.jsx
```diff
@@ -169,6 +198,54 @@ const BrandConciergeSection = ({ instanceFieldName }) => {
         >
           Collect sources
         </FormikCheckbox>
+        <View>
+          <LabeledValue label="Transfer cookies" value="" width="size-5000" />
+          <Content>
```
jonsnyder: Always show at least one text field, and then use the first item in the array for the label, and the last for the description. The content tag renders too big and different than the rest

jonsnyder: Done in 08753165 — the Transfer cookies list now always shows at least one field (delete disabled on the last one), and the label/description are carried by the first/last FormikTextField's own props instead of a separate LabeledValue/Content block.

packages/reactor-extension/src/view/configuration/brandConciergeSection.jsx
```diff
@@ -169,6 +210,55 @@ const BrandConciergeSection = ({ instanceFieldName }) => {
...
+                    />
+                    <ActionButton
+                      data-test-id={`deleteTransferCookie${index}Button`}
+                      isQuiet
+                      variant="secondary"
+                      isDisabled={transferCookies.length <= 1}
+                      onPress={() => {
+                        arrayHelpers.remove(index);
+                      }}
+                      aria-label={`Remove transfer cookie ${index + 1}`}
```
carterworks: vertical alignment of the delete button is incorrect. You can probably copy most of the code from the config overrides report suite section.

Image

Image

packages/reactor-extension/src/view/configuration/brandConciergeSection.jsx
```diff
@@ -169,6 +210,55 @@ const BrandConciergeSection = ({ instanceFieldName }) => {
...
+                      description={
+                        index === transferCookies.length - 1
+                          ? "Additional first-party cookie names to always transfer to Brand Concierge conversation requests, in addition to the ones transferred by default."
+                          : undefined
+                      }
+                      name={`${instanceFieldName}.conversation.transferCookies.${index}`}
+                      width="size-4600"
+                      marginEnd="size-100"
+                    />
+                    <ActionButton
```
carterworks: Should we support data elements?

carterworks (changes requested):

packages/reactor-extension/src/view/configuration/brandConciergeSection.jsx
```diff
@@ -11,35 +11,64 @@ governing permissions and limitations under the License.
...
 const STREAM_TIMEOUT_MS = 10000;
 const STREAM_TIMEOUT_SECONDS = STREAM_TIMEOUT_MS / 1000;
 
+// Whether transferCookies is entered as individual cookie names (FORM) or as a
+// single data element that resolves to the whole array (DATA_ELEMENT).
+const FORM = "form";
+const DATA_ELEMENT = "dataElement";
+
+const TRANSFER_COOKIES_DESCRIPTION =
+  "Additional first-party cookie names to always transfer to Brand Concierge conversation requests, in addition to the ones transferred by default.";
```
carterworks, line 48: can multiple cookies be specified in a single data element when selecting the "Data Element" option? If so, are they comma separated?

Image

carterworks (changes requested):

packages/reactor-extension/src/view/configuration/brandConciergeSection.jsx
jonsnyder: You are right I should add information that the data element should resolve to an array of strings.

PR #1571, opened 2026-09-10
FEAT: Add v2 Edge API implementation; Node SDK improvements

Spencer-Smith:
## Changed Packages

- [x] core
- [x] node
- [ ] reactor-extension 

## Description

- Enables core package to make requests to the v2 Edge AI when credentials are provided
- Requires credentials when using node SDK (has a workaround for testing/initial setup)
- Disables "pending" consent in Node SDK
- Updates implementation details in Node SDK

## Related Issue

Related to Universal JS Migration

## Motivation and Context

Allows requests to be made to v2 Edge API which is authenticated. This ensures integrity of recorded events for users as well as access to v2 features/improvements (I'm not aware of them, but I suppose they have to exist). Also changes/fixes a few things in the Node SDK that were determined during team review (require credentials, remove "pending" consent until we have customer signal that it is desired, update library name for node sdk in implementationDetails context).

N/A

## Documentation

To be included in [Node SDK documentation](https://jira.corp.adobe.com/browse/PDCL-16447)

## Types of changes

## Checklist:


packages/core/src/core/config/createCoreConfigs.js
```diff
@@ -29,4 +33,12 @@ export default () =>
     orgId: string().unique().required(),
     onBeforeEventSend: callback().default(noop),
     edgeConfigOverrides: validateConfigOverride,
+    // Opt-in OAuth Server-to-Server credentials — when present, requests
+    // use the authenticated Server API (v2) instead of v1.
+    edgeCredentials: objectOf({
+      clientId: string().nonEmpty().required(),
+      clientSecret: string().nonEmpty().required(),
```
carterworks, line 40: debug logging and the `onInstanceConfigured` both log the config in full. We may want to consider finding a way to redact the client secret from these logs (like a regex-exclusion list) so that they aren't naively exposed in customer implementations.

carterworks (changes requested):

packages/core/src/utils/redactSensitiveValues.js
```diff
@@ -0,0 +1,48 @@
...
+ * @param {RegExp} [sensitiveKeyPattern]
+ * @returns {any}
+ */
+export default (value, sensitiveKeyPattern = DEFAULT_SENSITIVE_KEY_PATTERN) => {
+  // clone() can't round-trip undefined (JSON.stringify(undefined) isn't a
+  // string); callers pass a possibly-missing config/options unconditionally.
+  if (value === undefined) {
+    return undefined;
+  }
+  const redacted = clone(value);
```
carterworks, line 45: This `clone` makes me uneasy. Because it uses `JSON.parse(JSON.stringify())`, it can lose information. Does it accidentally `undefined` the `onBeforeEventSend` and other hooks and monitors?
I'd prefer something more surgical that copies objects with keys and `...` spread-copy the rest 
Kinda like this (but better)
```js
const redactEdgeCredentials = (value) => {
  if (
    !value?.edgeCredentials ||
    !Object.prototype.hasOwnProperty.call(
      value.edgeCredentials,
      "clientSecret",
    )
  ) {
    return value;
  }

  return {
    ...value,
    edgeCredentials: {
      ...value.edgeCredentials,
      clientSecret: "[REDACTED]",
    },
  };
};
// …
const redactOptionsIfConfigure = (data) =>
  data.commandName === "configure"
    ? { ...data, options: redactEdgeCredentials(data.options) }
    : data;
And for computed configuration:
const redactedData = {
  ...data,
  config: redactEdgeCredentials(data.config),
};
```

packages/core/src/core/createLogger.js
```diff
@@ -44,23 +53,35 @@ export default ({ getDebugEnabled, console, getMonitors, context }) => {
...
       log(
         "info",
         `Executing ${data.commandName} command. Options:`,
-        data.options,
+        redactedData.options,
       );
     },
     logOnCommandResolved(data) {
-      notifyMonitors("onCommandResolved", data);
+      const redactedData = redactOptionsIfConfigure(data);
```
carterworks, line 78: this should be inside the `log` function and the `notifyMonitors` function, instead of inside each hook. that way you only need to call it twice.

carterworks: you also need to put it in `prepareLibraryInfo`

carterworks (changes requested):

packages/core/src/utils/redactSensitiveValues.js
Spencer-Smith: Fixed in 9484ca92 — replaced the JSON-clone/regex approach with a surgical `redactEdgeCredentials` util that only touches `edgeCredentials.clientSecret` via a shallow spread-copy (basically your suggestion). Everything else, including functions like `onBeforeEventSend`, survives as the exact same reference — added a test proving that. Also moved the redaction into `log()`/`notifyMonitors()` so it's centralized in two spots instead of duplicated per-hook.

packages/core/src/core/createLogger.js
Spencer-Smith: Done in 9484ca92 — redaction now lives inside `log()` (maps over every logged arg) and `notifyMonitors()` (redacts `data.config`/`data.options` once), so it's called from 2 places instead of inside each hook. Also added it to `prepareLibraryInfo` — good catch, `getLibraryInfo()` was returning the secret in plaintext with no redaction at all.

packages/core/src/core/config/createCoreConfigs.js
Spencer-Smith: Addressed in 9484ca92 — clientSecret is now redacted before it reaches the console, any monitor callback (onInstanceConfigured, onBeforeCommand, etc.), or getLibraryInfo()'s result.

packages/core/src/core/createLogger.js
```diff
@@ -10,18 +10,26 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+  // without redacting each one individually.
   const notifyMonitors = (method, data) => {
     const monitors = getMonitors();
     if (monitors.length > 0) {
-      const dataWithContext = { ...context, ...data };
+      const dataWithContext = {
+        ...context,
+        ...data,
+        config: redactEdgeCredentials(data.config),
+        options: redactEdgeCredentials(data.options),
```
carterworks, line 31: This adds `data` and `options` to every single monitoring hook (new functionality). Something to be aware of. I suppose it is fine.

packages/core/src/core/createLogger.js
```diff
@@ -33,17 +41,10 @@ export default ({ getDebugEnabled, console, getMonitors, context }) => {
   const log = (level, ...rest) => {
     notifyMonitors("onBeforeLog", { level, arguments: rest });
```
carterworks, line 42: `onBeforeLog` is still getting the unredacted config options.

carterworks (changes requested):

carterworks (approved):

PR #1572, opened 2026-09-15
test(browser): isolate browser integration tests

carterworks:
## Changed Packages
- [ ] core
- [x] browser (test config only)
- [ ] reactor-extension

## Description

Experiment. Sets the `browser/integration` Vitest project to `isolate: true` (was `false`).

### Why
The `Generate Coverage Report` job runs all ~500 test files in one long browser process with `FILE_PARALLELISM=false`. With `isolate: false`, state can leak across files (unawaited alloy commands, MSW handlers), degrading the shared process. Under that degradation, react-spectrum overlay integration tests intermittently exceed the 30s per-test timeout — e.g. the recent `configOverrideSection.spec.jsx` timeout on the coverage job.

Locally, in a CPU-limited container (1 vCPU, full `reactor-extension/integration` project with coverage), ~1 in 6 runs shows a whole-process slowdown (a normally ~18s file jumps to ~77s) and a react-spectrum test times out at 30s. `isolate: true` resets per-file state and is the first lever to test.

### Goal
Measure the effect on the CI coverage job. This is a targeted, single-flag change so the CI signal is clean; it is not a final fix. Note `reactor-extension/integration` is already `isolate: true`.


carterworks:
Negative result: with browser/integration isolate:true, the coverage job still timed out on the same test (configOverrideSection.spec.jsx:292, 30000ms) on the first run. Per-project file isolation does not fix the flake — reactor-extension/integration is already isolate:true and still degrades. The driver is a process-wide slowdown across the ~500-file single-process coverage run, not reloadable per-file module state. Closing this experiment.

